PR #27, opened 2026-05-19
status_data: postMessage-backed read/write API for STATUS.html (window.bbStatus)

SawyerHood:
## Motivation

A manager's STATUS.html lives in `~/.bb/thread-storage/<thread-id>/STATUS.html` and renders in an unsandboxed iframe in the bb secondary panel. Today it can only display static markup — nothing it does can survive a reload. This change gives it a tiny persistence surface so authors can build interactive views: todo lists, checkbox toggles, dismissible alerts, anything that wants to remember state across renders.

## Architecture

```
[iframe STATUS.html]
  └─ window.bbStatus.{read,write}
     │   (postMessage to window.parent)
     ▼
[bb app parent]
  └─ HtmlFilePreview postMessage listener
     │   validates event.source === iframe.contentWindow
     │   fetch
     ▼
[bb server]
  └─ GET /api/v1/threads/:id/thread-storage/content?path=...
  └─ PUT /api/v1/threads/:id/thread-storage/content?path=...
     │   path validation + 1 MiB cap + content-type check
     │   queueCommandAndWait
     ▼
[host daemon]
  └─ host.write_file (atomic write under declared rootPath)
     │
     ▼
~/.bb/thread-storage/<thread-id>/STATUS_DATA.json
```

The `srcdoc` iframe has a null/opaque origin and cannot resolve relative URLs or CORS-call the server directly, so it must round-trip through its parent (which is same-origin to the bb server and already authenticated).

## Security notes

- **Source-window check.** The parent listener compares `event.source` against `iframeRef.current.contentWindow` and drops anything else. Origin is null for srcdoc iframes, so source identity is the only signal we can trust.
- **Path validation.** The PUT route only accepts a bare filename — no `/`, no `\`, no leading `.`, no `..`, no NUL byte, and no `length > 255`. The daemon then double-checks that the resolved file is contained within the declared `rootPath` (parent-dir realpath to defeat `/tmp` → `/private/tmp` symlink tricks). Bare-name validation is stricter than the GET route because writes mutate state and we do not want callers planting files outside the storage root or alongside daemon tmp files.
- **Size limit.** 1 MiB (`THREAD_STORAGE_CONTENT_MAX_BYTES`). Checked from `Content-Length` first, then re-checked after reading the body. The daemon `HOST_WRITE_FILE_MAX_BYTES` mirrors the server cap so a stale server cannot ship a larger payload than the daemon will accept.
- **Auth.** Inherits the same authentication as the existing GET route — no new surface.
- **Atomic writes.** Daemon writes to a `.bb-write-<rand>.tmp` file and renames into place, so a concurrent reader never sees a torn write.

## Files changed

- `packages/host-daemon-contract/src/commands.ts` — adds `host.write_file` command + result schema, bumps `HOST_DAEMON_PROTOCOL_VERSION` to 16, exports `HOST_WRITE_FILE_MAX_BYTES`.
- `packages/host-daemon-contract/src/index.ts` — re-exports the new command/constant.
- `apps/host-daemon/src/command-handlers/file-write.ts` — new atomic write-under-root helper.
- `apps/host-daemon/src/command-handlers/host-files.ts` — `writeHostFile` thin wrapper.
- `apps/host-daemon/src/command-dispatch.ts` — wires the new command.
- `packages/server-contract/src/api-types.ts` — `THREAD_STORAGE_CONTENT_MAX_BYTES` + `ThreadStorageContentWriteResponse`.
- `packages/server-contract/src/public-api.ts` — adds `$put` to the `/threads/:id/thread-storage/content` endpoint.
- `packages/server-contract/src/index.ts` — re-exports the new constant/type.
- `apps/server/src/routes/threads/data.ts` — registers `PUT /threads/:id/thread-storage/content` with bare-filename validation, content-type negotiation (JSON / text), size cap, and JSON-validity check.
- `apps/server/src/internal/command-result-owners.ts` — registers the new command type.
- `apps/server/src/services/threads/default-template/STATUS.html` — ships `window.bbStatus.read/write` inline + a documented snippet block.
- `apps/app/src/lib/iframe-status-bridge.ts` — new pure helpers (`parseIframeStatusBridgeRequest`, `executeIframeStatusRead`, `executeIframeStatusWrite`, `handleIframeStatusRequest`).
- `apps/app/src/components/secondary-panel/FilePreview.tsx` — `HtmlFilePreview` mounts a postMessage listener tied to its iframe ref when an `iframeStatusBridgeThreadId` is provided.
- `apps/app/src/components/secondary-panel/ThreadStorageFilePreview.tsx` — threads the thread id through `ThreadStorageFilePreview` → `SecondaryPanelFilePreview` → `FilePreview`.
- `apps/app/src/views/thread-detail/ThreadDetailView.tsx` — passes `threadId` to `ThreadStorageFilePreview`.
- `packages/templates/src/templates/bb-guide-status.md` — new `bb guide status` chapter (HTML helper, HTTP routes, limits, worked todo-list example).
- `packages/templates/src/templates/bb-guide-overview.md` — adds `status` to the chapter list.
- `apps/cli/src/commands/guide.ts` — registers the new chapter.

## Tests added

- `apps/host-daemon/src/command-handlers/host-files.test.ts` — 6 cases for `writeHostFile`: utf8 happy path, overwrite, base64 decode, escape-root rejection, 1 MiB cap, missing `rootPath`.
- `packages/host-daemon-contract/test/contract.test.ts` — schema parse/reject and result-shape coverage for `host.write_file`.
- `apps/server/test/public/public-thread-data.test.ts` — 5 cases for the PUT route: json happy path, bare-name rejections (slashes, traversal, dotfile, empty, backslash), 413 oversize, invalid-JSON rejection, text/plain happy path.
- `apps/app/src/lib/iframe-status-bridge.test.ts` — 12 cases for the message parser, read/write executors, and overall handler error pathways.
- `apps/app/src/components/secondary-panel/FilePreview.test.tsx` — adds 3 component-level cases: relays a read message, ignores messages from non-iframe sources, reports `ok: false` on server failure.
- `apps/server/test/threads/manager-storage-templates.test.ts` — asserts the bundled STATUS.html now contains `window.bbStatus` and the new snippet block.
- `packages/templates/test/templates.test.ts` — confirms `bbGuideStatus` lists, renders, and is referenced from the overview.
- `apps/cli/src/__tests__/command-output.test.ts` — `bb guide status` prints non-empty content; updates the unknown-chapter error to expect `status` in the list.

## Validation

```
pnpm exec turbo run typecheck --filter=@bb/server --filter=@bb/app --filter=@bb/templates --filter=@bb/cli --filter=@bb/server-contract --filter=@bb/host-daemon-contract --filter=@bb/host-daemon
# → 7/7 packages green

pnpm exec turbo run test --filter=@bb/server --filter=@bb/app --filter=@bb/templates --filter=@bb/cli --filter=@bb/server-contract --filter=@bb/host-daemon-contract --filter=@bb/host-daemon --force
# → all packages green (server 673 passed, app 547 passed, host-daemon 275 passed, etc.)

pnpm exec turbo run build --filter=@bb/server
# → dist/default-template/STATUS.html contains window.bbStatus (8 occurrences)

git diff --check
# → clean
```

I did not run the dev server end-to-end manually — the iframe ↔ parent ↔ server flow is exercised through the listener + bridge unit tests and the server route integration tests above. The full round-trip on a live manager thread should be sanity-checked before relying on it for non-trivial state.

## Test plan

- [ ] `pnpm exec turbo run typecheck` across the affected filters
- [ ] `pnpm exec turbo run test` across the affected filters
- [ ] `pnpm exec turbo run build --filter=@bb/server` and inspect `dist/default-template/STATUS.html`
- [ ] Manual smoke test in dev: write a STATUS.html that calls `window.bbStatus.write({hello: 'world'})` then `window.bbStatus.read()` and verify `~/.bb-dev/thread-storage/<id>/STATUS_DATA.json` contains the payload
- [ ] `bb guide status` prints the new chapter


SawyerHood:
Architecture pivot — throwing this out in favor of a different design where the iframe `postMessage`s an event to the **manager thread** (as a new user turn), and the manager applies the update via its existing file-write tools. Avoids the server-side file-write route + STATUS_DATA.json overhead. New PR coming as a draft.

PR #28, opened 2026-05-19
status: window.bbStatus.tell — iframe postMessage bridge to the manager thread (prototype, draft)

SawyerHood:
## Motivation

Make `STATUS.html` interactive — buttons, checkboxes, forms — without giving
the iframe direct write access to the manager thread's storage. This PR
replaces the PR #27 design (server PUT route + `STATUS_DATA.json` +
`window.bbStatus.read/write`); none of that code is carried over. The new
design keeps the manager in the loop on every interactive event: slower per
event than direct file I/O, but the manager has unlimited file/tool access
and decides exactly how to react.

## Architecture

```
[iframe STATUS.html]
  └─ postMessage({ type: 'bb-status:tell', text }) ─> [bb app parent]
                                                       └─ POST /threads/:id/send (existing route)
                                                                                                  └─ Manager wakes with new user turn
                                                                                                                                                                                                └─ Manager edits STATUS.html (or anything else)
```

The iframe doesn't write files. It notifies the manager: "user did X in
STATUS.html — please react." The manager reads STATUS.html before deciding
what to do.

## Reused route

`POST /threads/:id/send` — `apps/server/src/routes/threads/actions.ts:74-90`.
This is the same route powering `bb thread tell <id> "..."`
(`apps/cli/src/commands/thread/actions.ts:228`) and the in-app composer
(`apps/app/src/hooks/mutations/thread-runtime-mutations.ts:290`). The new
listener calls the existing typed wrapper `sendThreadMessage()` at
`apps/app/src/lib/api.ts:809`, with `mode: "auto"` and the text wrapped as
`{ type: "text", text }`. No new server route, no new schema field.

## Security notes

- **Source-window scoping**: the parent listener compares `event.source` to
  `iframeRef.current?.contentWindow` by identity. Replies are sent only to
  that exact `MessageEventSource` (not by origin — STATUS.html is loaded via
  `srcdoc` and has the opaque `"null"` origin, so `targetOrigin: "*"` is the
  appropriate choice given the identity scoping).
- **Size cap**: 4 KiB byte-counted via `TextEncoder` before the request hits
  the server. Larger payloads reject at the bridge.
- **No path traversal**: the iframe never names a file — it only sends free
  text that becomes a user turn. The manager decides what to write, where.
- **No tool access exposed**: the bridge dispatches one specific endpoint
  (send user message). Nothing else is reachable from the iframe.

## Files changed

- `apps/app/src/lib/iframe-status-bridge.ts` (new) — pure validation +
  dispatch helpers; UTF-8 byte counting; result builder; `handleBbStatusMessage`
  that ties parse + send + reply together.
- `apps/app/src/components/secondary-panel/FilePreview.tsx` — `HtmlFilePreview`
  gains a postMessage listener (with cleanup) when `state.kind === "html"`
  carries a `managerThreadId`. Default sender calls `sendThreadMessage` from
  `@/lib/api` with `mode: "auto"`.
- `apps/app/src/components/secondary-panel/ThreadStorageFilePreview.tsx` —
  threads `managerThreadId` through `SecondaryPanelFilePreview` and emits it
  on the `kind: "html"` state.
- `apps/app/src/views/thread-detail/ThreadDetailView.tsx` — passes the route
  `threadId` to `ThreadStorageFilePreview`.
- `apps/server/src/services/threads/default-template/STATUS.html` — inlines
  the `window.bbStatus` helper before `</body>` and adds a "Send a message
  to the manager" entry in the snippet-library comment block.
- `packages/templates/src/templates/bb-guide-status.md` (new) — `bb guide
  status` chapter: authoring patterns, message-text advice, 4 KiB cap,
  iframe origin behaviour, minimal helper snippet.
- `packages/templates/src/templates/bb-guide-overview.md` — appends `status`
  to the chapter list.
- `apps/cli/src/commands/guide.ts` — registers `status: "bbGuideStatus"`.

## Tests added

- `apps/app/src/lib/iframe-status-bridge.test.ts` (15 tests) — envelope
  detection, shape validation, byte cap (incl. multi-byte UTF-8 boundary),
  dispatch success/failure, no-thread fallback, id echo (numeric + null),
  ignored non-tell envelopes.
- `apps/app/src/components/secondary-panel/FilePreview.test.tsx` (+2 tests)
  — wired listener forwards to the mocked `sendThreadMessage` with the right
  args and posts the ack back to the iframe window; cleanup on unmount;
  events whose `source` isn't the iframe are ignored.
- `packages/templates/test/templates.test.ts` — `bbGuideStatus` renders with
  the expected anchors; overview lists the new chapter.
- `apps/cli/src/__tests__/command-output.test.ts` — `bb guide status`
  prints; unknown-chapter error mentions `status` in the available list.
- `apps/server/test/threads/manager-storage-templates.test.ts` — bundled
  default STATUS.html contains `window.bbStatus` and the `tell(text)`
  method plus the bridge message-type constants.

## Validation run

- `pnpm exec turbo run typecheck --filter=@bb/server --filter=@bb/app --filter=@bb/templates --filter=@bb/cli` → 4/4 ok.
- `pnpm exec turbo run test --filter=@bb/server --filter=@bb/app --filter=@bb/templates --filter=@bb/cli --force` → templates 11, cli 183, server 668 (+4 pre-existing skipped), app 549. All green.
- `pnpm exec turbo run build --filter=@bb/server` → ok; `dist/default-template/STATUS.html` includes `window.bbStatus` (4 occurrences).
- `git diff --check` → clean.

## Where to read about it

`bb guide status` (the new chapter) — covers iframe authoring,
`window.bbStatus.tell(text)` usage, message-text guidance ("phrase the text
so the manager can act without ambiguity"), 4 KiB cap, and a minimal helper
snippet if you re-author STATUS.html from scratch.

## Open questions / future work

This prototype intentionally does NOT handle:

- **Latency feel** — `tell()` triggers a full manager wake-up, which can take
  seconds. The UI does nothing during that time. A spinner / pending state
  in STATUS.html is the author's responsibility today; a built-in pending
  affordance might be worth it.
- **Queueing** — rapid clicks each schedule a separate user turn. Whether
  the manager should debounce, coalesce, or queue them is unresolved.
  There is no per-iframe rate limit beyond the 4 KiB cap per call.
- **Multiple iframes in one thread** — STATUS.html is currently the only
  iframe authored to use this bridge, but the listener is per-`HtmlFilePreview`
  instance. If two manager-storage HTML files were ever previewed
  simultaneously they would share the global `message` event but each
  filter on its own `iframeRef`. The contract for "which iframe owns the
  thread" is not formalized.
- **Response routing beyond ok/error** — the ack is binary success/failure.
  There is no way for the iframe to await a manager *response* (e.g. "todo
  marked done"). The current path is: iframe sends `tell()`, manager
  decides, manager edits STATUS.html, iframe reloads from new contents.
  Round-tripping structured replies would need a separate channel.


SawyerHood:
Closing — prototype superseded by the alternate direction we explored in #27/#30. Not pursuing the iframe→manager-tell bridge for now; if we revisit, will open a fresh PR.

PR #29, opened 2026-05-19
fix: forward prompt attachments to Claude Code adapter

SawyerHood:
## Summary

- File attachments uploaded in the bb UI never reached Claude Code provider sessions: the bridge's `extractInputText` filter silently dropped every `localImage`, `localFile`, and `image` entry before building the SDK user message.
- Replace the text-only filter with a multimodal `buildUserMessageContent` that reads each staged attachment from disk and emits the right Claude Agent SDK content blocks (`text`, base64 `image`, base64 PDF, plain-text `document`, or a path-bearing text marker for unsupported binary types).
- Text-only turns still produce a plain `string` content, so cache hits and existing flows are unchanged.

## Root cause

`packages/agent-runtime/src/claude-code/bridge/bridge.ts:1293-1305` (pre-fix). The upload, server intake, daemon staging, runtime dispatch, and adapter all forward attachments correctly — the bridge then throws away everything that is not `{ type: "text" }` before calling `SdkSession.pushInput`. Codex and Pi at least preserve a text marker; Claude Code was the odd one out.

Full bug report (trace, comparison vs other providers, file:line): `~/.bb/thread-storage/thr_6ww2pwqsd7/reports/attachment-claude-code-bug.md`.

## Test plan

- [x] `pnpm exec turbo run typecheck` across all packages
- [x] `pnpm exec turbo run test --filter=@bb/agent-runtime` (511 tests pass — 14 new)
- [x] Full repo test run (`pnpm exec turbo run test`) — all 26 packages green
- [x] New unit tests for `buildUserMessageContent` cover text-only, missing input, base64 image (PNG/JPEG), URL image, plain-text document, PDF document, unsupported binary fallback, missing-file fallback, and the legacy string code path
- [x] New bridge integration tests assert the SDK receives `content: string` for text-only turns and the correct `ContentBlockParam[]` for `localImage` and `localFile` turns
- [ ] Manual repro in dev (left to reviewer) — drag an image into a Claude Code manager chat and confirm the model sees it

### DIFF
--- a/packages/agent-runtime/src/claude-code/bridge/__tests__/bridge.test.ts
+++ b/packages/agent-runtime/src/claude-code/bridge/__tests__/bridge.test.ts
@@ -297,6 +297,18 @@ async function readNextPromptText(call: ClaudeQueryCall): Promise<string> {
   return content;
 }
 
+type ClaudePromptContent = SDKUserMessage["message"]["content"];
+
+async function readNextPromptContent(
+  call: ClaudeQueryCall,
+): Promise<ClaudePromptContent> {
+  const result = await call.prompt[Symbol.asyncIterator]().next();
+  if (result.done) {
+    throw new Error("Expected Claude prompt input");
+  }
+  return result.value.message.content;
+}
+
 function createResultUsage(): SdkResultUsage {
   return {
     cache_creation: {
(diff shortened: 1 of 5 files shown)


SawyerHood:
Direction change — closing in favor of a simpler approach. Instead of inlining attachments as native Claude content blocks, Claude Code's adapter will pass attachments as text markers naming the path, matching Codex's behavior. Same worker (thr_6ww2pwqsd7) will produce the new PR; references-not-contents is the unifying principle across providers.

PR #40, opened 2026-05-20
fix: nested managers regression in sidebar grouping/rendering

SawyerHood:
## Summary
- PR #35's recursive nested-manager rendering shipped intact (grouping pass + recursive `ManagerThreadGroupRow` are correct, and the existing tests pass), but the manager-row indent cap stopped at depth 1: every manager at depth 1 *and* depth 2+ rendered at `pl-14`. A user with a 3-level chain (root → Layer 1 → Layer 2) saw Layer 2 managers visually appear as siblings of Layer 1 managers — "nested managers not rendering," from their perspective.
- The block was a one-line oversight: the constants block in `ProjectRow.tsx` already documented "Indent caps at depth 2", but `ManagerRowDepth = 0 | 1` and `MANAGER_DEPTH_CAP = 1` only supported one extra level.
- Extend `ManagerRowDepth` to `0 | 1 | 2`, bump the cap, add `SIDEBAR_DEEPLY_NESTED_MANAGER_ROW_PADDING_CLASS = "pl-20"` and the matching `left-[5.5rem]` connector hairline + continuation, and route depth-2+ manager rows + their child group lines through the new tier.
- Add a `ProjectList` regression test that builds a root → nested → deeply-nested chain and asserts each row renders at its own indent tier (`pl-8`, `pl-14`, `pl-20`). Pre-fix the deeply-nested row had `pl-14`; post-fix it has `pl-20`.

## Test plan
- [x] `pnpm exec turbo run typecheck --filter=@bb/app` — clean
- [x] `pnpm exec turbo run test --filter=@bb/app` — 460 / 460 (was 459 pre-fix)
- [x] `git diff --check` — clean
- [x] Verified the bug live: `dev-browser` against the running prod app showed Layer 2 manager rows at `pl-14` (same as Layer 1) under project `proj_55xtp72bhe`. The new failing-then-passing test reproduces that scenario.

### DIFF
--- a/apps/app/src/components/sidebar/ProjectList.test.tsx
+++ b/apps/app/src/components/sidebar/ProjectList.test.tsx
@@ -591,6 +591,88 @@ describe("ProjectList", () => {
     expect(await screen.findByText("Nested Child Task")).toBeTruthy();
   });
 
+  it("indents a depth-2 manager row deeper than its depth-1 parent", async () => {
+    // Three-level manager chain: root (depth 0) → nested (depth 1) →
+    // deeply-nested (depth 2). Each level must render at a distinct indent so
+    // the hierarchy is visually legible; without the depth-2 indent, depth-2
+    // managers visually appear as siblings of their depth-1 parent.
+    const rootManager: ProjectThreadListEntry = {
+      ...makeThreadListEntry("project-1", 1),
+      id: "thr_root_manager",
+      title: "Root Manager",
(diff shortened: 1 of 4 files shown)


SawyerHood:
Closing — PR #35 (the nested-managers UI feature this was a follow-on fix to) is being reverted via #41. Without #35 in main this fix has nothing to fix. Not needed standalone.

PR #58, opened 2026-05-30
fix: stop host-daemon from resurrecting destroyed environments (native watcher crash)

brsbl:
# Fix: host-daemon resurrecting destroyed environments (native file-watcher crash) + watch-lifecycle hardening

Fixes the native file-watcher crash that was repeatedly killing the desktop app, plus related watch-lifecycle hardening.

## Root cause
`requireWorkspaceEnvironment → RuntimeManager.ensureEnvironment` re-provisioned and re-subscribed an FSEvents watcher for any environment named by a `workspace.*` command, with **no guard against already-destroyed envs**. With ~300 destroyed worktrees, every `workspace.status` poll resurrected a dead env + watcher → continuous FSEvents churn → null-pointer segfault in `@parcel/watcher` (`watcher.node`, `FSEventsCallback → DirTree::add/find`) → whole-app crash (`EXC_BAD_ACCESS`). Crash reports: `bb-2026-05-29-145501.ips`, `bb-2026-05-29-183342.ips`.

## Changes
- **Tombstone** destroyed envs so they're never re-watched.
- **Idempotent** `environment.destroy`.
- **`reconcileLiveEnvironments`** on every (re)connect, driven by a new required `liveEnvironmentIds` field in the session-open response — drops watchers/runtimes for idle envs the server no longer lists as live, **keeps envs with active work**.
- **Bounded** `WorkspaceStatusWatcher` retry (60 attempts, resets on success; no tight loop).
- **P2-A heal-gap fix:** reconcile now also lifts a tombstone for any env that reappears in `liveEnvironmentIds`, so an idle env can't get stuck tombstoned-but-`ready` after a failed teardown.

## Validation
- Full-repo typecheck green.
- Tests green: host-daemon, host-watcher, contract, db, server — plus new tests for each scope above (incl. P2-A).

## Known / deferred (not in this PR)
- **P2-B:** the session-open contract is now required+strict on both sides, so a mixed-version (old daemon ↔ new server) reconnect fails session-open. Fine for the bundled app, which restarts server+daemon together. Accepted.
- **P3:** a `thread.start` can briefly lift a tombstone racing a just-processed destroy; self-heals on the next reconcile, no FSEvents leak. No action.
- **Deferred:** upgrade `@parcel/watcher` past 2.5.6 and fix the darwin-x64-vs-arm64 prebuild mismatch. The native segfault can't be caught from JS; eliminating the churn (this PR) is the real fix. Separate follow-up.

---

### What was actually crashing the app (fixed)
1. **Daemon resurrecting dead environments → file-watcher crash** *(the main one)* — see Root cause above. Fixed by this PR.
2. **Two duplicate backends fighting over the database** — two orphaned `pnpm run dev` stacks had been running since May 22, each a full second bb server+daemon pointed at the same data dir + DB. Multiple backends contending on one SQLite file is the likely cause of a separate database-layer crash (`better_sqlite3` segfault, `bb-2026-05-30-004416.ips`). *Operational fix (outside this PR): both stacks killed.*
3. **Database bloat** — bug #1 generated a flood of command records that were never pruned; `host_daemon_commands` reached 174k rows / 388 MB, pushing `bb.db` to 727 MB. *Operational fix (outside this PR): pruned terminal commands + VACUUM → 177 MB.*

### Looked alarming but harmless
- The "Failed to reprime app data change cache" ENOENT flood = references to deleted thread-storage folders. Not a crash.
- Stale-data warnings: a `thread.rename` on a provider-less thread; the status app posting to a deleted thread (404). Cosmetic.

### DIFF
--- a/apps/host-daemon/src/app.test.ts
+++ b/apps/host-daemon/src/app.test.ts
@@ -132,6 +132,7 @@ function createFetchRecorder(
           heartbeatIntervalMs: 30000,
           leaseTimeoutMs: 90000,
           trackedThreadTargets: [],
+          liveEnvironmentIds: [],
         },
         { status: 201 },
       );
@@ -570,6 +571,7 @@ describe("createHostDaemonApp", () => {
             heartbeatIntervalMs: 30000,
             leaseTimeoutMs: 90000,
             trackedThreadTargets: [],
+            liveEnvironmentIds: [],
           },
           { status: 201 },
         );
--- a/apps/host-daemon/src/app.ts
+++ b/apps/host-daemon/src/app.ts
@@ -717,6 +717,17 @@ export async function createHostDaemonApp(
     onTerminalMessage: (message) => terminalManager.handleMessage(message),
     onSessionOpened: (session) => {
       sessionState.value = session.sessionId;
+      void runtimeManager
+        .reconcileLiveEnvironments(new Set(session.liveEnvironmentIds))
+        .catch((error) => {
+          options.logger.warn(
+            {
+              sessionId: session.sessionId,
+              ...runtimeErrorLogFields(error),
+            },
+            "Failed to reconcile live environments after session opened",
+          );
+        });
       runtimeManager.replaceTrackedThreadStorageTargets(
         session.trackedThreadTargets,
       );
--- a/apps/host-daemon/src/command-dispatch-support.ts
+++ b/apps/host-daemon/src/command-dispatch-support.ts
@@ -220,6 +220,16 @@ export async function requireWorkspaceEnvironment(
     return existing;
   }
 
+  // A destroyed environment's worktree is gone. Reconnecting it here would
+  // re-provision a dead path and re-subscribe an FSEvents watcher on every
+  // poll, so refuse instead of resurrecting it.
+  if (runtimeManager.isEnvironmentDestroyed(args.environmentId)) {
+    throw new ExpectedCommandDispatchError(
+      "environment_destroyed",
+      `Environment ${args.environmentId} has been destroyed`,
+    );
+  }
+
   return runtimeManager.ensureEnvironment({
     environmentId: args.environmentId,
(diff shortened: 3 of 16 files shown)


brsbl:
**Safety review: GO — no P0/P1.** Tracking the reviewer's P2/P3 follow-ups here (documentation only; no code change in this PR). Also appended to the PR description's follow-ups section.

### Safety-review follow-ups (review came back GO — no P0/P1)

- **P2-A — reconcile does not heal a stuck tombstone (idle managed-worktree only, recoverable):** `destroyEnvironment` (`runtime-manager.ts`) tombstones *before* the teardown that can throw (`destroyedEnvironmentIds.add(...)` then `runtime.shutdown()` / `workspace.destroy()`). If teardown throws anything other than `path_not_found`, the command fails and the server reverts the env `destroying → ready`, but the daemon stays tombstoned — so every `workspace.status` / `workspace.diff` for that idle env returns `environment_destroyed` until a `thread.start`/terminal lifts the tombstone via `ensureEnvironment`. `reconcileLiveEnvironments` only *adds* tombstones (it iterates `entries`, and a tombstoned env has no entry), so reconnect does not heal it. **Suggested fix:** in `reconcileLiveEnvironments`, also remove from `destroyedEnvironmentIds` any id present in `liveEnvironmentIds`; or only tombstone after teardown succeeds. **Impact:** idle managed-worktree only, recoverable, never affects active threads.
- **P2-B — mixed-version session-open is incompatible by design:** the session-open response `liveEnvironmentIds` field is now required + strict on both sides, so an old-daemon ↔ new-server (or vice-versa) reconnect fails session-open. Fine for the bundled desktop app, which restarts server + daemon together (the hot-swap quits + relaunches the whole app), but noted for any independent/rolling deploy.
- **P3 (minor) — `thread.start`/terminal lifts the tombstone unconditionally** via `ensureEnvironment`. A `thread.start` racing a just-processed `destroy` can lift the tombstone; it self-heals on the next reconcile and causes no FSEvents leak (`createEntry` provisions before subscribing).

PR #68, opened 2026-06-07
Prevent provider idle watchdog during open commands

brsbl:
## Summary
- prevent the provider idle watchdog from selecting active turns while a foreground `commandExecution` item is still open
- keep open-item diagnostics on emitted watchdog events, including command/cwd and latest item activity when present
- render watchdog rows with the most useful open-item detail so genuine idle stops point at the stuck work
- refresh the generated SDK browser bundle after the watchdog event contract change
- give the SDK node-entry smoke test an explicit timeout budget so full CI load does not fail a valid dynamic import/build check

## Why
The prior failure was a false provider-idle stop while the provider was waiting on a long-running foreground command. Provider silence is expected during an open foreground command; command-specific lifecycle policy should own stale command progress instead of the provider watchdog marking the turn detached.

## Verification
- `pnpm exec turbo run test --filter=@bb/server -- test/services/threads/provider-turn-watchdog.test.ts`
- `pnpm exec turbo run test --filter=@bb/thread-view -- test/build-thread-timeline.test.ts`
- `pnpm exec turbo run test --filter=@bb/sdk --force`
- `pnpm exec turbo run typecheck --filter=@bb/domain --filter=@bb/db --filter=@bb/server --filter=@bb/thread-view --filter=@bb/sdk`
- `pnpm exec turbo run typecheck --filter=@bb/sdk`
- GitHub checks passed on head `354b4e827`: CI, Version Lockstep, Ubuntu package smoke, macOS package smoke


ymichael:
deleted the watchdog

PR #82, opened 2026-06-10
fix: regenerate stale app-runtime browser bundle

SawyerHood:
Main's CI (and every open PR) is failing `@bb/sdk` `app-runtime-browser-bundle.test.ts > keeps the committed bundle in sync with the runtime source graph`.

f1838409c ("fix: seed composer with project execution defaults") added `defaultExecutionOptions` to `projectWithThreadsResponseSchema`, which is part of the sdk app-runtime bundle's module graph, without regenerating the committed bundle.

This is just the output of `pnpm --filter @bb/sdk generate:app-runtime-browser-bundle` — no source changes.


SawyerHood:
Superseded by #83 — the bundle is no longer checked in, so the stale copy is deleted rather than regenerated.

PR #88, opened 2026-06-11
Fix blank in-app browser page for localhost SPA dev servers

brsbl:
## Problem
Opening a local SPA dev server (Vite, Next, CRA, …) in bb's in-app browser shows a **blank page**: the main-frame HTML loads but the app never mounts.

## Cause
The loopback/LAN request firewall (`desktop-browser-view.ts` → `shouldBlockBrowserRequest`) allows a same-origin loopback subresource only when the **requesting frame's origin** equals the committed main-frame origin. That origin is derived from `details.frame.origin` — but Electron reports a **blank origin** for a document's *initial* subresource requests (before it has run script and committed its origin). So a Vite page's own `/@vite/client` and `/src/main.tsx` module requests come back with `requestingFrameOriginKey === null` and get cancelled → no JS → blank page. A static page (no subresources) never hits this, which is why the existing fixtures didn't catch it.

## Fix
Add `resolveRequestingFrameLocalOriginKey({ origin, url, isTopFrame })` to the (Electron-free) policy module: prefer the frame's reported origin, and **for a top frame whose origin is blank, fall back to the frame's committed URL**. The fallback is **top-frame-only**, so a sub-iframe presenting a blank origin can never be mistaken for the trusted main frame. The `onBeforeRequest` firewall now uses it (passing `details.frame?.url` and `details.frame?.parent === null`).

## Tests
`@bb/desktop` typecheck + browser policy / view-manager / main-ipc suites green, including:
- new `resolveRequestingFrameLocalOriginKey` unit tests (origin used when present; URL fallback for a blank-origin top frame; **no** fallback for a sub-iframe; non-loopback URL → null; nothing available → null);
- an end-to-end firewall regression: a same-origin SPA subresource resolved via the URL fallback is **allowed**, while a blank-origin sub-iframe reaching the loopback origin is still **blocked**.

## Note
The root cause (Electron blanking `frame.origin` during initial load) is inferred from the blank-page signature + the firewall logic — runtime confirmation is loading a Vite dev server in the desktop in-app browser after this change. The fix is also defensively correct: resolving the requesting frame's origin robustly is right regardless.

### DIFF
--- a/apps/desktop/src/desktop-browser-policy.ts
+++ b/apps/desktop/src/desktop-browser-policy.ts
@@ -356,6 +356,39 @@ export function localRequestOriginKey(url: string): string | null {
   return `${protocolClass}|${parsed.originHost}|${parsed.port}`;
 }
 
+export interface ResolveRequestingFrameLocalOriginKeyArgs {
+  /** The requesting frame's reported origin (`details.frame?.origin`). */
+  origin: string | undefined;
+  /** The requesting frame's committed URL (`details.frame?.url`). */
+  url: string | undefined;
+  /** Whether the requesting frame is the top frame (`frame.parent === null`). */
+  isTopFrame: boolean;
+}
+
+/**
+ * Resolves the comparable local-origin key for the frame that initiated a
+ * request. Prefers the frame's reported `origin`; for a **top** frame whose
+ * origin is not yet populated — Electron reports an empty origin for a
+ * document's initial subresource requests, before it has run script and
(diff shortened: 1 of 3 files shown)


brsbl:
Folded into #85. As part of catching the thread-forking branch up to latest `main`, this localhost in-app-browser SPA fix is now cherry-picked onto that branch (commit `9c33a02c4`), so both fixes ship in one PR.

PR #129, opened 2026-06-12
Server lifecycle: event transition table + CAS single-writer

ymichael:
## What

Thread and environment status changes now flow through a single owner each: a pure `(status, event) → status` transition table in `@bb/domain` plus one CAS-guarded writer in `@bb/db` (`UPDATE … SET status WHERE id = ? AND status = ? RETURNING`). Callers stop choosing target statuses and instead report **what happened** (14 events: `turn.started`, `provision.failed`, `stop.completed`, …); each event declares its supersession predicates (`notDeleted`, `notStopRequested`, …) once in the table; illegal or superseded transitions return a typed no-op that gets logged — never thrown, never silent.

Replaces: ~21 transition call sites picking statuses ad hoc, the throwing `transitionThreadStatus` API + `tryTransition` catch-shims, `ALLOWED_TRANSITIONS`, 3 module-level in-flight Sets (now one `InFlightRpcGuard`), and 5 direct `.set({status})` sites on environments — including `applyProvisionedEnvironment`, which smuggled status through a metadata update.

**Behavior-neutral by design**: the tables encode *observed* behavior, including the questionable parts, marked `// observed:`. Tightening is a separate follow-up with its own QA.

## Why

~30 race fix-commits in 2 months all had one shape: a late async result applied after the world moved on, each fixed with one more scattered guard (90 of them at peak). Status is now writable by exactly two functions; staleness is structurally impossible to forget and observable when it happens.

## Review guide — where the judgment calls live

1. **The inventories** (the most important review artifact): header comments of `packages/domain/test/thread-lifecycle.test.ts` and `packages/domain/test/environment-lifecycle.test.ts`. Every transition site: from→to, assigned event, observed guards, and the classification of all 58+ status guards (became-predicate / non-lifecycle / suspicious).
2. **The `// observed:` cells** in `packages/domain/src/thread-lifecycle.ts` and `environment-lifecycle.ts` — current behaviors preserved verbatim that look wrong, e.g. `provisioning + turn.completed → idle` (a late turn-completion flips a revived thread to idle), and "a late provision success revives an in-flight destroy".
3. **Four latent bugs found and deliberately NOT fixed** (queued in the roadmap): generated-title renames silently dropped for fast-finishing threads; schedule dispatch path has no `stopRequestedAt` guard; queued-message dispatch never re-checks `deletedAt`/`stopRequestedAt` after claiming; `thread-send` applies `error→active` before daemon ack.
4. **Stronger caller-side guards kept** (where the inventory's weakest-common-set predicates are narrower than a specific caller's needs) — listed per-site in the migration commit bodies.

## Behavior deltas (race-window only, both deliberate)

- A duplicate `turn.interrupted` for an already-idle thread no longer re-runs best-effort idle pruning.
- `interruptActiveThreads` no longer rolls back the whole batch when one thread's status moved between selection and apply — that thread no-ops with a log.

## Validation

- Workspace typecheck 29/29; `@bb/domain` 17, `@bb/db` 27, `@bb/server` 77 test files green; integration suite 22/22 (`--force`).
- Race coverage at both layers: writer-level supersession/CAS-conflict tests (incl. a proxy-interleave harness for the conflict case, unreachable through the service API under better-sqlite3), and service-level tests through real paths (late start during in-flight stop, late provision for a tombstoned thread, redelivered stale turn-completion, double stop finalization, destroy completing after revive).
- Exit criteria: `status:` inside `.set(` appears only in the two writers + creation defaults; old API zero hits outside inventory comments; transition-protecting guard survivors: threads 4, environments 7, each commented.
- Not done: live dev-app smoke (stop mid-provisioning, delete mid-turn, daemon-kill mid-command).

The 11 commits tell the story in order: inventory+table → writer → three thread migration chunks → in-flight guard → environment trio → race tests → guard audit.


ymichael:
Superseded by #149, which contains these 11 commits plus the tightening this PR's description deferred: `stopping` replaces the `stopRequestedAt` field, the environment B* decoupling, designed (non-`// observed:`) transition tables, and fixes for all four of the deferred latent bugs listed in this PR's review guide. Closing in favor of #149.

PR #158, opened 2026-06-15
Start thread detail data queries eagerly

ymichael:
## Summary
- add an eager non-visual ThreadDetailDataOwner mounted from AppLayout for thread routes
- move thread detail adjunct loading to normal query hooks instead of bootstrap prefetch flags
- remove git work-status / pull-request loading from the eager owner
- defer environment work-status and PR requests until the initial timeline load has settled, so git status cannot sit on the first timeline render path
- keep visible page observers from immediately duplicate-refetching freshly owned data while preserving popout baseline behavior

## Validation
- pnpm exec turbo run test --filter=@bb/app -- src/hooks/cache-owners/cache-owner-registry.test.ts
- dev-browser timing check on /threads/thr_9z4ej99csc showed secondary non-git thread-detail requests starting together at about +232-238ms after the first API request


ymichael:
Closing per request.

PR #177, opened 2026-06-16
Fix Codex thread persistence for resume

ymichael:
## Summary

- start Codex app-server threads with `ephemeral: false` so bb's idle-reaped thread-scoped Codex processes can be resumed by provider thread id
- remove the obsolete `persistExtendedHistory` request field from Codex `thread/start` and `thread/resume`
- add adapter regression assertions for the durable start flag and stale field removal

## Context

A follow-up on an idle-reaped Codex thread failed with `no rollout found for thread id ...`. bb relies on Codex materializing the provider thread so a later `thread/resume` can reload it after the daemon reaps the live provider process. The current Codex app-server schema exposes `ephemeral` as the durability control, while `persistExtendedHistory` is no longer part of the generated/live schema.

I also checked pingdotgg/t3code: its generated Codex schema likewise uses `ephemeral` and has no `persistExtendedHistory` field. t3code relies on the default rather than setting `ephemeral: false` explicitly; bb is stricter here because it has an idle reaper that intentionally shuts down the live process and later resumes by provider thread id.

## Validation

- `pnpm exec turbo run test --filter=@bb/agent-runtime -- src/codex/adapter.test.ts src/runtime.process-lifecycle.test.ts`
- `git diff --check`

### DIFF
--- a/packages/agent-runtime/src/codex/adapter.test.ts
+++ b/packages/agent-runtime/src/codex/adapter.test.ts
@@ -546,9 +546,11 @@ describe("codex provider adapter", () => {
         approvalPolicy: "never",
         sandbox: "danger-full-access",
         cwd: "/tmp/worktree",
+        ephemeral: false,
         experimentalRawEvents: true,
       },
     });
+    expect(JSON.stringify(cmd)).not.toContain("persistExtendedHistory");
     expect(JSON.stringify(cmd)).not.toContain("baseInstructions");
     expect(JSON.stringify(cmd)).not.toContain("developerInstructions");
   });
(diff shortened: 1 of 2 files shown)


ymichael:
Direct Codex app-server probe (`codex-cli 0.139.0`) results:

- `thread/start` alone can return `ephemeral: false` plus a rollout path before the rollout file/state row exists; a fresh `thread/resume` by id fails until the thread is materialized.
- Using `thread/inject_items` as a no-inference materialization step, the PR payload (`ephemeral: false`, no `persistExtendedHistory`) created a rollout file/state row and resumed successfully from a fresh app-server process.
- The `ephemeral: true` control returned `path: null`, created no Codex `threads` row, and fresh resume failed with the same class of error: `no rollout found for thread id ...`.
- On current Codex, the old bb payload (`persistExtendedHistory: false`, no `ephemeral`) also materializes/resumes after `thread/inject_items`; so this probe supports explicit non-ephemeral start as the right guard and stale-field cleanup, but it is not a clean current-version before/after reproduction.

I also checked the failing local threads (`019ed18e-d915-7991-b398-0e1eab5cec88`, `019ed1cd-f925-7ab2-bc1d-2b7091fef58f`): neither has a rollout file, shell snapshot, nor a row in Codex `state_5.sqlite.threads`, which explains why bb cannot resume them by provider thread id after the idle reaper kills the live process.

PR #201, opened 2026-06-17
Timeline performance: incremental delta updates + response caching

ymichael:
## Summary

The thread timeline was slow and wasteful during streaming and on large threads. On every appended event the client refetched and the server rebuilt the **entire** visible window from scratch — and while a turn is streaming that window is rendered expanded (up to **~950 KB**), re-sent ~5×/second. Over one long turn that's **~214 MB** of mostly-unchanged data across the wire, plus a full deep-compare + re-render each time, repeated for every mounted timeline.

This PR makes live updates **incremental** and adds **server-side caching**, cutting streaming traffic ~97% with no change to what the user sees. Each item below is a separate commit.

## What's in here

1. **Incremental "row-patch" updates for live timelines** _(the main change)._ The client tells the server which revision it already holds (`afterSequence`); the server reprojects the window — so the tricky cases (turn collapse, window eviction, interrupts, background tasks that finish later) stay correct **by construction** — diffs it against what the client has, and returns **only the changed rows**. The client merges them in place, so React re-renders just the rows that changed. Falls back to a full window whenever a delta can't be safely computed.

2. **Server-side response cache.** The timeline endpoint rebuilt from scratch on every request (no caching at all). Added a small bounded cache keyed on the thread's high-water sequence + render options, so idle / repeat / duplicate-mount fetches return instantly instead of repaying the ~130–260 ms build.

3. **Stop refetching immutable turn details.** Appended events were invalidating expanded turn-detail panels, which can't change once a turn is finished. They're now left alone.

4. **Truncate giant inline outputs/diffs.** A few tool outputs are enormous (up to ~1 MB) and dominate payload + render. The window now caps them with a marker; the full content is still available when you expand the turn.

5. **Fix a stale-timeline bug.** An earlier perf change set `staleTime: Infinity` on the timeline query, which silently defeated refetch-on-mount — a thread revisited after events arrived while it was off-screen could stay frozen on stale rows until the next live event. Removed it; revisits refresh again (cheaply now, thanks to 1 & 2).

## Before / after (measured on the dev DB)

**Streaming a long turn** (huge thread, 942-event turn):

| | before (full window every event) | after (row-patch) |
|---|---|---|
| bytes per update | median **495 KB**, peak **889 KB** | median **16 KB** |
| total over the turn | **214 MB** | **6.8 MB** (−96.8%) |
| rows re-rendered | all ~350 | only the ones that changed |

**Re-opening / background-mounted threads:**

| build | before (cold) | after (cache hit) |
|---|---|---|
| typical thread | 64 ms | ~0 ms |
| large thread | 172 ms | ~0 ms |
| huge thread | 239 ms | ~0 ms |

## Correctness & tests

- **Golden test** for the incremental path: `applyDelta(previous)` is asserted byte-for-byte equal to a fresh full window across appends, turn-completion collapse (rows removed + reordered), and no-op updates — exercised through the real route + projection.
- Unit tests for the diff/merge logic and its stale-base fallback; plus tests for the cache, the truncation, and the turn-details invalidation change.
- Green: server-contract 29, server 619, affected app 54, cli 123; typecheck + lint clean across all packages.

## Backward compatibility

Fully compatible in both directions. The new request field (`afterSequence`) is optional and old servers ignore it; the new response fields are additive and clients don't runtime-validate the response, so an old client against a new server behaves exactly as before, and a new client against an old server just never engages the delta path (graceful full fetches). The WebSocket protocol is unchanged.

## Not in this PR

- **Gating hidden side-chat timelines** is intentionally deferred: doing it naively can make a side-chat that still has content look "unused" and delete it (it's coupled to a delete-safety check). Tracked as a follow-up. The incremental updates above already make those background refetches cheap.
- Optional polish noted in the plan: enable HTTP response compression, and shrink the delta further (its floor today is the row-order list).

Full design, investigation, and measurements: `plans/timeline-rendering-performance.md`.

### DIFF
--- a/apps/app/src/hooks/cache-owners/cache-invalidation-groups.ts
+++ b/apps/app/src/hooks/cache-owners/cache-invalidation-groups.ts
@@ -113,6 +113,24 @@ export function getThreadTimelineInvalidationQueryKeys({
       ];
 }
 
+/**
+ * Timeline-window-only invalidation for realtime `events-appended` /
+ * `thread-created` / `thread-deleted`. Deliberately excludes the
+ * turn-summary-details prefix: a completed turn's expanded detail is a fixed
+ * `sourceSeqStart..sourceSeqEnd` range and never changes once the turn is done,
+ * so re-fetching every open detail panel on every appended-event batch is pure
+ * waste during streaming. Mutations that can rewrite history (fork/retry/edit)
+ * still use {@link getThreadTimelineInvalidationQueryKeys}, which invalidates
+ * both prefixes.
+ */
+export function getThreadTimelineWindowInvalidationQueryKeys({
+  threadId,
+}: ThreadScopedInvalidationArgs): QueryKey[] {
+  return threadId
+    ? [threadTimelineQueryKeyPrefix(threadId)]
+    : [allThreadTimelineQueryKeyPrefix()];
+}
+
 export function getThreadQueueContentInvalidationQueryKeys({
   threadId,
 }: ThreadScopedInvalidationArgs): QueryKey[] {
--- a/apps/app/src/hooks/cache-owners/realtime-cache-registry.ts
+++ b/apps/app/src/hooks/cache-owners/realtime-cache-registry.ts
@@ -94,7 +94,7 @@ import {
   getThreadPendingInteractionInvalidationQueryKeys,
   getThreadPromptHistoryInvalidationQueryKeys,
   getThreadQueueContentInvalidationQueryKeys,
-  getThreadTimelineInvalidationQueryKeys,
+  getThreadTimelineWindowInvalidationQueryKeys,
 } from "./cache-invalidation-groups";
 
 interface CollectCachedThreadIdsForEnvironmentArgs {
@@ -498,7 +498,9 @@ function dirtyThreadSearchQueries(): QueryKey[] {
 function dirtyThreadTimelineQueries({
   threadId,
 }: ThreadRealtimeDirtyContext): QueryKey[] {
-  return getThreadTimelineInvalidationQueryKeys({ threadId });
+  // Window only: completed turn-summary-details are immutable, so realtime
+  // event batches must not refetch open detail panels (see helper docs).
+  return getThreadTimelineWindowInvalidationQueryKeys({ threadId });
 }
 
 function dirtyThreadQueueContentQueries({
--- a/apps/app/src/hooks/cache-owners/thread-runtime-cache-owner.test.ts
+++ b/apps/app/src/hooks/cache-owners/thread-runtime-cache-owner.test.ts
@@ -27,6 +27,7 @@ function makeTimelineResponse(): ThreadTimelineResponse {
     activeWorkflow: null,
     pendingTodos: null,
     goal: null,
+    maxSeq: 0,
     timelinePage: {
       kind: "latest",
       segmentLimit: 20,
--- a/apps/app/src/hooks/queries/query-helpers.test.ts
+++ b/apps/app/src/hooks/queries/query-helpers.test.ts
@@ -153,6 +153,7 @@ function makeThreadTimelineResponse(
     activeWorkflow: null,
     pendingTodos: null,
     goal: null,
+    maxSeq: 0,
     rows,
     timelinePage: {
       kind: "latest",
--- a/apps/app/src/hooks/queries/thread-queries.ts
+++ b/apps/app/src/hooks/queries/thread-queries.ts
@@ -24,6 +24,7 @@ import type {
   ThreadTimelineResponse,
   TimelineTurnSummaryDetailsResponse,
 } from "@bb/server-contract";
+import { applyTimelineDelta } from "@bb/server-contract";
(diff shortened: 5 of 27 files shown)


ymichael:
## Before / after validation (re-run)

Environment: live packaged server on `:38886` (the **un-cached** build = "before"); W3 "after" measured in-process against this PR's code path on a copy of the dev DB. Tests: cache unit (7) ✅, route (59) ✅, typecheck ✅, eslint ✅.

### Idle latest window — before (live server)
| thread | payload | latency |
|---|---|---|
| typical (414 ev) | 14.3 KB | 18 ms |
| large (2,986 ev) | 40.5 KB | 122 ms |
| huge (12,206 ev) | 60.1 KB | 131 ms |

### W3 — cold build vs warm cache hit (this PR's change)
| thread | before (cold build) | after (warm hit) | speedup |
|---|---|---|---|
| typical | 77 ms | ~0.01 ms | ~14,000× |
| large | 218 ms | ~0 ms | ~41,000× |
| huge | 263 ms | ~0.01 ms | ~26,000× |

→ Idle / warm-repeat / double-mount refetches go from a full rebuild to effectively free. (Cold first load is unchanged by design — that's the deferred render lever.)

### Streaming active turn — before (motivates the deferred W4, **not** in this PR)
Same huge thread's 942-event turn, built as the latest turn:
| latest-turn state | rows | payload |
|---|---|---|
| collapsed (completed) | 61 | 65 KB |
| **active / streaming** | 353 | **951 KB** |

W3 intentionally does **not** cache the expanded active-turn window (its `maxSeq` changes every event), so streaming is unchanged here — that's what **W4 (row-patch delta)** targets: ~950 KB/refetch → < 1 KB median, ~214 MB → ~0.9 MB per turn. Tracked in the plan as the next item (gated on the patch-diff-fidelity spike).

_Note: the running `:38886` instance is the older packaged build, so the W3 win isn't visible by curling it directly — it's validated in-process + by the unit test (build invoked once per key). A full end-to-end check would require rebuilding/restarting that instance from this branch._

ymichael:
## Before/after validation — re-run (shipped code, dev DB)

**Streaming (W4)** — huge thread's 942-event active turn, 415 refetches, real `computeTimelineRowDelta`:
```
full window / refetch : median 495 KB,  max 889 KB
row-patch  / refetch  : median  16 KB,  p90 26 KB
cumulative / turn     : 214 MB  →  6.8 MB   (−96.8%)
```

**Idle / warm (W3)** — cold build → warm cache hit:
```
typical : 64 ms  → 0.002 ms
large   : 172 ms → 0.002 ms
huge    : 239 ms → 0.004 ms
```

**Correctness (W4)** — golden integration test asserts `applyDelta(prev)` == fresh full window across append / turn-collapse / no-op; unit tests cover the diff/apply algebra + stale-base fallback. server-contract 29 ✓, server 619 ✓, app 54 ✓, cli 123 ✓; typecheck + eslint clean.

Note: the running `:38886` instance is the older packaged build, so these are measured in-process against this branch's code path, not by curling that instance.

ymichael:
## Added: drop `staleTime: Infinity` on the live timeline query

While reviewing, found a related bug from `cb78596611`: `useThreadTimelinePages` set `staleTime: Infinity` on the latest-window query (alongside `refetchOnMount: true`). Because `Infinity` means "never stale," `refetchOnMount` became a no-op — **a thread revisited after events arrived while it was unmounted stays frozen on stale rows** until the next live event (and an idle thread may never refresh). It also didn't reduce streaming rerenders (those are invalidation-driven and refetch regardless of `staleTime`).

Removed the line so the query inherits the app-wide 2s `staleTime` + `refetchOnWindowFocus`, consistent with every other query. Safe to restore now that a revisit/focus refetch is a cheap **delta (W4)** or a **cached/no-op response (W3)** instead of a full-window rebuild — the perf reason the hack existed is gone. Kept the rest of `cb78596611` (its memoization/component-split rerender wins are real and complementary). `useThreadTimelinePages` tests + app typecheck green.

ymichael:
## Rebased on `main` + validated against a real thread

Rebased onto current `main` (resolved conflicts with main's new `activeWorkflow` timeline field and the `AbortSignal` threading on `getThreadTimeline`). Re-ran everything green: server-contract 30, server 684, app 62 (affected), cli 133; typecheck + lint clean.

**Validated against `thr_j8dr5jauvz`** ("Host Daemon Crash", `proj_ij9ap69xzw`) — 682 events / 9 turns — by replaying its **entire real event stream** one event at a time through the shipped delta logic:

**Correctness (the important one):**
```
events replayed        : 677
mismatches             : 0      ← applyDelta(previous) == a fresh full projection, every step
null fallbacks         : 0
steps exercising removal: 7     ← real turn collapses (rows the client held disappearing)
steps exercising upsert : 386
```
So on this thread's real history — including real turn-completion collapses — the incremental patch reconstructs the exact full timeline at every step.

**Performance:**
```
per update : full median 67 KB / max 169 KB  →  delta median 2.8 KB / max 20 KB
over the thread's history : 48.6 MB → 2.4 MB  (−95.2%)
W3 cache (idle window, 27 rows): cold 94 ms → warm 0.007 ms
```

PR #242, opened 2026-06-18
feat(environments): archive grace period + Restore environment

ymichael:
## Problem

Two gaps around a thread losing its environment:

1. **Accidental archive bricks a thread.** Archiving the *last* live thread in a managed environment immediately tore down its git worktree — no undo, uncommitted work gone.
2. **No way back from a gone environment.** A thread whose environment is `destroying`/`destroyed` showed a dead-end "Environment is no longer available" banner.

## What this does

**Archive grace period (the brick fix).** Archiving the last live thread now leaves the environment `retiring` for a configurable window (`managedEnvironmentRetireGraceMs`, default **10s**) before the worktree is destroyed. The gate lives in `advanceEnvironmentCleanup` and is durable across restart (keyed on the row's `updatedAt`; no in-memory timer). It only applies when a **revivable archived thread** remains, so a deleted thread's orphaned environment is still cleaned up immediately.

**Lossless undo.** Within the window:
- Un-archiving the thread fires the existing `retire.cancelled` → environment back to `ready`, worktree (and uncommitted work) intact.
- A 10s **"Thread archived — Undo"** toast offers the same one click. The durable banner Unarchive remains as a fallback.

**Restore environment.** `POST /threads/:id/restore-environment` reprovisions a fresh environment for a thread whose environment is gone:
- The daemon's `createWorktree` now checks out an **existing branch in place** instead of `-B`-resetting it, so committed work (which survives `git worktree remove`) is recovered.
- The thread is re-seeded into `idle` (no automatic turn); routing handles `retiring` (revive in place), `destroying` (409, retry), `unmanaged` (409), and pruned envs (409).
- A "Restore environment" action in the read-only banner ("Cleaning up…" while destroying, active once destroyed).

T3-style per-turn checkpointing was evaluated and left **out of scope** (rationale in the plan): the grace period makes accidental archive lossless without new storage, and it conflicts with T3's own delete-refs-on-archive hygiene.

Design notes and exit criteria: `plans/environment-restore-and-archive-grace-period.md`.

## Testing

- Typecheck clean across server, db, host-workspace, server-contract, app, integration-tests.
- Unit/integration: server 692, db 305, host-workspace 163, app 761 — all pass.
- New regression coverage: grace-window defer-then-destroy, deleted-thread immediate cleanup, unarchive revive, banner Restore action (enabled vs "Cleaning up…"), and an **end-to-end integration test** that commits work → archives (destroy) → restores → asserts the committed file reappears in the fresh worktree on the same branch.

## Follow-ups (not blocking)

- Optional: hide the Restore button for unmanaged environments in the UI (the server already returns 409).

Not needed: `SideChatTabContent`'s "no longer available" copy is a *not-found thread* label, not an environment-gone surface — there is nothing to restore there.

### DIFF
--- a/apps/app/src/components/promptbox/banner/ThreadPromptContextBanner.test.tsx
+++ b/apps/app/src/components/promptbox/banner/ThreadPromptContextBanner.test.tsx
@@ -105,6 +105,54 @@ describe("ThreadPromptContextBanner", () => {
     expect(markup).not.toContain("Provision");
   });
 
+  it("renders an enabled Restore action once the environment is destroyed", () => {
+    const markup = renderToStaticMarkup(
+      <ThreadPromptContextBanner
+        gitSection={null}
+        gitSectionPending={false}
+        archivedSection={null}
+        environmentGoneSection={{
+          status: "destroyed",
+          onRestore: noop,
+          restorePending: false,
+        }}
+        parentThreadSection={null}
+        childThreadsSection={null}
+        pullRequestSection={null}
+        expandedSection={null}
+        onToggleSection={noop}
+      />,
+    );
+
+    expect(markup).toContain("Restore environment");
+    expect(markup).toContain("<button");
+    expect(markup).not.toContain('disabled=""');
+  });
+
+  it("shows a disabled cleaning-up Restore action while the environment is destroying", () => {
+    const markup = renderToStaticMarkup(
+      <ThreadPromptContextBanner
+        gitSection={null}
+        gitSectionPending={false}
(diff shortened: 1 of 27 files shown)


ymichael:
Superseded by #1016, which re-lands this change against the current codebase (this branch had drifted too far to rebase: the app moved to the SDK transport, single-thread archive became the archive-all cascade, and `createWorktree` gained remote-base fetching). #1016 also adds the now-required SDK/CLI surfaces and the host-daemon protocol bump.

PR #245, opened 2026-06-18
[codex] Split turn rollups at user boundaries

brsbl:
## Summary

- Track human user messages from other turns as hard boundaries for completed turn summaries.
- Split completed `Worked for...` summary segments at those boundaries so background parent turns cannot wrap output across a later user message.
- Keep enough summary detail only for crossed turns, then sort projected rows by source sequence after segmentation.
- Add a regression that reproduces the observed shape: `$ottonomous:review` parent turn, overlapping background worker turn, later user `new-turn`, then more parent output.

## Root Cause

Completed turn summaries were projected as one monolithic turn row anchored at the parent turn start. When a background/worker turn kept the parent turn open, a later human user message could land before the parent emitted more assistant/work output. The parent `Worked for...` row could then appear above that user message while containing later turn content.

## Validation

- `pnpm exec turbo run test --filter=@bb/thread-view`
- `pnpm exec turbo run typecheck --filter=@bb/thread-view`

### DIFF
--- a/packages/thread-view/src/apply-turn-message-detail.ts
+++ b/packages/thread-view/src/apply-turn-message-detail.ts
@@ -108,6 +108,7 @@ function applyTurnMessageDetail(
   const includeMessages =
     turn.status === "pending" ||
     turnMessageDetail === "full" ||
+    (turn.externalUserBoundarySeqs?.length ?? 0) > 0 ||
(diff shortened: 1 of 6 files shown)


ymichael:
Superseded by #303, which keeps this summary-boundary behavior but moves the split metadata out of `EventProjectionTurn`.

PR #254, opened 2026-06-19
Fix flaky timeline scroll-to-bottom restores

ymichael:
Summary:
- Cancel a pending saved-row restore when an explicit scroll-to-bottom request happens, so a later ResizeObserver settle pass cannot pull the timeline back up.
- Add a latched user-detached-from-bottom state so streaming/layout drift can save bottom intent, while actual manual scrolls still save the visible row.
- Flush the final scroll anchor from the effect-captured scroll element during unmount, since refs can already be nulled.
- Add regression tests for stale saved-row restore undoing scroll-to-bottom, streaming drift on unmount, and fast unmount after manual wheel scroll.

Validation:
- pnpm exec turbo run test --filter=@bb/app -- src/components/ui/bottom-anchored-scroll-body.scroll-preservation.test.tsx

### DIFF
--- a/apps/app/src/components/ui/bottom-anchored-scroll-body.scroll-preservation.test.tsx
+++ b/apps/app/src/components/ui/bottom-anchored-scroll-body.scroll-preservation.test.tsx
@@ -3,7 +3,10 @@
 import { cleanup, fireEvent, render } from "@testing-library/react";
 import { getDefaultStore } from "jotai";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
-import { BottomAnchoredScrollBody } from "@/components/ui/bottom-anchored-scroll-body";
+import {
+  BottomAnchoredScrollBody,
+  useBottomAnchoredScroll,
+} from "@/components/ui/bottom-anchored-scroll-body";
 import { threadTimelineScrollAnchorAtomFamily } from "@/lib/thread-timeline-scroll-anchor";
 
 // Real externals only: the ResizeObserver/rAF used by the scroll body are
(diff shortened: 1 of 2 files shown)


ymichael:
Temporarily reverted the branch changes in b13269af8 so the PR branch has no net diff from main while testing scroll behavior. The original fix commits remain in history and can be restored by reverting this revert commit.

ymichael:
Updated the branch with a smaller candidate after baseline testing: the PR now keeps main's scroll-preservation capture semantics and only cancels a pending saved-row restore on explicit scroll-to-bottom. This is ready for another manual test pass.

ymichael:
Added the latched-detach candidate in a6c581203. This version should preserve bottom when streaming/layout drift leaves the scrollport temporarily off-bottom, but still preserve a row after actual user scroll, including fast switch-away before a scroll event. Ready for the next manual test pass.

ymichael:
Added temporary console logging in 8721f510d. Please filter DevTools console for `[bb-scroll-anchor]` while reproducing. Most useful events:

- `save-row` vs `save-bottom-geometry` vs `save-bottom-sticky-intent`
- `user-scroll-intent` and `scroll-user-detached-from-bottom`
- `scroll-off-bottom-without-user-intent`
- `mount-restore-row-anchor`, `restore-row-applied`, and `pending-row-restore-*`
- `scroll-to-bottom-requested/applied`

The payload includes `threadId`, scroll metrics, `shouldStickToBottom`, `userDetachedFromBottom`, and whether recent user intent was present.

ymichael:
The provided logs identified the likely culprit: while scrolling, we correctly emitted `save-row`, but the passive unmount cleanup later emitted `save-bottom-geometry` for the same thread with `shouldStickToBottom: false` and `wasUserDetachedFromBottom: true`. That means the cleanup write happened after teardown/clamping changed the captured scroll element geometry, overwriting the correct row anchor.

I pushed ce650e60b to move that final anchor flush into a layout-effect cleanup (`source: layout-cleanup`) and removed the passive cleanup write. Please test again with `[bb-scroll-anchor]` filtered; the key check is that after a manual scroll + switch away, the final save for that thread should be `save-row` from `layout-cleanup`, not `save-bottom-geometry`.

PR #368, opened 2026-06-25
[codex] Fix ACP composer bootstrap fallback

### LINKED ISSUE #365: Composer bootstrap should survive ACP model-list failures and use last execution model
## Summary

A bb thread can lose its model/agent display in the composer and appear impossible to follow up on when the provider model-list lookup fails, even though the thread has a valid historical execution model stored in `client/turn/requested` events.

Observed in bb `0.0.22` with an ACP opencode thread.

## User-facing symptom

In the thread UI, a completed/idle thread showed no model/agent selection and the composer could not be used normally for follow-up. The thread itself still had valid execution history and could accept a follow-up when a model was supplied explicitly.

Affected local thread:

- Thread: `thr_prdv686jqi`
- Title: `Docs Review For Staleness`
- Project: `proj_m5i9ek5e6a` (`omi-upstream`)
- Environment: `env_c2vvz77jh2` (`/Users/dazheng/workspace/omi-upstream`)
- Provider: `acp-opencode`

## Evidence

Thread row:

```sql
select id,title,status,project_id,environment_id,provider_id,model_override,reasoning_level_override
from threads
where id='thr_prdv686jqi';
```

Result:

```text
id              title                      status  project_id       environment_id  provider_id   model_override  reasoning_level_override
--------------  -------------------------  ------  ---------------  --------------  ------------  --------------  ------------------------
thr_prdv686jqi  Docs Review For Staleness  idle    proj_m5i9ek5e6a  env_c2vvz77jh2  acp-opencode
```

Project defaults had drifted to a different provider:

```sql
select * from project_execution_defaults where project_id='proj_m5i9ek5e6a';
```

```text
project_id       provider_id  model         service_tier  reasoning_level  permission_mode
---------------  -----------  ------------  ------------  ---------------  ---------------
proj_m5i9ek5e6a  acp-cursor   composer-2.5  default       medium           readonly
```

But the execution model was present in the thread event log:

```sql
select sequence,type,substr(data,1,1000) as data
from events
where thread_id='thr_prdv686jqi'
  and type in ('client/turn/requested','client/turn/start','client/thread/start')
order by sequence desc
limit 10;
```

Relevant rows included:

```json
{
  "direction":"outbound",
  "source":"tell",
  "initiator":"user",
  "request":{"method":"turn/start","params":{}},
  "input":[{"type":"text","text":"New review comments on the PR","mentions":[]}],
  "target":{"kind":"new-turn"},
  "execution":{
    "model":"zai-coding-plan/glm-5.2",
    "permissionMode":"workspace-write",
    "reasoningLevel":"medium",
    "serviceTier":"default",
    "source":"client/turn/requested"
  }
}
```

So `resolveExistingThreadExecutionPlan()` has enough durable state to derive the default model from `getLastExecutionOptions()`.

Logs showed the separate failure in provider model lookup:

```text
Failed to resolve provider models
providerId="acp-opencode"
message="Runtime shutting down"
```

Stack included:

```text
resolveSystemExecutionOptions
buildThreadComposerBootstrapResponse
```

The relevant code path in the installed bundle is:

```js
async function buildThreadComposerBootstrapResponse(deps, threadId) {
  const thread = requirePublicThread(deps.db, threadId);
  const defaultExecutionOptions = (await tryResolveExistingThreadExecutionPlan(deps, {
    executionSource: "client/turn/requested",
    input: {},
    threadId
  }))?.defaultView ?? null;
  const composerEnvironmentId = shouldResolveThreadComposerExecutionOptions({ thread }) ? thread.environmentId : null;
  const executionOptions = composerEnvironmentId ? await resolveSystemExecutionOptions(deps, {
    environmentId: composerEnvironmentId,
    providerId: thread.providerId
  }) : null;
  return { defaultExecutionOptions, queuedMessages, executionOptions, ... };
}
```

`defaultExecutionOptions` should be sufficient to show/use the previously selected model, but `executionOptions` resolution can fail because it depends on a live provider model-list call.

## Reproduction Shape

1. Create/use an ACP-backed thread, e.g. `acp-opencode`, with model `zai-coding-plan/glm-5.2`.
2. Let bb store the model in `client/turn/requested.execution` but not in `threads.model_override`.
3. Change project defaults to another provider, e.g. `acp-cursor` / `composer-2.5`.
4. Make `provider.list_models` for the original provider fail, e.g. host daemon returns `Runtime shutting down`.
5. Open the thread UI.

Expected:

- Composer bootstrap still returns `defaultExecutionOptions` from the last turn.
- UI shows the historical provider/model or a degraded but usable composer state.
- Follow-up remains possible using the last known execution model.
- The model picker can be disabled/degraded if live model options are unavailable.

Actual:

- UI loses model/agent display and appears not follow-up capable.
- Server logs show model-list failure bubbling through composer bootstrap.

## Additional Verification

A controlled explicit-model follow-up to the affected thread succeeded:

```bash
bb thread tell thr_prdv686jqi "bb diagnostic follow-up probe; please respond with 'probe ok' only" --mode queue --model zai-coding-plan/glm-5.2
```

It completed with an assistant response of `probe ok`. So the thread/provider session was still follow-up capable when given the historical model explicitly. The failure is in the composer/bootstrap/defaulting path, not inherent thread corruption.

## Proposed Fix

Make `buildThreadComposerBootstrapResponse` degrade gracefully when live provider option resolution fails.

Suggested behavior:

- Keep resolving `defaultExecutionOptions` via `tryResolveExistingThreadExecutionPlan()` first.
- Wrap `resolveSystemExecutionOptions()` in a non-fatal try/catch for host/provider/model-list failures.
- Return `executionOptions: null` or a degraded payload plus warning metadata when provider model listing is unavailable.
- Do not block composer rendering/follow-up if `defaultExecutionOptions` exists.
- Optionally persist the concrete model used at ACP thread creation or expose the last execution model directly in the thread summary/UI state, so ACP/project-default drift is less fragile.

Regression test idea:

- Thread has `provider_id='acp-opencode'`, `model_override=null`.
- Project defaults are for `acp-cursor`.
- Latest `client/turn/requested` has `execution.model='zai-coding-plan/glm-5.2'`.
- Stub `resolveSystemExecutionOptions` / provider model listing to throw `Runtime shutting down`.
- Assert composer bootstrap still returns `defaultExecutionOptions.model === 'zai-coding-plan/glm-5.2'` and does not 500 / disable follow-up.

## Related Smaller Smell

`acp-opencode` thread rename also logged as unsupported:

```text
Provider "acp-opencode" does not support thread rename.
Live thread rename command failed
```

That appears non-fatal and separate, but it is another ACP capability mismatch worth keeping non-fatal.

Git-on-my-level:
## Summary

Fixes composer bootstrap for ACP-backed threads when known-agent discovery is degraded. If `known_acp_agents.status` fails, the requested known ACP provider is now preserved in the provider list so existing ACP threads can keep resolving their provider/model state instead of falling back to another project default provider.

Closes #365.

## Root Cause

`resolveSystemExecutionOptions()` intentionally tolerated host/provider discovery failures by omitting known ACP agents. That is fine for generic provider lists, but it breaks thread-specific composer bootstrap when the thread provider itself is a known ACP provider such as `acp-opencode`. The requested provider disappeared from `executionOptions.providers`, so the composer could not reliably display or continue with the thread's historical execution model.

## Changes

- Include the requested known ACP provider in execution options even when the installed-known-agent probe is unavailable.
- Keep model loading degraded through the existing `modelLoadError` path.
- Add a regression test covering an `acp-opencode` thread with last-turn model state, stale project defaults for `acp-cursor`, and failing known-agent/model-list RPCs.

## Validation

- `pnpm exec turbo run test --filter=@bb/server -- --run test/public/public-thread-data.test.ts`
- `git diff --check`

### DIFF
--- a/apps/server/src/services/system/execution-options.ts
+++ b/apps/server/src/services/system/execution-options.ts
@@ -101,6 +101,22 @@ function listConfiguredSystemProviderInfos(
   return providers;
 }
 
+function includeRequestedKnownAcpProvider(
+  providers: ProviderInfo[],
+  providerId: string | undefined,
+): ProviderInfo[] {
+  if (
+    providerId === undefined ||
+    providers.some((provider) => provider.id === providerId)
+  ) {
+    return providers;
+  }
+  const knownAgent = findKnownAcpAgentForProviderId(providerId);
+  return knownAgent === undefined
+    ? providers
+    : [...providers, buildKnownAcpProviderInfo(knownAgent)];
+}
+
 function canOmitKnownAcpAgentsForError(error: unknown): error is ApiError {
   return (
     error instanceof ApiError && (error.status === 502 || error.status === 504)
@@ -319,6 +335,7 @@ export async function resolveSystemExecutionOptions(
     await earlyModelResultPromise?.catch(() => undefined);
     throw error;
   }
+  providers = includeRequestedKnownAcpProvider(providers, query.providerId);
   const requestedProvider = query.providerId
     ? providers.find((provider) => provider.id === query.providerId)
     : undefined;
--- a/apps/server/test/public/public-thread-data.test.ts
+++ b/apps/server/test/public/public-thread-data.test.ts
@@ -2562,6 +2562,96 @@ describe("public thread data routes", () => {
     });
   });
 
+  it("keeps ACP thread composer defaults when provider discovery is degraded", async () => {
+    await withTestHarness(async (harness) => {
+      const { host, session } = seedHostSession(harness.deps, {
+        id: "host-composer-acp-degraded",
+      });
+      seedPrimaryHost(harness.deps, host.id);
+      const providerResponder = registerHostRpcResponder(harness, {
+        hostId: host.id,
+        sessionId: session.id,
+        handle: (request) => {
+          if (
+            request.command.type === "known_acp_agents.status" ||
+            request.command.type === "provider.list_models"
+          ) {
+            return {
+              ok: false,
+              errorCode: "command_failed",
+              errorMessage: "Runtime shutting down",
+            };
+          }
+          throw new Error(`Unexpected RPC command ${request.command.type}`);
+        },
+      });
+      const { project } = seedProjectWithSource(harness.deps, {
+        hostId: host.id,
+      });
+      upsertProjectExecutionDefaults(harness.deps.db, {
+        projectId: project.id,
+        providerId: "acp-cursor",
+        model: "composer-2.5",
+        reasoningLevel: "medium",
+        permissionMode: "readonly",
+        serviceTier: "default",
+      });
+      const environment = seedEnvironment(harness.deps, {
+        hostId: host.id,
+        projectId: project.id,
+      });
+      const thread = seedThread(harness.deps, {
+        projectId: project.id,
+        environmentId: environment.id,
+        providerId: "acp-opencode",
+      });
+      seedThreadRuntimeState(harness.deps, {
+        environmentId: environment.id,
+        inputText: "New review comments on the PR",
+        model: "zai-coding-plan/glm-5.2",
+        permissionMode: "workspace-write",
+        providerThreadId: "ses_opencode",
+        reasoningLevel: "medium",
+        threadId: thread.id,
+      });
+
+      const response = await harness.app.request(
+        `/api/v1/threads/${thread.id}/composer-bootstrap`,
+      );
+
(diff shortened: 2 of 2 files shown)


SawyerHood:
I tested this manually and it does seem to fix it! Are you fine if I merge?

Git-on-my-level:
Yes go ahead it's ready

PR #485, opened 2026-07-03
Close side tabs before desktop window

Git-on-my-level:
## Summary
- route the desktop Cmd/Ctrl+W menu command through the renderer before closing the native window
- close the active secondary-panel file/browser/new/terminal/side-chat tab when one is open
- preserve native window close behavior when no renderer handler consumes the request

## Validation
- pnpm exec turbo run test --filter=@bb/desktop
- pnpm exec turbo run typecheck --filter=@bb/desktop --filter=@bb/app --force
- git diff --check


SawyerHood:
This lgtm! I ran into a few bugs and I also aligned the behavior to work like it does in Codex

PR #491, opened 2026-07-03
Expose plugin commands in prompt typeahead

brsbl:


SawyerHood:
Are you trying to support Codex plugins? We already slash command complete them if they have skills. 

A Codex plugin is just a collection of skills + MCPs so I think the generic way to fix this is to just `/command` complete MCPs

brsbl:
Closing per thread takeover — discarding this implementation and restarting with a spec-first pass on @-mentions for BB-installed and provider plugins (new thread thr_msje8pynyr).

PR #450, opened 2026-07-01
feat: Add search to model picker

barelyhuman:
Fixes: https://github.com/ymichael/bb/issues/496

https://github.com/user-attachments/assets/06cfa07e-79a0-4372-bebf-51c8d074c106

PR #500, opened 2026-07-04
Add oh-my-pi (omp) agent provider

Git-on-my-level:
## Summary

Adds **omp** (oh-my-pi) as a fourth built-in agent provider alongside codex, claude-code, and pi.

omp is a **divergent fork of vanilla pi** — different auth store (`~/.omp/agent/agent.db`, SQLite + multi-credential), Bun-first runtime with Rust native bindings, and its own RPC protocol. bb's existing `"pi"` provider bundles the pre-rename `@mariozechner/pi-coding-agent@0.70.5`, which cannot read omp's auth DB (verified: `AuthStorage.hasAuth()` returns false for every provider against a real omp install). Rather than migrate that dependency (broken at import time by omp's Bun-only APIs), this PR adds omp as a **separate provider** that drives the user-installed `omp` CLI over its RPC mode via a thin framing bridge — **no omp SDK is bundled, so there is no version drift**, and omp owns auth/models entirely.

image
image

## Changes

- **Provider registry** (`agent-providers`): `"omp"` entry, full-only perms, pi-mirroring caps; `AgentRuntimeOmpSkillRoot`.
- **Adapter** (`omp/adapter.ts`): bb JSON-RPC ↔ omp `AgentSessionEvent` translation — turn lifecycle + assistant text streaming. Unrecognized lifecycle/UI events are dropped silently (no timeline noise).
- **Bridge** (`omp/bridge/bridge.ts`): JSON-RPC 2.0 ↔ omp `{type:...}` RPC framing; spawns the user's `omp --mode rpc` (no shell); answers omp `extension_ui_request` frames so turns don't block; resets child state on exit so the bridge recovers after an omp crash; preserves text alongside attachment input; surfaces prompt/steer dispatch failures instead of hanging.
- **UI**: `OmpIcon` (pi glyph), provider-icon wiring, model picker **groups by provider** (`groupModelOptionsByProvider`) when a catalog spans 2+ `provider/` prefixes.
- **Host-daemon**: bridge bundle manifest entry.
- **Server/tests**: catalog-driven (no source changes); + tests mirroring pi.
- **Docs**: agent-runtime README, bb-guide template (+ regen), `configuration.md` (`BB_OMP_BINARY`).

## Verified

- All touched packages typecheck (agent-runtime, agent-providers, server, app, host-daemon, templates).
- 211 unit tests pass across the touched suites (no regressions); 10 omp adapter tests.
- Live end-to-end against `omp 16.2.2`: `model/list` returns the user's full catalog (157 models across cursor/openai-codex/xai-oauth/zai), and a full turn streams assistant text and completes cleanly (`turn/completed`, zero `provider/unhandled`).
- Subagent review completed; all three reported bridge defects (attachment input loss, post-crash bridge deadlock, swallowed dispatch errors) addressed.

## Test plan

- `bb provider models omp` → lists the user's omp models grouped by provider.
- `bb thread spawn --provider omp --model <provider/model> ...` → turn runs and completes.
- omp must be installed and authenticated (`omp` then `/login`); override binary via `BB_OMP_BINARY`.

## Known limitations (follow-ups)

- Permission modes: `full` only (omp has no bb readonly/workspace-write analog).
- Event coverage: core streaming + turn lifecycle; full parity with the pi adapter's translator (compaction/reasoning/token accounting, subagent materialization) is deferred.
- Host-tool / host-URI callbacks into omp are a future enhancement.

## Review update

- Fixed OMP resume session identity so bb persists the actual OMP session file as providerThreadId while keeping forwarded events routed to the bb thread id.
- Added a regression for the thread/resume command shape.

## Additional validation


SawyerHood:
I'm all for adding more providers! Our Pi provider is pretty out of date right now (I don't think anyone uses it). 

I think omp uses ACP so to simplify the surface area we could just support it via this. This is how our OpenCode integration works as well!

It isn't well documented now but you should actually be able to add omp to editor by just adding a line to the ~/.bb/config.json, ask your agent how to do this!

I think then we can make a config that auto detects if you have omp installed and auto configures it via acp (again look at how we handle open code)

SawyerHood (changes requested):
Try to use ACP as the connector please. Also in the future open an issue before opening a PR

SawyerHood:
Superseded by #515, which keeps the acp-omp known-agent approach and icon from this PR (thanks @Git-on-my-level — credited as co-author there) in a smaller change: the bridge now prefers session/set_config_option but keeps the session/set_model fallback (preserving opencode selection from #345 and the dynamic-acp integration fixture), and the large-catalog reasoning-discovery cliff omp exposed is fixed separately in the same PR. The model-picker grouping and eviction-race fixes were split out as follow-ups — grouping needs a redesign on top of the #450 nav-row abstraction, and the eviction fix wants to be based on threadOperationCounts to cover the full thread-start span.

PR #520, opened 2026-07-07
Add CSV file previews

georgecollier-nqu:
## Summary
- add CSV detection alongside the existing markdown/image/PDF preview routing
- render CSV text previews as a scrollable table with row numbers, sticky headers, fixed-width columns, and truncation for long cell values
- keep the existing Preview/Raw toggle behavior and copy labeling for CSV files
- add focused unit coverage and a Ladle story for the CSV preview path

CleanShot 2026-07-07 at 15 54 58@2x

### DIFF
--- a/apps/app/src/app.css
+++ b/apps/app/src/app.css
@@ -281,6 +281,41 @@
   }
 
   /*
+   * Opt out of the global thin scrollbar into always-visible classic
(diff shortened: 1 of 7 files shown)


SawyerHood:
LGTM! There were a few quirks:
- The left column would break out of the CSV viewer
- The horizontal scrollbar was very flaky on mac for me

I patched these up and will merge

PR #521, opened 2026-07-07
Add live voice waveform recording pill

harveyfullstack:
## What

Replaces the prompt box's cramped dual-button voice control (an `AudioLines` pulse button plus a tiny `X`) with a dedicated recording pill that shows a live audio waveform while you dictate. When recording starts, the editor collapses and the pill crossfades in; when it ends, the editor comes back. Escape now cancels an active recording, matching the pill's `X`.

- While recording or transcribing, the box morphs to `rounded-3xl` and the editor/header/footer region collapses (CSS grid `1fr` to `0fr` + opacity) as the pill row expands in, on a shared springy transition with `motion-reduce` guards.
- The pill has a cancel (`X`), the live waveform, and a confirm button that shows a check while recording and a spinner while transcribing (disabled during transcription), with `sr-only` `aria-live` status text.
- Escape cancels the recording from anywhere in the composer, in both the recording and transcribing states.

voice1
image

## Key pieces

- **`WaveformVisualizer.tsx`** (new) draws a scrolling bar waveform on a `<canvas>` from a `MediaStream`, via a WebAudio `AnalyserNode` and a `requestAnimationFrame` loop. It computes per-frame RMS (noise floor, gain, gamma curve), samples every other frame, and only redraws on sample frames. Size-derived geometry and 2D context state are set once in `measure()` (mount + `ResizeObserver`), since resizing the canvas resets context state. It respects `prefers-reduced-motion` via the shared `useMediaQuery` hook, falls back to a static idle waveform when it cannot animate (no `AudioContext`, no track, reduced motion, or transcribing), and cleans up fully on unmount. It clones the audio track for analysis because some mobile browsers feed silence to WebAudio when `MediaRecorder` owns the original track.
- **`VoiceRecordingBar.tsx`** (new) is the pill layout described above.
- **`PromptBoxInternal.tsx`** threads the `MediaStream` into `PromptVoiceConfig`, restructures the composer to crossfade the editor against the voice bar via collapsing grid rows (`COLLAPSING_GRID_CLASS`), and adds the Escape handler. The handler is a capture-phase `window` keydown listener with `stopPropagation()`, so it wins over the composer's own `useEscapeToHide` (which would otherwise hide the whole box when the prompt is empty) and never leaks to the collapsed editor.
- **`usePromptVoice.ts` / `useVoiceInput.ts`** expose the live `MediaStream` as state (set on `getUserMedia`, cleared on stop) so the visualizer can consume it.

The large `PromptBoxInternal.tsx` diff is mostly indentation churn: the existing editor/footer JSX was wrapped in the new collapsing-grid container, so git renders unchanged logic as a rewrite.

## Validation

- `pnpm exec turbo run typecheck --filter=@bb/app` clean.
- Promptbox voice tests green: 42 tests across `PromptBoxInternal.test.tsx` (incl. 3 new Escape-cancel tests proving the capture-phase handler fires, consumes the event, and wins over a stand-in composer dismiss), `VoiceRecordingBar.test.tsx` (new), and `FollowUpPromptBox.test.tsx`.
- `WaveformVisualizer` is exercised through typecheck and is mocked in the bar test (jsdom has no canvas 2D context); the RAF/WebAudio loop has no unit harness.

## Follow-up

- The crossfade hardcodes `duration-[180ms] ease-[cubic-bezier(0.16,1,0.3,1)]` in a few places. Those values already exist as shared tokens (`HEIGHT_TRANSITION_DURATION_MS` / `HEIGHT_TRANSITION_EASE_CSS` in `components/ui/height-transition.tsx`). Consolidating onto them is a reasonable follow-up, but it is not a drop-in: Tailwind JIT cannot interpolate a constant into an arbitrary class, so it means either moving to inline `style` (and handling reduced motion in JS) or adding a shared motion-timing class.


georgecollier-nqu:
Does it keep text visible if there's already text in the message composer? I think it should, because the behaviour is that it appends the text atm and if it hides the message composer I feel like it's gonna overwrite everything I've typed

harveyfullstack:
@georgecollier-nqu Yes, same underlying behavior as before! This is purely for slicker visuals that match more closely to Claude + ChatGPT.

SawyerHood:
10/10 no notes this looks great!

SawyerHood:
Well @georgecollier-nqu's point is fair. I guess we could make this take up the bottom part of the promptbox so you can see what  you already typed, but this is already a huge improvement over what we had

PR #595, opened 2026-07-10
Repair legacy Codex subagent timelines

brsbl:
## Fix legacy Codex subagent timelines

This repairs existing threads that were recorded before #589. Those timelines can show Codex native subagent work as unrelated root turns, and then route later child-scoped errors or interactions to the wrong turn.

It complements #589: that PR fixed forward ingestion, but intentionally did not rewrite persisted `provider/unhandled` rows. This PR makes those legacy rows project correctly without mutating stored history.

## Reproduce the original bug

Use a build immediately before #589 (`1912e204f`, the parent of `379a00d5c`), so new events still take the old ingestion path.

1. Create a normal Codex thread with these execution settings:

   | Setting | Value |
   | --- | --- |
   | Model | `gpt-5.6-sol` |
   | Reasoning | `xhigh` |
   | Permission | `full` |
   | Service tier | `default` |

2. Send this prompt (it must use native Codex subagents, not `bb thread spawn`):

   ```text
   Spawn two native subagents in parallel and keep the root turn active until both return.
   - renderer_fixes: inspect packages/thread-view and run one focused test.
   - storage_fixes: inspect packages/agent-runtime and run one focused test.
   Then summarize both results in the root turn.
   ```

3. In the persisted event stream, find Codex `provider/unhandled` rows whose `rawEvent.params.item` is `subAgentActivity` with `kind: "started"`. The following child `turn/started` rows have no `parentToolCallId`.

4. Re-open the thread timeline. The child work appears as root-level turns instead of nested delegation rows. While a child is active, a steer or other child-scoped event can make the active-turn resolver expect that child while the host still has the root turn active.

The model is not part of the parser contract; `gpt-5.6-sol` at `xhigh` is the incident configuration that rapidly produces two overlapping native subagents.

## Root cause

Before #589, Codex native `subAgentActivity` notifications were persisted as opaque `provider/unhandled` events. Their child turns share the parent provider thread and have no persisted `parentToolCallId`. The timeline projector therefore treats each child `turn/started` as a new root.

In the reported thread, two starts (`/root/renderer_fixes` and `/root/storage_fixes`) created two such orphan child turns. Later child-scoped `system/error` rows stayed top-level too, which made the rendered timeline and active-turn state disagree with the still-running root turn.

## Minimal fix

At replay time only, recognize the exact legacy Codex raw-event shape and synthesize a `spawnAgent` lifecycle in memory. When the matching legacy child turn starts, stamp its recovered `parentToolCallId`; all of its transcript and errors then flow through the existing delegation renderer. A terminal child turn completes or interrupts the synthesized lifecycle.

The turn-summary detail endpoint retains only the recoverable legacy child lineage long enough for the same projection repair to run. New events, stored event rows, server/daemon ownership, and database migrations are unchanged.

This fixes the bug because the projector once again has the parent edge it requires before grouping turns. Explicit recovered parent IDs also avoid the stale FIFO link that can otherwise misattach a child after another legacy subagent is interrupted.

## Verification

- Replayed the 2,058-event incident stream: two completed/interrupted delegation rows, no legacy child turn left at the top level.
- Added projection regressions for concurrent children, an interrupted first child, explicitly parented child turns, accepted human roots, and child-scoped errors.
- Added an API regression for expanded turn-summary details.
- `pnpm exec turbo run test --filter=@bb/thread-view --force` — 342 passed.
- `pnpm exec turbo run test --filter=@bb/server --force` — 986 passed.
- `pnpm exec turbo run typecheck --filter=@bb/thread-view --force`
- `pnpm exec turbo run typecheck --filter=@bb/server --force`

### DIFF
--- a/apps/server/src/services/threads/timeline.ts
+++ b/apps/server/src/services/threads/timeline.ts
@@ -256,6 +256,127 @@ function tryReadClientTurnRequestedRequestId(
   return event.requestId;
 }
 
+function collectAcceptedRootTurnIds(
+  rows: readonly StoredEventRow[],
+): ReadonlySet<string> {
+  const isRootTargetByRequestId = new Map<ClientTurnRequestId, boolean>();
+  for (const row of rows) {
+    if (row.type !== "client/turn/requested") {
+      continue;
+    }
+    const event = parseStoredEvent(row);
+    if (event.type !== "client/turn/requested") {
+      continue;
+    }
+    isRootTargetByRequestId.set(
+      event.requestId,
+      event.target.kind === "new-turn" || event.target.kind === "thread-start",
+    );
+  }
+
+  const turnIds = new Set<string>();
+  for (const row of rows) {
+    if (row.type !== "turn/input/accepted") {
+      continue;
+    }
+    const event = parseStoredEvent(row);
+    if (
+      event.type === "turn/input/accepted" &&
+      event.scope.kind === "turn" &&
+      isRootTargetByRequestId.get(event.clientRequestId)
+    ) {
+      turnIds.add(event.scope.turnId);
+    }
+  }
+  return turnIds;
+}
+
+function isLegacyCodexSubagentStarted(row: StoredEventRow): boolean {
+  if (
+    row.type !== "provider/unhandled" ||
+    row.scopeKind !== "turn" ||
+    row.turnId === null ||
+    row.providerThreadId === null
+  ) {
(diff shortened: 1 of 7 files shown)


brsbl:
@ymichael, when you get a chance, could you take a look? This repairs the pre-#589 persisted Codex subagent projection path and includes the exact rapid repro plus the incident replay evidence.

PR #737, opened 2026-07-16
Bundle official plugins into the app; retire the remote marketplace

SawyerHood:
## What

Official plugins (`github`, `docs`, `memory`, `tasks`) now build and bundle **into the packaged app** alongside the builtins. They show up in Settings → Plugins → Browse as a local store and install with one click from the bundled copy — no network, no GitHub Release downloads, no separate publish pipeline. Installed official plugins are pinned to the bundled copy and update by riding normal app releases.

## How

- **Registry** (`builtin-registry.ts`): new `OFFICIAL_PLUGINS` tier — bundled like builtins but `autoInstall: false` (store-only, never auto-installed). `copy-builtin-plugins.ts` now packages all ten plugins into `<server dist>/builtin-plugins/`; `tasks` joins the store for the first time.
- **Install identity**: a store install registers as `source: builtin:<name>` with `provenance: catalog` (+ `catalogEntryId`). Startup reconciliation re-points installed officials to the new bundled copy on app upgrades, skips ones the user never installed, and respects remove tombstones. `bb plugin install github|docs|memory|tasks` works by bare name.
- **Catalog service** is now a static in-process catalog over the bundled manifests (displayName/description/icon from each plugin's `bb` manifest, category from the registry). Install refuses server-side when the experiment is off or the entry is incompatible/unavailable.
- **Removed**: remote catalog fetch + 6h refresh + etag caching, the `plugin_catalog` table (migration `0074` drops it), `plugin-catalog.json`, `github-release-source.ts` and the tarball download/SHA-verify branches, `installFromCatalog` and its catalog-only `InstallContext` fields, the `plugin-catalog-changed` change kind, `POST /plugin-catalog/refresh` + the Browse Refresh button + `bb plugin catalog` command group, and `.github/workflows/publish-official-plugins.yml`.
- **Kept unchanged**: generic `path:` / `npm:` / `git:` direct installs and their update machinery.
- **Experiment gate** now keys off provenance instead of source kind, so store-installed officials stay behind the "Plugins" experiment (like catalog installs did) while auto builtins remain exempt; the install route's gate bypass matches.
- Docs/guide/skill surfaces and the plugin-SDK generated types regenerated; release runbook rewritten (officials now just ship with app releases).

## Deliberate non-migration

Existing installs from the old marketplace (npm rows with GitHub-Release registry URLs) are left as-is: they keep running from their cached artifact, their update checks report `unavailable`, and reconcile logs a skip instead of hijacking the id. One-off manual migration steps for the only live machine are in the plan (`bb plugin remove <id> && bb plugin install <entry>`, or a settings-preserving SQL re-point).

## Shared UI (follow-on commit)

With the officials bundled, they now take the same `@bb/shared-ui` workspace dependency the first-party `plugins/*` use instead of vendoring shadcn component copies: ~68 vendored files (~6.4k lines) deleted across github/docs/memory/tasks with no call-site behavior changes (all drift was import-path noise or stale shadcn generations with matching APIs). Dead cva/clsx/tailwind-merge/radix deps dropped; the plugin-registry + `npx shadcn add` path for out-of-repo authors is untouched. Also adds the missing `bb-app#build` → `@bb/plugin-sdk#build` turbo edge (the tasks app bundle value-imports the sdk root) and registers all bundled plugin dirs in `bb-app#build` inputs.

## Verification

- `turbo typecheck` + `turbo test` green across server, app, cli, sdk, server-contract, db, domain, plugin-sdk, templates (one pre-existing flaky connect host-management timeout passes in isolation).
- New tests: registry invariant (declared plugin id matches each bundled manifest), official-plugin lifecycle (no auto-install, catalog-provenance install, re-point on version bump, tombstone across restarts, experiment gating), static catalog service, route behavior (refresh 404s).
- Packaged smoke: built `bb-app`, booted the packaged server on a scratch data dir — builtins auto-load, officials don't; Browse search lists all four from bundled manifests; one-click install of Docs runs from the bundled copy (`provenance: catalog`, `source: builtin:docs`); remove → installable again → reinstall; zero requests to github in server logs.


SawyerHood:
**Dual model review completed** (one Claude Fable 5 thread, one Codex gpt-5.6-sol thread on high reasoning, both reviewing the full branch):

- **Fixed from the GPT-5.6 review**: bundled plugins no longer show the "Updates & source" card / check-now flow (they're pinned to the app bundle); stale `tasks@bb-official` install spells, `catalog.refresh` in the SDK reference, and the experiment-exemption wording were scrubbed.
- **Fixed from the Fable review**: a legacy retired-marketplace row no longer breaks `checkForUpdates` for *every* plugin — it now degrades to a per-row `unavailable` result with a "reinstall from Browse" hint (regression test added); GitHub-Release claims removed from the plugin-authoring skill and the release-process pointer; tasks' dead radix deps dropped and shared-ui peer declarations aligned with `plugins/*`.
- **Flagged for sign-off, intentionally kept** (Fable finding 4): the shared-ui migration converges a few cosmetic divergences onto the canonical components — the tasks dialog/drawer scrim goes from `bg-foreground/40` to the standard `bg-black/40`, and the docs New-folder/Add-vault dialogs become bottom drawers on compact viewports, matching every other first-party plugin.

Both reviewers verified the reconcile/tombstone lifecycle, provenance-based experiment gating, contract/SDK/CLI consistency, and the drizzle migration as sound.

PR #808, opened 2026-07-19
fix(sdk): accept legacy workspace-write spawn input

brsbl:
## Summary

- accept the deprecated `workspace-write` alias only at the SDK spawn input boundary
- forward that alias unchanged for server-side normalization while retaining the narrow stored/output permission type
- cover transport, contract normalization, public type exclusion of `readonly`, and generated plugin/template declarations

## Validation

- `pnpm exec turbo run test --filter=@bb/sdk --filter=@bb/server-contract --force`
- `pnpm exec turbo run typecheck --filter=@bb/sdk --filter=@bb/server-contract --filter=@bb/plugin-sdk --filter=@bb/templates --force`
- `pnpm exec turbo run build --filter=@bb/sdk --filter=@bb/server-contract --filter=@bb/plugin-sdk --filter=@bb/templates --force`
- bundled/template declaration generation checks
- browser QA: Prompt Shaper helper spawned with `workspace-write`, completed normally, and returned the enhanced draft without HTTP 400


SawyerHood:
I'm going to close this out! We do not want to do this, the sdk is pre-1.0 so I'm not adding a bunch of backwards compat until plugins are out

PR #843, opened 2026-07-22
Classify workflow notifications as system messages

brsbl:
Workflow completion notifications are now trusted system messages, while the thread table of contents remains limited to human and agent messages.

## What changed

- Added the structured `workflow-finished` system-message kind.
- Added a narrow experimental backend plugin API for server-stamped workflow notifications.
- Preserved system provenance while notifications wait in an active thread's queue; editing a queued notification clears that provenance.
- Rendered workflow-specific chrome only when the stored initiator and kind identify a genuine workflow notification.
- Built the conversation outline from human and agent provenance, naturally omitting system rows from the ToC.

## Why

The Workflows plugin previously delivered completion notices through the ordinary thread-send API. That made them indistinguishable from user-authored text, so trusted-looking workflow chrome would have needed to rely on a spoofable text prefix. Those rows could also enter the conversation outline as user messages.

This change moves trust to structured server metadata. A user can still type text that resembles a workflow notice, but it remains an ordinary user message and is rendered and indexed accordingly.

## Validation

- Typecheck: 7 packages passed.
- App: 1,899 tests passed.
- Server: 98 relevant tests passed.
- Database: 345 tests passed.
- Workflows: 213 tests passed.
- Plugin SDK: 74 tests passed.
- Thread view: 345 tests passed.
- Lint: zero errors.


SawyerHood:
Superseded by #844, which implements workflow completion delivery as agent-only context and steers active origin threads. #844 has now been merged, so I’m closing this PR without merging it.

PR #846, opened 2026-07-22
Add plugin timeline annotation host contracts

brsbl:
## Summary

Add the narrow BB host and plugin SDK contracts required by selection-bound timeline annotations:

- let message actions opt into the floating selection menu without appearing in the per-message action bar
- capture rendered-text UTF-16 selectors, quote context, and line-fragment geometry from user and assistant prose
- expose experimental native timeline DOM hooks without enabling annotation inside embedded plugin chats
- give lifecycle-managed content scripts experimental plugin-scoped RPC, realtime, connection-state, and composer navigation capabilities
- let thread-panel plugins experimentally reveal a native message, including bounded older-history loading, before resolving its exact text range
- update the SDK harness, generated declarations, templates, plugin-authoring documentation, and API audit ledger

Every new public member uses an `experimental_` prefix. Native DOM hooks use the equivalent `data-bb-experimental-*` namespace.

## Why

This unblocks the separate [Timeline Comments plugin PR](https://github.com/brsbl/bb-plugins/pull/8) without moving plugin-owned annotation data or UI into BB core. BB supplies only the selection, lifecycle transport, DOM, and reveal primitives that an external plugin cannot infer safely.

## Validation

- 105 focused app tests
- 75 plugin SDK tests
- 14 plugin-authoring documentation tests
- 13 template and external-scaffold tests
- scoped app/server/SDK/templates typecheck passed
- app lint passed with zero errors and the repository's existing warnings only
- generated SDK declarations and templates are current
- `git diff --check`


brsbl:
Closing as stale. The branch is retained, so this can be reopened if the timeline annotation contracts are still wanted.

PR #847, opened 2026-07-22
Refine thread title generation prompt

georgecollier-nqu:
Titles are kind of bad at the moment, because they do not consider the user intent.

For a prompt like:

'Using Axiom CLI, debug Frank's issue on prod', it will tend to generate a title like:

'Use Axiom CLI On Prod'

When the title really ought to be:

'Debug Frank's issue on prod'

I'd also updated from title case to sentence case; I think it is way more readable but feel free to change that if you disagree.


SawyerHood:
Seems like a positive ship to me!

PR #861, opened 2026-07-22
Retry timed-out provider initialization once

brsbl:
## Summary

- classify JSON-RPC request timeouts with a typed error
- retry provider startup once, with a fresh process and adapter, only when `initialize` times out
- suppress the retry during shutdown and cover the recovery path with a lifecycle regression test

## Why

A cold Codex app-server process can occasionally miss the 30-second `initialize` deadline. The PartySocket upgrade protects the server-to-daemon transport, but this failure is inside the daemon-to-provider JSON-RPC startup path.

## Verification

- `pnpm exec turbo run test --filter=@bb/agent-runtime --force` (795 tests)
- independent review-loop re-review: CORRECT

### DIFF
--- a/packages/agent-runtime/src/runtime-json-rpc.ts
+++ b/packages/agent-runtime/src/runtime-json-rpc.ts
@@ -48,6 +48,16 @@ export class ProviderResponseEncodeError extends Error {
   }
 }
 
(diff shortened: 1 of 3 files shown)


brsbl:
Closing as stale. The branch is retained, so this can be reopened if the retry work is still wanted.

PR #865, opened 2026-07-24
Let New project select an execution machine

amadad:
## Summary

- let **New project** choose among enrolled machines when more than one is available
- browse and create the local-path source on the selected host, while disabling offline machines
- preserve the existing native folder-picker flow for single-machine installs and existing add/update-source behavior
- document the machine-first project creation path

## Why

The project-create contract and remote folder browser already support a host ID, but the New project UI always targeted the primary host. Multi-machine users therefore had to leave the UI and use `bb project create --machine ...`.

This keeps the host decision in the project path dialog and passes the selected host through the existing create contract. It does not change the server, database, HTTP API, or host-daemon protocol.

## Verification

- `pnpm exec turbo run test --filter=@bb/app` — 285 files, 2,083 tests passed
- `pnpm exec turbo run lint --filter=@bb/app` — 0 errors
- manual browser QA with two connected machines and one offline machine: switched machines, browsed the selected host, created a project, and confirmed its source stored the selected `hostId`; the offline machine remained disabled

### DIFF
--- a/apps/app/src/components/dialogs/ProjectPathDialog.test.tsx
+++ b/apps/app/src/components/dialogs/ProjectPathDialog.test.tsx
@@ -1,0 +1,119 @@
+// @vitest-environment jsdom
+
+import { cleanup, fireEvent, render, screen } from "@testing-library/react";
+import type { Host } from "@bb/domain";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { ProjectPathDialog } from "./ProjectPathDialog";
+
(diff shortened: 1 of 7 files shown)


SawyerHood:
Superseded by #875.

Thanks for this — the change is real and it's in. I reviewed it against a local checkout and found that the single-machine gate counted every enrollment rather than the connected ones, so one long-offline machine silently replaced the native folder picker, so rather than round-trip here I folded the fix into #875 alongside the other three open contributor PRs (#865, #868, #869, #870).

PR #868, opened 2026-07-25
Fix Docs YAML frontmatter parsing

amadad:
## Summary

- parse fenced YAML frontmatter through one shared Docs boundary
- use a string frontmatter `title` (or the document H1) for navigation and mentions
- keep frontmatter out of the rich editor and previews while preserving it byte-for-byte on save
- keep frontmatter-managed filenames stable instead of auto-renaming them

## Problem

Docs previously derived a note title from the first raw Markdown line and passed the whole file to Tiptap. Wiki pages therefore appeared as repeated `---` entries, rendered metadata as document content, and risked rewriting frontmatter after an edit.

## Verification

- `pnpm exec turbo run test --filter=bb-plugin-simple-notes --force` — 46 passed
- `pnpm exec turbo run typecheck --filter=bb-plugin-simple-notes --force` — passed
- `pnpm exec turbo run build --filter=bb-app --force` — passed
- `pnpm exec turbo run smoke:tarball --filter=bb-app --force` — passed
- exercised against a 433-document Markdown vault — zero delimiter-derived titles

### DIFF
--- a/official-plugins/docs/README.md
+++ b/official-plugins/docs/README.md
@@ -27,6 +27,10 @@ provider, and directive are all Docs.
   `.markdown` files, so it can be selected under Settings → File openers or
   chosen from a file link's Open with menu. Workspace and absolute host files
   retain compare-and-swap saves even when they are outside a Docs vault.
+- **YAML frontmatter:** fenced frontmatter supplies the document title when it
+  has a string `title`, stays out of the rendered body and search preview, and
+  is preserved byte-for-byte when the rich editor saves body changes. Docs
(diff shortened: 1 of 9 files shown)


SawyerHood:
Superseded by #875.

Thanks for this — the change is real and it's in. I reviewed it against a local checkout and found four issues — the biggest being that any text between two `---` lines was treated as frontmatter, so a document opening with a thematic break had its first section hidden from the editor, so rather than round-trip here I folded the fix into #875 alongside the other three open contributor PRs (#865, #868, #869, #870).

PR #869, opened 2026-07-25
fix(app): stop remark-directive from eating colons in assistant messages

bhuvansingla:
## Problem

remark-directive (enabled for plugin directives like `::inline-vis{...}`) also parses inline `:name` and block `:::name` directives. So any colon before a letter or digit — `9:30`, `key:value`, `:D` — became a directive node.

remarkMessageDirectives only handled leaf (`::name`) directives. Text and container directives fell through to mdast-util-to-hast, which renders an unknown directive as an empty `<div>`. Inside a paragraph, that block element split the line and dropped the directive's text:

    source:   at 9:30 today

    before:   at 9
              ⏎  today      ← :30 gone, paragraph broken

    after:    at 9:30 today

## Fix

Handle all three directive kinds. Leaf directives mount as before; text and container directives are rewritten back to their literal source (text inline, container as a paragraph).

Display-only change — the persisted message text was already correct.

## Tests

- Sentence with `9:30`, `a:b`, `:D` renders verbatim in a single `<p>`, no injected block element (fails before the fix, passes after).
- `:::note` container renders as literal source text.
- `reconstructDirectiveSource` uses the right marker per kind (`:` / `:::`).

### DIFF
--- a/apps/app/src/components/ui/markdown-message-directives.test.ts
+++ b/apps/app/src/components/ui/markdown-message-directives.test.ts
@@ -81,4 +81,9 @@ describe("normalizeDirectiveAttributes / reconstructDirectiveSource", () => {
       reconstructDirectiveSource("inline-vis", { file: 'demo"x".html' }),
     ).toBe('::inline-vis{file="demo\\"x\\".html"}');
   });
+
+  it("uses the supplied marker for non-leaf directive kinds", () => {
+    expect(reconstructDirectiveSource("30", {}, ":")).toBe(":30");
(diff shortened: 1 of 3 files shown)


SawyerHood:
Superseded by #875.

Thanks for this — the change is real and it's in. I reviewed it against a local checkout and found two stale test assertions and an out-of-scope container-directive rewrite that stopped nested plugin directives from mounting, so rather than round-trip here I folded the fix into #875 alongside the other three open contributor PRs (#865, #868, #869, #870).

PR #870, opened 2026-07-25
Demote skill-tier override dedup logs to debug

amadad:
Every catalog resolution (each thread turn and project-page load) logs one line per deduped skill; with a 71-skill inherited root that is ~30K journald lines/day of expected-precedence noise.

Measured on a deployed instance: skill-scan log volume fell from ~1,300 lines/hour to ~9 lines/hour after deploying this change.

### DIFF
--- a/apps/server/src/services/skills/injected-skills.ts
+++ b/apps/server/src/services/skills/injected-skills.ts
@@ -581,7 +581,7 @@ function excludeOverriddenBuiltins(
     if (!userClaimedNames.has(source.name)) {
       return true;
     }
-    logger.info(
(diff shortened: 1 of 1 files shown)


SawyerHood:
Superseded by #875.

Thanks for this — the change is real and it's in. I reviewed it against a local checkout and found three tests left red — the capturing logger in the injected-skills test stubs `debug` out, so the assertions were looking at an empty `infos` array, so rather than round-trip here I folded the fix into #875 alongside the other three open contributor PRs (#865, #868, #869, #870).

PR #876, opened 2026-07-28
Keep the mobile submit tap ahead of keyboard dismissal

amadad:
## Summary

Prevent the default focus transfer on the submit button's `pointerdown` so the prompt editor keeps focus through a tap. The click still owns the commit.

## Why

On iOS, focus transfer happens before `click`. Moving focus from the editor to the submit button starts keyboard dismissal, which resizes the app shell while the tap is still resolving — so the submit races a viewport resize and can be lost. Keeping the editor focused for the duration of the tap removes the race; genuine focus changes elsewhere still dismiss the keyboard normally.

## Scope

The handler is deliberately narrow — it only calls `preventDefault()` when all of the following hold:

- the pointer is the primary button (`event.button === 0`)
- the editor exists and is not destroyed
- the editor is already focused

Keyboard activation is unaffected: `Enter`/`Space` on the button dispatch `click` without `pointerdown`, so they never reach this path.

## Verification

- `apps/app` — `PromptBoxInternal.test.tsx`, 67/67 passing, including a new regression test asserting `pointerdown` is default-prevented, `document.activeElement` is still the editor, and `onSubmit` still fires once on the subsequent click
- `pnpm exec turbo run typecheck --filter=@bb/app` clean

### DIFF
--- a/apps/app/src/components/promptbox/PromptBoxInternal.test.tsx
+++ b/apps/app/src/components/promptbox/PromptBoxInternal.test.tsx
@@ -1601,6 +1601,34 @@ describe("PromptBoxInternal compact layout", () => {
     ).toBe(true);
   });
 
+  it("keeps the editor focused through a pointer submit", async () => {
+    const onSubmit = vi.fn();
+    render(
+      <PromptBoxInternal
+        {...createPromptBoxProps({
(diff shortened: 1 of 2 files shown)


SawyerHood:
Great catch! TY

PR #914, opened 2026-07-30
feat(plugins): add native tool activity labels

### LINKED ISSUE #912: Plugin SDK: native activity labels for agent tools
## Problem

Plugin tools currently render generic timeline text such as `Ran tool repository_context`. A plugin can only customize that with a `bb.app` frontend, which is unavailable to managed Git installs when a frontend build cannot run.

## Proposal

Add an experimental `experimental_activity` field to `bb.agents.registerTool()`. It supplies native `running` and `completed` labels. BB renders these labels in its existing tool timeline; registrations that omit the field retain the current generic fallback.

## Scope

- SDK contract, server contribution plumbing, and timeline rendering
- Validation and regression tests
- API-audit entry while the field remains experimental

This is a focused prototype contribution; no plugin frontend or tool-execution behavior changes.

mywwave:
Closes #912

## Summary
- add `experimental_activity: { running, completed }` to `bb.agents.registerTool()`
- snapshot labels server-side into native plugin tool-call events; no plugin frontend or daemon protocol change is required
- render labels only while pending and after successful completion, preserving BB's normal identity/error/approval fallback for all other states
- update portable SDK declarations, fake-host parity, the plugin-authoring skill, and the experimental API audit

## Validation
- `pnpm --filter @bb/plugin-sdk test -- src/__tests__/public-types.test.ts src/testing/__tests__/fake-plugin-host.test.ts`
- `pnpm --filter @bb/server test -- test/services/plugins/plugin-agent-tools.test.ts`
- `pnpm --filter @bb/server test -- test/internal/internal-events-tool-calls.test.ts`
- `pnpm --filter @bb/thread-view test -- test/build-thread-timeline.test.ts test/timeline-row-title.test.ts test/tool-activity-projection.test.ts`
- `pnpm exec turbo run typecheck --filter=@bb/plugin-sdk --filter=@bb/thread-view --filter=@bb/server --filter=@bb/templates`


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I started this review under the `review` rule.

I will review security, code quality, performance, architecture, and safe test evidence. I will post an update after each phase.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Performance review is complete.

It found two medium risks. Event ingestion rebuilds and scans the plugin tool list for each eligible lifecycle event. Labels also have no size limit, although the server stores both labels in two events for every tool call.

I will include exact lines and fixes in the final review.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Security review is complete.

It found no direct security defect. Plugins already run as trusted server code, and the UI renders these labels as plain text.

The review found one defense-in-depth concern. The server accepts an existing daemon-supplied `activity` value, although the change defines this value as server-owned metadata.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

End-to-end test review is complete.

I did not start a server from this fork because the contributor code is untrusted. GitHub also reports `action_required` for the CI and Version Lockstep workflows, so no check results exist.

I reviewed the event, contract, projection, and title paths by source inspection. The final review will state this test limit.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Code quality and architecture review is complete.

It found a blocking contract issue. The new field enters the daemon ingress schema, although the server claims ownership of it. This also changes a daemon wire contract without a protocol update.

It also found unbounded, non-truncated labels. The same activity shape appears in several package contracts and view interfaces.

apps/server/src/services/plugins/plugin-api.ts
```diff
@@ -1022,6 +1027,21 @@ export function createPluginApi(options: {
...
         );
       }
+      const experimentalActivity = tool.experimental_activity;
+      if (experimentalActivity !== undefined) {
+        if (
+          typeof experimentalActivity !== "object" ||
+          experimentalActivity === null ||
+          typeof experimentalActivity.running !== "string" ||
+          typeof experimentalActivity.completed !== "string" ||
+          experimentalActivity.running.trim().length === 0 ||
```
SawyerHood: 🚨 `slopcop/review` — Bound and truncate activity labels

These checks reject empty labels but accept labels of any size. The server copies both labels into two stored events for every call.

The title path also renders this segment without truncation. A large label can increase storage, transport, parse cost, and timeline width.

Add a small documented maximum. Mirror it in the fake host and schemas. Also set `truncate: true` on the title segment.

packages/domain/src/provider-event.ts
```diff
@@ -311,6 +311,10 @@ export const threadEventItemSchema = z.discriminatedUnion("type", [
     server: z.string().optional(),
     tool: z.string(),
     arguments: z.record(z.string(), z.unknown()).optional(),
+    /** Server-enriched labels for a native plugin tool's timeline row. */
+    activity: z
```
SawyerHood: 🚨 `slopcop/review` — Keep server-owned labels out of daemon input

`threadEventSchema` also defines `hostDaemonWireEventSchema`. This field therefore lets daemon input set timeline labels before server enrichment.

The early returns in `withPluginToolActivity` keep those values for unknown, MCP, and unlabeled tools.

Use a separate stored-event schema, or reject and remove `activity` at ingress. If this wire field stays, increment the protocol version.

apps/server/src/internal/events.ts
```diff
@@ -275,6 +276,39 @@ function toStoredEvent(args: ToStoredEventArgs): AppendDaemonEventInput {
...
+): HostDaemonEventEnvelope {
+  const event = envelope.event;
+  if (
+    (event.type !== "item/started" && event.type !== "item/completed") ||
+    event.item.type !== "toolCall" ||
+    event.item.server !== undefined
+  ) {
+    return envelope;
+  }
+  const activity = findPluginAgentTool(event.item.tool)?.record
```
SawyerHood: 🚨 `slopcop/review` — Avoid a full plugin scan for each event

`findAgentTool` calls `collectAgentTools`. That function sorts loaded plugins and allocates a set and an array before its linear search.

This path now runs for each eligible start and completion event, including native tools without labels. Batch cost grows with events and plugins.

Keep a name-indexed map current across plugin changes. A smaller fix can build one activity map for each event batch.

SawyerHood (changes requested):
🚨 **SLOP COP** 🚨 · `review`

Review result: changes requested.

The blocking issue is contract ownership. `activity` now enters `threadEventSchema`, which also validates daemon wire events. Daemon input can therefore set server-owned timeline labels.

Use a separate schema for enriched stored events. Another option is to reject and remove `activity` at daemon ingress. Increment the daemon protocol if this field stays in the wire contract.

Two medium issues also need work:

- Event ingestion rebuilds and scans the full plugin tool list for each eligible lifecycle event.
- Activity labels have no size limit and render without title truncation.

The repeated activity shape is a real cross-package contract. Export one schema and type where package boundaries permit reuse.

The security review found no direct exploit. Plugins already run as trusted server code, and the UI renders labels as text.

I did not run code from this untrusted fork. GitHub reports `action_required` for both workflows, so no CI evidence exists.

Add tests for daemon-supplied labels, MCP tools, unknown tools, label limits, and approval or failure fallback behavior.

PR #916, opened 2026-07-31
Add seeding props to experimental_NewThreadComposer

SawyerHood:
## The bug

`experimental_NewThreadComposer` exposed only `defaultProjectId` and `initialPrompt` as seeds. A plugin that stores the submitted `NewThreadRequest` and re-opens it for editing (e.g. bb-plugin-slopcop's `RuleEditor`, where the composer's submit IS the save) got a composer reset to project defaults on every mount — so re-saving after an unrelated edit silently overwrote the stored provider, model, permission mode, and environment. No warning, no diff.

## The fix

Additive, optional seed props on `NewThreadComposerProps`:

- `defaultProviderId`, `defaultModel`, `defaultReasoningLevel`, `defaultServiceTier`, `defaultPermissionMode` — fed into `useThreadCreationOptions` as initial values, taking precedence over the project's remembered defaults and falling back to today's behavior when omitted.
- `defaultEnvironment` — a new pure mapping (`new-thread-environment-seed.ts`) inverts `resolveRootComposeThreadEnvironment`: args → encoded environment-picker selection + branch pick, including the managed-worktree base branch and unmanaged existing/new branch specs. Unrepresentable variants are documented on the prop (`project-default` seeds nothing; `personal` without `hostId` and an `unmanaged` non-null `path` cannot be encoded).

Semantics (documented in the prop doc comments): they are seeds, not controlled values; the props are value-compared each render, and any change re-seeds every selection — including user-touched ones — via the creation-options `resetKey`, following the established `defaultProjectId` re-seed-during-render pattern. So switching between two saved records in one mounted composer reloads that record's values.

Supporting fix in `useThreadCreationOptions`: a `resetKey` change cleared touched fields in a layout effect *after* the render's `executionInputSources` had been computed, so the reset render reported stale `explicit` provenance. The provenance memo now treats a pending reset as "nothing touched", mirroring the render-time rule `renderedThreadSelections` already applies.

`RootComposeView` is untouched. The SDK testing harness's fake composer now echoes the seeds back on submit so plugin tests can cover store-then-restore.

## Round-trip test

The property that prevents the data loss: feeding a submitted request back in as the `default*` props and submitting untouched must reproduce an equivalent request.

- `new-thread-environment-seed.test.ts` — pure round trip of every environment variant (reuse, managed worktree named/default base, unmanaged existing/new/no branch) plus the documented unrepresentable cases.
- `PluginNewThreadComposer.test.tsx` — renders the real composer (prompt box + query hooks mocked) and asserts: (1) a seeded, untouched submit `toEqual`s the stored request; (2) after a user touches the model, swapping in a second record's seeds re-seeds everything and submits that record exactly; (3) with no seed props, behavior is unchanged (project defaults).

## Also

- Bundled type declarations regenerated via the repo scripts (`build-bundled-dts.mjs`, `generate-templates.mjs`).
- bb-plugin-authoring skill: new props + store-then-restore pattern documented.
- `docs/api_to_audit.md`: audit entry #6 for the seeding props.

`turbo lint`/`typecheck` clean for app, plugin-sdk, server, templates; full `@bb/app` (2197 tests) and `@bb/plugin-sdk` suites pass.


SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I will review this pull request for security, code quality, and performance. I will also test the user flow.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Security review complete. I found no injection path, secret exposure, authorization bypass, or unsafe trust change.

The composer checks host and reuse seeds against current project data. The existing submission resolver still controls the final environment request.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Performance review complete. I found no material performance issue.

The new render work processes six small seed values and one fixed-shape environment value. It does not add a large loop or repeated request.

apps/app/src/components/plugin/PluginNewThreadComposer.tsx
```diff
@@ -106,6 +113,32 @@ export function PluginNewThreadComposer({
     setSeededDefaultProjectId(defaultProjectId);
     setPickedProjectId(defaultProjectId ?? null);
   }
+
+  // --- Execution/environment seeds ---------------------------------------
+  // The `default*` props are seeds, value-compared so a plugin can pass a
+  // fresh `defaultEnvironment` object each render. A signature change means
+  // the caller switched records: it re-seeds every selection (via the
+  // creation-options resetKey below), including ones the user had touched.
+  const seedSignature = JSON.stringify([
```
SawyerHood, line 126: 🚨 `slopcop/review` — Include the project seed in the branch reset identity

This signature does not include `defaultProjectId`. A project change can reset selections but keep `branchSeedOverridden` from the prior record.

The second record then loses its branch seed when all other seed values match. Please include the project seed in this identity.

apps/app/src/components/plugin/PluginNewThreadComposer.tsx
```diff
@@ -192,17 +225,23 @@ export function PluginNewThreadComposer({
...
     preferenceProjectId: projectId,
-    resetKey: projectId,
+    resetKey: `${projectId} ${seedSignature}`,
     resolveProviderRouting,
-    initialProviderId: projectDefaults?.providerId,
-    initialModel: projectDefaults?.model,
-    initialServiceTier: projectDefaults?.serviceTier,
-    initialReasoningLevel: projectDefaults?.reasoningLevel,
-    initialPermissionMode: projectDefaults?.permissionMode,
+    initialProviderId: defaultProviderId ?? projectDefaults?.providerId,
```
SawyerHood, line 242: 🚨 `slopcop/review` — Preserve the execution source data for seeded values

This prop overrides project defaults in the UI. However, the hook treats the value as an untouched default and returns empty `executionInputSources`.

The server ignores `providerId` and `model` when their source keys are absent. It can replace these visible seeds with project defaults.

Please accept and restore the saved source map, or mark each provided seed as explicit. Add a round-trip test with explicit sources.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Code quality review complete. I found one blocking correctness issue and two branch seed defects.

The blocking issue loses execution source data. The server can then replace visible provider and model seeds with project defaults.

I added three inline comments with the exact failure paths and test requests.

apps/app/src/components/plugin/PluginNewThreadComposer.tsx
```diff
@@ -271,15 +310,44 @@ export function PluginNewThreadComposer({
...
     environmentValue: effectiveEnvironmentValue,
     projectId,
   });
+  // The seeded branch backs the picker only while the user has not picked or
+  // cleared one AND the environment still matches the seed — a manual pick,
+  // clear, or environment change makes the composer behave exactly as if the
+  // seed were absent.
+  const selectedBranch =
+    pickedBranch ??
+    (!branchSeedOverridden &&
```
SawyerHood, line 377: 🚨 `slopcop/review` — Stop the old branch seed after an environment change

This callback changes the environment without setting `branchSeedOverridden`. If the user changes away and returns, the old branch seed becomes active again.

Please wrap environment and reuse changes to clear the seed. Add a test for the away-and-return path.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

End-to-end smoke test complete on commit `e6236780`.

I installed the Cascade example in a fresh development instance. I opened its host composer, entered a prompt, and submitted it.

The composer created a second thread. The route showed the new thread and no alert.

This test covers the basic route. It does not cover the blocking execution source round trip.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Request changes.

The seed props discard the saved `executionInputSources`. The server can then replace visible provider and model seeds with project defaults.

I also found two branch seed defects:

- A project change can keep the prior branch override state.
- An environment change can restore an old branch seed after the user returns.

Security and performance reviews found no issues.

The 12 focused tests pass. The app and plugin SDK typechecks pass. The live Cascade composer smoke test also passes.

The current tests use empty source data. Add explicit source data to the round-trip test and fix all three inline findings.

SawyerHood:
Addressed all three findings. Each fix has a test I verified fails without it.

**1. Seeded values discarded by the server (blocking).** Confirmed against `resolveRequestedCreateExecutionValue` in `apps/server/src/services/threads/project-execution-defaults.ts`: when `sources` is present but a field's key is absent, the requested value is dropped and re-derived from the project's stored defaults. The composer always sent `executionInputSources` (possibly `{}`), so seeded `providerId`/`model` were silently replaced — exactly the data loss this PR exists to fix. Worse, `buildExecutionInputSources` never emits a `providerId` source for `component-local` scope at all (the standing gap in audit item #2), so the composer's provider could never survive `spawn`.

Fix: each seeded field now reports `"explicit"` provenance in the submitted request. `"explicit"` is correct whether or not the user then changed the value — either way it is a concrete choice the composer displayed, not a default the server should re-derive. Unseeded fields keep the hook's own provenance, so a caller passing no seed props sends exactly what it sent before. The round-trip test now asserts the full source map rather than `{}`.

**2. Branch reset identity missing the project.** Fixed — `defaultProjectId` is now part of the seed signature. New test: clear the branch on record one, switch to a record differing only by project, and the second record keeps its own seeded base branch.

**3. Stale branch seed after an environment round trip.** Fixed — user environment changes (including the worktree picker, which routes through the same setter) retire the seed permanently. New test covers away-and-return: back in worktree mode the base branch falls to `{ kind: "default" }` instead of the retired `release` seed.

One deliberate exception: the host-driven "managed worktree unavailable → fall back to local" effect keeps the raw setter. It is a system correction, not a user pick, so it should not retire the seed.

Also updated the `defaultProviderId` doc comment to state the provenance guarantee, since it is load-bearing for anyone relying on the round trip.

Full suites green: 2199 `@bb/app` tests, 75 `@bb/plugin-sdk`; lint and typecheck clean across app, plugin-sdk, server, templates.

PR #924, opened 2026-08-01
Let a plugin replace the sidebar thread list

SawyerHood:
**Layer 1 of 5.** Base: `main`.

Adds `app.slots.experimental_threadList` — the first **exclusive** plugin slot. One registered list at a time fills the sidebar's scroll area, and `chrome: "list-and-actions"` additionally claims the New-thread + search row above it.

### Why it is different from every other slot

Every other `app.slots.*` member is additive. Two thread lists cannot share one scroll area, so this one needs arbitration:

- The built-in list stays the default. Installing a plugin never takes the sidebar.
- The user picks in **Settings → Appearance → Sidebar**, stored client-local under `bb.sidebar.threadListProvider`.
- A preference naming an unregistered provider resolves to the built-in list **without clearing the stored value**, so a disabled or reloading plugin gets its list back when it returns.
- A crashing component renders the built-in list plus one toast — not the usual "plugin crashed" chip, which in place of a whole sidebar would strand the user with no way to reach their threads.

### The plugin gets the list, not the chrome

The New-thread button, the search field, the plugin nav rows (Tools, Docs, Tasks), and the footer stay host-rendered in every sidebar. Two of those regions are *other plugins''' surfaces*, so a replaced list must not be able to remove them. A list that wants its own controls puts them at the top of its own scroll area and filters by the `searchQuery` prop rather than shipping a second search field.

### Notes for review

- `chrome` is defaulted once at the collector boundary (`CollectedPluginThreadList`), so no read site re-applies it.
- Docs ship with the change per AGENTS.md: `docs/api_to_audit.md` entry, the plugin-authoring skill, and the design spec (`docs/plugin-sidebar-thread-list.md`) + build plan the rest of the stack follows.

Verified: `typecheck`, `lint`, and the full `@bb/app` / `@bb/plugin-sdk` / `@bb/server` suites pass on this layer alone.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the Slop Cop. I am reviewing this pull request for security, code quality, performance, duplication, architecture, and end-to-end behavior.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Performance phase: pass.

The provider uses the existing plugin slot store. The host adds no data query. Selection scans only registered thread-list slots. The fallback list does not mount during normal plugin use. The generation key remounts only after plugin reload. Focused app and SDK tests passed.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Security phase: pass.

Plugins already run as trusted, same-origin code. The change adds no privilege, injection, authentication, network, file-system, route, or daemon risk. Invalid and missing stored providers fall back to the built-in list.

The reviewer also found two functional contract issues. Plugin search lacks host keyboard-result controls. The viewport documentation also promises coarse-pointer detection, but the implementation passes only the width result. I will include both in the final review.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Code-quality and end-to-end phase: findings.

Browser tests passed provider discovery, selection, local storage, live replacement, search-text delivery, crash fallback, and the crash toast. The development server and browser stopped cleanly.

The review found three issues. Two new public SDK types lack the required experimental prefix. Two new documents conflict with the implemented contract. One exported crash-state helper has no caller and describes obsolete architecture.

packages/plugin-sdk/src/app-contract.ts
```diff
@@ -87,6 +87,30 @@ export interface PluginPendingInteractionProps {
...
+ * Props passed to an `experimental_threadList` component — the sidebar's
+ * scrolling thread area, replaced wholesale by one plugin.
+ */
+export interface PluginThreadListProps {
+  /** The thread the route currently shows; null on non-thread routes. */
+  activeThreadId: string | null;
+  /** The project the route currently shows; null when none is selected. */
+  activeProjectId: string | null;
+  /** True on phone-width viewports and coarse pointers. */
+  isCompactViewport: boolean;
```
SawyerHood, line 100: 🚨 `slopcop/review` — `isCompactViewport` does not match this public contract.

The implementation passes `useSidebar().isCompactViewport`, which uses the width query only. It does not include coarse pointers. A wide touch device therefore receives `false`. Either combine the width and coarse-pointer values before this prop crosses the SDK boundary, or remove the coarse-pointer promise from the contract.

docs/t3sidebar-plugin-plan.md
```diff
@@ -0,0 +1,373 @@
...
+
+Create `plugins/t3sidebar/` in the shape of `plugins/automations/`:
+a `package.json` with a `bb` block, `src/server.ts`, and `app.tsx`.
+
+## Stage B1: cards, and nothing else
+
+**What you build.** One flat list. Sort by `createdAt`, newest first, and
+never re-sort. Draw the three-line card: project and status on line one,
+title on line two, branch and glyphs on line three. Register with
+`chrome: "list-and-actions"` so you own the header row, with a project scope
```
SawyerHood, line 169: 🚨 `slopcop/review` — This plan describes an unsupported and conflicting contract.

The registration has no `chrome` field, and `AppSidebar` always keeps the host search controls. The new specification also says that the earlier plugin-owned action row is gone. Update this plan to match the implemented contract. Also update the specification status, which currently says that none of this work exists.

apps/app/src/components/plugin/PluginSlotMount.tsx
```diff
@@ -51,6 +51,26 @@ function releaseSlotInstanceOwnedState(instanceKey: string): void {
...
+ * Whether a slot instance is currently disabled by a crash. Read by hosts
+ * that hide their own UI while a plugin owns a region: the boundary's
+ * fallback restores the region's CONTENT, but only the host can restore the
+ * chrome it stopped rendering.
+ *
+ * Safe to read during render: the set only changes in `componentDidCatch`,
+ * which is followed by the boundary's own re-render, and crashed instances
+ * stay crashed for the session.
+ */
+export function isPluginSlotInstanceCrashed(
```
SawyerHood, line 64: 🚨 `slopcop/review` — This crash-state helper has no caller.

The comment describes a host that hides chrome while a plugin owns a region. This implementation always renders the host chrome. Remove this obsolete export and comment unless a current caller needs them.

packages/plugin-sdk/src/app-contract.ts
```diff
@@ -87,6 +87,30 @@ export interface PluginPendingInteractionProps {
...
+   * Call after the user opens a thread. Closes the mobile sidebar drawer and
+   * is a no-op on desktop, so always call it.
+   */
+  onNavigate: () => void;
+  /**
+   * The host search field's current text, or "" when the field is closed.
+   * The host owns that field, so a plugin list filters by this rather than
+   * shipping a second search box.
+   */
+  searchQuery: string;
```
SawyerHood, line 111: 🚨 `slopcop/review` — Plugin search cannot use the host keyboard controls.

The plugin receives only `searchQuery`, but `AppSidebar` still handles ArrowUp, ArrowDown, and Enter through `threadSearchNavigationItems`. Only the unmounted built-in `ProjectList` can populate that array. A fresh plugin search cannot select results. A provider change during active search can let Enter open a stale hidden built-in result. Add a result and active-index contract, or let the plugin own the search control. Add a keyboard-flow test.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Verdict: request changes. GitHub does not permit this account to request changes on its own pull request.

Blocking findings:

1. Plugin search cannot use the host keyboard controls. The plugin receives only the query. Only the built-in list supplies navigation results.
2. `isCompactViewport` promises coarse-pointer support. The implementation supplies only the width result.
3. The new plan and specification conflict with the implemented chrome and search contract.

Optional cleanup:

- Remove the unused `isPluginSlotInstanceCrashed` export and its obsolete comment.

Security review passed. Performance review passed. I found no duplicate implementation that needs a refactor.

Browser tests passed provider discovery, selection, storage, live replacement, query delivery, crash fallback, and the crash toast.

Focused app, SDK, documentation, and type checks passed. GitHub CI passed. One unrelated local server installation test failed.

PR #932, opened 2026-08-01
Force the host daemon to exit when a shutdown does not

SawyerHood:
## Summary

A MacBook Air stopped connecting to bb after the server moved to host-daemon protocol 69. The daemon had installed the matching bb-app package and logged `Installed the server-matched bb-app package; restarting the daemon.`, but the process never exited. `launchd` uses `KeepAlive`, so it only restarts the job after the process exits — it saw a live PID and did nothing. The stale process looped on `protocol version 68 does not match 69` for over an hour.

`stop()` memoizes its promise, so the later self-update attempt could not retrigger the restart either. The same failure happened once before, on 2026-07-21.

The restart depends on the event loop draining. Any leftover socket, timer, watcher, or child process keeps the daemon alive forever, and there was no backstop.

## Change

`createDaemon` now arms a force-exit watchdog at the start of `stop()`:

- The timer is **unref'd**, so it does not hold the event loop open. A clean shutdown that drains still exits by itself, exactly as before.
- If the process is still alive after 15 seconds, the watchdog logs the shutdown reason plus `process.getActiveResourcesInfo()`, then calls `process.exit(0)`. The service manager restarts the daemon.
- The timer starts before the cleanup awaits, so it also covers a shutdown step that hangs.

This covers every shutdown path: self-update, signals, session-close, and startup-failure.

`forceExit` is injected. Only `startHostDaemon` supplies `process.exit`, so in-process callers and tests cannot kill the test runner.

The `activeResources` field in the new error log will name the handle kinds that blocked the exit the next time this happens.

## Tests

Two new tests in `daemon.test.ts` cover a hung cleanup step and a successful cleanup that leaves the event loop alive.

- `pnpm exec turbo run test --filter=@bb/host-daemon` — 473 tests pass
- `pnpm exec turbo run typecheck --filter=@bb/host-daemon` — clean

## Notes

No wire format changed, so `HOST_DAEMON_PROTOCOL_VERSION` stays at 69.

### DIFF
--- a/apps/host-daemon/src/daemon.ts
+++ b/apps/host-daemon/src/daemon.ts
@@ -20,6 +20,12 @@ export interface CreateDaemonOptions {
   shutdownRuntimes?: () => Promise<void>;
   onStart?: () => Promise<void>;
   signalSource?: SignalSource;
+  /**
+   * Ends the process when a shutdown does not. Only the real process
+   * entrypoint supplies this; in-process callers such as tests leave it unset.
+   */
+  forceExit?: (code: number) => void;
+  shutdownExitGraceMs?: number;
 }
 
 export interface HostDaemon {
@@ -30,6 +36,22 @@ export interface HostDaemon {
 }
 
 const TERMINATION_SIGNALS: NodeJS.Signals[] = ["SIGINT", "SIGTERM"];
+
+/**
+ * How long a shutdown may take before the process is ended by force. A restart
+ * after a self-update only happens once the process really exits, so a hung
+ * shutdown step or an undrained event loop must not keep the daemon alive.
+ */
+export const DEFAULT_SHUTDOWN_EXIT_GRACE_MS = 15_000;
+
+function listActiveResources(): string[] {
+  const getActiveResourcesInfo = (
+    process as NodeJS.Process & {
+      getActiveResourcesInfo?: () => string[];
+    }
+  ).getActiveResourcesInfo;
+  return getActiveResourcesInfo ? getActiveResourcesInfo() : [];
+}
 
 export function createDaemon(options: CreateDaemonOptions): HostDaemon {
   let started = false;
@@ -51,10 +73,32 @@ export function createDaemon(options: CreateDaemonOptions): HostDaemon {
     listeners.clear();
   }
 
+  // Unref'd so a drained event loop still ends the process on its own; the
+  // timer only fires when something is still holding the loop open.
+  function armShutdownExitWatchdog(reason: string): void {
+    const forceExit = options.forceExit;
+    if (!forceExit) {
+      return;
+    }
+
+    const graceMs =
+      options.shutdownExitGraceMs ?? DEFAULT_SHUTDOWN_EXIT_GRACE_MS;
+    const timer = setTimeout(() => {
+      options.logger.error(
+        { reason, graceMs, activeResources: listActiveResources() },
+        "Host daemon shutdown did not end the process; forcing exit so the service manager can restart it.",
+      );
+      forceExit(0);
+    }, graceMs);
+    timer.unref?.();
+  }
+
   async function stop(reason: string): Promise<void> {
     if (stopPromise) {
       return stopPromise;
     }
+
+    armShutdownExitWatchdog(reason);
 
     stopPromise = (async () => {
       unregisterSignalHandlers();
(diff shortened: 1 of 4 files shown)


SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I started the security, code quality, performance, architecture, and end-to-end review phases.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Security phase complete. I found no direct privilege or remote-code issue. The server already had authority to stop the daemon.

The fixed deadline can exit before all provider children stop. This can leave child processes after the service manager restarts.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Code quality and architecture phase complete. I found one medium issue.

The watchdog always exits with code 0. Startup failure and daemon lock loss require code 1.

The watchdog can change these failures into successful exits. Please pass the intended exit code into the watchdog.

I found no useful shared helper and no wire protocol change.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Performance phase complete. The timer uses little CPU and memory. It arms one time and does not keep the event loop active.

However, shutdown stops environment runtimes one at a time. Three provider timeouts can use the full 15-second deadline.

Parallel cleanup or a longer deadline can prevent a forced exit during expected cleanup.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

End-to-end phase complete. A browser test does not apply because this pull request changes no route or user interface.

A real child process used the watchdog with a blocked shutdown step. It logged active resources and exited with code 0.

The full host daemon test suite passed. All 473 tests passed, and the type check passed.

apps/host-daemon/src/daemon.ts
```diff
@@ -51,11 +73,33 @@ export function createDaemon(options: CreateDaemonOptions): HostDaemon {
...
+    }
+
+    const graceMs =
+      options.shutdownExitGraceMs ?? DEFAULT_SHUTDOWN_EXIT_GRACE_MS;
+    const timer = setTimeout(() => {
+      options.logger.error(
+        { reason, graceMs, activeResources: listActiveResources() },
+        "Host daemon shutdown did not end the process; forcing exit so the service manager can restart it.",
+      );
+      forceExit(0);
```
SawyerHood, line 91: 🚨 `slopcop/review` — Preserve the failure exit code

This watchdog always calls `forceExit(0)`. The startup failure path later sets `process.exitCode = 1`.

The lock loss path also uses `process.exit(1)`. A blocked shutdown lets this timer report those failures as success.

Pass an exit code to the watchdog. Use code 1 for startup failure and lock loss.

apps/host-daemon/src/daemon.ts
```diff
@@ -31,6 +37,22 @@ export interface HostDaemon {
 
 const TERMINATION_SIGNALS: NodeJS.Signals[] = ["SIGINT", "SIGTERM"];
 
+/**
+ * How long a shutdown may take before the process is ended by force. A restart
+ * after a self-update only happens once the process really exits, so a hung
+ * shutdown step or an undrained event loop must not keep the daemon alive.
+ */
+export const DEFAULT_SHUTDOWN_EXIT_GRACE_MS = 15_000;
```
SawyerHood, line 45: 🚨 `slopcop/review` — Ensure the grace period covers sequential runtime cleanup

`runtimeManager.shutdownAll()` stops environment runtimes one at a time. Each provider runtime can wait five seconds before `SIGKILL`.

Three such runtimes can use this full deadline. The forced exit can skip later child cleanup and the final event flush.

Run environment shutdowns in parallel or increase the deadline above the bounded cleanup time.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

I found two medium issues. I did not find a direct security defect.

1. The watchdog always exits with code 0. It can hide startup failure and lock loss as successful exits.
2. The 15-second deadline can end sequential runtime cleanup. It can skip later child cleanup and the final event flush.

The service files always restart the daemon. Therefore, the first issue does not prevent a managed restart.

The timer arms once and does not keep the event loop active. It adds no important CPU or memory cost.

I found no useful shared helper. I found no host wire change, so the protocol version does not need a change.

Validation passed at the exact head SHA.

- The host daemon type check passed.
- All 44 host daemon test files passed.
- All 473 host daemon tests passed.
- A real child process logged active resources and exited after its blocked shutdown.
- A browser test did not apply because this change has no route or user interface.

I left inline comments for both findings. I submit this as a comment because neither issue blocks the managed restart.

PR #941, opened 2026-08-03
Refine Extensions defaults and plugin ordering

brsbl:
## What changed

- Open Extensions on Plugins from both the root route and the sidebar entry path.
- Default the Skills agent filter to BB, keep it selected even when the current library has no BB skills, and place BB Official skills first.
- Order installed plugins with enabled plugins first, then inactive plugins; enabled BB Official plugins lead their group, with name direction and plugin ID providing stable tie-breaks.
- Preserve user-selected filters, sort direction, and pagination after defaults are applied.
- Render the skills.sh Browse grid incrementally with one stable loading slot per skill; each slot becomes a fully populated card only after that card description and repository stats settle.
- Increase Browse card bylines from the `text-2xs` token to the sanctioned `text-xs` token.
- Expand regression coverage for Extensions entry routing, Skills defaults, Plugins pagination and mode switching, and per-card skills.sh loading.

## Why

The Extensions landing page and collection defaults should surface the most relevant BB-owned resources first without overwriting later user choices. Installed plugin ordering should make active plugins easiest to scan while remaining stable when names collide. Browse skill cards should appear progressively as coherent units, without either blocking the entire grid or visibly reshaping individual cards as secondary skills.sh requests complete.

The reported page-2 freeze came from the shared resource-pagination race already fixed on current main in `0a84c3708`: a stale page-size synchronization effect could overwrite a newer page selection. This PR verifies that shared fix through the Plugins integration rather than adding a plugin-specific workaround.

## Validation

- Route-memory regression: 7 tests passed.
- Skills view and per-card skills.sh loading regressions: 24 tests passed.
- App and shared UI Turbo typechecks passed.
- App and shared UI lint passed with 0 errors; the app reports 147 existing warnings.
- Running PR desktop verified Plugins opens first and Browse transitions from initial search loading into the card grid.
- Deterministic regression resolves one card while another remains pending and verifies only the complete card appears.
- Review loop completed with no remaining P0-P2 findings.

### DIFF
--- a/apps/app/src/views/thread-detail/ThreadDetailSecondaryContent.tsx
+++ b/apps/app/src/views/thread-detail/ThreadDetailSecondaryContent.tsx
@@ -20,7 +20,10 @@ import { DETAIL_GRID_CLASS } from "@/components/ui/detail-card.js";
 import { useAtomValue } from "jotai";
 import { cn } from "@bb/shared-ui/lib/utils";
 import { ThreadSecondaryPanel } from "@/components/secondary-panel/ThreadSecondaryPanel";
-import { secondaryPanelWidthPercentAtom } from "@/components/secondary-panel/threadSecondaryPanelAtoms";
+import {
+  secondaryPanelWidthPercentAtom,
+  threadSecondaryPanelResizingAtom,
+} from "@/components/secondary-panel/threadSecondaryPanelAtoms";
 import {
   ThreadMetadataCard,
   ThreadMetadataContent,
@@ -282,6 +285,7 @@ function ThreadDetailSecondaryContentBody({
         <ThreadSecondaryPanel
           {...threadSecondaryPanelProps}
           browserDeck={browserDeck}
+          onPanelResize={setLiveSecondaryWidthPercent}
           renderAsDrawer={false}
           isConversationCollapsed={isConversationCollapsedActive}
           onToggleConversationCollapse={onToggleConversationCollapse}
@@ -362,6 +366,13 @@ function ThreadDetailSecondaryContentBody({
     );
   }
 
+  const headerSidebarSurfaceWidth =
+    isSecondaryPanelOpen && !renderAsDrawer
+      ? isConversationCollapsedActive
+        ? "100%"
+        : `${liveSecondaryWidthPercent}%`
+      : "0%";
+
   return (
     <div
       className={cn(
@@ -374,7 +385,19 @@ function ThreadDetailSecondaryContentBody({
         actions stay anchored to the window edge instead of riding the timeline
         panel's width as the secondary panel opens and closes.
       */}
-      {header}
+      <div className="relative shrink-0 bg-surface-scrim">
+        <div
+          aria-hidden
+          data-testid="thread-secondary-panel-header-surface"
+          className={cn(
+            "pointer-events-none absolute inset-y-0 right-0 bg-sidebar",
+            !isSecondaryPanelResizing &&
+              `transition-[width] ${PANEL_COLLAPSE_TRANSITION_CLASS}`,
+          )}
+          style={{ width: headerSidebarSurfaceWidth }}
+        />
+        <div className="relative [&>header]:bg-transparent">{header}</div>
+      </div>
       {/*
         When collapsed we keep the resizable PanelGroup mounted: the timeline
         lifts to 0% and the panel to 100% via the layout effect. Nothing
--- a/apps/app/src/views/thread-detail/ThreadDetailSecondaryContent.tsx
+++ b/apps/app/src/views/thread-detail/ThreadDetailSecondaryContent.tsx
@@ -120,6 +120,12 @@ function ThreadDetailSecondaryContentBody({
   const persistedSecondaryWidthPercent = useAtomValue(
     secondaryPanelWidthPercentAtom,
   );
+  const isSecondaryPanelResizing = useAtomValue(
+    threadSecondaryPanelResizingAtom,
+  );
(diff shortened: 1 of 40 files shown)


brsbl:
Integrated the thread/right-sidebar UI patch in d7c1313ec, scoped to apps/app only. This preserves the existing collapse icon, makes tab close actions hover/focus-only, keeps maximized panes free of the collapsed rail while preserving Restore pane, and applies themed sidebar surfaces/fades across right-panel content. Review-loop fixes also keep the header surface aligned during live resize and prevent a maximized, conversation-collapsed pane from hiding Restore. Validation: 7 focused files / 51 tests passed; @bb/app Turbo typecheck passed; Prettier and git diff checks passed; app lint passed with 0 errors (147 existing warnings); focused Ladle stories registered from this worktree. The PR dev app is being relaunched after validation.

brsbl:
Added the automation UX follow-up in 6359592ac. Pending one-time automations now trail other active entries; completed one-time automations are treated as inactive and trail other inactive entries, while the selected name/project order remains intact inside each category. Automation detail removes the duplicate Read only badge, uses the permission control as the access source of truth, adds disabled-cursor tooltips pointing to Edit with chat, and always renders the full model label (covered with 5.5 Sol). Validation: Automations plugin 55/55 tests; focused app 38/38 tests; @bb/app and bb-plugin-automations Turbo typechecks; Prettier/diff checks; app lint 0 errors (147 existing warnings). Stories: Overview page, Detail states, and Provider identities updated/verified as registered. No merge performed.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I will check security, code quality, performance, architecture, and the main user flow.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Security review complete.

I found no security problem in the changed client code. The pull request does not change server routes, daemon messages, data access, or authorization.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Performance review complete.

I found one important performance regression. The new live panel-width state updates the full thread view on each resize event.

I also found small, bounded costs in the skill-card loading changes and new sort logic.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I confirmed two concrete UI regressions. A narrow automation detail view can hide its model label. The new disabled selector can also hide the selected value from keyboard users.

I also confirmed the live resize performance problem. Several small helpers and predicates repeat existing code, but they do not block this change.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

End-to-end review complete.

The local app opened `/tools` at `/tools/plugins`. The Installed and Browse tabs worked. Enabled official plugins stayed first in both sort directions.

The browser reported no page error. The app typecheck passed. All 2,302 app tests passed.

plugins/automations/detail-view.tsx
```diff
@@ -698,15 +708,12 @@ export function AutomationDetailView({
                     label: (
                       <DisabledAutomationSelector
                         label="Provider and model"
+                        disabledReason="Use Edit with chat to change the provider and model."
```
SawyerHood, line 711: 🚨 `slopcop/review` — The model label disappears in narrow automation details.

The 34rem container query hides `[data-promptbox-full-label]`. This selector no longer provides `compactValue`, so only the icon and chevron remain. Please restore the model label as `compactValue`.

plugins/automations/detail-view.tsx
```diff
@@ -332,6 +334,25 @@ function DisabledAutomationSelector({
...
   );
+
+  return (
+    <TooltipProvider delayDuration={250}>
+      <Tooltip>
+        <TooltipTrigger asChild>
+          <span
+            className="inline-flex shrink-0 cursor-not-allowed rounded-md focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 focus-visible:ring-offset-background"
+            tabIndex={0}
+            aria-label={`${label}. ${disabledReason}`}
```
SawyerHood, line 345: 🚨 `slopcop/review` — The focus label omits the selected value.

This span becomes the keyboard focus target because the child button is disabled. Its label gives the reason but omits the provider, model, or permission. Include `accessibleValue ?? value` in this label.

apps/app/src/views/thread-detail/ThreadDetailSecondaryContent.tsx
```diff
@@ -282,6 +291,7 @@ function ThreadDetailSecondaryContentBody({
         <ThreadSecondaryPanel
           {...threadSecondaryPanelProps}
           browserDeck={browserDeck}
+          onPanelResize={setLiveSecondaryWidthPercent}
```
SawyerHood, line 294: 🚨 `slopcop/review` — Live resize now rerenders the full thread view for each pointer update.

Each resize sends a new float into top-level React state. This rerenders the thread body and timeline during each drag frame. The split host pays the same cost but never reads the value. Use a CSS property or attach this callback only outside the hosted view.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Review complete.

Three findings need fixes:

- Restore the compact model label in narrow automation detail views.
- Include the selected value in each disabled selector focus label.
- Avoid a full thread view rerender on each panel resize update.

The security review found no issue.

The architecture scan found small duplication in disabled-control tooltips and official-plugin checks. These items do not block this pull request.

The `@bb/app` typecheck passed. All 2,302 app tests passed. The browser verified Extensions routes, tabs, and plugin order without a page error.

This is a comment-only review. I did not approve or request changes.

PR #942, opened 2026-08-03
Ask before the desktop app attaches to a running bb

SawyerHood:
## What

The desktop app attached to any compatible bb on its port silently. If you had started bb in a terminal, the app opened and gave no sign that it was using that other copy.

Now it asks first, and it can stop the other copy for you.

| A bb it can identify | An older bb it cannot |
| --- | --- |
| Quit this bb / Quit other bb / Connect | Quit this bb / Connect |

## How

- A running `bb-app start` writes `<dataDir>/bb-app-runtime.json` (pid, server URL, version, start time, surface) and removes it on exit.
- The server probe now returns the server's `dataDir`, so the desktop app can find that file and describe what it found: address, data directory, version, and how long ago it started.
- New modal dialog (`existing-server-dialog.ts`, following the existing `server-url-dialog.ts` preload/IPC pattern, because the local-view CSP blocks scripts). Closing the window or pressing Escape quits rather than attaching — attaching must be deliberate.
- Local development stays silent: the prompt is skipped when the app is unpackaged or `BB_DESKTOP_APP_URL` is set, so the `pnpm dev` attach is unchanged.
- A bb older than the runtime file cannot be identified, so the stop option and its warning disappear and the copy explains why.
- New `npx bb-app stop` gives the CLI and agents the same ability.
- `verified-process-stop.ts` holds the shared verify-then-signal logic (`ps` check, SIGTERM, then SIGKILL). The desktop's `owned-runtime-supervisor.ts` now uses it instead of its own copy.

## Safety

Both stop paths confirm the recorded pid against the `ps` command line before signalling, so a stale file left by a crash cannot stop an unrelated process. Verification accepts the entry path or its file name: Node resolves `argv[1]` to an absolute path, but `ps` reports the command line as typed, so a relative launch never contains the absolute form. A runtime file describing a different server than the one that answered is ignored.

## Tests

`turbo run lint typecheck test build` for `@bb/desktop`, `bb-app`, and `@bb/config`. New coverage: stale-file matching, SIGKILL escalation, refusal to signal a recycled pid, relative-path verification, dialog markup and escaping, and the `dataDir` probe field.

Also verified by hand against a real bb on an isolated data directory and ports: the runtime file appeared with the right contents, `bb-app stop` stopped the server and daemon and freed both ports, and a second stop reported no running bb and exited 0. That run is what caught the relative-path verification bug.

## Notes

- Docs updated: `packages/bb-app/README.md` and a "Stopping A Running bb" section in `docs/configuration.md`.
- No host-daemon wire change, so `HOST_DAEMON_PROTOCOL_VERSION` is untouched.
- Not verified: the dialog inside a packaged Electron build. The gate needs `app.isPackaged`, so the markup and rendering were checked instead. The primary button uses `AccentColor`, matching the existing Set Server URL dialog.
- Follow-up filed as BB-81: stop starting a local server at all when a remote target is configured.

### DIFF
--- a/apps/desktop/src/foreign-runtime.ts
+++ b/apps/desktop/src/foreign-runtime.ts
@@ -1,5 +1,6 @@
 import {
   bbAppRuntimeVerifyTokens,
+  clearOwnBbAppRuntimeFile,
   readBbAppRuntimeFile,
   type BbAppRuntimeFile,
 } from "@bb/config/app-runtime-file";
@@ -17,6 +18,7 @@ import type { VerifiedProcessOps } from "@bb/config/verified-process-stop";
  */
 export interface ForeignRuntimeDetails {
   dataDir: string;
+  entryPath: string;
   pid: number;
   startedAt: string;
   surface: string;
@@ -29,14 +31,17 @@ export interface ReadForeignRuntimeDetailsArgs {
 }
 
 export interface StopForeignRuntimeArgs {
-  dataDir: string;
+  /** The record the person saw and approved, not a fresh read. */
+  details: ForeignRuntimeDetails;
+  killTimeoutMs: number;
   processOps?: VerifiedProcessOps;
   timeoutMs: number;
 }
 
 export type StopForeignRuntimeResult =
-  | { kind: "no-runtime-file" }
   | { kind: "not-running" }
+  | { kind: "replaced" }
+  | { kind: "still-running"; pid: number }
   | { kind: "stopped" }
   | { kind: "unverified"; pid: number };
 
@@ -74,6 +79,7 @@ export async function readForeignRuntimeDetails(
 
   return {
     dataDir: args.dataDir,
+    entryPath: runtimeFile.entryPath,
     pid: runtimeFile.pid,
     startedAt: runtimeFile.startedAt,
     surface: runtimeFile.surface,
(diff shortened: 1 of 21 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I now review this pull request for security, code quality, performance, architecture, and end-to-end behavior.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Code-quality and architecture review complete. I found one runtime-file owner defect.

A second launcher can replace and then delete the first launcher's runtime record. The dialog code also repeats the existing HTML-dialog pattern.

The focused test set passed with 254 tests. The three Turbo type checks also passed.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Performance review complete. The normal probe path adds little work.

The worst stop path can wait about 45 seconds. It uses three separate 15-second wait budgets.

The browser smoke test showed the expected title, details, and three actions.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Security review complete. I found two process-stop safety defects.

The runtime record supplies both the target PID and its verification token. A stale PID can therefore match another bb process.

The stop action also reads the record again after approval. It can target a process that the dialog did not show.

apps/desktop/src/foreign-runtime.ts
```diff
@@ -0,0 +1,107 @@
...
+    startedAt: runtimeFile.startedAt,
+    surface: runtimeFile.surface,
+    version: runtimeFile.version,
+  };
+}
+
+export async function stopForeignRuntime(
+  args: StopForeignRuntimeArgs,
+): Promise<StopForeignRuntimeResult> {
+  const runtimeFile = await readBbAppRuntimeFile(args.dataDir);
```
SawyerHood: 🚨 `slopcop/review` — Stop only the process that the user approved

The dialog shows one record, but this function reads the file again after an unlimited wait for user input. A new launcher can replace the file during that wait. This code then stops a PID that the dialog never showed. Pass the approved identity into this function. Reject the action if a new read differs.

packages/config/src/verified-process-stop.ts
```diff
@@ -0,0 +1,141 @@
...
+  const exited = await processOps.waitForExit({
+    pid: args.pid,
+    timeoutMs: args.timeoutMs,
+  });
+  if (exited || !processOps.isRunning(args.pid)) {
+    return { kind: "stopped", usedKill: false };
+  }
+
+  processOps.kill(args.pid, "SIGKILL");
+  await processOps.waitForExit({
```
SawyerHood: 🚨 `slopcop/review` — Do not report a failed SIGKILL as success

This code ignores the final wait result and always returns `stopped`. A process can remain alive after the wait. Both callers then report success, and the CLI removes its runtime record. Return a `still-running` result and keep the record when the final wait fails.

packages/config/src/app-runtime-file.ts
```diff
@@ -0,0 +1,97 @@
...
+ * Substrings that prove a `ps` command line belongs to the launcher this file
+ * describes. Node resolves `argv[1]` to an absolute path before the launcher
+ * records it, but `ps` reports the command line as it was typed. A launcher
+ * started with a relative path therefore never shows the absolute form, so the
+ * file name is offered as a second spelling.
+ */
+export function bbAppRuntimeVerifyTokens(
+  runtimeFile: BbAppRuntimeFile,
+): string[] {
+  return [runtimeFile.entryPath, basename(runtimeFile.entryPath)];
```
SawyerHood: 🚨 `slopcop/review` — Do not use a basename substring as process identity

A stale record can point to a reused PID whose command contains `bb-app.js`. Another bb on a different data directory will pass this check. The stop path can then send SIGKILL to that bb. Use a strong argv path check and compare the recorded start time with the process start time.

packages/bb-app/src/launcher.ts
```diff
@@ -2960,5 +3056,6 @@ export async function runBbApp(
     throw error;
   } finally {
     removeSignalForwarding();
+    await clearBbAppRuntimeFile(context.dataDir);
```
SawyerHood: 🚨 `slopcop/review` — Do not remove another launcher's runtime record

A second `bb-app start` can replace the live file before its server bind. When that start fails, this unconditional clear removes the record while the first bb still runs. Then `bb-app stop` cannot find it. Clear only when a new read still has `process.pid`. Also reject a live owner before the write.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

I request changes for four process-stop safety defects.

- A second launcher can remove the live launcher's runtime record.
- A basename substring can verify the wrong process after PID reuse.
- The stop action can target a process that the dialog did not show.
- A failed final stop wait still reports success and removes the runtime record.

The Electron dialog uses a sandbox, strict IPC data, escaped text, and a restrictive content policy. I found no defect there.

The normal probe path adds little work. The worst stop path can use three separate 15-second wait budgets.

The new dialog repeats much of the server URL dialog. A shared HTML-dialog helper can reduce this repeated code.

Validation succeeded:

- The focused test set passed with 254 tests.
- Turbo type checks passed for all three affected packages.
- The Turbo desktop build passed.
- The browser smoke test showed the expected title, details, and three actions.
- `git diff --check` passed.

The full packaged test could not run. The local package lacks a required native optional dependency.

PR #946, opened 2026-08-03
Limit the permission mode a machine will run

### LINKED ISSUE #931: Sandbox configuration
Hi y'all, I am confused on the best way to run and manage a sandbox machine. I setup `bb` first on my computer in auto permissions, then decided to provision a remote VM to allow for running an agent in full access mode, added that machine to my `bb` account, etc.

But then, the remote `bb` on the VM is able to create `bb` threads on my computer? And also `bb` threads with full access on my computer? Tested by having it do it as a proof of concept.

Ideally the remote `bb` has a more scoped token that only allows it to create `bb` threads on the VM. Or at most `bb` threads on other machines that respect some max level of permissions (restricted on my personal computer, full on the VM).

Wdyt? Am I using `bb` incorrectly?

SawyerHood:
Closes #931.

A paired machine could start Full Access threads on any other machine on the account. Each machine now carries a **permission limit** — the highest permission mode any thread on it may run with — so a sandbox VM can stay at Full Access while a personal laptop stays lower. Every machine ships at Full Access, so nothing changes until an owner lowers one.

## Enforcement

The server resolves every thread on a machine down to its limit, in two places:

- `thread-execution-plan.ts` — covers turns, thread creation, and the displayed default, so the stored execution fact and the UI agree.
- `thread-commands.ts` — just before the daemon command, so a message queued *before* the limit dropped cannot replay at the old mode.

Clamping only ever lowers. A mode under the limit that the provider does not support stays a provider-capability error rather than becoming a silent upgrade, and a provider that supports nothing under the limit is refused outright (`host_permission_ceiling_conflict`). ACP providers, which support Accept Edits and Full Access but not Auto, resolve to Accept Edits under an Auto limit.

Verified against a live daemon: with the machine at Accept Edits, `bb thread spawn --permission-mode full` recorded `client/turn/requested = accept-edits` in the events table; at Auto it recorded `auto`; uncapped it recorded `full`.

## Who can change it

Only an owner session, from the new machine page. The route is refused for machine credentials at the bb connect gate (`isHostManagementMutation`) and again at the server (`assertHostManagementAllowed`), and it is deliberately absent from `@bb/sdk` and the `bb` CLI.

That last part is a conscious exception to the "every end-user feature ships an SDK and CLI surface" rule in `AGENTS.md`: this control is what stops one machine from escalating another, so no agent surface should be able to raise it. Reading the limit stays open everywhere (`GET /hosts`, `bb machine show --json`) so an agent can see the rules it works under.

## UI

- **Machine page** (`/settings/machines/:hostId`, new): the limit as radio cards with each mode's description, plus per-machine state that was scattered across Machines, Updates, and provider settings — projects, provider CLI versions, update status with retry, rename, and remove.
- **Machines list**: rows link to that page and show each machine's limit read-only in an aligned column, so "which machines are capped?" is answerable at a glance.
- **Permission pickers**: read the routed machine's limit from `GET /system/execution-options` (already routed by `hostId`/`environmentId`) and disable anything above it, with the reason in place of the description. A stored Full Access preference displays as the mode that will actually run.

## Notes for review

- **No daemon protocol bump.** The daemon receives the same `permissionMode` values it always did; only the server's choice of value changed, so an older daemon paired to a newer server is still capped correctly.
- Migration `0083` adds `hosts.max_permission_mode` with default `'full'`. Re-enrollment does not reset it — verified by restarting the dev server and daemon with a limit set.
- The issue's other half — a machine being able to create threads on *other* machines at all — is untouched. Scoping machine credentials to their own host would break legitimate cross-machine orchestration unless it is opt-in, so it is left for a separate decision.
- The last-mile clamp for a message queued before the limit dropped is covered by the code path and the primary clamp's tests, but that race was not staged live.

## Tests

New: clamp helper (5, including the never-raise rule), ceiling resolution and the provider conflict, the route boundary including machine credentials, the execution-options ceiling, the machine page (4), and the list's read-only limit and row link.

Repo typecheck 56/56. `@bb/app` 2235, `@bb/server` 1318/1319, `@bb/db` 355, `@bb/domain` 125, `@bb/cli` 368, `@bb/connect` 88, `@bb/server-contract` 42, `@bb/templates` 14. The single server failure is a pre-existing umask-dependent file-mode assertion in `internal-skill-trees.test.ts`, reproduced on an unmodified tree.

Docs updated in the same change: the in-CLI machines guide, the `bb-cli` skill, and `docs/configuration.md`.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I will review security, code quality, performance, architecture, and end-to-end behavior.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Performance review complete.

The review found one low-impact extra database read in the thread execution path. All new queries use indexed point lookups.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

End-to-end review complete.

I changed the machine limit to Accept Edits in a fresh development instance. The new-thread composer then disabled both higher modes.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Security review complete.

The review found one local trust-boundary concern. Remote machine credentials cannot change the permission limit.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The review found a blocking read-path error for providers that only support Full Access. It also found small duplication in provider capability checks.

apps/server/src/services/hosts/permission-ceiling.ts
```diff
@@ -0,0 +1,84 @@
...
+): PermissionMode {
+  const ceiling = getHostPermissionCeiling(deps, args.hostId);
+  const supported = supportedPermissionModes(args.providerId);
+  const clamped = clampPermissionModeToCeiling({
+    ceiling,
+    permissionMode: args.permissionMode,
+    ...(supported ? { supportedPermissionModes: supported } : {}),
+  });
+  if (clamped === null) {
+    throw new ApiError(
```
SawyerHood: 🚨 `slopcop/review` — Return a safe read result when no permission mode fits.

This plain `ApiError` escapes `tryResolveExistingThreadExecutionPlan`. A PI thread on an Accept Edits machine makes both execution-options read routes return 400. Please make this a handled capability conflict, while the send path still returns the error.

apps/server/src/routes/hosts.ts
```diff
@@ -135,6 +135,23 @@ export function registerHostRoutes(
     return context.json(requireNonDestroyedHostWithStatus(deps, updated.id));
   });
 
+  // Owner-session only, and deliberately absent from the SDK and the `bb` CLI:
+  // this ceiling is what stops one paired machine from running privileged work
+  // on another, so an agent on any machine must not be able to raise it.
+  patch(routes.updatePermissionCeiling, (context, payload) => {
+    assertHostManagementAllowed(context);
```
SawyerHood, line 153: 🚨 `slopcop/review` — This check proves only that the caller lacks a machine header.

A direct local request has no gate header, so this route accepts it. Thus, a local process can raise any machine limit through `BB_SERVER_URL`. Please require a positive owner signal, or document the local server process as a trusted owner.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

This review found one blocking correctness problem.

- A Full-Access-only provider causes execution-option read routes to return 400 after the machine limit drops.
- The local permission-limit route accepts callers without a positive owner signal.
- The first-load placeholder briefly enables modes above the machine limit.
- The thread hot path performs one avoidable thread read.
- Provider capability resolution now has duplicate logic.

The server and app type checks passed. Focused permission tests passed. Browser checks confirmed that Accept Edits disables both higher modes.

The full server suite passed 1,317 tests. One unrelated install-script test failed alone, and one timed-out host test passed alone.

GitHub does not permit an author to request changes on their own pull request. I am leaving this blocking review as a comment.

PR #888, opened 2026-07-29
Separate Automations from Extensions

brsbl:
## What

- Rename the experiment-gated management surface to **Extensions** and keep only Skills and Plugins inside it.
- Keep the existing **Automations** product name and move its plugin-owned page beside Threads at `/plugins/automations/automations`.
- Group every plugin-contributed nav page under a first-level **Plugins** sidebar section. The group shares the main sidebar scroll, is hidden when empty, shows all contributed pages, and persists its collapsed state.
- Redirect legacy `/tools/automations` and `/automations` collection, browse, detail, and edit URLs to the canonical Automations plugin routes.
- Clarify plugin detail pages:
  - Put Release directly below About, and make it a quiet metadata run instead of a bordered fact table.
  - Lift Compatibility out of the fact list into a warning-toned row with its own details action, so an exceptional blocked update does not read as another fact.
  - Widen the page and render Includes as a single `ResourceDetailCollection` — one surface, with each capability category as a quiet row-group header over its items.
  - Lay out each item as name plus description on one baseline, so long ids and long summaries wrap as a hanging definition instead of stranding text mid-row.
  - Give Activity the same row-group language, so the two neighbouring sections read as one list vocabulary instead of two.
  - Separate runtime Activity into plugin health, background services, scheduled jobs, and handler errors, with visible status text and leading status icons.
- Add production-backed Automations overview, browse, state, and detail stories.
- Reorganize the Sidebar stories so Plugin pages and Extensions are focused, non-overlapping stories.
- Isolate plugin-detail story fixtures and seed Release metadata locally so the story does not depend on backend plugin-source requests.
- Cover the Release states that previously had no story: bundled, update available, update failed, and compatibility blocked.

## Why

Extensions is the management surface for skills and plugin installations. Plugin-contributed working pages are a different kind of navigation: they belong beside Threads, under an explicit Plugins group that can grow without crowding fixed sidebar chrome. Automations remains a familiar user-facing concept while moving out of the management hub.

Plugin details now distinguish two questions that the previous page mixed together: **Includes** explains what the plugin adds for the user, while **Activity** explains whether its background work is healthy or needs attention.

[...]

PR #949, opened 2026-08-03
Put plugin pages back in flat sidebar rows

SawyerHood:
## What

Revert the sidebar layer of #888.

- Delete `PluginNavSidebarSection.tsx`. There is no collapsible **Plugins** group any more.
- `AppSidebar` mounts `PluginNavSidebarItems` again as flat rows in fixed chrome above `SidebarContent`, not inside the thread scroll area.
- Remove `pluginNavSidebarCollapsedAtom` and its `bb.sidebar.pluginNavCollapsed` storage key.
- Remove the separate **Extensions** button from `ProjectListActionButtons`.
- Restore `seedLeadingNavPanelKeys()`, so the built-in row stays on top for users who already reordered their rows.

Two changes on top of the revert:

- The built-in row is named **Extensions**, not **Tools**. Its internal id stays `"tools"`, so an order or hidden list a user already saved keeps naming the same row.
- The Automations row is no longer filtered out while the `toolsHub` experiment is on. Extensions no longer contains Automations, so the filter would leave the page with no sidebar entry.

## Why

The section wrapper broke drag reorder for the rows inside it. The flat list also matches what users expect: plugin pages sit beside **New thread**, with hidden rows in the **More (N)** disclosure.

## Validation

- `pnpm exec turbo run test --filter=@bb/app --force` — 307 files, 2327 tests passed.
- `pnpm exec turbo run lint --filter=@bb/app` — 0 errors (149 pre-existing warnings).
- Manual QA on a dev instance with Docs, GitHub, Memory, Tasks, Automations, and Workflows installed.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull request for security, code quality, performance, architecture, and user-path behavior.

PR #950, opened 2026-08-03
Align sidebar nav row menus with the search icon

SawyerHood:
## What

The plugin nav rows (**Tasks**, **Automations**, **Extensions**) drew their 3-dot trigger 4px left of the search icon on the **New thread** row above.

## Why

The hover-actions wrapper used `right-1`, and the trigger button adds its own `m-1`. The two insets stacked to 8px. The search icon sits flush at the container edge, and thread-row menus use `inset-0` plus the same `m-1`, so both center 22px from the sidebar edge while the nav rows centered at 26px.

## How

Drop the wrapper to `right-0` and let the trigger's `m-1` supply the whole inset. The row button's `pr-7` still clears the 28px actions block.

## Test

- The 6 `PluginNavSidebarItems` tests pass.
- Checked the rows in the local dev app.

### DIFF
--- a/apps/app/src/components/plugin/PluginNavSidebarItems.tsx
+++ b/apps/app/src/components/plugin/PluginNavSidebarItems.tsx
@@ -557,7 +557,10 @@ function SidebarNavRowChrome({
             data-sidebar-hover-actions-open={isActionsOpen ? "true" : undefined}
             className={cn(
               SIDEBAR_HOVER_ACTIONS_CLASS,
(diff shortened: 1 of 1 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I started the review for commit `7496b1064591719b370db6af167f27a2c57e741d`.

I will check security, code quality, performance, architecture, and the user interface.

PR #934, opened 2026-08-02
Polish split workspace panels and browser chrome

brsbl:
## Summary

- unify split, sidebar, and panel chrome so active context is clear and inactive splits recede
- refine semantic panel controls, split icons, queued attachments, and shared tab styling
- quiet the browser navigation with a compact idle state that remains recoverable by pointer and keyboard
- align side chat surfaces with the contextual right panel and add representative story coverage

## Why

The multi-split workspace accumulated competing borders, tab treatments, and controls. This pass reduces visual noise while preserving the hierarchy between the selected thread, its focused split, and the attached side panel.

## Verification

- pnpm exec turbo run typecheck --filter=@bb/app
- pnpm exec turbo run typecheck --filter=@bb/shared-ui
- focused Vitest suites covering split headers, browser chrome/deck, tab pills, sidebar controls, side chat, and embedded chat
- focused ESLint checks
- git diff --check
- visual review in the dev app and Ladle across representative split and side-panel configurations

PR #951, opened 2026-08-03
Replace the collapsed-conversation rail with a panel header toggle

SawyerHood:
## Summary

- delete `ConversationCollapsedRail` and restore the conversation from the right panel's own header control
- collapse the two macOS traffic-light reserve constants into one now that both surfaces share a `px-4` base inset
- drop the panel's `border-l` while the conversation is collapsed so it no longer doubles the sidebar's `border-r`

## Why

The rail was a 32px placeholder whose only job was to carry the restore action. `resolveConversationCollapseControl` already resolves to "Restore conversation" while collapsed, but the panel header suppressed the control in that state so the rail could own it. Rendering it in both states keeps the expand/restore pair in one slot and removes a whole surface.

Two consequences fell out of the deletion:

**Traffic-light geometry.** The panel's reserve was `pl-[68px]`, sized to sit *behind* the rail. Without the rail that lands leading content at 84px — under the pinned sidebar trigger, which is BB-46. Both the page header and the collapsed panel now start from the same 16px `px-4` inset, so `MACOS_COLLAPSED_HEADER_RESERVE_CLASS` and `MACOS_COLLAPSED_PANEL_TRAFFIC_LIGHT_RESERVE_CLASS` merge into `MACOS_COLLAPSED_TOP_LEFT_RESERVE_CLASS` at `pl-[104px]`. The paired-geometry test was already stale — it hardcoded `RAIL_WIDTH = 36` after #934 shrank the rail to `w-8`, so it asserted a 120px target the app missed by 4px.

**Double seam.** Collapsing drops the timeline panel to 0% and the resize handle to 0px, so the panel's `border-l border-border-seam` landed directly on the app sidebar's `border-r border-border-seam` and rendered as one 2px line. The panel gives up its border while collapsed; the sidebar owns that boundary. The split-workspace host was never affected — there the panel already omits its border and the host's resize handle is the seam.

**Behavior dropped:** the rail showed a spinner while the agent ran with the conversation hidden. That signal is gone, by request.

## Verification

- Vitest across `secondary-panel/`, `thread-detail/`, and `bb-desktop.test.ts` — 38 files, 279 tests
- new `ThreadSecondaryPanel.collapseControl.test.tsx` pins both halves of the disclosure pair; without a rail this control is the only way back from a collapsed conversation, so losing it strands the user. Confirmed it fails when the old gating is restored.
- ESLint on changed files — 0 errors
- manual QA in the dev app, measuring the seam pixel columns at the sidebar/panel boundary:

| state | seam columns |
| --- | --- |
| before fix, collapsed | 319, 320 — 2px |
| after fix, collapsed | 319 — 1px |
| after fix, conversation restored | 319 — 1px |


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

SlopCop started this review. I will check security, code quality, performance, architecture, and the changed user flow.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

End-to-end phase complete on the exact pull request SHA.

I opened the right panel and used the header control to collapse the conversation. The control changed to `Restore conversation` without movement.

The hidden conversation became inert. The same control restored the conversation.

The app typecheck passed. All 52 focused tests passed.

PR #953, opened 2026-08-03
Add automation breadcrumbs and pane full-screen controls

brsbl:
Adds the next focused layer on top of #941.

Changes:
- Reuses the Extensions breadcrumb renderer for Automations installed, browse, detail, edit, loading, and missing-data routes.
- Makes pane Full Screen the immediate click action and exposes supported split move actions on hover.
- Uses Exit Full Screen for the maximized-state tooltip and restores the plain X for thread-pane close.
- Adds focused route, interaction, split-wiring, and Ladle coverage.

Verification:
- 54 focused app tests passed.
- pnpm exec turbo run typecheck --filter=@bb/app passed on Node 22.22.1.
- BB desktop app verified Automations Installed and Browse breadcrumbs and navigation.

Stack parent: #941.

PR #954, opened 2026-08-03
Expand top-to-bottom split resize target

brsbl:
## Summary

- make split divider hit regions explicit and symmetric across both orientations
- preserve the one-pixel visual seam and existing resize persistence behavior
- add focused top-to-bottom drag regression coverage

## Verification

- `pnpm exec turbo run test --filter=@bb/app -- --run src/views/thread-detail/SplitThreadArea.test.tsx` (33 passed)

### DIFF
--- a/apps/app/src/views/thread-detail/SplitThreadArea.test.tsx
+++ b/apps/app/src/views/thread-detail/SplitThreadArea.test.tsx
@@ -282,11 +282,14 @@ function stackingLayer(element: HTMLElement): number {
   return 0;
 }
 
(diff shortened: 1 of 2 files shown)


brsbl:
Final stack update:

- Rebased onto PR #953 at 76d9d32c2; top head is a6463b88d.
- Top-to-bottom split dividers retain the one-pixel visual seam and now expose the same effective 12px drag target as left-to-right splits.
- Focused top-stack pane suite: 46 tests passed, including both resize orientations and the aligned pane-header controls.
- Full combined package matrix passed locally; BB desktop QA confirmed both orientations resize easily.
- No unresolved review threads.

Not merged.

PR #956, opened 2026-08-03
Inherit parent thread visibility on child threads

SawyerHood:
## Summary

A child thread now inherits its parent's visibility when the create request does not state one. The sub-agents that a hidden thread delegates stay hidden with it.

- `resolveCreateThreadVisibility` (`apps/server/src/services/threads/thread-create.ts`) resolves in this order: an explicit request value, then side chats stay `hidden`, then a hierarchy child takes its parent, then a root is `visible`. Resolution moved after the parent validates.
- `ThreadCreateServiceRequest.visibility` is now required, so the value is filled once at the create boundary instead of being re-defaulted in `createThreadRecord`.
- Parent notifications now key on thread origin, not visibility. The new `isParentNotifiableChildThread` predicate keeps forks and side chats silent — including legacy rows that hold an origin next to a parent id — while a hidden delegated child reports its turns and blockers again. Without this, inheritance would leave a hidden parent with no signal from the children it just started.

The four gates updated: `internal/events.ts` (parent turn notification, root turn completion), `internal/interactive-requests.ts` (needs-attention), `thread-lifecycle.ts` (command failure).

Discoverable surfaces updated in the same change: the `bb thread spawn --visibility` help, the `bb-guide-threads` template (regenerated), the `bb-cli` skill, and the `createThreadRequest` contract comment.

## Tests

- `@bb/server` — 1322 of 1323 pass. The one failure, `internal-skill-trees.test.ts`, is a pre-existing file-mode 0644 vs 0664 umask mismatch that also fails on a clean tree.
- `@bb/cli`, `@bb/sdk`, `@bb/server-contract`, `@bb/templates` — all pass.
- `@bb/integration-tests` — 25 files, 55 tests pass. New: `tests/integration/fake/smoke/hidden-delegation.test.ts` drives a hidden parent, an inherited hidden child, and the parent notification through the real server and host daemon. Confirmed non-vacuous: with inheritance removed it fails on `expected 'visible' to be 'hidden'`.

Updated tests that encoded the old rule now use a fork or side-chat child for the negative case. The needs-attention negative uses a fork, because approval requests from side chats are rejected earlier and never reach that gate.

## Risks

A hidden agent-delegated child created before this change was silent toward its parent and now posts notifications. That only affects threads created with an explicit `visibility: hidden` plus a `parentThreadId`.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Performance review complete.

I found no actionable performance issue. The change reuses the loaded parent row and adds no database query. The new notification check uses only thread fields.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I found two follow-ups:

- Four plugin and workflow guides still state the old hidden-thread notification rule.
- `isParentNotifiableChildThread` accepts thread values without origin fields. Such values can incorrectly pass the notification check.

The create-boundary visibility resolver uses the existing parent row. It does not duplicate a database read.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

End-to-end review complete.

I started the exact pull request build. I created a hidden parent through the CLI. Its child inherited `hidden`. The browser omitted both threads from the sidebar. A second child used `--visibility visible`, and the browser showed it.

The affected server tests, the new integration test, and the server, CLI, and contract type checks also passed.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Security review complete.

The parent lookup still rejects cross-project, missing, archived, deleted, and side-chat parents. The daemon route still authenticates its session and host ownership. I found no direct authorization bypass or new thread-data access.

The review found two trust-related follow-ups. Some agent guides still promise that hidden children suppress parent notifications. The CLI also prints the notification promise only for visible children, although hidden hierarchy children now notify.

apps/server/src/services/skills/builtin-skills/bb-cli/SKILL.md
```diff
@@ -129,11 +129,14 @@ isolated|reuse`, or anchor with `--source-seq-end`. Permission mode inherits
...
-  attention or native parent notifications. `bb thread list` excludes them by
+  attention. `bb thread list` excludes them by
   default; pass `--include-hidden` when a hidden worker must be discovered.
-  Direct-ID lifecycle and messaging operations remain available. Visible is
-  the default. Promote or hide an existing thread with `bb thread update <id>
---visibility visible|hidden`.
+  Direct-ID lifecycle and messaging operations remain available. A root thread
+  is visible by default; a child thread inherits its parent's visibility, so a
+  hidden thread's subagents are hidden too. Pass `--visibility` to override the
+  inherited value. A hidden child still reports its turns and blockers to its
```
SawyerHood, line 137: 🚨 `slopcop/review` — Update all guides that describe hidden-child notifications.

This text has the new rule, but four agent-facing sources still promise that hidden threads suppress parent notifications: `apps/server/src/services/skills/builtin-skills/bb-plugin-authoring/SKILL.md`, `packages/templates/src/templates/bb-guide-plugins.md`, `plugins/workflows/README.md`, and `plugins/workflows/skills/workflows/SKILL.md`. An agent can therefore select `hidden` under the old contract and receive unexpected parent turns. Please update these discoverable sources with this change.

apps/server/src/services/threads/thread-parent.ts
```diff
@@ -16,6 +16,23 @@ export function isAgentDelegatedChildThread<
...
+/**
+ * Whether a child thread reports its turns and blockers to its parent. Forks
+ * and side chats are user-initiated branches the user reads directly, so their
+ * origin excludes them; legacy rows keep that origin next to a parent id. A
+ * hidden child still reports, because a hidden parent delegates work too and
+ * needs the result.
+ */
+export function isParentNotifiableChildThread<
+  T extends Pick<Thread, "parentThreadId"> &
+    Partial<Pick<Thread, "originKind" | "childOrigin">>,
```
SawyerHood: 🚨 `slopcop/review` — Require both origin fields in this helper input.

The `Partial` type lets a caller pass only `parentThreadId`. The expression then treats both missing origin fields as `null` and classifies a legacy fork as parent-notifiable. Current callers pass full thread rows, but a future projection can reopen the fork-notification bug without a type error. Require `originKind` and `childOrigin`, and add a legacy `childOrigin: "fork"` test.

apps/cli/src/commands/thread/spawn.ts
```diff
@@ -215,7 +215,10 @@ export function registerSpawnCommand(
       [],
     )
     .option("--section <id>", "Create the thread in a section")
-    .option("--visibility <visibility>", "Thread visibility: visible or hidden")
+    .option(
+      "--visibility <visibility>",
+      "Thread visibility: visible or hidden (a child inherits its parent)",
```
SawyerHood, line 220: 🚨 `slopcop/review` — Keep the post-spawn message consistent with this new rule.

The output logic at line 340 still prints "You will be notified" only when `thread.visibility === "visible"`. A hidden hierarchy child now reports to its parent, so the CLI omits the promise for a notification that will occur. This can cause an agent to poll or duplicate work. Remove that visibility gate or base the message on the parent relationship.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

I completed the security, code quality, architecture, performance, and end-to-end reviews.

I found three non-blocking issues:

1. Four plugin and workflow guides still state that hidden threads suppress parent notifications.
2. The CLI post-spawn message still promises notifications only for visible children.
3. `isParentNotifiableChildThread` accepts missing origin fields and can fail open for a future partial projection.

I left inline comments with the exact locations and suggested changes.

The parent validation remains correct. It rejects invalid and cross-project parents before it reads visibility. The daemon request route keeps its session and host checks. I found no direct security bypass.

The visibility resolver uses the parent row that validation already loaded. It adds no database query or loop. I found no performance problem or useful large refactor.

Validation passed:

- 75 affected server tests.
- The new hidden-delegation integration test.
- Type checks for `@bb/server`, `@bb/cli`, and `@bb/server-contract`.
- A live CLI and browser flow on the exact head SHA. The hidden child stayed absent from the sidebar. An explicit visible child appeared.

PR #958, opened 2026-08-03
Delete completed plans and refresh stale docs

SawyerHood:
## Summary

Audited all 99 checked-in Markdown files against the code, then removed the plans whose work shipped and corrected the docs that no longer matched the tree.

**Deleted — 13 plans and design docs for shipped work**

| File | Why |
| --- | --- |
| `plans/tools-hub-plugin-detail-handoff.md` | PR #888 merged 2026-08-02; the doc still said "do not merge" |
| `plans/bb-connect-v1.md` | `bb connect` shipped |
| `plans/bb-connect-architecture.html` | Architecture plan for shipped connect |
| `plans/hosted-bb-architecture.html` | The connect plan said to delete it once the work landed |
| `plans/multi-machine-v1.html` | `bb machine` and the pair dialogs shipped |
| `plans/secret-request-plugin-and-pending-interaction-api.md` | `plugins/secrets` shipped |
| `plans/plugin-system-design.md` | The plugin system shipped |
| `plans/notes-plugin-api-spec.html` | The Docs plugin API shipped |
| `plans/askuserquestion-option-preview.md` | The server now offers the `preview` option field |
| `plans/loops-skills-redesign-build.md` | Its own status said all stages built; `host.write_skill` shipped |
| `docs/t3sidebar-plugin-plan.md` | `plugins/t3sidebar` shipped |
| `artifacts/simple-notes-implementation-plan.md` | The Docs plugin shipped |
| `artifacts/simple-notes-ui-mocks.html` | Mocks for the same shipped plugin |

Four source comments pointed at the deleted connect plan (`packages/tunnel-contract`, `packages/connect-db` x2, `apps/connect/scripts/spike-client.mts`). Those pointers are gone; the comments still say what they said.

`plans/` now holds only open work.

**Updated — 6 files**

- `docs/plugin-sidebar-thread-list.md` — said "Nothing here is implemented yet", but all five `experimental_*` members ship in `@bb/plugin-sdk`. Status corrected; the dead link to the deleted plan now points at `plugins/t3sidebar`.
- `plans/plugin-system-qa-catalog.md` — setup told you to enable "Settings → Experiments → Plugins"; the experiment is now Extensions (`toolsHub`). Added a rename table for `marketplace/plugins/github` → `official-plugins/github` and the deleted `small-ux-pack` and `linear` examples.
[...]

PR #961, opened 2026-08-03
Delete low-quality tests found in a codebase-wide sweep

SawyerHood:
## What

AGENTS.md says to write only tests that verify where bugs are possible. This PR sweeps every test file against that rule and deletes the tests that fail it.

Scope: all 888 test files, 7,910 test cases, 314,511 lines.

Result: **14 tests deleted across 11 files. 366 lines removed, 1 line added** (an import rewrite). 888 → 886 files. 7,910 → 7,897 cases.

## How candidates were chosen

Two passes, both adversarial:

1. **Mutation check.** For every candidate, an agent opened the source file under test, proposed realistic one-line bugs (inverted condition, dropped `await`, swapped arguments, off-by-one, removed field, early return), and had to show that none of them turn the test red. Any test that caught a mutation stayed.
2. **Skeptic pass.** An independent agent then tried to *save* every flagged test, with instructions to default to KEEP. It rejected 24 of the 38 flags.

Contract tests, migration tests, permission and parsing tests, error paths, and sole-coverage tests were out of scope by rule. Each of the 14 survivors was then re-verified by hand against the real source.

## What was deleted

| Test | File | Why |
|---|---|---|
| `gives every %s example a semantic launcher icon` | `apps/app/src/components/create-via-prompt-examples.test.ts` | `icon` is typed `IconName` on a static config and only field-copied. `icon.length > 0` is type-guaranteed. |
| `does not render argument hint placeholders…` | `…/promptbox/editor/prompt-editor-serialization.test.ts` | Calls a `return null` stub whose declared return type is the literal `null`, with zero production callers. |
| `environment_not_ready destroyed cleanup` | `apps/app/src/lib/lifecycle-errors.test.ts` | Byte-identical to the case above it. Only the `name` string differs. |
| *whole file* — `Tools Hub public export contracts` | `apps/app/src/views/tools-public-exports.test.ts` | AST export-name snapshot. No behavior change can fail it. Real removals are caught by typecheck. |
| `is stable for identical inputs` | `apps/server/test/services/threads/timeline-cache.test.ts` | Pure template-join key builder. No realistic mutation fails it while the two sibling tests stay green. |
| `is false for a root thread with no parent` | `apps/server/test/threads/thread-parent.test.ts` | Byte-for-byte duplicate of the preceding test. |
| `pushInput queues messages before start` | `…/claude-code/bridge/__tests__/sdk-session.test.ts` | Zero assertions. Dominated by the rejection test 25 lines below. |
| `renames a thread`, `stops a thread`, `steers an active turn` | `packages/agent-runtime/src/runtime.lifecycle.test.ts` | Zero assertions. Verified by injecting throws: each mutation reddened asserting twins elsewhere, never these. |
| `has correct process config`, `uses the configured bridge bundle directory…` | `packages/agent-runtime/src/pi/adapter.test.ts` | Same assertions on the same path as `provider-registry.test.ts`, which asserts strictly more. |
| *describe block* — `server-to-daemon watch payload sizes` | `packages/host-daemon-contract/test/payload-size.test.ts` | Imports no source. Measures gzip bytes of a hand-rolled untyped literal. Schema coverage is in `contract.test.ts`. |
| *whole file* — `real provider registry integration` | `tests/integration/real/provider-registry.test.ts` | Same setup as `provider-turns.test.ts` with a strict subset of its assertions. Also saves a real-provider CI run. |

The sibling `daemon-to-server` block in `payload-size.test.ts` was kept: it asserts that grouping shrinks the payload, which is a real invariant.

## Tests

- `turbo run typecheck` passes for `@bb/app`, `@bb/server`, `@bb/agent-runtime`, `@bb/host-daemon-contract`, `@bb/integration-tests`.
- `turbo run test` passes for `@bb/app` (307 files), `@bb/agent-runtime` (41), `@bb/host-daemon-contract` (3).
- `@bb/server` passes 152 of 153 files. `internal-skill-trees.test.ts` fails identically at HEAD on the dev machine, because `umask 0002` gives file mode `0664` where the test expects `0644`. It is unrelated to this change.
- The real-provider integration suite was not run locally. It needs live provider credentials.

## Risk

Low. The change deletes test code only. No source file is touched.

One judgment call to flag: `tools-public-exports.test.ts` was created deliberately as a frozen export surface. It matched the implementation-mirror category exactly, so it was deleted. Its documentation reference in `plans/tools-hub-plugin-detail-handoff.md` disappeared with #958, so no stale pointer remains.

## Follow-up

`promptMentionArgumentHintPlaceholder` in `prompt-mention-extension.ts` is now unreferenced. Removing the function is a separate change.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the Slop Cop. I started the review for this pull request.

I will review security, code quality, architecture, performance, and applicable end-to-end behavior.

PR #963, opened 2026-08-04
Fill the screen in the iOS standalone PWA

### LINKED ISSUE #871: iOS/iPadOS standalone PWA: app shell renders shorter than the viewport (bottom gap, content under status bar)
## Summary

When BB is added to the iOS/iPadOS home screen and launched as a standalone PWA, the whole app shell renders shorter than the viewport: dead space along the bottom, and content pushed up so the top runs under the status bar. It is not a cosmetic offset — the shell never fills the screen in either orientation.

The same build is completely fine in a normal Safari tab. It only reproduces when launched from the home-screen icon.

## Environment

- BB 0.33.0, self-hosted, reached over Tailscale Serve (HTTPS, `atum-vps.<tailnet>.ts.net`)
- Reproduces on **both** iPhone and iPad, current iOS/iPadOS
- Reproduces against unmodified upstream UI — `git diff origin/main <our branch> -- apps/app packages/shared-ui` is empty on our fork, so no local styling is involved
- Long-standing; not a recent regression

## Steps to reproduce

1. Open BB in Safari on an iPhone or iPad.
2. Share → Add to Home Screen.
3. Launch from the home-screen icon (standalone mode only engages from the icon, not from a Safari tab).

**Expected:** shell fills the screen, content clears the status bar and home indicator.
**Actual:** shell is shorter than the screen. Empty band at the bottom; content shifted up and colliding with the status bar at the top.

## The useful signal

**The bottom gap is visibly wider in portrait than in landscape.**

Portrait reserves more room for the home indicator than landscape does, so a gap whose height tracks orientation points at `env(safe-area-inset-bottom)` being accounted for twice — once by the standalone viewport itself and again by padding inside the layout.

Consistent with that: in a Safari tab the insets are effectively zero, there is nothing to double-reserve, and the layout is correct. That is why this only ever shows up in standalone.

Rotating the device does not correct it, so this is not a stale height captured at mount — it is wrong at every layout pass, by an amount that tracks the inset.

## Where I think it comes from

Two independently-computed viewport heights are nested inside an ancestor that clips rather than scrolls:

- `apps/app/src/components/ui/sidebar.tsx:1175` — `SidebarInset` `<main>` sets `min-h-svh`
- `apps/app/src/components/layout/AppLayout.tsx:1082` — its only child independently sets `h-[100dvh]` **plus** all four `env(safe-area-inset-*)` paddings
- `apps/app/src/app.css:10-24` — `html.bb-app-shell-root` / `body.bb-app-shell` are locked to `height: 100%; overflow: hidden`, so any excess is clipped and unreachable rather than scrollable

`apps/app/index.html` sets `viewport-fit=cover` with `apple-mobile-web-app-status-bar-style: black-translucent`, so content is deliberately painted under the status bar and the inset padding is the only thing pushing it back. That makes the shell's height arithmetic load-bearing in standalone in a way it is not in a browser tab.

One of those two heights looks redundant. I suspect the child should fill its parent rather than re-assert a full-viewport height, but I could not determine which one is actually wrong without measurements from the device.

## What I could not verify

I have not confirmed which of the two heights is miscomputing, because `env(safe-area-inset-*)` cannot be emulated in Chrome DevTools or any headless runner — the only surface that reproduces it is a physical device in standalone mode. I did not want to guess at a fix and submit it untested, which is why this is an issue rather than a PR.

Happy to run a diagnostic build on real hardware and report `100svh` / `100lvh` / `100dvh` / `innerHeight` / `visualViewport.height` and the four resolved inset values in both orientations, if that would help pin it down. Also glad to turn this into a PR if you can say which of the two heights is meant to be authoritative.

SawyerHood:
Fixes #871.

## The bug

Launched from the iOS home-screen icon, the app shell rendered shorter than the screen: a dead band at the bottom, and content pushed up under the status bar. The same build was fine in a Safari tab.

## Root cause

Measured on a physical iPhone in standalone mode (393x852 screen, 59pt top inset, 34pt bottom inset):

| Value | Reported |
|---|---|
| `innerHeight` / `visualViewport` | 793 |
| `100svh` / `100dvh` / `height: 100%` | 793 |
| `100lvh` / `100vh` | **852** |

`793 = 852 - 59`. iOS standalone subtracts the **top** inset from the small and dynamic viewports, but leaves the large viewport at the full screen height. The shell painted 793pt starting at y=0, so it stopped 59pt above the bottom of the screen.

The original issue theorized a double-counted *bottom* inset. That was not it — the insets are applied once and correctly. `100lvh` is the only unit that reports the real screen.

## The fix

- `app.css`: in standalone only, `html` is `height: 100lvh`. Scoped to standalone because in a Safari tab `lvh` is the toolbars-retracted height, which is taller than the visible area.
- `sidebar.tsx` / `AppLayout.tsx`: the shell had three independent viewport measurements (two `min-h-svh`, one `h-[100dvh]`) nested inside ancestors that clip. They agree in a browser tab and disagree in standalone. Each now fills its parent, so one place measures the screen.
- `RootComposeSecondaryContent.tsx`: the pinned panel toggle is `fixed`, so it positioned against a viewport whose origin is under the translucent status bar. At `top-2.5` it landed on the battery. It now carries the safe-area insets, and its top offset tracks the button box size (28px fine, 36px coarse) so it centers on the same 48px chrome axis as the sidebar trigger in both pointer modes.
- Drops the `peer-data-[variant=inset]` min-height rule. Nothing in the app sets that variant.

## Testing

Verified on a physical iPhone in standalone mode: the shell now spans the full 852pt, `innerHeight` and `100dvh` correct themselves to 852, and the toggle clears the status bar.

Desktop and desktop web, measured in a headless browser:

| Check | 1440x900 | 1024x768 |
|---|---|---|
| Sidebar trigger center Y | 24 | 24 |
| Panel toggle center Y | 24 | 24 |
| Shell height vs `innerHeight` | 900 / 900 | 768 / 768 |
| Document overflow | 0 | 0 |
| Sidebar panel height | 900 | 768 |

`display-mode: standalone` is false on desktop web, so the `100lvh` rule never applies there. The insets resolve to 0 off-device, so the toggle computes to exactly `top: 10px; right: 16px` — unchanged from before. Electron uses a different trigger branch (`AppLayout.tsx:230`) that this does not touch.

`pnpm exec turbo run test --filter=@bb/app`: 307 files, 2330 tests, all passing. Typecheck passes.

## Risks

`env(safe-area-inset-*)` cannot be emulated in any headless runner, so the standalone path is covered by device testing rather than automated tests. The macOS drag-strip cutout reads the same position constant as the toggle, so their footprints stay paired.

### DIFF
--- a/apps/app/src/components/ui/sidebar.tsx
+++ b/apps/app/src/components/ui/sidebar.tsx
@@ -451,7 +451,11 @@ const SidebarProvider = React.forwardRef<
               } as React.CSSProperties
             }
             className={cn(
-              "group/sidebar-wrapper flex min-h-svh w-full has-[[data-variant=inset]]:bg-sidebar",
+              // Fill the shell root (html/body/#root are height:100%) instead of
+              // re-measuring the viewport. On iOS standalone, viewport units and
+              // the safe-area insets disagree, and app.css clips the difference
+              // into an unreachable band at the bottom of the screen.
+              "group/sidebar-wrapper flex h-full min-h-0 w-full has-[[data-variant=inset]]:bg-sidebar",
               className,
             )}
             ref={ref}
@@ -638,7 +642,7 @@ const Sidebar = React.forwardRef<
         <div
           data-sidebar="gap"
           className={cn(
-            "relative hidden h-svh w-(--sidebar-width) bg-transparent transition-[width] duration-200 ease-linear md:block",
+            "relative hidden h-full w-(--sidebar-width) bg-transparent transition-[width] duration-200 ease-linear md:block",
             "group-data-[collapsible=offcanvas]:w-0",
             "group-data-[side=right]:rotate-180",
             variant === "floating" || variant === "inset"
@@ -649,7 +653,7 @@ const Sidebar = React.forwardRef<
         <div
           data-sidebar="panel"
           className={cn(
-            "fixed inset-y-0 z-10 flex h-svh w-(--sidebar-width) flex-col bg-sidebar text-sidebar-foreground transition-[left,right,width] duration-200 ease-linear",
+            "fixed inset-y-0 z-10 flex h-full w-(--sidebar-width) flex-col bg-sidebar text-sidebar-foreground transition-[left,right,width] duration-200 ease-linear",
             side === "left"
               ? "left-0 group-data-[collapsible=offcanvas]:left-[calc(var(--sidebar-width)*-1)]"
               : "right-0 group-data-[collapsible=offcanvas]:right-[calc(var(--sidebar-width)*-1)]",
@@ -1173,8 +1177,8 @@ const SidebarInset = React.forwardRef<
       ref={ref}
       data-sidebar="inset"
       className={cn(
-        "relative flex min-h-svh min-w-0 flex-1 flex-col bg-background",
-        "peer-data-[variant=inset]:min-h-[calc(100svh-theme(spacing.4))] md:peer-data-[variant=inset]:m-2 md:peer-data-[variant=inset]:ml-0 md:peer-data-[variant=inset]:rounded-xl md:peer-data-[variant=inset]:shadow",
+        "relative flex h-full min-h-0 min-w-0 flex-1 flex-col bg-background",
+        "md:peer-data-[variant=inset]:m-2 md:peer-data-[variant=inset]:ml-0 md:peer-data-[variant=inset]:rounded-xl md:peer-data-[variant=inset]:shadow",
         className,
       )}
       {...props}
--- a/apps/app/src/components/ui/sidebar.tsx
+++ b/apps/app/src/components/ui/sidebar.tsx
@@ -594,7 +594,9 @@ const Sidebar = React.forwardRef<
                 suppressMobileOpenAnimation ? "true" : undefined
               }
               className={cn(
-                "group fixed inset-y-0 z-40 flex h-dvh w-(--sidebar-width-mobile) flex-col bg-sidebar text-sidebar-foreground outline-none",
+                // Fixed: a percentage height would resolve against the short
+                // initial containing block, so it reads the shell unit directly.
+                "group fixed inset-y-0 z-40 flex h-(--bb-shell-height) w-(--sidebar-width-mobile) flex-col bg-sidebar text-sidebar-foreground outline-none",
                 "[&[data-sidebar-suppress-open-animation=true][data-state=open]]:![animation:none]",
                 side === "left" ? "left-0" : "right-0",
                 variant === "floating" || variant === "inset"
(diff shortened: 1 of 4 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I started the review for security, code quality, performance, architecture, and end-to-end behavior.

I will post an update after each phase.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Security phase complete.

The security agent found one potential privacy issue. Some modal screens can remain shorter than the new standalone shell. I will validate this finding during synthesis.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Performance phase complete.

The performance agent found no issue. The change adds no effects, listeners, render paths, or animations. Relevant tests and the app type check passed.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

End-to-end phase complete.

I tested mobile and Chromium standalone app modes. The standalone rule matched. A 59-pixel top inset and 34-pixel bottom inset applied correctly. The shell, backdrop, mobile sidebar, and content reached the same bottom edge.

I could not run mobile Safari or a physical iOS device.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The agent found two incomplete sidebar surfaces and one hardware-dependent viewport concern. I confirmed the fixed panel issue with a simulated 852-pixel screen and a 793-pixel viewport.

A shared screen-height CSS property can remove the duplicate viewport rules.

apps/app/src/components/ui/sidebar.tsx
```diff
@@ -649,7 +653,7 @@ const Sidebar = React.forwardRef<
         <div
           data-sidebar="panel"
           className={cn(
-            "fixed inset-y-0 z-10 flex h-svh w-(--sidebar-width) flex-col bg-sidebar text-sidebar-foreground transition-[left,right,width] duration-200 ease-linear",
+            "fixed inset-y-0 z-10 flex h-full w-(--sidebar-width) flex-col bg-sidebar text-sidebar-foreground transition-[left,right,width] duration-200 ease-linear",
```
SawyerHood: 🚨 `slopcop/review` — The fixed sidebar still uses the shorter viewport

`h-full` cannot inherit the new `100lvh` shell height here because this panel uses fixed positioning. Its percentage height uses the initial containing block.

I reproduced a 793-pixel panel inside an 852-pixel shell. The sidebar left a 59-pixel bottom band.

Define one mode-aware screen-height CSS property. Apply it directly to fixed full-height surfaces, including the compact sidebar.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

I found one P2 issue.

The fixed sidebar panel still uses the shorter viewport. I reproduced a 793-pixel panel inside an 852-pixel shell.

The compact sidebar still uses `h-dvh`. A physical iPhone must confirm that path and the visual viewport hook.

Use one mode-aware screen-height CSS property. Apply it to the shell and all fixed full-height surfaces.

Security review found the same exposed bottom band. Performance review found no issue.

Validation passed:

- `git diff --check`
- The `@bb/app` type check
- 307 test files and 2,330 tests
- Mobile and Chromium standalone browser smoke tests

The browser test used 59-pixel top and 34-pixel bottom safe areas. I could not run mobile Safari on a physical device.

PR #968, opened 2026-08-04
Align full-stack regression coverage

brsbl:
## Summary

- update the sidebar-tab full-screen regression to assert the shipped Full Screen / Exit Full Screen interaction and pressed state
- keep the same-slot restoration guarantee covered after the unified control rename

## Verification

- focused @bb/app pane/sidebar/browser suite: 9 files, 67 tests passed
- full-stack branch app: thread and browser-tab expansion/restoration, browser chrome, nested split resize, side-panel availability, and new-thread terminal surface verified

Stack parent: #967. Do not merge independently.

### DIFF
--- a/apps/app/src/components/tools/Automations.stories.tsx
+++ b/apps/app/src/components/tools/Automations.stories.tsx
@@ -6,6 +6,7 @@ import {
   type AutomationCollectionMode,
 } from "bb-plugin-automations/overview-view";
 import type {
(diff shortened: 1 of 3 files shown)


brsbl:
This empty, closed layer was removed from the GitHub stack after its regression-only change was absorbed into #967 during rebasing. The active top layer is now #970, position 9 of official stack #971. No code was dropped.

PR #990, opened 2026-08-04
Skip the local bb server when the desktop targets a remote server

SawyerHood:
Closes BB-81.

## What changed

The desktop app no longer starts a bb server on this Mac when the saved target
is a bb Connect server or a custom URL. It starts one again when you select
"This Mac".

Three things kept the local server alive. Each one now has a remote path:

1. **The Connect session cookie.** The app enrolls itself once as a connect
   machine (mint a code on the local server, redeem it at the apex), then mints
   its own cookie at the gate. The credential lives in Electron `userData`,
   encrypted with the OS keychain through `safeStorage`. When the OS offers no
   encryption backend, the cache writes nothing and the app keeps asking the
   local server.
2. **The account server list.** The Server menu syncs from the gate with the
   same credential when no local server runs.
3. **System config.** A remote server cannot use the realtime socket, because
   only Electron's network stack carries the session cookie. A remote target
   polls `net.fetch` on start, on `did-become-active`, and every five minutes.

`packages/connect-client` now holds the gate client (credential shape, URL
helpers, server list, desktop session, machine redeem). The connect plugin and
the desktop app share it.

## On the credential

The task proposed caching the server's own pairing secret. This enrolls the app
as its own machine instead. Same file and same encryption, but the server's
secret never leaves `bb.db`, the app shows up in the dashboard machine list, and
revoking it there is enough. A gate `401`/`403` drops the cached copy, which
also closes the "bypasses the plugin gate" risk.

It costs one machine slot on the account, and only for users who select a remote
target.

## First launch

No cached credential means the local server mints the cookie, exactly as before,
and enrollment runs in the background for the next launch.

## Tests

New: credential cache (round trip, no-encryption no-op, undecryptable bytes),
machine enrollment (mint + redeem, missing handle, unpaired, machine limit),
both cookie sources, gate-backed server sync. `@bb/connect-client` covers the
URL helpers, the server list, and every redeem rejection code.

Whole-repo typecheck passes. Four `@bb/desktop` test files fail on Linux both
before and after this change; they need the Electron binary.

## Not covered

I could not run the packaged app: this worktree is Linux and the Electron binary
is macOS-only. The remote-target launch path wants a manual pass on a Mac.

### DIFF
--- a/apps/desktop/src/connect-credential-cache.ts
+++ b/apps/desktop/src/connect-credential-cache.ts
@@ -27,6 +27,12 @@ export interface CreateConnectCredentialCacheArgs {
 }
 
 export interface ConnectCredentialCache {
+  /**
+   * Whether a written credential survives a restart. False means the OS gave
+   * Electron no encryption backend, so callers must not enroll: the credential
+   * would live in memory only, and every launch would enroll another machine.
+   */
+  canPersist(): boolean;
   clear(): Promise<void>;
   read(): Promise<ConnectCredential | null>;
   write(credential: ConnectCredential): Promise<void>;
@@ -58,6 +64,9 @@ export function createConnectCredentialCache(
   }
 
   return {
+    canPersist() {
+      return args.encryption.isEncryptionAvailable();
+    },
     clear,
     async read() {
       if (!args.encryption.isEncryptionAvailable()) {
--- a/apps/desktop/src/connect-credential-cache.ts
+++ b/apps/desktop/src/connect-credential-cache.ts
@@ -1,0 +1,105 @@
+import { rm, readFile, writeFile } from "node:fs/promises";
+import { join } from "node:path";
+import {
+  connectCredentialSchema,
+  type ConnectCredential,
+} from "@bb/connect-client";
+
+export const CONNECT_CREDENTIAL_FILE_NAME = "connect-credential.bin";
+
+/** Electron's `safeStorage`, narrowed to what the cache uses. */
+export interface ConnectCredentialEncryption {
+  decryptString(encrypted: Buffer): string;
+  encryptString(plainText: string): Buffer;
+  isEncryptionAvailable(): boolean;
+}
+
+export interface ConnectCredentialCacheFs {
+  readFile(path: string): Promise<Buffer>;
+  rm(path: string, options: { force: true }): Promise<void>;
+  writeFile(path: string, data: Buffer): Promise<void>;
+}
+
+export interface CreateConnectCredentialCacheArgs {
+  encryption: ConnectCredentialEncryption;
+  fs?: ConnectCredentialCacheFs;
+  userDataPath: string;
+}
+
+export interface ConnectCredentialCache {
+  clear(): Promise<void>;
+  read(): Promise<ConnectCredential | null>;
+  write(credential: ConnectCredential): Promise<void>;
+}
+
+const defaultFs: ConnectCredentialCacheFs = {
+  readFile,
+  rm,
+  writeFile,
+};
+
+/**
+ * The desktop app's own connect machine credential, encrypted at rest with the
+ * OS keychain through Electron `safeStorage`. It lets the app mint a Connect
+ * session cookie and list account servers with no local bb server running.
+ *
+ * When the OS offers no encryption backend the cache turns into a no-op rather
+ * than writing a durable secret in the clear; the app then falls back to the
+ * local server for those two calls.
+ */
+export function createConnectCredentialCache(
+  args: CreateConnectCredentialCacheArgs,
+): ConnectCredentialCache {
+  const fsImpl = args.fs ?? defaultFs;
+  const filePath = join(args.userDataPath, CONNECT_CREDENTIAL_FILE_NAME);
+
+  async function clear(): Promise<void> {
+    await fsImpl.rm(filePath, { force: true });
+  }
+
+  return {
+    clear,
+    async read() {
+      if (!args.encryption.isEncryptionAvailable()) {
+        return null;
+      }
+      let encrypted: Buffer;
+      try {
+        encrypted = await fsImpl.readFile(filePath);
+      } catch {
+        return null;
+      }
+      let plainText: string;
+      try {
+        plainText = args.encryption.decryptString(encrypted);
+      } catch {
+        // A keychain the OS can no longer unlock, or a file from another
+        // machine: the bytes are useless, so drop them and re-enroll.
+        await clear();
+        return null;
+      }
+      let parsedJson: unknown;
+      try {
+        parsedJson = JSON.parse(plainText);
+      } catch {
+        await clear();
+        return null;
+      }
+      const parsed = connectCredentialSchema.safeParse(parsedJson);
+      if (!parsed.success) {
+        await clear();
+        return null;
+      }
+      return parsed.data;
+    },
+    async write(credential) {
+      if (!args.encryption.isEncryptionAvailable()) {
+        return;
+      }
+      await fsImpl.writeFile(
+        filePath,
+        args.encryption.encryptString(JSON.stringify(credential)),
+      );
+    },
+  };
+}
--- a/apps/desktop/src/connect-machine-enrollment.ts
+++ b/apps/desktop/src/connect-machine-enrollment.ts
@@ -1,0 +1,132 @@
+import { z } from "zod";
+import {
+  ConnectMachineRedeemError,
+  deriveConnectBaseUrl,
+  redeemMachineCredential,
+  type ConnectCredential,
+} from "@bb/connect-client";
+
+export const CREATE_MACHINE_CODE_RPC = "createMachineCode";
+
+const machineCodeRpcSchema = z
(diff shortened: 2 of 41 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The stack contains only #990.

I will review security, code quality, architecture, duplication, performance, tests, and the desktop path.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The performance review is complete.

It found three main risks. A missing keychain can create a new machine on each launch. Gate calls have no timeout. The one-hour remote session has no renewal path.

The focused type checks and 293 tests passed.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The security review is complete.

It found one boundary risk. The new client stores and reuses the returned server URL without URL or origin checks.

The credential storage and cookie settings passed review.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The end-to-end desktop test is complete.

Electron loaded a saved remote target through the desktop bridge. Browser automation found the bb user interface. The local server and host daemon ports stayed closed.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The code quality and architecture review is complete.

The shared client removes useful duplication. The review found an RPC error-envelope mismatch, a blocked local fallback, and a multi-server identity mismatch.

No host daemon protocol change is necessary.

apps/desktop/src/connect-machine-enrollment.ts
```diff
@@ -0,0 +1,132 @@
...
+        code: z.string().min(1),
+        expiresAt: z.number(),
+        serverUrl: z.string().min(1),
+      })
+      .strict(),
+  })
+  .strict();
+
+const rpcFailureSchema = z
+  .object({ ok: z.literal(false), error: z.string().optional() })
```
SawyerHood: 🚨 `slopcop/review` — Parse the real RPC error envelope

The server returns `error: { code, message }` for handler failures. This schema accepts only a string. Thus `not_paired` and `machine_limit` never match below. The test uses a wire shape that the server does not send. Parse the object and update the test.

packages/connect-client/src/redeem-machine.ts
```diff
@@ -0,0 +1,118 @@
...
+  const { credential, handle, serverUrl } = parsed.data;
+  // The gate echoes the server the code targeted. Without it there is nothing
+  // to point a session at, so treat the row as unusable.
+  if (serverUrl === null) {
+    throw new ConnectMachineRedeemError(
+      "invalid_response",
+      "Machine redeem returned no server URL",
+    );
+  }
+  const resolvedHandle = handle ?? handleFromServerUrl(serverUrl);
```
SawyerHood: 🚨 `slopcop/review` — Use the target server label

The redeem route returns the account primary handle here, while `serverUrl` names the target server. A non-primary bb therefore stores the wrong self handle. The server list then hides the primary remote bb and shows This Mac as remote. Derive the label from a validated `serverUrl`, and constrain that URL to the known apex.

apps/desktop/src/main.ts
```diff
@@ -919,11 +987,97 @@ async function ensureBuiltinRuntimeAttached(): Promise<boolean> {
...
+  const cookieStore = session.defaultSession.cookies;
+  if (cachedConnectCredential !== null) {
+    const cachedResult = await installConnectDesktopSession({
+      cookieStore,
+      mintCookie: createCredentialCookieSource({
+        credential: cachedConnectCredential,
+      }),
+      remoteServerUrl,
+    });
+    if (cachedResult.ok || cachedResult.code !== "unauthorized") {
```
SawyerHood: 🚨 `slopcop/review` — Use the local fallback after all cached failures

This condition returns for each failure except `unauthorized`. A 404, invalid body, or stalled gate call never reaches the local server. That result conflicts with the function comment and can wedge a renamed server. Continue to the local source after each failure. Clear the cache only for `unauthorized`.

apps/desktop/src/main.ts
```diff
@@ -989,11 +1128,12 @@ async function applyServerTarget(): Promise<void> {
     }
     bbAppLoaded = true;
     await loadWindowUrl({ url: target.server.url });
+    startRemoteSystemConfigSync(target.server.url);
```
SawyerHood, line 1198: 🚨 `slopcop/review` — Renew the remote session before expiry

The desktop cookie expires after one hour. Authentication runs only at startup or after a target switch. This five-minute config poll does not renew the cookie. Remote API calls will fail after expiry. Keep `expiresAt` and schedule renewal, or authenticate again after the first 401.

apps/desktop/src/connect-credential-cache.ts
```diff
@@ -0,0 +1,105 @@
...
+      }
+      const parsed = connectCredentialSchema.safeParse(parsedJson);
+      if (!parsed.success) {
+        await clear();
+        return null;
+      }
+      return parsed.data;
+    },
+    async write(credential) {
+      if (!args.encryption.isEncryptionAvailable()) {
```
SawyerHood, line 105: 🚨 `slopcop/review` — Do not enroll when the cache cannot persist

This return looks like a successful write. The main process keeps the new credential only in memory. The next launch reads `null` and enrolls another permanent machine. Repeated launches can reach the account machine limit. Expose cache availability or return a persistence result, then skip enrollment when storage is unavailable.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

The review is complete. The stack contains only #990.

I found five issues:

1. Medium: The enrollment client parses the wrong RPC error envelope.
2. Medium: A non-primary server stores the account primary handle as its self handle.
3. Medium: Most cached credential failures block the local fallback.
4. Medium: An unavailable keychain can create a new Connect machine after each launch.
5. Medium: The one-hour desktop session has no renewal path.

The shared `@bb/connect-client` package removes useful plugin duplication. Some desktop contract types still duplicate shared types, but they do not add release risk.

The focused Turbo type checks passed. All 293 focused tests passed. The desktop build passed. All GitHub CI checks passed.

The Electron test loaded a saved remote target through the desktop bridge. Browser automation found the bb interface. The local server and daemon ports stayed closed.

apps/desktop/src/connect-machine-enrollment.ts
SawyerHood: Fixed in 58f6b1b86. Confirmed the envelope: `plugin-service.ts` returns `error: { code: "handler_error", message }`, and `rpc.ts` throws the stable code as that message. The schema now parses the object and matches on `message`. The test uses the real wire shape.

packages/connect-client/src/redeem-machine.ts
SawyerHood: Fixed in 58f6b1b86. Confirmed in `apps/web/src/server/api.ts`: `handle` is `prof.handle`, while `serverUrl` uses `targetServer.subdomain`. The client now ignores `handle` and takes the label from `serverUrl`, which must sit under the apex the caller asked (same protocol, same port, hostname one label below). New tests cover a mismatched apex, a suffix lookalike, a plain-http downgrade, and a self-hosted apex.

apps/desktop/src/main.ts
SawyerHood: Fixed in 58f6b1b86, with one deliberate exception. Every cached-credential failure now falls through to the local server, except `network`. When the gate is unreachable, the local server calls the same gate, so starting one cannot help — and not starting one is the point of this change. The cache is cleared only for `unauthorized`, and the first failure is returned when the local server is also unavailable.

apps/desktop/src/main.ts
SawyerHood: Fixed in 58f6b1b86. `installConnectDesktopSession` now returns the cookie `expiresAt`. The app re-mints five minutes before expiry, and also on `did-become-active` when the session is inside that window, because a sleeping Mac holds the timer. A failed renewal retries on the same cadence rather than giving up. Switching to "This Mac" or a custom URL cancels it.

apps/desktop/src/connect-credential-cache.ts
SawyerHood: Fixed in 58f6b1b86. The cache exposes `canPersist()`, and `ensureDesktopMachineEnrolled` returns early when it is false, with a log line. Without a keychain the app keeps using the local server for session cookies, so it never enrolls a machine it cannot remember.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The stack contains PR #990 only. I found no draft or untrusted pull requests in this stack.

I will check security, code quality, architecture, performance, tests, and the desktop route.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

End-to-end phase complete.

The desktop QA launcher started the server and Electron shell at `467dd1416`. The app route loaded in Chromium with no page errors.

The focused test run passed 303 tests. The three focused package type checks also passed.

I could not test a live remote Connect session. The isolated QA data has no paired account or stored machine credential.

PR #1004, opened 2026-08-04
Install git plugin dependencies before bundling

SawyerHood:
**Layer 1 of 3.** Base of the stack — targets `main`.

A `git:` plugin could only use Node built-ins, the shimmed host packages, and whatever it committed into its own tree. `installGitSource` runs only `git clone`, `checkout`, and `rev-parse`, and `buildPluginApp` resolves dependencies from `rootDir/node_modules` — empty in a fresh clone.

`docs/configuration.md` already documented the opposite. This makes the doc true.

## What changed

bb now installs a git plugin's declared runtime dependencies, builds **both** bundles so those dependencies are inlined, and discards `node_modules` before the artifact is hashed and promoted.

## Why it is safe

No plugin code executes at any point:

- `npm install --ignore-scripts` downloads tarballs and writes files. It runs nothing. bb already does exactly this for `npm:` plugins.
- esbuild bundles by parsing, never by evaluating.

So the update-check path — where `plugin-updates.ts:195` stages git candidates — stays inert.

## Scope

- Dependency-free git plugins skip npm entirely and still install with only `git` on PATH.
- `path:` sources are untouched: the author owns that directory.
- Building the server bundle makes the git artifact self-contained, so `resolveServerEntry` no longer falls back to source for git — matching the intent already documented at `plugin-runtime.ts:818-820`.

## Consequence worth reviewing

A committed `dist/` is no longer authoritative for git; bb rebuilds over it. The artifact-metadata validator still guards `npm:` sources and now has direct unit coverage instead of being exercised through a git install.

## Tests

`@bb/server` plugin-install: 29 passed, including a new case proving a third-party dependency is inlined into the built bundle.

Pre-existing unrelated failure on this machine: `internal-skill-trees.test.ts` asserts file mode 0644 and gets 0664 (umask). Fails identically on a clean tree.

---

## Review round applied

Reviewed by `thr_hdxy2afasy` (gpt-5.6-sol). Two findings landed here.

**Update checks were not read-only.** `stageGitCandidate({promote: false})` reached `validateInstallDir`, so polling installed dependencies and bundled — resolving an author-declared `file:`/`git:` dependency on every check. Split `validateManifestOnly` out; checks read the manifest and stop. Covered by a test that fails without the fix (verified by reverting).

**`node_modules` pruning was unsafe.** esbuild only bundles statically reachable code, so a dependency reading a data file or `.wasm` at runtime would break. No longer pruned, which also keeps the `resolveServerEntry` source fallback usable.

Per your call, the dependency-free special case is gone — git installs always run npm — so the shared `@bb/domain` schema is untouched.

### DIFF
--- a/apps/server/src/services/plugins/managed-plugin-artifacts.ts
+++ b/apps/server/src/services/plugins/managed-plugin-artifacts.ts
@@ -11,7 +11,7 @@ import {
   type PluginProvenance,
   type PluginSourceIntent,
 } from "@bb/db";
-import { buildPluginApp } from "@bb/plugin-build";
+import { buildPluginApp, buildPluginServer } from "@bb/plugin-build";
 import { validatePluginArtifactMeta } from "./app-bundle.js";
 import {
   gitArtifactCacheDir,
@@ -94,6 +94,40 @@ export interface ManagedPluginArtifactsContext {
   activateManagedUpdate: (args: ActivateManagedUpdateArgs) => Promise<void>;
 }
 
+/**
+ * Install a git plugin's declared runtime dependencies into its staging dir.
+ *
+ * Nothing here executes plugin code: `--ignore-scripts` means npm only
+ * downloads tarballs and writes files, and esbuild bundles by parsing rather
+ * than evaluating. That keeps the update-check path (which stages git
+ * candidates via `plugin-updates.ts`) free of plugin-authored execution.
+ *
+ * Skipped entirely when the package declares no `dependencies`, so a
+ * dependency-free git plugin still installs with only `git` on PATH.
+ */
+async function installGitDependencies(args: {
+  rootDir: string;
+  manifest: PluginManifest;
+}): Promise<void> {
+  if (!args.manifest.hasDependencies) return;
+  await runInstallCommand(
+    "npm",
+    [
+      "install",
+      "--prefix",
+      args.rootDir,
+      "--ignore-scripts",
+      "--omit=dev",
+      "--omit=optional",
+      "--no-audit",
+      "--no-fund",
+    ],
+    {
+      notFoundHint: `"npm" was not found on PATH — git plugin "${args.manifest.id}" declares dependencies, which require npm to install`,
+    },
+  );
+}
+
 async function installNpmCandidate(args: {
   stagingPrefix: string;
   registry: string;
@@ -162,10 +196,18 @@ export function createManagedPluginArtifacts(
         );
       }
     }
-    // Frontend policy (design §5.1): path:/git: sources build dist/ at
-    // install time — a build failure fails the install, like a manifest
-    // error would. npm packages are never built here; they must ship a
-    // prebuilt dist whose metadata is compatible with this SDK.
+    // Dependency + bundle policy (design §5.1):
+    // - git: bb installs declared runtime deps (scripts disabled — nothing
+    //   executes), builds BOTH bundles so those deps are inlined, then
+    //   discards node_modules. The promoted artifact is self-contained, so
+    //   `resolveServerEntry` never needs the source fallback for git.
+    // - path: the author owns that directory. Never install into it and
+    //   never prune it; only the frontend is built.
+    // - npm: never built here; must ship a prebuilt dist whose metadata is
+    //   compatible with this SDK.
+    if (kind === "git") {
+      await installGitDependencies({ rootDir: args.rootDir, manifest });
+    }
     if (manifest.appEntry !== undefined) {
       if (kind === "npm") {
         const jsPresent = await stat(join(args.rootDir, "dist", "app.js"))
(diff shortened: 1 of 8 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The stack contains #1004, #1005, #1006, and #1015, from bottom to top.

I will review security, code quality, performance, architecture, and end-to-end behavior.

I did not skip a pull request.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The security phase completed across #1004, #1005, #1006, and #1015. I will attach each finding to its own pull request.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The performance phase completed across #1004, #1005, #1006, and #1015. I will attach each finding to its own pull request.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The code-quality and architecture phase completed across #1004, #1005, #1006, and #1015. I will attach each finding to its own pull request.

apps/server/src/services/plugins/managed-plugin-artifacts.ts
```diff
@@ -94,6 +94,40 @@ export interface ManagedPluginArtifactsContext {
...
+async function installGitDependencies(args: {
+  rootDir: string;
+  manifest: PluginManifest;
+}): Promise<void> {
+  await runInstallCommand(
+    "npm",
+    [
+      "install",
+      "--prefix",
+      args.rootDir,
```
SawyerHood, line 126: 🚨 `slopcop/review` — Block repository npm configuration during git installs

`--prefix` makes npm read `.npmrc` from the cloned repository. That file can replace registry, cache, and TLS settings. It can also put server environment values into request URLs. `--ignore-scripts` blocks lifecycle scripts, but it does not block this configuration. Do not let npm read repository configuration. Use BB-owned npm configuration and explicit registry and cache values.

apps/server/src/services/plugins/managed-plugin-artifacts.ts
```diff
@@ -188,6 +254,19 @@ export function createManagedPluginArtifacts(
...
     }
+    if (kind === "git") {
+      try {
+        await buildPluginServer(args.rootDir, deps.appVersion);
+      } catch (error) {
+        throw new Error(
+          `install failed: server bundle build for "${manifest.id}" failed: ${error instanceof Error ? error.message : String(error)}`,
+        );
+      }
+      // node_modules is deliberately retained. esbuild only bundles what it
```
SawyerHood, line 267: 🚨 `slopcop/review` — Avoid repeated full scans of retained dependencies

The retained dependency tree enters `hashInstallDir` and the promotion fsync path. Reuse paths hash the tree again. A three-package test produced 6,571 files. A warm hash took 1.53 seconds, and fsync took 0.50 seconds. Avoid repeated full-tree scans. Use a compact install manifest or separate generated dependencies from the source hash.

apps/server/src/services/plugins/managed-plugin-artifacts.ts
```diff
@@ -162,10 +203,35 @@ export function createManagedPluginArtifacts(
...
+    source: string;
+    refuseEngineMismatch: boolean;
+  }): Promise<PluginManifest> {
+    const manifest = await validateManifestOnly(args);
+    const kind = sourceKind(args.source);
+    const managed = kind === "git" || kind === "npm";
+    // Dependency + bundle policy (design §5.1):
+    // - git: bb installs declared runtime deps (scripts disabled — nothing
+    //   executes), builds BOTH bundles so those deps are inlined, then
+    //   discards node_modules. The promoted artifact is self-contained, so
```
SawyerHood: 🚨 `slopcop/review` — Correct the node_modules policy comment

This comment says that BB discards `node_modules`. It also says that git plugins never need the source fallback. Lines 265 through 268 state the opposite. Remove these stale claims so the code has one clear policy.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Result: I found two medium issues and one quality issue.

- Security: a repository `.npmrc` can control npm network settings and put environment values into request URLs.
- Performance: retained `node_modules` adds full hashing and per-file fsync work.
- Quality: two policy comments conflict. An old JSDoc block also describes the wrong function.

The plugin install file passed 28 tests. The plugin update file passed 14 tests.

The full server suite had 16 unrelated timeout failures under concurrent load. The changed server tests still passed.

No later pull request fixes these findings. See the inline comments for exact locations.

apps/server/src/services/plugins/managed-plugin-artifacts.ts
SawyerHood: Fixed the other two. Leaving this one, with reasoning.

Excluding `node_modules` from `hashInstallDir` would change what the artifact hash *means* — it currently covers the promoted directory, and `promoteImmutableDir` uses it to decide whether a cached tree can be reused. Making the hash cover only some of the tree, late in a stack, risks a subtle reuse bug in exchange for time on a path that is already dominated by the network: the same macOS run measured a real GitHub install at 12.6s end to end.

Two things also bound the cost. It is install-and-update-apply only — update *checks* no longer reach this path at all (#1004), so routine polling does not pay it. And a warm hash of 1.53s against a 12.6s install is roughly 12%.

If it does become a problem, the right fix is probably hashing the dependency *inputs* (`package.json` + lockfile) rather than the resolved tree, since the tree is derived. That is a real design change and wants its own PR rather than a late edit here. Happy to file it.

PR #1005, opened 2026-08-04
Fetch the plugin build toolchain on demand

SawyerHood:
**Layer 2 of 3.** Depends on #1004.

`esbuild` (~12MB Go binary) and `@tailwindcss/oxide` (napi addon) were shipped in the server, CLI, `bb-app`, and desktop app — roughly 20MB of platform binaries per artifact, for a capability most users never invoke. Only `git:` and `path:` plugins with a frontend are ever built locally.

## Why it is cheap

Both were **already dynamic imports** (`build-plugin-app.ts:466`, `build-plugin-server.ts:132`). Only the resolution target changes.

## How it resolves

`resolvePluginBuildToolchain` prefers a locally resolvable toolchain — the monorepo, tests, any machine that already has the packages — and only otherwise installs a pinned set into `<dataDir>/plugins/toolchain-<pins>/`. So development and CI pay no download, and only a shipped artifact fetches.

The cache directory is keyed by the pinned versions, so a bump installs beside the old set rather than mutating a directory a concurrent build may be importing from. The fetch runs `--ignore-scripts` over bb's own packages only.

`buildPluginApp`/`buildPluginServer` take resolved specifiers as an optional third argument defaulting to bare names, so monorepo build scripts are unchanged. The server memoizes per data dir so concurrent installs share one fetch; a failed fetch is not cached.

## Removals

Packages and `--external` flags from server, CLI, and `bb-app`; the per-arch `@esbuild/*` optional dependencies, native packaging assertions, and `prepare-native-modules.cjs` handling from desktop; the shared runtime-externals entries in `build-utils.mjs`.

`esbuild` stays a **devDependency** of desktop (its own main-process bundle) and of plugin-build/CLI (types and tests). Neither is shipped.

## Verification

Probed a real cold fetch into an empty directory: esbuild transforms TypeScript and the native oxide addon loads, despite `--ignore-scripts`. That was the main technical risk.

`@bb/plugin-build` 3 files passed (incl. new toolchain tests) · `@bb/cli` 44/44 · `@bb/server` only the pre-existing umask failure · `@bb/desktop` `electron-builder-config.test.ts` fully green.

Pre-existing unrelated: 4 desktop test files fail to load with "Electron failed to install correctly" on this Linux box. Confirmed identical on a stashed tree.

---

## Review round applied

Reviewed by `thr_hdxy2afasy` (gpt-5.6-sol). Four findings landed here.

**Frontend builds would have failed in every shipped artifact.** `buildTailwindCss` resolved `tailwindcss/theme.css` via `createRequire(import.meta.url)` — bb's own bundle in a packaged server, where tailwindcss is not installed. Every test passed because they all took the local-resolution branch. `PluginBuildToolchain` now carries `tailwindCssDir`; `createRequire` is gone from the file.

**The toolchain argument is now required**, not defaulted to bare specifiers. That default was the mechanism that let a caller silently depend on ambient resolution.

**Pins are verified before local resolution is accepted**, so an unpinned ambient esbuild or Tailwind cannot bypass the pinned set. Versions are read by walking up from the resolved entry, since all four pins hide `./package.json` behind an exports map.

**Cross-process safety:** installs go to a private staging directory and promote by rename, so a racing server and CLI never run two npms in one prefix. A promoted tree is re-verified, so a half-deleted cache with a live marker is refetched.

Adds the end-to-end fetched-toolchain build test that was the missing coverage, gated behind `BB_TEST_TOOLCHAIN_FETCH=1`.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The stack contains #1004, #1005, #1006, and #1015, from bottom to top.

I will review security, code quality, performance, architecture, and end-to-end behavior.

I did not skip a pull request.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am reviewing this stack under the review rule.

The stack contains #1005, #1006, and #1015. I will post each finding on its correct PR.

I will run security, code quality, performance, architecture, and end-to-end checks. I will use parallel bb subagents.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Security phase complete.

Medium: the fetched toolchain fixes four direct versions, but it has no lockfile or expected integrity graph.

`@tailwindcss/node@4.3.0` still selects newer compatible transitive releases. The server later imports that fetched code in process.

Before this change, `pnpm-lock.yaml` fixed the shipped dependency graph. A later fetch can now produce different executable code for the same cache key.

Embed a generated lockfile, run `npm ci --ignore-scripts`, and include the lockfile hash in the cache key.

I also checked command injection, local `.npmrc` control, path traversal, and the staging race. I found no additional security defect.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Performance phase complete.

Medium: the npm child process has no time limit.

A stalled registry can hold the shared promise and all later plugin builds for the same data directory.

Set an `execFile` time limit and a kill signal. The existing catch path will then clear the memoized promise.

packages/plugin-build/src/toolchain.ts
```diff
@@ -0,0 +1,250 @@
...
+        staging,
+        "--ignore-scripts",
+        "--no-audit",
+        "--no-fund",
+        "--no-package-lock",
+        ...Object.entries(PLUGIN_TOOLCHAIN_PINS).map(
+          ([name, version]) => `${name}@${version}`,
+        ),
+      ],
+      { maxBuffer: 1024 * 1024 * 16 },
```
SawyerHood, line 219: 🚨 `slopcop/review` — Add a timeout to the toolchain fetch

`execFile` has no timeout here. A stalled registry keeps the memoized promise pending. All later plugin builds then wait until the process restarts. Use the same five-minute bound as other npm installs. Also add an `ENOENT` message that names npm.

packages/plugin-build/src/toolchain.ts
```diff
@@ -0,0 +1,250 @@
...
+        "the downloaded plugin build toolchain is incomplete or misversioned",
+      );
+    }
+    await writeFile(
+      join(staging, ".bb-toolchain.json"),
+      `${JSON.stringify({ pins: pinKey() }, null, 2)}\n`,
+    );
+    await mkdir(dirname(dir), { recursive: true });
+    try {
+      await rename(staging, dir);
```
SawyerHood, line 233: 🚨 `slopcop/review` — Repair a damaged cache before promotion

If `dir` exists but `isInstalled(dir)` returns false, this rename fails with `ENOTEMPTY`. Each retry downloads the toolchain and fails again. The user must then remove the cache manually. Move the invalid directory aside, or remove it before promotion.

packages/plugin-build/src/toolchain.ts
```diff
@@ -0,0 +1,250 @@
...
+    await run(
+      "npm",
+      [
+        "install",
+        "--prefix",
+        staging,
+        "--ignore-scripts",
+        "--no-audit",
+        "--no-fund",
+        "--no-package-lock",
```
SawyerHood, line 214: 🚨 `slopcop/review` — Pin the complete fetched toolchain

Exact top-level versions do not pin transitive packages. `--no-package-lock` lets each machine resolve a different dependency tree. The server then loads these modules and native binaries. A later transitive release can change executable code. Ship a lockfile and use `npm ci --ignore-scripts` with an explicit registry.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Result: I found one high issue, two medium issues, and several low issues.

- Security: the fetched toolchain pins top-level versions but does not pin the complete dependency tree.
- Reliability: the npm fetch has no timeout.
- Reliability: a damaged cache directory prevents every later promotion.
- Quality: cache checks can resolve packages above the cache directory.
- Maintenance: old toolchain directories have no cleanup path.

The CLI suite passed 44 test files. The plugin-build suite passed 10 tests and skipped one network test.

No later pull request fixes these findings. See the inline comments for the three primary issues.

packages/plugin-build/src/toolchain.ts
```diff
@@ -0,0 +1,250 @@
...
+      parsed === null ||
+      (parsed as { pins?: unknown }).pins !== pinKey()
+    ) {
+      return false;
+    }
+  } catch {
+    return false;
+  }
+  // The marker is written last, but a half-deleted cache can outlive it.
+  return toolchainFrom(createRequire(join(dir, "noop.js"))) !== null;
```
SawyerHood, line 165: 🚨 `slopcop/review` — Cache verification can escape the cache directory.

`createRequire` searches ancestor `node_modules` directories. A marker-only cache can therefore accept packages outside its cache directory.

This also lets an empty staging install pass when an ancestor contains the pinned packages.

Require every resolved package path to stay under the expected `node_modules` directory.

packages/plugin-build/src/toolchain.ts
```diff
@@ -0,0 +1,250 @@
...
+    );
+    await mkdir(dirname(dir), { recursive: true });
+    try {
+      await rename(staging, dir);
+    } catch {
+      // Another process promoted first, or the rename crossed a device.
+      // Either way its tree is equivalent — verify and use it.
+      if (!(await isInstalled(dir))) throw new Error(errorPromoting(dir));
+    }
+  } finally {
```
SawyerHood, line 239: 🚨 `slopcop/review` — A non-empty corrupt cache cannot repair itself.

`rename(staging, dir)` cannot replace a non-empty invalid directory. The catch checks the same invalid target and then throws.

I reproduced this with one junk file. Every retry failed after a successful npm fetch.

Safely move or remove the invalid target before promotion. Preserve the existing winner verification for concurrent installs.

apps/desktop/test/electron-builder-config.test.ts
```diff
@@ -255,10 +237,12 @@ describe("electron-builder signing config", () => {
       JSON.parse(packageJsonText),
     );
 
-    expect(packageJson.optionalDependencies).toMatchObject({
-      "@esbuild/darwin-arm64": "^0.28.0",
-      "@esbuild/darwin-x64": "^0.28.0",
-    });
+    expect(Object.keys(packageJson.optionalDependencies ?? {})).not.toEqual(
```
SawyerHood, line 240: 🚨 `slopcop/review` — This assertion permits one binary dependency to return.

The negated `arrayContaining` fails only when both package names exist. One restored esbuild dependency still passes this test.

Assert each forbidden name separately, or require an empty optional dependency map. Include the removed oxide and lightningcss package names.

packages/plugin-build/src/toolchain.test.ts
```diff
@@ -0,0 +1,124 @@
...
+
+        // Proves `tailwindcss/theme.css` resolved from the fetched package.
+        expect(css.length).toBeGreaterThan(0);
+        expect(css).toContain("--");
+      },
+      600_000,
+    );
+
+    it("reuses an already-fetched toolchain without reinstalling", async () => {
+      const local = await resolvePluginBuildToolchain(baseDir);
```
SawyerHood, line 119: 🚨 `slopcop/review` — This cache-hit test never uses the cache.

Local package resolution returns before the code reads `baseDir`. The final length check only verifies the monorepo toolchain.

Seed a cache, set `ignoreLocal: true`, and assert that `onFetchStart` does not run.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

I posted four inline findings. Two affect cache correctness, and two weaken regression coverage.

The largest design issue sits in `toolchainFrom`. It verifies package versions but does not verify the expected package root.

Return one scoped toolchain result from that function. Reuse it for cache checks, staging checks, and the final result.

The review also found repeated test helpers and repeated CLI toolchain resolution. These are refactor opportunities, not separate defects.

packages/plugin-build/src/toolchain.ts
```diff
@@ -0,0 +1,250 @@
...
+        "--ignore-scripts",
+        "--no-audit",
+        "--no-fund",
+        "--no-package-lock",
+        ...Object.entries(PLUGIN_TOOLCHAIN_PINS).map(
+          ([name, version]) => `${name}@${version}`,
+        ),
+      ],
+      { maxBuffer: 1024 * 1024 * 16 },
+    );
```
SawyerHood, line 220: 🚨 `slopcop/review` — The fetched dependency graph is not reproducible.

The four direct versions are exact, but their transitive ranges still move. The cache key does not record those selected versions or integrity values.

Embed a lockfile, use `npm ci --ignore-scripts`, and add the lockfile hash to the cache key.

packages/plugin-build/src/toolchain.ts
```diff
@@ -0,0 +1,250 @@
...
+    const staged = toolchainFrom(createRequire(join(staging, "noop.js")));
+    if (staged === null) {
+      throw new Error(
+        "the downloaded plugin build toolchain is incomplete or misversioned",
+      );
```
SawyerHood, line 225: 🚨 `slopcop/review` — A stalled npm process wedges all plugin builds.

`execFile` has no time limit. The server shares this pending promise, so later builds wait until the process restarts.

Set a time limit and a kill signal. Map `ENOENT` to a clear missing-npm error.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Final review summary for PR #1005.

I found four medium defects.

- Cache verification can accept packages from an ancestor directory.
- A non-empty corrupt cache cannot repair itself.
- The fetched transitive dependency graph has no fixed integrity set.
- A stalled npm process can block all shared plugin builds.

I also found two regression-test gaps. Inline comments contain the exact locations and fixes.

The architecture scan found repeated toolchain verification and test helpers. A scoped toolchain resolver can remove this duplication.

Validation passed for five package typechecks, 11 plugin-build tests, the real fetch test, CLI tests, desktop tests, and builds.

I submitted only comments. I did not approve or request changes.

PR #1009, opened 2026-08-04
Finish extension detail and automation editing UX

brsbl:
Completes the inherited skill pagination and wheel navigation, vertically centers shared plugin capability cells, subdues description disclosures, and finishes automation prompt editing with independently loaded permission modes, cached options, aligned provider icons, cancel/save behavior, and content-hugging selectors.\n\nFocused coverage: detail-page recipes (40 tests), automation server harness (14 tests), Turbo typechecks for @bb/app and bb-plugin-automations. Desktop QA artifacts cover the Tasks capability table, plugin action menu, and automation detail.

### DIFF
--- a/plugins/automations/app.tsx
+++ b/plugins/automations/app.tsx
@@ -248,13 +248,14 @@ function useAutomation(route: DetailRoute): {
 function useAutomationExecutionOptions(
   route: DetailRoute,
   enabled: boolean,
+  executionKey: string,
 ): {
   options: AutomationExecutionOptionsResponse | null;
   error: string | null;
 } {
   const rpc = useRpc<typeof automationRpcContract>();
   const { projectId, automationId } = route;
-  const requestKey = `${projectId}:${automationId}`;
+  const requestKey = `${projectId}:${automationId}:${executionKey}`;
   const requestedKeyRef = useRef<string | null>(null);
   const [state, setState] = useState<{
     options: AutomationExecutionOptionsResponse | null;
@@ -293,6 +294,7 @@ function useAutomationExecutionOptions(
 function useAutomationPermissionOptions(
   route: DetailRoute,
   enabled: boolean,
+  executionKey: string,
 ): {
   options: AutomationPermissionOptionsResponse | null;
   error: string | null;
@@ -300,7 +302,7 @@ function useAutomationPermissionOptions(
 } {
   const rpc = useRpc<typeof automationRpcContract>();
   const { projectId, automationId } = route;
-  const requestKey = `${projectId}:${automationId}`;
+  const requestKey = `${projectId}:${automationId}:${executionKey}`;
   const requestedKeyRef = useRef<string | null>(null);
   const [attempt, setAttempt] = useState(0);
   const [state, setState] = useState<{
@@ -602,13 +604,22 @@ function DetailView({
   const navigate = useBbNavigate();
   const { automation, error, missing, refetch } = useAutomation(route);
   const [editingRequested, setEditingRequested] = useState(initialEditing);
+  const editingExecutionKey =
+    automation?.execution.mode === "agent"
+      ? JSON.stringify({
+          providerId: automation.execution.providerId,
+          environment: automation.execution.environment,
+        })
+      : "not-agent";
   const executionOptionsState = useAutomationExecutionOptions(
     route,
     editingRequested && automation?.execution.mode === "agent",
+    editingExecutionKey,
   );
   const permissionOptionsState = useAutomationPermissionOptions(
     route,
     editingRequested && automation?.execution.mode === "agent",
+    editingExecutionKey,
   );
   const editing = editingRequested && permissionOptionsState.options !== null;
   const overviewState = useOverview();
--- a/plugins/automations/app.tsx
+++ b/plugins/automations/app.tsx
@@ -21,6 +21,7 @@ import { toast } from "sonner";
 import type {
   AutomationResponse,
   AutomationExecutionOptionsResponse,
+  AutomationPermissionOptionsResponse,
   AgentExecutionUpdate,
   AutomationRunListResponse,
   AutomationRunResponse,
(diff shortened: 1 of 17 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I will review this pull request under the `review` rule.

The stack contains these pull requests, from bottom to top:

- #1009 — Finish extension detail and automation editing UX
- #1010 — Refresh sidebar tab overflow from intrinsic content

I found no draft pull requests or authors without write access in this stack.

I will review security, code quality, performance, architecture, and duplicate code. I will run one end-to-end test on #1010.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Performance review is complete.

I found one P1 request-effect failure and one P2 duplicate provider-discovery cost. I will include exact locations in the final review.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Security review is complete.

I found no security issue in #1009. The RPC keeps strict schemas, project binding, and permission-mode validation.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Code-quality review is complete.

I found one high-severity editor failure and two medium-severity cancel or retry failures. I also found duplicate option hooks and route logic.

I will include exact locations in the final review.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

End-to-end review is complete on the #1010 stack head.

The automation edit route returned both option RPC responses with status 200. The page still showed zero text areas after 2.5 seconds.

This result confirms the #1009 request-effect failure.

plugins/automations/detail-view.tsx
```diff
@@ -741,12 +748,23 @@ function AgentAutomationDefinition({
             }
           />
         </div>
+        <Button
+          type="button"
+          variant="ghost"
+          size="sm"
+          className="ml-auto shrink-0"
+          disabled={pending || dirty}
```
SawyerHood, line 756: 🚨 `slopcop/review` — P2: Cancel stops after the user changes a field.

The `dirty` condition disables Cancel after any prompt, model, or permission change. A failed save also leaves the user unable to discard the draft.

Keep Cancel available while the form is dirty. Reset the local draft state before you close the editor.

plugins/automations/app.tsx
```diff
@@ -256,28 +257,96 @@ function useAutomationExecutionOptions(
...
+            requested: true,
+          });
+        }
       },
     );
     return () => {
       active = false;
     };
-  }, [automationId, enabled, projectId, rpc]);
+  }, [automationId, enabled, projectId, rpc, state]);
```
SawyerHood: 🚨 `slopcop/review` — P1: This effect cancels its own request.

The effect sets `requested` to true. The `state` dependency then starts cleanup and sets `active` to false before the RPC completes. The result cannot update the options. The same fault exists in the permission-options effect at line 347. The edit route stays in read mode even when both RPC calls return status 200.

Remove the complete state object from both dependency lists. Use stable request state or a request reference. Add a delayed-RPC test through `DetailView`.

plugins/automations/app.tsx
```diff
@@ -538,11 +607,16 @@ function DetailView({
...
   const executionOptionsState = useAutomationExecutionOptions(
     route,
-    editing && automation?.execution.mode === "agent",
+    editingRequested && automation?.execution.mode === "agent",
+  );
+  const permissionOptionsState = useAutomationPermissionOptions(
+    route,
+    editingRequested && automation?.execution.mode === "agent",
   );
+  const editing = editingRequested && permissionOptionsState.options !== null;
```
SawyerHood, line 624: 🚨 `slopcop/review` — P2: A permission-options error prevents a retry.

A failed permission request leaves `editingRequested` true and `editing` false. The editor does not show its Cancel action. A second Edit action also cannot change the true request state. The user must reload or leave the route.

Show a retry action, clear the request state after an error, or render the editor during option load.

plugins/automations/src/service.ts
```diff
@@ -438,6 +443,33 @@ export function createAutomationService(args: {
...
+        );
+      }
+      const environment = execution.environment;
+      const routing =
+        environment.type === "reuse"
+          ? { environmentId: environment.environmentId }
+          : environment.type === "host" && environment.hostId !== undefined
+            ? { hostId: environment.hostId }
+            : {};
+      const providers = await bb.sdk.providers.list(routing);
```
SawyerHood, line 458: 🚨 `slopcop/review` — P2: Edit now runs duplicate provider discovery.

The client starts this request with `executionOptions`. The model request already resolves the provider list before model discovery. This additional list call repeats host RPC work and provider executable checks for each edit.

Coalesce or cache provider discovery for the route. Another option is to share the provider metadata between both option responses.

plugins/automations/app.tsx
SawyerHood: 🚨 `slopcop/review` — Later commits fixed this self-cancel fault. I retract this finding. A separate cancel-and-reopen fault remains in the new request logic.

plugins/automations/app.tsx
SawyerHood: 🚨 `slopcop/review` — Later commits added a permission retry path. I retract this finding.

plugins/automations/src/service.ts
SawyerHood: 🚨 `slopcop/review` — I retract this as a finding. The extra call is a deliberate latency tradeoff that makes permission options available before model discovery.

plugins/automations/app.tsx
```diff
@@ -247,37 +248,101 @@ function useAutomation(route: DetailRoute): {
...
       },
       (error: unknown) => {
-        if (active) setState({ options: null, error: errorText(error) });
+        if (active) {
+          requestedKeyRef.current = null;
+          setState({ options: null, error: errorText(error) });
+        }
       },
     );
     return () => {
```
SawyerHood, line 286: 🚨 `slopcop/review` — P2: Cancel during model discovery prevents the next edit from loading models.

Cleanup sets `active` to false but keeps `requestedKeyRef` on this request key. If permission options finish first, the user can enter and cancel the editor while model discovery continues. The late model result is ignored. Reopening uses the same key, so the guard prevents a replacement request and model options remain unavailable.

Clear the request key when cleanup rejects an incomplete result. Add a delayed model-RPC test that cancels and reopens the editor.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Current head reviewed: `6e78c7fa4aeafd108450434c5fa9707326217d0b`.

I found two P2 issues:

- Cancel during model discovery can prevent model options from loading on the next edit.
- Cancel becomes unavailable after any draft change, so the user cannot discard the draft.

Security review found no issue. Performance review found no remaining issue.

Architecture review found duplicate lazy option hooks and duplicate environment-routing logic. The execution-options response also keeps an unused `permissionModes` field.

The author fixed the earlier request-cancel and permission-retry findings during this review. I retracted those comments.

Validation on the current stack passed 43 focused app tests, 56 automation tests, and both Turbo typechecks. Most current CI checks pass.

The one allowed end-to-end run used the earlier top head. It reproduced a fault that the author later fixed.

I used a comment-only review, as required.

PR #1020, opened 2026-08-05
Unify the extensions collection toolbars

brsbl:
Plugins, Skills, and Automations had each grown their own filter and sort treatment. This pulls them onto one set of shared primitives.

## Filters

- **Automations** — the separate Projects and Status dropdowns become one grouped `ResourceFilterMenu`. Each group keeps its own state and handler, so filtering behavior is identical; only the affordance merges.
- **No more "All" row.** An empty selection already means "no filter", so `ResourceMultiSelectMenu` loses `allOptionLabel` entirely rather than keeping dead API.
- **Installed plugins** gain a Type filter (`bb Official` / `User`), mapping `builtin`/`catalog` → bb Official.
- **Plugins Browse** — category pills move into the toolbar as a multi-select. The pill row ended 170px short of every other band; folding it in removes that ragged strip and gives Browse the same filter language as Installed.
- **Skills** type labels renamed to `bb Official` and `Included in plugin`.

## Control styling

Toolbar filter and sort keys are now siblings of the search input beside them — same 32px box, same `--input` border, same radius — instead of floating chips in a recessed track. Engaged uses the app's one selection surface, `bg-state-active` (the same `CONTEXT_SELECTION_SURFACE_CLASS` behind selected sidebar rows, active tab pills, and focused split panes). Icon buttons carry no resting fill, which let a hardcoded background come off `ResourceTemplateBrowseCard`.

Sort triggers are `ArrowUpDown` on all three pages; Automations previously swapped in a direction-flipping arrow.

## Layout

Every band — description, tabs, toolbar, results, footer — carries the same `pr-1` scrollbar gutter, so they share one content width. Previously the description was capped at `max-w-2xl` and inset `px-1`, and the toolbar ran 4px wider than the results.

The Automations prompt footer now sits inside the box while editing, matching read-only geometry; it used to float below on `mt-1`.

## Behavior change worth flagging

The Skills library now opens **unfiltered** (142 skills) rather than pre-filtered to bb Official (3). That follows from "nothing selected means all types", but it is the one user-visible behavior change here.

## Verification

- `turbo run test --filter=@bb/app` — 312 files / 2347 tests pass
- `turbo run typecheck` — clean for `@bb/app`, `@bb/shared-ui`, `bb-plugin-automations`
- ESLint clean on changed files
- Every page and state driven in the running app; light/dark states covered by a new `control-weight` story, and the Nord palette checked via `bb theme set nord` to confirm the `--canvas`/`--ink`-derived tokens tint correctly


SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

The stack contains only PR #1020. I found no later pull requests after the required second scan.

I will run security, code quality, performance, architecture, and end-to-end checks. I will post each phase result here.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Security review complete. I found no security problem.

I checked text injection, plugin provenance labels, permission controls, filter state, and the generated button registry source. The changes do not alter authorization, requests, routes, storage, or the host protocol.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Performance review complete. I found no user-visible performance defect.

The Skills page now searches more items because it opens without a source filter. Current work remains within one frame. A future search index could remove repeated string work if the library grows.

Pagination still limits the DOM. The new menu state stays local, and the new filter work occurs only after a user action.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

End-to-end review complete. I found no functional or visual defect in the tested paths.

I tested the Skills Type filter, the installed Plugins Type filter, the Browse Category filter, and the grouped Automations filter. I also inspected the Automations toolbar and open menu.

The app suite passed 2,347 tests. The Automations suite passed 58 tests. The App, Shared UI, and Automations type checks passed.

apps/app/src/components/ui/control-weight.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+    <div className="flex items-center gap-3">
+      <span className="w-28 shrink-0 text-xs text-muted-foreground">
+        {label}
+      </span>
+      {children}
+    </div>
+  );
+}
+
+/** The toolbar's recessed track, so engaged reads in its real context. */
```
SawyerHood, line 54: 🚨 `slopcop/review` — Replace the removed toolbar track in this story.

This story still renders the recessed 28px track that this PR removes. The shipped keys use separate 32px outline buttons with an input border. Build the story from the real trigger classes so it checks the design that users receive.

apps/app/src/components/tools/SkillsCollection.tsx
```diff
@@ -332,7 +333,10 @@ export function SkillsOverview({
   const visibleSkills = useMemo(() => {
     const filtered = skills.filter((skill) => {
       const source = skillSourceFilterId(skill);
-      if (source !== null && !sourceFilters.includes(source)) {
+      if (
+        sourceFilters.length > 0 &&
```
SawyerHood, line 337: 🚨 `slopcop/review` — Add a User source option before this filter excludes user skills.

Any selected Type now removes every user and project scope because `skillSourceFilterId` returns `null` for them. The menu only offers bb Official and Included in plugin. Add a User value and map all user-authored scopes to it.

packages/shared-ui/src/components/ui/button.tsx
```diff
@@ -30,6 +29,17 @@ const buttonVariants = cva(
         icon: "h-9 w-9",
       },
     },
+    compoundVariants: [
```
SawyerHood, line 32: 🚨 `slopcop/review` — Regenerate the plugin starter template.

This button source also feeds `packages/templates/src/generated/plugin-starter-files.generated.ts`. The current PR updates the registry output but not this second output. Both `@bb/templates` tests and type checks fail their generation check. Run `node packages/templates/scripts/generate-templates.mjs` and commit the result.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Code-quality and architecture review complete. I found one blocking build defect and two user-facing consistency defects.

The plugin starter template is stale. Both `@bb/templates` tests and type checks fail. The Skills Type menu omits user-authored skills. The new control story still shows the removed toolbar track.

The new grouped menu centralizes the main toolbar structure. Its option toggle logic still duplicates `ResourceMultiSelectMenu`, so a shared helper can reduce future drift. The Browse empty state also still says “this category” after the move to multiple selections.

I posted three inline comments for the actionable changed lines.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Final review result: changes are required, but I am leaving a comment-only review as instructed.

The current PR state still fails the `@bb/templates` generation check. Regenerate and commit `packages/templates/src/generated/plugin-starter-files.generated.ts`.

Please also add a User value to the Skills Type filter. Update the new control story to use the shipped 32px outline keys.

Security review found no problem. Performance review found no user-visible defect at the current scale. Browser tests passed across Skills, Plugins, Browse, and Automations.

The app suite passed 2,347 tests. The Automations suite passed 58 tests. The relevant App, Shared UI, and Automations type checks passed.

The grouped menu is a good shared base. Extract the duplicated option-toggle logic before it causes behavior drift.

brsbl:
Note on where the review-loop fixes landed: a staged review of this PR and #1038 was run against the combined diff (`main...bb/align-filter-controls`), so findings were not attributed per layer. Ten of the sixteen are against code introduced here — `ResourceFilterMenu`, `pluginTypeFilterId`, the `selectedLabel` call sites — but all the fixes are committed on #1038 (`77951057f`) rather than split back into this branch.

That is deliberate: `gh stack merge` lands every layer up to the chosen one atomically, so `main` never sees this PR's code without #1038's fixes, and splitting them would have meant a rebase cascade through four layers, three of them owned by another thread. The exposure is only if this PR is ever merged on its own, outside the stack — don't do that.

PR #1024, opened 2026-08-05
Fix provider usage limit normalization

ymichael:
## Summary

- expose Claude model-scoped weekly usage buckets such as Fable
- isolate malformed optional Claude rows and suppress unsupported surface-scoped or duplicate labels
- classify Codex weekly limits from the provider window duration instead of assuming the primary window is always a session limit
- clarify provider hierarchy using the shared settings row, badge, and provider icon patterns
- reuse Ladle's global router in the Automations breadcrumb story instead of nesting routers
- bump the host-daemon protocol so enrolled hosts pick up the normalized provider usage behavior

## Testing

- `pnpm exec turbo run test --filter=@bb/host-daemon --force` (473 tests)
- `pnpm exec turbo run test --filter=@bb/host-daemon-contract --force` (49 tests)
- `pnpm exec turbo run test --filter=@bb/app --force -- src/components/settings/UsageLimitsSettingsSection.test.tsx`
- `pnpm exec turbo run test --filter=@bb/app --force -- src/components/tools/Automations.stories.test.tsx`
- `pnpm exec turbo run typecheck --filter=@bb/host-daemon --filter=@bb/host-daemon-contract`

### DIFF
--- a/apps/host-daemon/src/provider-usage.ts
+++ b/apps/host-daemon/src/provider-usage.ts
@@ -107,11 +107,13 @@ function codexPlanLabel(planType: string | null | undefined): string | null {
 
 function codexWindow(
   window: z.infer<typeof codexUsageWindowSchema> | null | undefined,
-  label: string,
+  fallbackLabel: string,
 ): ProviderUsageWindow | null {
   if (!window) {
     return null;
   }
+  const label =
+    window.limit_window_seconds === 604_800 ? "Weekly limit" : fallbackLabel;
   return {
     label,
     usedPercent: clampPercent(window.used_percent),
@@ -251,10 +253,26 @@ const claudeUsageWindowSchema = z.object({
   resets_at: z.string().nullish(),
 });
 
+const claudeScopedUsageLimitSchema = z
+  .object({
+    kind: z.string(),
+    scope: z
+      .object({
+        model: z
+          .object({ display_name: z.string().trim().min(1).nullish() })
+          .nullish(),
+      })
+      .nullish(),
+    percent: z.number().nullish(),
+    resets_at: z.string().nullish(),
+  })
+  .passthrough();
+
 const claudeUsageResponseSchema = z
   .object({
     five_hour: claudeUsageWindowSchema.nullish(),
     seven_day: claudeUsageWindowSchema.nullish(),
+    limits: z.array(claudeScopedUsageLimitSchema).nullish(),
   })
   .passthrough();
 
@@ -359,6 +377,30 @@ function claudeWindow(
   };
 }
 
+function claudeScopedWindows(
+  limits: z.infer<typeof claudeScopedUsageLimitSchema>[] | null | undefined,
+): ProviderUsageWindow[] {
+  // `limits` repeats the aggregate session/week rows and adds model buckets.
+  // Only the model-scoped weekly rows are additive to the legacy top-level data.
+  return (limits ?? []).flatMap((limit) => {
+    const label = limit.scope?.model?.display_name;
+    if (
+      limit.kind !== "weekly_scoped" ||
+      label == null ||
+      limit.percent == null
+    ) {
+      return [];
+    }
+    return [
+      {
+        label,
+        usedPercent: clampPercent(limit.percent),
+        resetsAt: normalizeIsoTimestamp(limit.resets_at),
+      },
+    ];
+  });
+}
+
 function normalizeClaudeUsage(
   raw: unknown,
   credentials: ClaudeCredentials,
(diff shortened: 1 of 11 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I will review this pull request against the configured security, quality, performance, architecture, and test rules.

The stack contains only PR #1024. I skipped no pull requests.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Security review complete. I found no security vulnerability.

The new provider label reaches React as text. React escapes the label. Percentage values stay between zero and 100. The change does not alter credential access or token output. The protocol version increase protects the changed daemon result meaning.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Code quality review complete. I found one high-risk robustness issue and two medium label risks.

One invalid Claude scoped limit makes the parser reject all valid Claude usage rows. The Codex fallback can give two different windows the same weekly label. Claude rows can also share a model label when their surfaces differ.

I will include verified findings in the final review.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Performance review complete. I found no new material hot-path cost.

The new Claude array has no size limit. A large provider response can increase the daemon payload and the settings row count. The existing provider fetch path also has no daemon cache, but this pull request does not add that behavior.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

End-to-end review complete. I started the development app at commit `1d334f5dc`.

The Usage limits page showed the Codex weekly limit. It also showed the Claude Fable scoped limit. Both values remained after a manual reload. The browser reported no warning or error during that reload.

apps/host-daemon/src/provider-usage.ts
```diff
@@ -251,10 +253,26 @@ const claudeUsageWindowSchema = z.object({
...
+    percent: z.number().nullish(),
+    resets_at: z.string().nullish(),
+  })
+  .passthrough();
+
 const claudeUsageResponseSchema = z
   .object({
     five_hour: claudeUsageWindowSchema.nullish(),
     seven_day: claudeUsageWindowSchema.nullish(),
+    limits: z.array(claudeScopedUsageLimitSchema).nullish(),
```
SawyerHood: 🚨 `slopcop/review` — An invalid scoped row hides valid aggregate usage

The strict array parser rejects the full response if one optional limit item has an unexpected shape. `normalizeClaudeUsage` then drops valid `five_hour` and `seven_day` rows. Parse each scoped item separately and skip invalid items. Add a test that keeps the aggregate rows after one invalid scoped item.

apps/host-daemon/src/provider-usage.ts
```diff
@@ -359,6 +377,30 @@ function claudeWindow(
...
+    if (
+      limit.kind !== "weekly_scoped" ||
+      label == null ||
+      limit.percent == null
+    ) {
+      return [];
+    }
+    return [
+      {
+        label,
```
SawyerHood: 🚨 `slopcop/review` — Preserve the surface scope or prevent duplicate model rows

The payload includes `scope.surface`, but this code uses only `model.display_name`. Two surfaces for the same model will create the same label. The UI uses that label as the React key and can merge the rows. Include the surface in a stable identity, or reject scoped rows that include a surface.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Review result: I found two correctness risks.

- An invalid optional Claude scoped row makes the parser drop valid session and weekly rows.
- Two Claude surfaces for one model can create duplicate labels and duplicate React keys.

I found no security vulnerability. I found no material performance regression. The new transform has linear cost.

The architecture scan found no duplicate provider normalizer. The host daemon remains the correct owner for provider translation. The protocol increase to version 71 matches the changed wire meaning.

Validation passed. The host daemon passed 471 tests. The contract package passed 49 tests. Both focused type checks passed. The development Usage limits page showed the Codex weekly row and the Claude Fable row after a manual reload.

The stack contains only PR #1024.

apps/host-daemon/src/provider-usage.ts
ymichael: Addressed in 05eceb160. Each optional `limits[]` entry now recovers independently to `null`, so a malformed scoped row is skipped without invalidating the aggregate session/weekly data. Added a regression test that preserves both aggregate rows and a valid Fable row alongside a malformed item.

apps/host-daemon/src/provider-usage.ts
ymichael: Addressed in 05eceb160. The boundary now accepts only absent/null surface scope, skips unsupported surface-specific rows independently, and deduplicates model labels case-insensitively before they reach React. Added coverage for both a non-null surface scope and duplicate Fable labels.

PR #1029, opened 2026-08-05
Support .worktreeinclude files for new worktrees

SawyerHood:
## Summary

A new worktree checks out tracked files only, so a local `.env` or a credentials directory stays behind in the source checkout. Today the only escape hatch is to hand-write the copy inside `.bb-env-setup.sh`.

This adds a repo-root `.worktreeinclude` file that lists what a worktree needs, matching the semantics Claude Code already ships:

```gitignore
.env
.env.*
!.env.example
certs/
```

bb copies every matching **untracked** file from the source checkout after `git worktree add` and before `.bb-env-setup.sh`, so the setup script can read the copied files.

## Implementation

- `packages/host-workspace/src/worktree-include.ts` — `copyWorktreeIncludeFiles`. Matching runs through `git ls-files --others --ignored --exclude-from=.worktreeinclude -z`. Passing the include file as the only exclude source makes git's own matcher decide every pattern, so directory patterns, `**`, and `!` negation behave exactly as in `.gitignore` — and the feature needs no new dependency.
- `packages/host-workspace/src/provisioning.ts` — `copyIncludedFiles` runs inside `createWorktree` and writes each copy and each skip into the provisioning transcript.
- Files only. bb skips a symlinked source instead of copying its target, and refuses any destination that leaves the worktree through a committed symlink.
- Nothing here fails provisioning. A pattern that matches nothing, an unreadable file, or a failed copy is reported and the thread still starts.

Two deliberate differences from Claude Code: bb accepts any untracked file rather than only gitignored ones, because listing a path is already explicit intent; and bb reports into the provisioning transcript instead of a debug log, so the result is visible in the app.

## Tests

6 new tests in `packages/host-workspace/test/worktree-include.test.ts` cover pattern matching, `!` negation, symlink refusal, comment-only files, ordering against the setup script, and the missing-file case.

- `pnpm exec turbo run test --filter=@bb/host-workspace --force` — 178 passed
- `pnpm exec turbo run typecheck --filter=@bb/domain --filter=@bb/host-workspace --filter=@bb/templates --filter=@bb/server` — passed

## Docs

`docs/worktrees.md`, `docs/platform-support.md`, the `bb guide environments` template (regenerated), and the `bb-cli` skill.

## Notes

No `HOST_DAEMON_PROTOCOL_VERSION` bump: this adds no field to any session payload, WebSocket message, or host RPC command.

Large directories such as `node_modules` are copied file by file, which is slow. The docs point users at `.bb-env-setup.sh` for dependency installation instead.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Stack: PR #1029 only. The second stack check found no later pull request. I skipped no pull requests.

I will run security, code quality, architecture, performance, and end-to-end checks.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Security review completed.

The review found a high-impact destination symlink defect at `packages/host-workspace/src/worktree-include.ts:136`. The real `createWorktree` path followed a tracked destination symlink and wrote the copied secret outside the worktree.

The same copy step can replace a tracked destination file when the source checkout and target base branch differ. I will include the exact fix and evidence in the final review.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Performance review completed.

The matcher runs `git ls-files` without a pathspec. It walks the complete untracked tree for every new worktree, even when the include file names one small file. A 3.3 GB checkout produced 109,768 candidate paths and took 0.48–0.87 seconds with a warm cache.

Large result sets also create unbounded transcript text and can exceed the existing 16 MB Git output buffer. I will include the actionable performance finding in the final review.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Code-quality and architecture review completed.

The review confirmed the destination symlink and tracked-file overwrite defect. It also found three contract gaps: Git listing failures become a false zero-copy success, transcript output has no size limit, and cancellation does not stop the file-copy loop.

The repository scan found no substantial duplicate implementation. The new helper is the correct ownership boundary, but it needs the safety and failure controls above.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

End-to-end review completed.

This change has no browser route, so a dev-browser test cannot exercise it. I tested the real host boundary instead.

Turbo ran the `@bb/host-workspace` test and typecheck tasks. All 178 tests passed, including the six worktree-include tests. The integration test created a real Git worktree and verified that the setup script received the copied file.

A separate real `createWorktree` fixture reproduced the destination symlink escape. The passing suite does not cover that target-side case.

packages/host-workspace/src/worktree-include.ts
```diff
@@ -0,0 +1,148 @@
...
+    return {
+      ran: true,
+      copied: [],
+      skipped: [`${args.targetPath}: ${describeError(error)}`],
+    };
+  }
+
+  const copied: string[] = [];
+  const skipped: string[] = [];
+  for (const relativePath of relativePaths) {
```
SawyerHood, line 164: 🚨 `slopcop/review` — P2: Stop the copy loop after cancellation

The abort signal controls the Git command only. A large copy continues after cancellation, and cleanup later deletes all copied data.

Check the signal during each loop step. Throw the existing `provision_cancelled` error so the caller keeps the correct behavior.

packages/host-workspace/src/worktree-include.ts
```diff
@@ -0,0 +1,148 @@
...
+ * patterns. `--others` limits the walk to untracked paths, and `--ignored`
+ * with `--exclude-from` (and no `--exclude-standard`) makes the include file
+ * the only exclude source, so git's own gitignore matcher decides every
+ * pattern — including directory patterns, `**`, and negation.
+ */
+async function listMatchingFiles(
+  sourcePath: string,
+  signal: AbortSignal | undefined,
+): Promise<string[]> {
+  const result = await runGit(
```
SawyerHood, line 80: 🚨 `slopcop/review` — P2: Avoid the full untracked-tree scan for narrow patterns

This command has no pathspec. It walks the complete untracked tree before it returns one narrow match.

A 3.3 GB checkout had 109,768 paths. The warm scan took 0.48–0.87 seconds, compared with 0.023 seconds for a pathspec scan.

Use safe pathspec pruning for patterns that permit it. Add a time limit for the remaining full scans.

packages/host-workspace/src/worktree-include.ts
```diff
@@ -0,0 +1,148 @@
...
+      "ls-files",
+      "--others",
+      "--ignored",
+      `--exclude-from=${WORKTREE_INCLUDE_FILE_NAME}`,
+      "-z",
+    ],
+    { cwd: sourcePath, allowFailure: true, signal },
+  );
+  if (result.exitCode !== 0) {
+    return [];
```
SawyerHood: 🚨 `slopcop/review` — P2: Report Git listing failures

This return converts every Git error into a false zero-match result. It also hides the existing 16 MB output-buffer error.

Provisioning then reports a successful zero-file copy. Throw the error or return a clear skip reason with the Git error text.

The read helper also hides non-ENOENT file errors at line 47. Report those errors as the documentation promises.

packages/host-workspace/src/provisioning.ts
```diff
@@ -417,6 +432,66 @@ export async function createWorktree(
...
+  }
+
+  for (const skipped of result.skipped) {
+    emitOutput(args.onProgress, "worktree-include", `Skipped ${skipped}`);
+  }
+  if (result.copied.length > 0) {
+    emitOutput(
+      args.onProgress,
+      "worktree-include",
+      `Copied ${result.copied.length} file(s): ${result.copied.join(", ")}`,
```
SawyerHood: 🚨 `slopcop/review` — P2: Limit transcript path lists

A broad pattern can put thousands of paths in this one entry. The skipped-path loop also emits one entry for each file.

The daemon keeps and sends the full transcript. Limit each list, show the first paths, and report the remaining count.

packages/host-workspace/src/provisioning.ts
```diff
@@ -389,6 +398,12 @@ export async function createWorktree(
       keySuffix: "target",
       cwd: args.targetPath,
     });
+    await copyIncludedFiles({
```
SawyerHood, line 401: 🚨 `slopcop/review` — P1: Increase the host daemon protocol version

This behavior runs in the host daemon. An old enrolled daemon stays connected at protocol version 71 and silently omits this copy step.

The new transcript entries also cross the daemon connection. Increase `HOST_DAEMON_PROTOCOL_VERSION` and update its contract test.

packages/host-workspace/src/worktree-include.ts
```diff
@@ -0,0 +1,148 @@
...
+        skipped.push(`${relativePath}: symlink`);
+        continue;
+      }
+      await fs.mkdir(path.dirname(targetFile), { recursive: true });
+      const parentRealPath = await fs.realpath(path.dirname(targetFile));
+      if (!isInside(targetRealPath, parentRealPath)) {
+        skipped.push(`${relativePath}: destination escapes the worktree`);
+        continue;
+      }
+      await fs.copyFile(sourceFile, targetFile);
```
SawyerHood: 🚨 `slopcop/review` — P1: Do not follow an existing destination symlink

The parent check does not validate `targetFile`. `fs.copyFile` follows a destination symlink and writes outside the worktree.

I reproduced this through `createWorktree` with a tracked symlink on the base branch. It replaced a host file with the copied secret.

The same call replaces an existing tracked regular file. The documentation promises that bb never does this.

Use an exclusive or no-follow destination open. Skip every destination that exists, and add a target-side symlink test.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Review result: I found two P1 defects and four P2 defects.

Stack: PR #1029 only. I skipped no pull requests.

P1 findings:

- The copy follows a tracked destination symlink. A real `createWorktree` test wrote a copied secret outside the worktree.
- The change does not increase `HOST_DAEMON_PROTOCOL_VERSION`. Old enrolled daemons stay connected and silently omit the feature.

P2 findings:

- Git listing and pattern-file read errors become false zero-match results.
- Cancellation does not stop the file-copy loop.
- Transcript path output has no size limit.
- Narrow patterns still scan the complete untracked tree.

The repository scan found no substantial duplicate implementation. The helper has the correct package owner.

Turbo test and typecheck tasks passed for `@bb/host-workspace`. All 178 tests passed.

The feature has no browser route. I tested the real Git worktree and setup-script path instead.

packages/host-workspace/src/worktree-include.ts
SawyerHood: Not taking this one. The measured cost is 0.48–0.87s warm on a 3.3 GB checkout, and it is paid once, only when the repo actually commits a `.worktreeinclude`, inside an operation that already runs `git worktree add` over the same tree and then usually a dependency install. Against that baseline the scan is not the bottleneck.

Safe pathspec pruning is also narrower than it looks. A pattern only yields a usable path prefix when it is anchored to a directory segment. The dominant real-world config is `.env`, which is unanchored and matches at any depth, so it prunes to nothing — the optimization would miss exactly the case it is meant to help while adding a second matching path that can disagree with git's matcher. Handing every pattern to `git ls-files --exclude-from` is what makes the semantics correct by construction, and I would rather not trade that for a speedup that does not apply to the common case.

A time limit would change a slow copy into a silent partial copy, which is worse than slow for something that provisions secrets.

The other five findings are fixed in 5ccf70351.

PR #1030, opened 2026-08-05
Add first-run onboarding

SawyerHood:
Two-step modal on first launch: confirm which coding agent CLIs are installed and which plan pays for inference, then add the user's projects.

## Design

bb has no inference of its own, so screen one is a **confirmation**, not a chooser, whenever the machine already has a usable agent. Only a machine with nothing installed is asked to install something.

| Host state | Screen |
| --- | --- |
| ≥1 agent connected | Read-only list: brand mark, account, plan badge, Connected |
| CLIs present, none authed | Same list; amber rows expand to that agent's own login command with an "I've signed in" re-check |
| Nothing installed | Only the three CLIs with a managed installer, each with Install |

bb never drives another tool's login — it shows the command and re-checks, so credentials only ever pass through the agent itself. Escape and the close button both exit to the app; outside-click is blocked so a stray click cannot dismiss setup mid-install. Dismissing counts as done.

## Repo discovery

Walks `$HOME` and **stops the moment a directory contains `.git`** — everything below a repo root belongs to that repo, so it never enters `node_modules` or build trees.

| Approach | Dirs visited | Time |
| --- | --- | --- |
| `find -maxdepth 4 -name .git -prune` | 5,115 | ~100 ms |
| Stop-at-repo walk | **77** | **~5 ms** |

Cold page cache is the real cost, not CPU, so the walk is time-boxed with per-call deadlines (a dead NFS mount blocks `opendir` with no abort signal) and reports `truncated` rather than hanging.

Repos an agent has already run in rank first and arrive pre-checked, joined from Claude Code's `~/.claude.json` and Codex's app-server `thread/list`. Both are best-effort ranking hints, filtered exactly like the walk — must exist, not under a dot-directory, must contain `.git` — and never a discovery source. Unfiltered, Codex's history is mostly bb's own ephemeral worktrees; `gitInfo.originUrl` collapses many worktrees of one repo into a single signal.

## Plan detection survives a rate-limited usage API

Plan and account are read from local credentials *before* the usage HTTP call, so the `error` variant now carries them instead of discarding what bb already knows. Verified against a live 429: the row still shows `Max (20x)`. This also improves Settings → Usage limits, which can now name the plan while reporting usage unavailable.

## Showing once, and re-triggering

`onboardingCompletedAt` is the only gate. Migration 0085 backfills installs that already have a project, so upgrades are not re-prompted — which keeps a null timestamp meaning exactly one thing and lets **Settings → General → Setup guide** and **`bb settings replay-onboarding`** re-trigger it. Deliberately not conflated with "is bb set up": that is answered live by the agents query.

## Telemetry

`onboarding_started` / `step_completed` / `step_skipped` / `completed` / `dismissed`. "How many installs have no compatible CLI" is `onboarding_started` where `agent_state = none`; `detected_agent_count` gives the distribution. Properties are categorical or counts only, and the route validates against a zod union, so no path, project name, or email can reach analytics.

## Also

- Composer preselects a **connected** agent instead of always `providers[0]` (static catalog order, which always resolved to Codex — wrong for anyone who only has Claude Code).
- Native-picker vs in-app-dialog decision moves from `useQuickCreateProject` into `useLocalPathPicker.openPathEntry` so onboarding shares it; its tests moved with it.
- Pi's `openai-codex` default model is now `gpt-5.6-sol`.

**Protocol 72 → 73**: adds `workspace.discover_repos`, widens the provider-usage `error` variant.

## Tests

Full monorepo typecheck (57/57) and tests green on every touched package. New coverage for the walk's skip rules (stops at repo roots, skips dot-dirs and build dirs, recency filter, no-agent-history) and for the migration backfill on both an upgrade and a fresh install.

Manually exercised against a live dev instance: fresh install, upgrade backfill, re-trigger via Settings and CLI, dismissal persistence, folder picking into the selection list, and real plan data for all three providers.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am reviewing this change for security, code quality, performance, architecture, and end-to-end behavior.

The stack contains only #1030. No pull request requires a skip.

I will post a comment after each review phase.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

End-to-end phase complete.

I started the development server at the exact head SHA. The browser test covered the full two-step guide. It verified agent detection, project selection, project creation, reload persistence, Settings replay, dismissal, and dismissal persistence.

The browser reported no page errors or console errors during the tested actions. I found no end-to-end defect.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The review found actionable defects. The migration misses configured upgrades without an `app_settings` row. Unknown usage state can appear as connected. The install action can fail silently. Replay can add duplicate projects. Repository discovery misses linked worktrees because their `.git` marker is a file.

The review also found incomplete reuse and discoverability. The usage identity logic differs between onboarding and Usage limits. The new command does not update all required guide and skill surfaces. I will add exact findings to the final review.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Performance phase complete.

The four-second agent poll is too expensive. Each read runs CLI health checks, known-agent checks, package-registry requests, and provider usage requests. It continues on the projects step. The root host also runs a CLI health check after onboarding completion.

Repository discovery also has unbounded concurrency, uncapped enrichment work, late file-handle cleanup, and sequential work that can increase total latency.

I will include precise findings in the final review.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Security phase complete.

I found no trust-boundary, command-execution, path-escape, analytics privacy, or injection defect.

I found three low availability risks in repository discovery. A late `opendir` result can leak a directory handle. The walk has no concurrency limit. Repeated requests also start an uncached Codex process.

packages/db/drizzle/0085_glamorous_black_panther.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE `app_settings` ADD `onboarding_completed_at` text;--> statement-breakpoint
+-- Existing installs predate first-run onboarding, so stamp them as already
+-- onboarded rather than showing a setup flow to someone mid-project. Having a
+-- real (non-personal) project is the signal: it can only exist if the user has
+-- already set bb up. Doing this here, once, rather than in the client keeps
+-- `onboarding_completed_at IS NULL` meaning exactly one thing afterwards —
+-- "show the flow" — so Settings can re-trigger it by clearing the column.
+UPDATE `app_settings`
```
SawyerHood, line 15: 🚨 `slopcop/review` — This backfill misses configured upgrades without an app settings row.

The app creates this row only after a settings update. An existing user can have projects and default settings without this row. This UPDATE then changes no row. The default timestamp stays null, so the first-run guide opens after the upgrade.

Use an INSERT with an ON CONFLICT update. Add a migration test for a project database without an app settings row.

apps/host-daemon/src/command-handlers/discover-repos.ts
```diff
@@ -0,0 +1,417 @@
...
+    if (handle === null) {
+      if (Date.now() > deadline) truncated = true;
+      return;
+    }
+
+    const children: string[] = [];
+    let isRepo = false;
+    try {
+      for await (const entry of handle) {
+        if (!entry.isDirectory()) continue;
```
SawyerHood: 🚨 `slopcop/review` — This check misses linked worktrees and submodules.

Their .git marker is a file, not a directory. This early continue prevents repo detection. The walk then enters the complete worktree and returns no project.

Check the .git name before the directory test. Use the marker file time when HEAD is not a direct path.

apps/app/src/components/onboarding/OnboardingHost.tsx
```diff
@@ -0,0 +1,189 @@
...
+      // remote, not their path, so the remote is the join key.
+      const existingRemotes = new Set(
+        (projects ?? [])
+          .map((project) => project.gitRemoteUrl)
+          .filter((remote): remote is string => remote !== null),
+      );
+      // Sequential: project creation touches the host workspace, and a burst of
+      // parallel creates would race on the same daemon.
+      for (const repo of repos) {
+        if (repo.originUrl !== null && existingRemotes.has(repo.originUrl)) {
```
SawyerHood: 🚨 `slopcop/review` — Replay can create duplicate projects.

This check compares raw remote strings only. SSH and HTTPS forms for the same repository do not match. A local repository without a remote never matches. The replay command then creates another project for the same host path.

Compare host paths and normalized remote URLs before project creation.

apps/server/src/services/system/onboarding.ts
```diff
@@ -0,0 +1,238 @@
...
+      return "connected";
+    case "unauthenticated":
+      return "unauthenticated";
+    case "expired":
+      return "expired";
+    case "not_installed":
+      // The usage probe disagrees with the CLI health check (different lookup
+      // paths). Trust the health check, which is what install acts on.
+      return "unauthenticated";
+    default:
```
SawyerHood, line 92: 🚨 `slopcop/review` — A missing usage result becomes connected.

The service converts a failed provider.usage RPC to null. This default branch then marks every installed CLI as connected. The guide can continue and telemetry records a connected agent without proof.

Handle the error variant explicitly. Keep an absent result as unknown or omit the started event.

apps/host-daemon/src/command-handlers/discover-repos.ts
```diff
@@ -0,0 +1,417 @@
...
+async function withDeadline<T>(
+  operation: Promise<T>,
+  deadline: number,
+): Promise<T | null> {
+  const remaining = deadline - Date.now();
+  if (remaining <= 0) return null;
+  let timer: NodeJS.Timeout | undefined;
+  try {
+    return await Promise.race([
+      operation,
```
SawyerHood: 🚨 `slopcop/review` — A late directory result can leak its file handle.

When the timer wins, opendir still runs. A later result has no owner and no close call. Slow network directories can leak handles on each discovery request.

Close a late handle after the race loses. Keep that cleanup specific to open directory operations.

apps/app/src/hooks/queries/system-queries.ts
```diff
@@ -209,6 +215,35 @@ export function useHostProviderCliStatus({
...
+) {
+  return useQuery<OnboardingAgentOverview>({
+    queryKey: onboardingAgentsQueryKey(),
+    queryFn: ({ signal }) => sdk.system.onboardingAgents({ signal }),
+    enabled: options.enabled ?? true,
+    // Onboarding polls so a terminal install/login lands without a refresh.
+    // Other readers (the composer's provider default) just want one answer.
+    ...(options.poll === false
+      ? { staleTime: 60_000 }
+      : { refetchInterval: 4000 }),
```
SawyerHood: 🚨 `slopcop/review` — This four-second poll runs expensive host and network checks.

Each read runs provider health checks, package-registry requests, known-agent checks, and up to three provider usage requests. The poll also continues on the projects step. A two-minute pause can create hundreds of processes and many external requests.

Use a slower interval. Enable it only on the agent step. Prefer a small, cheap status check for the poll.

apps/app/src/components/onboarding/OnboardingHost.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const installRunner = useProviderCliInstallRunner();
+  // Stamped in an effect rather than during render: `Date.now()` in a render
+  // body is impure and would drift on every re-render.
+  const startedAt = useRef<number | null>(null);
+
+  const settings = configQuery.data?.generalSettings;
+  const primaryHostId = primaryHost?.id ?? null;
+  const cliStatusQuery = useHostProviderCliStatus({
+    hostId: primaryHostId,
+    enabled: primaryHostId !== null,
```
SawyerHood: 🚨 `slopcop/review` — This host check runs after onboarding completion.

OnboardingHost stays at the app root. This condition starts the full provider CLI health check on each app start when a primary host exists. Completed users still pay for CLI and package-registry work.

Enable this query only when the guide must show.

apps/host-daemon/src/command-handlers/discover-repos.ts
```diff
@@ -0,0 +1,417 @@
...
+      const head = await withDeadline(
+        stat(join(dir, ".git", "HEAD")),
+        deadline,
+      );
+      lastActivityMs = head?.mtimeMs ?? 0;
+      repos.push({ path: dir, lastActivityMs });
+      return;
+    }
+
+    await Promise.all(
```
SawyerHood: 🚨 `slopcop/review` — This recursion has no concurrency limit.

A wide home directory can start hundreds or thousands of opendir calls together. The daemon can reach file limits and affect active agent work. Errors disappear as partial results without a truncated flag.

Use a small worker pool and set truncated when a resource limit stops the search.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Review result: I found eight inline defects. I found no security defect.

The most important findings are:

- The migration misses upgrades without an app settings row.
- The four-second poll creates excessive process and network work.
- Completed users still pay for a provider CLI health check on app start.
- A failed usage probe can appear as a connected agent.
- Onboarding replay can create duplicate projects.
- Repository discovery misses linked worktrees and submodules.
- Repository discovery has unbounded directory concurrency.
- A late directory result can leak a file handle.

Additional issues remain. Usage limits ignores plan data on the new error variant. An unavailable install action can fail without feedback. The UI ignores the truncated discovery field. The new CLI command misses required guide and built-in skill updates.

Validation passed for nine package typechecks. Targeted tests passed for the app, database migration, host daemon, SDK, and contracts. All current GitHub checks pass.

The browser test passed the complete first-run flow. It covered detection, project selection, persistence, replay, and dismissal. It found no console or page errors.

PR #1031, opened 2026-08-05
Add keyboard shortcuts to cycle the model and reasoning level

SawyerHood:
## Summary

Two new keyboard shortcuts change the composer's model and reasoning level without opening the picker.

| Chord | Action |
| --- | --- |
| `Alt+M` | Select the next model of the composer's provider |
| `Alt+T` | Select the next reasoning level of the selected model |
| `Mod+Shift+M` | Open the picker (unchanged) |

Both chords wrap around at the end of the list. Both share the `when` conditions of `modelPicker.toggle`, so they act on the same picker that chord acts on: the focused pane, and the composer that holds the caret. Both appear in Settings → Keyboard and accept an override.

Alt is unused elsewhere in bb, in the browser, and in both desktop menus. These chords shadow nothing, and nothing shadows them. A test asserts that no other default binding uses Alt.

`Ctrl+P` and `Shift+Tab` were the first candidates. Both conflict: `Ctrl+P` is `file.quickOpen` (`Mod+P`) off macOS, and `Shift+Tab` is reverse tab navigation. Alt avoids both.

## macOS Option composition

macOS composes `Option+<letter>` into a different character. `Option+M` reports `key` as `"µ"`, not `"m"`. The matcher compared only `key`, so an Alt binding could never fire on a Mac.

This change adds the physical key (`KeyboardEvent.code`) to `AppShortcutInput`, and `normalizeAppShortcutInputKey` resolves Alt chords through it for the letter and digit rows. Chords without Alt still match by `key`, so a custom keyboard layout keeps its behavior.

The same fix corrects the Settings shortcut recorder. It stored the composed character, so a recorded Alt chord could never fire again. No default binding used Alt before now, so nothing exposed the fault.

`code` is required on `AppShortcutInput`, so the desktop `before-input-event` bridge passes it. Electron's input object carries `code`.

## Tests

- `turbo run typecheck` — passes for app, server, domain, templates, and desktop.
- `turbo run test --filter=@bb/app` — 2361 tests pass. New tests cover the rotation and scope helpers, Option+M matching, and Option+M recording.
- `turbo run test --filter=@bb/server` — passes, except a pre-existing failure in `test/internal/internal-skill-trees.test.ts` that also fails on unmodified `main`.
- `turbo run test --filter=@bb/desktop` — 209 tests pass. Four files fail to import with "Electron failed to install correctly". This is a local environment fault, not a code fault.
- `turbo run lint` — 0 errors.

Manual QA in the dev app confirms the behavior:

```
Alt+T:  High -> Extra High -> Max -> Low -> Medium -> High   (wraps)
Alt+M:  5.6-Luna -> 5.5 -> 5.4
```

A model switch reconciles the reasoning level. A move to a model without a Max level lowers the level to Extra High.

## Risks

- The daemon protocol is unchanged. Keybindings travel from the server to the app, not to the host daemon.
- `Alt+M` does nothing when focus sits outside the composer on a single-pane surface. This is not new. `Mod+Shift+M` does nothing in the same state, and the cycle chords inherit that rule.


SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

The stack contains only #1031. I did not skip a pull request.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Security review complete. I found no security defect.

The embedded browser cannot dispatch these commands. The physical-key match still requires the Alt modifier.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Performance review complete. I found no material performance defect.

The new match work costs a few microseconds per Alt chord. The cycle lookup scans only a small local option list.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

End-to-end review complete on the top pull request.

Alt+M cycled seven models and returned to the start. Alt+T cycled six reasoning levels and returned to the start.

I found one defect. Neither shortcut works while the model picker is open.

apps/server/src/services/system/app-keybindings.ts
```diff
@@ -179,6 +179,19 @@ export const DEFAULT_APP_KEYBINDINGS: AppKeybindings = [
...
     none: [],
   }),
+  // Rotate the composer's model and reasoning level without opening the picker,
+  // scoped exactly like `modelPicker.toggle` above. Alt is otherwise unused by
+  // bb, the browser, and both desktop menus, so these chords shadow nothing.
+  // macOS composes Option+<letter> into another character, so they match on the
+  // physical key — see `normalizeAppShortcutInputKey` in @bb/domain.
+  binding("modelPicker.cycleModel", "m", { alt: true }, {
+    all: ["mainSurface", "promptAvailable"],
+    none: ["modalOpen", "terminalFocus", "browserFocus"],
```
SawyerHood, line 189: 🚨 `slopcop/review` — The cycle shortcuts stop when the picker opens.

The model picker sets `modalOpen`. This condition blocks both new bindings before their handlers run.

I confirmed this failure in the development app. Alt+M and Alt+T work and wrap when the picker is closed.

Add scoped bindings for `modelPickerOpen`, like `modelPicker.toggle`, or permit this modal explicitly. This also prevents Option characters from reaching the search field.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

I found four additional issues:

- A chord can type an Option character when no next value exists.
- The new normalization can break non-US Alt overrides outside macOS.
- The cycle scope duplicates the existing toggle scope.
- The documented default shortcut table omits both new commands.

I found no other duplicate cycle implementation.

apps/app/src/components/pickers/modelPickerCycle.ts
```diff
@@ -0,0 +1,52 @@
...
+  isSplitPane: boolean;
+  /** Whether this composer is the pane's primary composer. */
+  isPrimaryComposer: boolean;
+  /** The caret sits inside THIS picker's composer. */
+  caretInThisComposer: boolean;
+  /** The caret sits inside a DIFFERENT composer of the same focused pane. */
+  caretInOtherComposerOfPane: boolean;
+}
+
+export function shouldModelPickerCycle(scope: ModelPickerCycleScope): boolean {
```
SawyerHood: 🚨 `slopcop/review` — Reuse the existing model-picker scope decision.

This interface and decision repeat `ModelPickerToggleInput` and `resolveModelPickerToggle`. Their scope rules can now change separately.

Extract one scope type. Then resolve a cycle through the same function with `open: false`.

packages/domain/src/app-keybindings.ts
```diff
@@ -131,10 +135,26 @@ const SHIFTED_KEY_BASES: Readonly<Record<string, string>> = {
...
+
 export function normalizeAppShortcutInputKey(input: AppShortcutInput): string {
   if (input.key === " " || input.key === "Spacebar") {
     return "Space";
   }
+  // macOS composes Option+<letter> into another character — Option+M reports
+  // key "µ" — so an Alt chord can never be matched by `key` there. The physical
+  // code is stable, so Alt chords resolve through it on every platform. Chords
+  // without Alt keep matching by `key`, which respects the user's layout.
+  if (input.altKey) {
```
SawyerHood: 🚨 `slopcop/review` — Preserve character-based Alt shortcuts on non-Mac keyboard layouts.

This branch uses the physical key on every platform. A Windows AZERTY Alt+A event has `key: "a"` and `code: "KeyQ"`.

Existing Alt+A overrides will stop after this change. New Alt defaults also move away from their displayed key.

Use `code` only when `key` is not a plain letter or digit. That condition still handles macOS Option characters.

apps/server/src/services/system/app-keybindings.ts
```diff
@@ -179,6 +179,19 @@ export const DEFAULT_APP_KEYBINDINGS: AppKeybindings = [
     all: ["mainSurface", "modelPickerOpen"],
     none: [],
   }),
+  // Rotate the composer's model and reasoning level without opening the picker,
+  // scoped exactly like `modelPicker.toggle` above. Alt is otherwise unused by
+  // bb, the browser, and both desktop menus, so these chords shadow nothing.
+  // macOS composes Option+<letter> into another character, so they match on the
+  // physical key — see `normalizeAppShortcutInputKey` in @bb/domain.
+  binding("modelPicker.cycleModel", "m", { alt: true }, {
```
SawyerHood, line 187: 🚨 `slopcop/review` — Update the documented default shortcut table.

`docs/configuration.md` lists the complete defaults, but it omits both new commands. The customization guide points users to that table.

Add the Next model and Next reasoning level rows with Alt+M and Alt+T.

apps/app/src/components/pickers/ModelReasoningPicker.tsx
```diff
@@ -557,29 +561,68 @@ export function ModelReasoningPicker({
...
+  // Both cycle chords rotate the COMMITTED provider's lists, never a previewed
+  // tab's, so the shortcut means the same thing whether the popover is open or
+  // shut. A preview in progress is dropped so the visible tab keeps matching the
+  // committed selection.
+  useAppCommandHandler(
+    "modelPicker.cycleModel",
+    ({ target }) => {
+      if (!shouldModelPickerCycle(resolveCommandScope(target))) return false;
+      const next = nextCycleValue(modelOptions, modelValue);
+      if (next === null) return false;
```
SawyerHood: 🚨 `slopcop/review` — Consume an in-scope chord when no next value exists.

A `false` result makes the command provider return before `preventDefault()`. On macOS, Option+M then inserts `µ` into the prompt.

Option+T has the same failure for a model without reasoning options. A single-entry list also causes this path.

Return `true` after the scope check. Apply a change only when `next` is not null.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Review complete for the single-pull-request stack: #1031. I did not skip a pull request.

I found five issues:

1. The cycle shortcuts do not work while the model picker is open.
2. An in-scope chord can insert an Option character when no next value exists.
3. Physical Alt matching can break non-US keyboard overrides outside macOS.
4. The new cycle scope duplicates the existing toggle scope.
5. The complete default shortcut table omits both new commands.

The security review found no security defect. The performance review found no material performance defect.

End-to-end checks passed for the closed picker. Alt+M cycled seven models and wrapped. Alt+T cycled six levels and wrapped.

All four changed package type checks passed. The 28 focused app, server, and desktop tests passed.

The full domain suite passed 125 tests. The full desktop suite passed 229 tests.

The full server suite passed 1,323 of 1,324 tests. The unrelated machine installer test also failed when I ran it alone.

I used comment status only. I did not approve the pull request or request changes.

PR #1019, opened 2026-08-04
Give every app header the seam border

SawyerHood:
## Summary

The bottom seam under a header was a per-call-site decoration, not part of the shared chrome. Only `ThreadDetailHeader` passed `border-b border-border-seam-vertical/60`. Plugin panel headers and the full-page `AppLayout` header passed no border, so their title bar sat flush against the body.

This moves the seam into `AppPageHeader` as a shared `HEADER_SEAM_CLASS` constant applied in the base class list, and removes the duplicate string from `ThreadDetailHeader`.

## Impact

`AppPageHeader` has three call sites. All three now carry the seam:

| Surface | Before | After |
|---|---|---|
| Split-pane plugin panels (Docs, Automations) | no seam | seam — the fix |
| Project, archived, settings, and tools routes | no seam | seam — new |
| Plugin panel routes with splits off | no seam | seam — new |
| Thread header | seam | seam — no change |
| Root and compose views | no header | no header |

The second row is a deliberate widening: those routes now match the thread header.

## Tests

- New `AppPageHeader.test.tsx` guards the seam for a header with no `className`, and for one that passes its own classes.
- Removed the old seam assertion from `ThreadDetailHeader.test.tsx`. It mocked `AppPageHeader`, so it could no longer see the real class.
- `turbo run typecheck --filter=@bb/app` passed.
- `turbo run test --filter=@bb/app` passed: 312 files, 2338 tests.
- `turbo run lint --filter=@bb/app`: 0 errors, no warnings in the changed files.

## QA

Checked in the dev app: the plugin panel and settings headers gained the seam, the thread header kept its single seam with no doubling, and the root view is unaffected.

## Risks

There is no opt-out prop. A future seamless header must pass `border-b-0` in `className`. I chose this over a `hideSeam` prop because no call site needs it today.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #991, opened 2026-08-04
Paginate skill detail content

brsbl:
## Summary
- constrain skill detail content to a viewport-relative viewer
- reuse the shared extension pagination control for previous/next navigation
- recalculate pages across content reflow, window resizing, and native full-screen transitions

## Verification
- pnpm exec turbo run test --filter=@bb/app -- --run src/components/tools/detail-page-recipes.test.tsx src/components/ui/resource-pagination.test.tsx (41 passed)
- pnpm exec turbo run typecheck --filter=@bb/app
- git diff --check
- branch dev app: verified first/last pages, narrow window reflow, and native full-screen reflow; focused test covers short content without pagination

## Stack
Layer 1, based on main. Follow-on fixes should branch from this head and target this branch.

PR #1036, opened 2026-08-05
Restore plugin pane chrome and the Docs plugin sidebar

SawyerHood:
## Summary

Four regressions left plugin panes with broken chrome. Three came from #991; one predates it. Each was reproduced in the dev app before the fix.

**Docs sidebar looked like the host's shared sidebar.** #991 restyled it to `bg-sidebar` and painted a fake background up into the app header. #1019 then gave every header its seam, which that background cut short. The sidebar sits below the header again, with the seam edge to edge and the plugin's own `bg-muted/20` tint.

**The Docs header toggle did nothing.** #991 deleted the shared sidebar store once its mounts hit zero. The header and the sidebar live in separate React subtrees and find each other only through that module-level map, so a deleted entry left each subtree on a different store object: two toggles rendered, and the header one was dead (measured 288px to 288px on click). Entries now live for the session. A reopened pane still starts clean, because the view resets the state on its first mount.

**Plugin panes lost full screen, move-to-side, and their corner reserve.** #991 excluded `plugin-panel` content from all three. Without the reserve, the host's pinned panel toggle landed on Close pane whenever a plugin pane held the workspace corner — measured overlap of 22 of 28px, with the toggle winning the click. Position alone decides the reserve now, and plugin panes carry the same pane controls as every other pane.

**A pane with no right panel lost its toggle and placeholder.** #736 shipped both. #962 disabled the button and deleted the placeholder; #991 removed the control outright. Both are restored: the toggle always works (including `Ctrl+J`), and a pane that publishes no panel shows `This pane has no right panel.` in a resizable panel whose width persists. Since that empty state has no chrome of its own, the corner button stays visible while it is open — otherwise no button could close it. The host exposes this as `pinsCornerToggle`, and pane headers reserve the corner from the same flag.

Two smaller fixes in the same area:

- The reserve no longer leaves an empty 28px slot when the host hides its toggle. This showed as dead space beside Exit Full Screen and predates #991 (it came in with #781).
- The pinned toggle moved from `right-2.5` to `right-4`, so it shares the pane header's `px-4` action axis. Measured after: Full Screen 1652–1680, Close pane 1684–1712, toggle 1716–1744 — an even 4px rhythm, with the toggle landing exactly in the reserved slot.

## Tests

- `turbo run test --filter=@bb/app`: 2364 pass (315 files).
- `turbo run test --filter=bb-plugin-simple-notes`: 59 pass.
- `turbo run typecheck` passes for `@bb/app` and `bb-plugin-simple-notes`. Lint reports 0 errors.
- New regression tests: the corner reserve for a plugin pane at the top right (verified failing against the old condition), the reserve dropping while a full-screen pane hides the toggle, and the empty panel state with a live toggle.
- Updated the tests from #962 and #991 that locked in the removed controls.

## QA

Checked in the branch dev app with the Docs, Tasks, and GitHub plugins installed: the header seam, a single working Docs toggle, plugin full screen and restore, no button collision, the empty-state panel opening and collapsing to 0px, and even header spacing.

## Risks

Restoring move-to-side for plugin panes was not requested directly. It came out in the same #991 block as full screen, so I put it back with the rest. Say the word if it should stay off.

### DIFF
--- a/apps/app/src/views/thread-detail/SplitWorkspaceSecondaryPanelHost.tsx
+++ b/apps/app/src/views/thread-detail/SplitWorkspaceSecondaryPanelHost.tsx
@@ -6,13 +6,15 @@ import {
   useState,
   type ReactNode,
 } from "react";
-import { useAtomValue } from "jotai";
+import { useAtomValue, useSetAtom } from "jotai";
 import {
   Panel,
   PanelGroup,
+  PanelResizeHandle,
   type ImperativePanelGroupHandle,
 } from "react-resizable-panels";
 import { Button } from "@bb/shared-ui/button";
+import { EmptyStatePanel } from "@bb/shared-ui/empty-state";
 import { Icon } from "@bb/shared-ui/icon";
 import { cn } from "@bb/shared-ui/lib/utils";
 import { HEADER_ICON_BUTTON_CLASS } from "@/components/layout/AppPageHeader";
@@ -23,11 +25,16 @@ import {
 } from "@/components/commands/AppCommandProvider";
 import { secondaryPanelWidthPercentAtom } from "@/components/secondary-panel/threadSecondaryPanelAtoms";
 import {
+  THREAD_SECONDARY_PANEL_MAX_SIZE_PERCENT,
+  THREAD_SECONDARY_PANEL_MIN_SIZE_PERCENT,
+} from "@/components/secondary-panel/ThreadSecondaryPanel";
+import {
   SecondaryPanelHostLayoutContext,
   type SecondaryPanelHostLayout,
 } from "@/components/secondary-panel/SecondaryPanelHostLayoutContext";
 import {
   PANEL_COLLAPSE_TRANSITION_CLASS,
+  PANEL_RESIZE_HIT_AREA_MARGINS,
 } from "@/components/secondary-panel/panelTransitionTokens";
 import { MACOS_APP_REGION_NO_DRAG_CLASS } from "@/lib/bb-desktop";
 import { PluginComposerHostProvider } from "@/components/plugin/plugin-composer-host";
@@ -62,7 +69,7 @@ export function SplitWorkspaceSecondaryPanelHost({
   // Null until the first pane publishes, so entering a split adopts the
   // focused content's persisted state.
   const [isPanelVisible, setIsPanelVisible] = useState<boolean | null>(null);
-  const isOpen = model !== null && (isPanelVisible ?? model.isOpen);
+  const isOpen = isPanelVisible ?? model?.isOpen ?? false;
   const lastTargetRef = useRef<{
     paneId: string;
     contentKey: string;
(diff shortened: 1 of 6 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The stack contains PR #1036 only. I will post a comment after each review phase.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Security phase complete.

I found no security issues. The new controls use host-owned pane IDs. The empty panel adds no untrusted HTML, URL, or filesystem input.

I also checked plugin trust boundaries, composer scope, store keys, and denial-of-service risks. `git diff --check` passed.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Performance phase complete.

I found one low-severity issue. `notesSidebarStores` keeps inactive pane and vault entries for the full application session.

The map can grow across many Docs vaults. A bounded cache or safe delayed cleanup would prevent this long-session heap growth.

I found no other performance issue. All GitHub checks passed.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I found one blocking lifecycle defect. An empty panel starts at zero size for a focused plugin pane.

The panel can call `onCollapse` during its first mount. That call changes the visibility sentinel from null to false.

A later thread pane then closes its persisted-open panel. The current panel mock does not reproduce this library callback.

The code should ignore collapse until the empty panel first reports a positive size. A regression test should cover the plugin-first path.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

End-to-end phase complete on the top PR.

I installed the Docs plugin in an isolated development instance. I opened Docs beside a new-thread pane.

Both panes showed Full Screen and Close pane controls. The Docs sidebar expanded to 288 pixels and showed its action toolbar.

The plugin pane entered and left full-screen mode. The host toggle opened and closed the empty right-panel state.

The visual check showed no control overlap. This path passed.

apps/app/src/views/thread-detail/SplitWorkspaceSecondaryPanelHost.tsx
```diff
@@ -248,12 +269,60 @@ export function SplitWorkspaceSecondaryPanelHost({
...
+                aria-label="Resize right panel"
+              />
+              <Panel
+                id="split-workspace-empty-secondary-panel"
+                collapsible
+                collapsedSize={0}
+                defaultSize={isOpen ? panelWidthPercent : 0}
+                minSize={THREAD_SECONDARY_PANEL_MIN_SIZE_PERCENT}
+                maxSize={THREAD_SECONDARY_PANEL_MAX_SIZE_PERCENT}
+                onCollapse={handleEmptyPanelCollapse}
```
SawyerHood, line 306: 🚨 `slopcop/review` — The initial collapse overwrites the visibility sentinel.

This panel starts at zero size when a plugin pane has focus. React Resizable Panels calls `onCollapse` for that initial size.

The callback changes `isPanelVisible` from null to false. A later persisted-open thread then closes instead of adopting its saved state.

Ignore collapse until this panel first reports a positive size. Add a test that calls `onCollapse` for the initial zero size.

official-plugins/docs/app.tsx
```diff
@@ -1261,6 +1256,11 @@ interface NotesSidebarStore {
   listeners: Set<() => void>;
 }
 
+// The header toggle and the sidebar view live in separate React subtrees and
+// find each other only through this map. Entries stay for the session: a
+// deleted entry lets one subtree keep the old store while the other creates a
+// new one, which silently breaks the header toggle. A reopened pane starts
+// clean anyway, because the view resets the state on its first mount.
 const notesSidebarStores = new Map<string, NotesSidebarStore>();
```
SawyerHood, line 1264: 🚨 `slopcop/review` — This map keeps inactive stores for the full application session.

Each distinct pane and vault key adds a state object and listener set. Unmounting both React subtrees no longer removes the entry.

Use safe delayed cleanup or a bounded cache. Cancel cleanup when either subtree mounts again.

apps/app/src/views/thread-detail/SplitWorkspaceSecondaryPanelHost.tsx
```diff
@@ -141,36 +148,47 @@ export function SplitWorkspaceSecondaryPanelHost({
...
+    setPanelWidthPercent(lastEmptyPanelSizeRef.current);
+  };
+  const handleEmptyPanelCollapse = () => {
+    setIsPanelVisible(false);
+  };
+
+  const toggleLabel = isOpen ? "Hide right panel" : "Show right panel";
+  // An open pane panel carries the toggle in its own chrome, and a full-screen
+  // pane hides it. The empty state has no chrome, so it keeps the corner.
+  const pinsCornerToggle = !isPaneMaximized && !(isOpen && model !== null);
```
SawyerHood: 🚨 `slopcop/review` — An open empty panel leaves a blank header action slot.

The empty panel contains the visible toggle. The separate right-edge plugin pane still reserves 28 pixels for that toggle.

Reserve this slot only when the closed panel places the toggle over the main pane. Add an open-empty-panel test.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Review complete for the one-PR stack: #1036.

I found one blocking defect and two low-severity defects.

1. The initial empty-panel collapse clears the null visibility sentinel.
2. The Docs sidebar store map grows for the full application session.
3. An open empty panel leaves a blank 28-pixel plugin header slot.

The first defect can close a persisted-open thread panel after a plugin pane first gains focus.

The fix should ignore empty-panel collapse until that panel first reports a positive size. A regression test must simulate the initial collapse callback.

I found no security issue.

The empty panel repeats collapse and resize behavior from `ThreadSecondaryPanel`. A shared lifecycle helper could prevent more behavior differences.

Local validation passed:

- The app test passed 38 tests.
- The Docs test passed 24 tests.
- Both package typechecks passed.
- All GitHub checks passed.
- The browser test passed the Docs sidebar, split chrome, full-screen, and empty-panel flows.

The GPT-5.6 final gate returned `REQUEST CHANGES`. I posted a comment-only review and did not use GitHub approval controls.

official-plugins/docs/app.tsx
SawyerHood: Thanks — I looked at this one and am leaving it as is. Deleting entries is what broke the toggle in the first place, so I would rather pay a small bound than reintroduce that failure mode.

The key is `${paneScope}:${vaultKey}`. Pane scopes come from live split-pane IDs, capped at 8 panes, and vault keys come from the user's own vault list. So the map tops out at a few dozen entries per session, each a small object plus an empty listener set. It grows only when the user opens new panes or vaults, not on a loop.

Delayed cleanup with cancellation would restore the race in a subtler form: the header and the sidebar mount in separate React subtrees with no ordering guarantee, so any window where the entry is gone can hand the two subtrees different store objects again. That is exactly the desync this PR fixes — a dead header toggle and a duplicate control, with no error to point at it.

If the map ever needs bounding, the safe shape is a cap on total entries with eviction of keys whose panes no longer exist in the split layout, checked against the live layout rather than against mount counts. That is worth doing when there is evidence of real growth; today it would add machinery and risk for a few dozen small objects.

PR #1037, opened 2026-08-05
Keep the provider prefix on Pi aggregator model IDs

### LINKED ISSUE #1033: Failed to resolve Pi model "deepseek/deepseek-v4-flash-0731"
Image

SawyerHood:
Fixes #1033.

## The bug

bb identifies a Pi model by `<provider>/<model id>`, but `toCanonicalPiModelId` skipped the prefix whenever the model id already contained a slash:

```ts
return modelId.includes("/") ? modelId : `${provider}/${modelId}`;
```

Aggregator providers name a model after the vendor that serves it. OpenRouter returns `provider: "openrouter"`, `id: "deepseek/deepseek-v4-flash-0731"`, so the canonical id collapsed to `deepseek/deepseek-v4-flash-0731`. `resolveConfiguredModel` then split it at the first slash and looked for `deepseek-v4-flash-0731` on the **direct DeepSeek** provider, which has only `deepseek-v4-flash` and `deepseek-v4-pro`. `thread.start` failed with `Failed to resolve Pi model`.

## The wider blast radius

The reported hard failure was the lucky case. When the parsed provider *does* have a model by that name, the request silently ran against the wrong vendor — different credentials, different billing, no error. OpenRouter's `openai/gpt-5.1-codex` resolved to the direct OpenAI provider.

Slash-carrying ids per provider in Pi's catalog: OpenRouter 273/274, Vercel AI Gateway 194/194, HuggingFace 50/50, NVIDIA 18/18, Together 16/16, Fireworks 16/16 (two slashes: `accounts/fireworks/models/…`), Cloudflare Workers AI 13/13, Groq 5/7.

Both configured aggregator defaults were also unreachable: `openrouter → openai/gpt-5.1-codex` never matched, and `vercel-ai-gateway → anthropic/claude-opus-4.8` misses the direct Anthropic catalog, which uses dashes.

## The fix

- **`model-list.ts`** — always add the prefix. Consumers split on the first slash only, which `sdk-session.ts` already did.
- **`sdk-session.ts`** — `resolveConfiguredModel` accepts both the canonical form and a bare provider-native id, so stored selections, project defaults, queued messages, and `bb thread spawn --model` keep working with no database migration. When both readings match different models, a provider the user has authenticated wins, so a bare aggregator id runs through the aggregator rather than an unreachable direct provider.
- **`adapter.ts`** — split the context-window lookup into per-provider and per-model-id maps. The single map let an aggregator raw id and a direct canonical id share a key. Pi's catalog holds 134 such pairs and some disagree: OpenRouter reports 1048575 tokens for `deepseek-v4-flash` where DeepSeek reports 1000000.
- **`HOST_DAEMON_PROTOCOL_VERSION` 73 → 74**, because the model strings on the wire change meaning. An enrolled daemon on 73 would otherwise return unprefixed ids.

## Tests

Five new cases in `adapter.test.ts` and `sdk-session.test.ts`. Reverting the one-line `toCanonicalPiModelId` change fails two of them:

```
AssertionError: expected [ …(3) ] to deeply equal [ …(3) ]   ← prefix dropped
AssertionError: expected 1000000 to be 1048575                ← context window from the wrong provider
```

`pnpm exec turbo run test --filter=@bb/agent-runtime` — 807 passed. Full-repo typecheck passes.

## Manual verification against real OpenRouter

Ran the dev app with an OpenRouter key. `bb provider models pi` returns 287 models, 280 of them prefixed `openrouter/…`, including the exact `openrouter/deepseek/deepseek-v4-flash-0731` from the issue.

Three threads, three real turns, all replied `ok`:

| Model string | Result |
|---|---|
| `openrouter/deepseek/deepseek-v4-flash-0731` | ok — the issue's failure |
| `deepseek/deepseek-v4-flash` (old bare form) | ok |
| `openrouter/openai/gpt-5.1-codex` | ok |

`~/.pi/agent/auth.json` held credentials for `openai-codex` only, with no direct `openai` or `deepseek` entry, so none of these could have reached a direct vendor. The recorded context window for the bare-id thread was 1048575 (OpenRouter's figure) rather than 1000000 (DeepSeek's), confirming the aggregator-preference path.

## Note for existing users

A thread that stored a prefix-free aggregator id will show the catalog default in the picker after this lands, because that exact string is no longer in the model list. The turn still runs correctly through the bare-id fallback, but the picker selection resets once. The alternative — emitting every aggregator id twice — adds ~273 duplicate rows to "More models" for every OpenRouter user.

### DIFF
--- a/packages/agent-runtime/src/pi/adapter.ts
+++ b/packages/agent-runtime/src/pi/adapter.ts
@@ -1607,15 +1607,17 @@ function resolvePiModelContextWindow(
 
   // Pi reports the provider and the provider-native model id separately, and an
   // aggregator model id such as "deepseek/deepseek-v4-flash" also names a
-  // direct provider's model. Match the pair first so the two do not swap.
+  // direct provider's model. A known provider therefore decides the answer on
+  // its own. Falling back to the id alone would hand a model another provider's
+  // window whenever the catalog lacks the pair, which happens for models the
+  // network refresh added and for custom models.
   const providerId = toOptionalString(lastAssistant?.provider);
   if (providerId) {
-    const contextWindow = modelContextWindowLookup.byCanonicalId.get(
-      toCanonicalPiModelId(providerId, modelId),
+    return (
+      modelContextWindowLookup.byCanonicalId.get(
+        toCanonicalPiModelId(providerId, modelId),
+      ) ?? null
     );
-    if (contextWindow !== undefined) {
-      return contextWindow;
-    }
   }
 
   return modelContextWindowLookup.byModelId.get(modelId) ?? null;
--- a/packages/agent-runtime/src/pi/adapter.ts
+++ b/packages/agent-runtime/src/pi/adapter.ts
@@ -613,7 +613,10 @@ function buildPiConfig(
   return Object.keys(config).length > 0 ? config : undefined;
 }
 
-type PiModelContextWindowLookup = ReadonlyMap<string, number>;
+interface PiModelContextWindowLookup {
+  byCanonicalId: ReadonlyMap<string, number>;
+  byModelId: ReadonlyMap<string, number>;
+}
 
 type PiModelContextWindowResolver = (
   lastAssistant: PiAssistantMessage | undefined,
@@ -622,25 +625,35 @@ type PiModelContextWindowResolver = (
 function buildPiModelContextWindowLookup(
   models: readonly PiContextWindowModel[],
 ): PiModelContextWindowLookup {
-  const lookup = new Map<string, number>();
+  const byCanonicalId = new Map<string, number>();
+  const byModelId = new Map<string, number>();
   for (const model of models) {
     const contextWindow = toPositiveNumber(model.contextWindow);
     if (contextWindow === undefined) {
       continue;
     }
-    const canonicalId = toCanonicalPiModelId(model.provider, model.id);
-    lookup.set(canonicalId, contextWindow);
-    if (model.id.includes("/")) {
-      lookup.set(model.id, contextWindow);
-    }
+    byCanonicalId.set(
+      toCanonicalPiModelId(model.provider, model.id),
+      contextWindow,
+    );
+    // Aggregator providers share model ids, so this map is ambiguous. It only
+    // serves messages that report no provider.
+    byModelId.set(model.id, contextWindow);
   }
-  return lookup;
+  return { byCanonicalId, byModelId };
 }
 
 function createPiModelContextWindowResolver(): PiModelContextWindowResolver {
   const models = getBuiltinProviders().flatMap((provider) =>
     getBuiltinModels(provider),
   );
+  return createPiModelContextWindowResolverFrom(models);
+}
+
+/** @internal Test seam: resolve against an explicit catalog. */
+export function createPiModelContextWindowResolverFrom(
+  models: readonly PiContextWindowModel[],
+): PiModelContextWindowResolver {
   const modelContextWindowLookup = buildPiModelContextWindowLookup(models);
   return (lastAssistant) =>
     resolvePiModelContextWindow(lastAssistant, modelContextWindowLookup);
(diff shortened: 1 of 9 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am reviewing this change for security, code quality, performance, architecture, duplication, and end-to-end behavior.

The stack contains only PR #1037: **Keep the provider prefix on Pi aggregator model IDs**.

I found no draft pull requests or untrusted authors in this stack.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Security phase complete.

I found one medium-risk issue. The resolver can silently change vendors when the selected provider lacks authentication.

For example, `anthropic/claude-sonnet-5` can route through OpenRouter. The Pi catalog has 78 similar canonical and raw ID collisions.

This route can send workspace data to a vendor that the user did not select. I will add line-specific details in the final review.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Quality and architecture phase complete.

The canonical ID fix and protocol bump are correct. However, I found one high-risk migration defect.

The app cannot match an existing bare aggregator ID after this change. Its recovery path selects the default model and persists that replacement.

I also found two provider-selection ambiguities and one context-window fallback defect. I will add the exact lines in the final review.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Performance phase complete.

I found no throughput, latency, or memory regression on a hot path. The new catalog scan averaged about 20 microseconds per session start.

The two lookup maps add about 50 KB per Pi process. The model-list prefix adds about 6 KB for an OpenRouter-only response.

The context-window fallback has a correctness defect, but it does not create a performance regression.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

End-to-end phase complete with one limitation.

I started the development app at the pull request head. The browser loaded the app, opened a new thread, and opened the Pi model tab.

The isolated QA data had no Pi credentials. Therefore, the Pi tab correctly showed no models, and I could not make an aggregator call.

The affected package tests passed: 856 tests across 44 files. Both affected package type checks also passed.

packages/agent-runtime/src/pi/bridge/sdk-session.ts
```diff
@@ -614,26 +614,54 @@ export class PiSdkSession {
...
   }
 
-  const provider = modelStr.slice(0, slashIdx);
-  const modelId = modelStr.slice(slashIdx + 1);
-  const model = modelRuntime.getModel(provider, modelId);
+  const bare = modelRuntime
+    .getModels()
+    .filter((candidate) => candidate.id === modelStr);
+  const model =
+    bare.find((candidate) =>
```
SawyerHood: 🚨 `slopcop/review` — Do not redirect a canonical ID to another provider.

If Anthropic lacks authentication, `anthropic/claude-sonnet-5` can select OpenRouter or Vercel here. Pi has 78 similar canonical and raw ID collisions.

This route can send workspace content and billing to a vendor that the user did not select. Slash-free IDs also use catalog order.

Keep an explicit provider authoritative. Accept a legacy raw ID only when the match is unique and clearly identified.

packages/agent-runtime/src/pi/adapter.ts
```diff
@@ -1592,17 +1605,20 @@ function resolvePiModelContextWindow(
...
+      toCanonicalPiModelId(providerId, modelId),
+    );
+    if (contextWindow !== undefined) {
+      return contextWindow;
+    }
   }
 
-  const canonicalId = toCanonicalPiModelId(providerId, modelId);
-  return modelContextWindowLookup.get(canonicalId) ?? null;
+  return modelContextWindowLookup.byModelId.get(modelId) ?? null;
```
SawyerHood, line 1623: 🚨 `slopcop/review` — Limit raw context lookup to messages without a provider.

The map comment says raw lookup only serves messages without a provider. This return also runs after a present provider misses canonical lookup.

A custom or newly refreshed model can then use another provider's context window. The catalog has 21 slash-free collisions with different values.

Return `null` when a present provider misses the canonical map.

packages/agent-runtime/src/pi/model-list.ts
```diff
@@ -88,11 +88,23 @@ export function buildPiAvailableModels(
...
+ *
+ * The model id keeps its own slashes, so consumers must split on the FIRST
+ * slash only.
+ */
 export function toCanonicalPiModelId(
   provider: string,
   modelId: string,
 ): string {
-  return modelId.includes("/") ? modelId : `${provider}/${modelId}`;
+  return `${provider}/${modelId}`;
```
SawyerHood, line 107: 🚨 `slopcop/review` — Preserve existing aggregator selections.

Changing every catalog ID leaves stored bare IDs unmatched in the app. The app then selects the catalog default.

Its recovery path marks that default as explicit. The server persists it when the user sends the next message.

Thus, one message can permanently change an existing thread's model and vendor. Map a stored bare ID to one unique canonical row first.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Review complete. I found three actionable defects.

- High: Existing aggregator threads lose their stored model. The app selects the default and persists that replacement after one message.

- High: A canonical model ID can silently route to another authenticated provider. This can change data access and billing.

- Medium: A canonical context-window miss falls back to an ambiguous raw ID. Custom and refreshed models can show a wrong context limit.

I found no hot-path performance regression. The canonical helper has no material duplicate, and the protocol version bump is correct.

An inverse canonical-ID parser could replace two manual first-slash parses. This refactor is optional.

All 856 affected package tests passed. Both affected package type checks passed.

The development app loaded and exposed the Pi picker. The isolated QA data had no Pi credentials, so I could not make an aggregator call.

I posted each defect on its exact changed line. I did not approve this pull request or request changes.

PR #1045, opened 2026-08-06
Restore the shared header for plugin panels on compact viewports

### LINKED ISSUE #1042: Mobile plugin panels render under the sidebar trigger
## Summary
On compact/mobile viewports, standalone `navPanel` plugin routes render their body under the fixed sidebar trigger. The trigger icon overlaps the plugin panel header/content.

This affects built-in Automations and custom plugins such as Helm Wiki, so it is not plugin-specific.

## Reproduction
1. Open BB on an iPhone-sized viewport.
2. Open a plugin nav panel, for example Automations.
3. Observe the top-left area.

The sidebar/menu trigger is drawn over the panel description/header. The plugin icon/title and body do not receive a clean header boundary.

## Expected
Standalone plugin panels should render the shared `AppPageHeader` on compact viewports. The header should reserve the sidebar trigger footprint, and the panel body should begin below the header.

## Observed
With the thread-splits experiment enabled, `AppLayout` suppresses the header when `pluginPanelMatch !== null`:

```ts
!(threadSplitsEnabled && pluginPanelMatch !== null)
```

That leaves `SidebarTriggerOverlay` visible while the plugin body starts at the top.

## Suggested fix
Do not suppress the shared header for standalone plugin routes. Thread pane headers should remain owned by the split workspace. Add a compact-viewport regression test covering a built-in `navPanel` route with the sidebar open and closed.

Local verification: changing `showHeader` to `!isThreadView && !isRootView` fixes the geometry; the targeted AppLayout test passes.

SawyerHood:
Fixes #1042

## Cause

`SplitThreadArea` bails to a bare `StandalonePaneContent` on compact viewports (`useIsCompactViewport()`, `max-width: 767px`), so no pane header renders. But `AppLayout` suppressed the shared header for every plugin panel route whenever thread splits were on, without checking the viewport.

The result on mobile: a `navPanel` route rendered zero header, and the fixed `SidebarTriggerOverlay` drew on top of the panel body. This hit built-in Automations and custom plugins alike.

## Fix

`AppLayout` now suppresses the shared header only when the split workspace actually owns one:

```ts
!(threadSplitsEnabled && !isCompactViewport && pluginPanelMatch !== null)
```

It uses the same `useIsCompactViewport()` hook that `SplitThreadArea` uses, so the two conditions stay complementary.

The issue suggested dropping the suppression completely (`!isThreadView && !isRootView`). That is not correct: on desktop the single-pane path still renders `AppPageHeader` in `NonThreadPaneContent`, so a full revert gives two headers.

## Test

New `apps/app/src/components/layout/AppLayout.plugin-panel-header.test.tsx`:

- Compact viewport shows the shared header with the plugin title.
- Regular viewport leaves the header to the split workspace.

The first case fails on the old code.

## Verification

- `turbo run test --filter=@bb/app` over `src/components/layout/` and `SplitThreadArea`: 77 tests pass.
- `turbo run typecheck --filter=@bb/app` passes.
- Manual QA in the dev app at a 390x844 viewport; before/after screenshots in the comments.

### DIFF
--- a/apps/app/src/components/layout/AppLayout.plugin-panel-header.test.tsx
+++ b/apps/app/src/components/layout/AppLayout.plugin-panel-header.test.tsx
@@ -1,0 +1,190 @@
+// @vitest-environment jsdom
+
+import { cleanup, render, screen } from "@testing-library/react";
+import type { ReactNode } from "react";
+import { MemoryRouter } from "react-router-dom";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { AppLayout } from "./AppLayout";
+
+const viewportState = vi.hoisted(() => ({ compact: false }));
+
+vi.mock("@bb/shared-ui/hooks/use-compact-viewport", () => ({
+  useIsCompactViewport: () => viewportState.compact,
+  CompactViewportOverrideProvider: ({ children }: { children: ReactNode }) => (
+    <>{children}</>
+  ),
+}));
+
+vi.mock("@/components/sidebar/AppSidebar", () => ({
+  AppSidebar: () => <aside data-testid="app-sidebar" />,
+}));
+
+vi.mock("@/hooks/useThreadSplitsEnabled", () => ({
+  useThreadSplitsEnabled: () => true,
+}));
+
+vi.mock("@/hooks/queries/system-queries", () => ({
+  useSystemConfig: () => ({
+    data: {
+      experiments: {
+        claudeCodeMockCliTraffic: false,
+        toolsHub: true,
+      },
+    },
+  }),
+}));
+
+vi.mock("@/lib/plugin-slots", () => ({
+  usePluginSlots: () => ({
+    navPanels: [
+      {
+        pluginId: "helm-wiki",
+        path: "wiki",
+        title: "Helm Wiki",
+        icon: "Book",
+      },
+    ],
+  }),
+}));
+
+vi.mock("@/components/plugin/PluginPanelHeader", () => ({
+  PluginPanelHeaderCenter: ({ panel }: { panel: { title: string } }) => (
+    <span data-testid="plugin-panel-header-center">{panel.title}</span>
+  ),
+  PluginPanelHeaderActions: () => null,
+}));
+
+vi.mock("@/components/project/ProjectActionsProvider", () => ({
+  ProjectActionsProvider: ({ children }: { children: ReactNode }) => (
+    <>{children}</>
+  ),
+}));
+
+vi.mock("@/components/thread/ThreadActionsProvider", () => ({
+  ThreadActionsProvider: ({ children }: { children: ReactNode }) => (
+    <>{children}</>
+  ),
+}));
+
+vi.mock("@/components/dialogs/ProjectPathDialog", () => ({
+  ProjectPathDialog: () => null,
+}));
+
+vi.mock("./AppPageHeader", () => ({
+  HEADER_ICON_BUTTON_CLASS: "header-icon-button",
+  AppPageHeader: ({
+    center,
+    actions,
+  }: {
+    center?: ReactNode;
+    actions?: ReactNode;
+  }) => (
+    <header data-testid="app-page-header">
+      {center}
+      {actions}
+    </header>
+  ),
+}));
+
+vi.mock("@/lib/iframe-drag-guard", () => ({
+  IframeDragGuardOverlay: () => null,
+}));
+
+vi.mock("@/lib/bb-desktop", () => ({
+  BROWSER_SIDEBAR_TRIGGER_INSET_CLASS: "",
+  CHROME_ROW_CLASS: "",
+  DEFAULT_DESKTOP_WINDOW_STATE: { isFullScreen: false },
+  MACOS_CHROME_CONTROL_AXIS_CLASS: "",
+  MACOS_CHROME_CONTROL_NO_DRAG_CLASS: "",
+  MACOS_CHROME_TRAFFIC_LIGHT_AXIS_NUDGE_CLASS: "",
+  MACOS_TRAFFIC_LIGHT_RESERVE_OFFSET_CLASS: "",
+  MACOS_WINDOW_DRAG_CLASS: "",
+  MACOS_WINDOW_NO_DRAG_CLASS: "",
+  getBbDesktopInfo: () => null,
+  shouldReserveMacosTrafficLights: () => false,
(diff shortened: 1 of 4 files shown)


SawyerHood:
## Manual QA

Dev app at a 390x844 viewport, route `/plugins/automations/automations` (built-in Automations `navPanel`).

**Before** — no header at all. The fixed sidebar trigger overlaps the panel description text at the top-left:

```
[⧉]anage scheduled bb work across projects and
folders. Automations run recurring or one-time tasks
without manual prompting.
```

**After** — the shared `AppPageHeader` renders, reserves the trigger footprint, and the panel body starts below it:

```
[⧉]  Automations  ›  Installed
──────────────────────────────────────────────────
Manage scheduled bb work across projects and
folders. Automations run recurring or one-time tasks
without manual prompting.
```

Also checked with the sidebar open on the same viewport: the compact sidebar is a full-width overlay sheet, so it never interacts with the header geometry.

Screenshot files are attached in the review thread.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the Slop Cop. I am reviewing this pull request under the review rule.

Stack, from bottom to top:

- #1045 — Restore the shared header for plugin panels on compact viewports

No pull request requires a skip. I will run security, code quality, performance, architecture, and end-to-end checks. The end-to-end check will run on #1045.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Security phase complete. I found no security issue.

The change only uses the compact viewport state to select existing header code. React renders the plugin title as text. Existing validation, plugin context, and error boundaries remain in place.

No blocker exists.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Performance phase complete. I found no performance or resource-use regression.

The change adds one Boolean check to an existing render path. It adds no route computation, event listener, layout loop, or frame work.

The focused Turbo tests passed 42 tests.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Code quality and architecture phase complete. I found no blocking issue.

Low-priority suggestion: centralize the effective split-workspace condition. `AppLayout` now uses the inverse of the compact condition in `SplitThreadArea`. A shared predicate could prevent future policy drift.

The focused `SplitThreadArea` suite passed 40 tests.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

End-to-end phase complete at the exact head SHA.

I installed the bundled Docs plugin in a fresh development profile. I opened its panel at 500 by 800 pixels.

The compact route showed one `Docs` header. The header height was 48 pixels. The sidebar control stayed inside that row. A regular 1100-pixel viewport also showed one header.

The new focused Turbo test also passed two tests.

No end-to-end blocker exists.

apps/app/src/components/layout/AppLayout.plugin-panel-header.test.tsx
```diff
@@ -0,0 +1,190 @@
...
+    viewportState.compact = true;
+    renderPluginPanelRoute();
+
+    expect(screen.getByTestId("app-page-header")).toBeTruthy();
+    expect(screen.getByTestId("plugin-panel-header-center").textContent).toBe(
+      "Helm Wiki",
+    );
+  });
+
+  it("leaves the header to the split workspace on regular viewports", () => {
```
SawyerHood, line 185: 🚨 `slopcop/review` — Consider a route-level test for header ownership.

This test mocks the shared header and does not render the real split workspace. A full route test could catch zero-header and duplicate-header regressions.

apps/app/src/components/layout/AppLayout.tsx
```diff
@@ -564,10 +564,15 @@ export function AppLayout({ children }: AppLayoutProps) {
...
+  // Plugin panel routes hand their header to the split workspace, which draws a
+  // pane header per pane. Compact viewports render the route as a single page
+  // surface with no pane chrome (see SplitThreadArea), so the shared header must
+  // come back — it reserves the sidebar trigger footprint, and without it the
+  // trigger overlays the panel body.
   const showHeader =
     !isThreadView &&
     !isRootView &&
-    !(threadSplitsEnabled && pluginPanelMatch !== null);
+    !(threadSplitsEnabled && !isCompactViewport && pluginPanelMatch !== null);
```
SawyerHood: 🚨 `slopcop/review` — Consider centralizing the effective split-workspace condition.

`AppLayout` now uses the inverse of the compact condition in `SplitThreadArea`. A shared predicate could prevent future policy drift.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

No blocking findings.

The stack contains only #1045. No pull request required a skip.

- Security: no issue found.
- Code quality: the fix is correct. I left two low-priority suggestions about shared policy and route-level coverage.
- Performance: no regression found.
- End-to-end: a real Docs panel showed one header at compact and regular widths. The compact sidebar control stayed inside the header row.
- Tests: the focused Turbo test passed two tests. All required GitHub checks passed.
- Final GPT-5.6 gate: `APPROVE`.

apps/app/src/components/layout/AppLayout.tsx
SawyerHood: Done in 92c73fe49. Extracted `useSplitWorkspaceActive()` (`threadSplitsEnabled && !isCompactViewport`) and read it on both sides of the header handoff — `AppLayout.showHeader` and `SplitThreadArea`'s standalone fallback plus `isSplitActive`.

The reconcile effect still uses `useThreadSplitsEnabled` alone: the layout atom is deliberately preserved across compact viewports so the arrangement returns when the viewport widens.

apps/app/src/components/layout/AppLayout.plugin-panel-header.test.tsx
SawyerHood: Not taking this one, for two reasons.

A route-level test would have to render `AppLayout` and `SplitThreadArea` together, which pulls in the sidebar, the query layer, the plugin host, and the iframe-mounted panel component. The mock surface needed to get there is large enough that the test would mostly assert the shape of its own mocks.

The coverage it would add is already in place from two directions: the unit test pins `AppLayout`'s side of the handoff at both viewports, and after 92c73fe49 both sides read one predicate (`useSplitWorkspaceActive`), so zero-header and duplicate-header are no longer independently reachable. Your own end-to-end phase also checked exactly this — one header at 500px, one at 1100px.

Happy to revisit if the handoff grows a third case.

Issue #1028, opened 2026-08-05
Import an existing ACP provider session as a thread (session/load with history persist)


mgpai22:
## What I want

Import an existing agent session as a bb thread, with the full history in the timeline and the original session as the live provider session. Nothing summarized, nothing re-ingested, nothing lost.

I run long-lived omp sessions over SSH. Some of them are weeks of accumulated context (a repo-understanding session that later turns treat as memory). I added the project to bb and want those sessions to continue as bb threads. Today the only path is to export the transcript and attach it to a fresh thread's first prompt. That costs a full re-ingest, loses tool calls and turn structure in the timeline, and the result is a copy, not the session.

## What already exists

Most of the plumbing is in place on both sides.

bb side:

- The ACP bridge probes `agentCapabilities.loadSession` at initialize (`packages/agent-runtime/src/acp/bridge/bridge.ts:1525`).
- On `resume` of a bb-known thread it already calls `session/load` with the stored `providerThreadId`, and falls back to `session/new` with a warning (`bridge.ts:1532-1552`).
- During load, every replayed `session/update` is dropped behind the `session.loading` gate (`bridge.ts:1717`). Correct for resume: bb already has that timeline.

Agent side (omp as the working example):

- `omp acp` declares `loadSession: true` (`packages/coding-agent/src/modes/acp/acp-agent.ts:520`).
- Its `loadSession` handler loads the managed session and replays the full history, tool calls with ids and arguments included (`acp-agent.ts:563`, `#replaySessionHistory` at `acp-agent.ts:2041`).
- The loaded session is the same session file that `omp --resume` uses. Continuation appends to it. Compaction state and context carry over exactly.

## The gap

There is no way to hand bb an external provider session id. `session/load` is only reachable for sessions that bb created itself. And the one code path that could ingest history discards it by design.

So the feature is narrow:

1. An import entry point that creates a thread with a caller-supplied `providerThreadId`, for providers whose agent advertises `loadSession`.
2. A persist mode for that first load: translate the replayed `session/update` stream through the normal event path instead of dropping it, marked as historical so it does not run turn logic.
3. A workspace check: the imported session has a cwd; bind the thread to a matching project source or environment, and refuse a mismatch.

## Proposed shape

```
bb thread import --project proj_x --provider acp-omp \
  --provider-session 019fb4b0-32f0-7000-b43c-69ac16a18236
```

[...]

PR #1047, opened 2026-08-06
Add bb thread import for existing ACP sessions

mgpai22:
Implements #1028.

Adds `bb thread import`, which adopts an existing external agent session as a bb thread: the replayed history lands in the timeline, and the thread continues that same provider session.

```
bb thread import --project proj_x --provider acp-omp \
  --provider-session 019fd3b5-baf1-7000-b953-df71341a47d3 \
  --cwd /path/to/repo
```

## How it works

The ACP bridge already calls `session/load` when resuming a bb-created thread, and drops the replayed `session/update` stream behind the `session.loading` gate. That is correct for resume, because bb already holds that history.

Import reuses the same call with a caller-supplied session id, and forwards the replay instead of dropping it. Replayed events are marked `historical`: they persist as timeline events but skip turn lifecycle effects, so the thread lands idle without running a turn. Live turns after that behave normally.

Also included:

- `supportsSessionImport`, derived from `agentCapabilities.loadSession` at initialize, with a live per-agent probe on top of the static ACP-family default. Providers without it are refused.
- cwd validation: the asserted session cwd must match the project source or an existing project workspace.
- A duplicate-binding guard: two threads cannot bind the same provider session (409, plus a bridge-side check).
- Contract, route, SDK verb, CLI command, and the bb-cli skill doc.

`HOST_DAEMON_PROTOCOL_VERSION` goes to 75. `thread.start` gained the `sessionImport` descriptor and `provider.list_models` gained the optional `supportsSessionImport` result field; an old daemon would drop the descriptor and silently start a fresh, history-less session.

## Verification

Tests cover the bridge (replay persisted, unsupported agent refused, load failure surfaced), adapter translation, and the route (cwd mismatch, unsupported provider) against real in-memory sqlite.

Live run against omp 17.2.9: a 2,404-message session imported as 1,994 persisted events (993 completed items, including tool calls and reasoning), thread idle, no model calls. Bogus session id and cwd mismatch both refuse with a clear error.

## Known limits

- Replayed history lands as one synthetic historical turn. ACP replay carries no turn boundaries, so adjacent assistant messages merge unless a user message or tool call separates them.
- The session cwd is caller-asserted. `session/load` takes cwd as input and it cannot be probed from the agent, so validation is against the project's own paths.
- No UI surface. The capability is exposed in the providers API for gating.
- Fork of an imported thread is unchanged, so still unsupported for ACP.

Two commits touch test infrastructure unrelated to the feature: a `@bb/server` vitest timeout raise, and hermeticity fixes in `install-machine-script` (a globally installed `bb-app` on PATH leaks into the test that asserts its absence) and `internal-skill-trees` (hardcoded 0644 vs a 0002 umask). Happy to split those out.


SawyerHood:
TY for the pr @mgpai22 will take a look tomorrow

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull request for security, code quality, architecture, performance, and practical end-to-end behavior.

apps/server/src/services/threads/thread-import.ts
```diff
@@ -0,0 +1,205 @@
...
+ * turns by provider session id (packages/agent-runtime/src/acp/bridge/bridge.ts).
+ * Binding a second bb thread to a provider session another live thread
+ * already binds would make the bridge misroute turns between the two, so
+ * refuse it up front.
+ */
+function requireUnboundProviderSession(
+  deps: Pick<ThreadImportDeps, "db">,
+  args: { hostId: string; providerSessionId: string },
+): void {
+  const existingThreadId = findLiveThreadIdByProviderThreadId(deps.db, {
```
SawyerHood: 🚨 `slopcop/review` — Make the provider-session claim atomic across environments.

This lookup only sees identity events that a completed start already stored. Two imports for different project workspaces can both pass before either identity event exists. Each environment owns a separate runtime and ACP bridge, so the process-local reservation does not protect this race. Both threads can then load and continue the same external session. Add a server-owned reservation with a unique key on `(hostId, providerId, providerSessionId)`. Claim it in the thread-create transaction. Release it after failed creation or permanent deletion. Add a concurrent test that uses two environments.

packages/sdk/src/areas/threads.ts
```diff
@@ -914,6 +933,13 @@ export function createThreadsArea(args: CreateSdkAreaArgs): ThreadsArea {
       );
     },
     get: getThread,
+    async import(input) {
```
SawyerHood, line 951: 🚨 `slopcop/review` — Preserve automatic plugin attribution for imports.

This SDK method defaults the origin to `sdk`. The plugin SDK wrapper adds plugin attribution only for `spawn` and `fork`. Thus, `bb.sdk.threads.import()` creates a thread without `origin: "plugin"` and `originPluginId`. Extend `wrapSdkForPlugin` with an import wrapper that uses the same rules. Add a plugin API test for default and explicit origins.

packages/agent-runtime/src/acp/adapter.ts
```diff
@@ -1228,7 +1362,7 @@ export function createAcpProviderAdapter(
   function buildSessionParams(
     command: Extract<
       AdapterCommand,
-      { type: "thread/start" | "thread/resume" }
+      { type: "thread/start" | "thread/resume" | "thread/import" }
```
SawyerHood, line 1365: 🚨 `slopcop/review` — Use the validated import directory for this command.

The server validates the request directory and uses it for the thread environment. This shared builder then selects `profile.cwd ?? command.cwd`. A custom ACP agent with a configured directory therefore sends a different path to `session/load`. The thread attaches to one project while the provider loads another directory. Use `command.cwd` for `thread/import`, or validate the effective profile directory at the server boundary. Add a custom-agent directory test.

packages/db/src/data/events.ts
```diff
@@ -2516,6 +2517,42 @@ export function getStoredProviderThreadIdAtOrBeforeSequence(
...
+ * Reverse lookup for provider-session uniqueness: the most recently touched
+ * non-deleted thread on this host whose event log has ever recorded this
+ * providerThreadId (via thread/identity or any provider-scoped event). Used
+ * to refuse importing an external ACP session that another live bb thread
+ * already binds, since the ACP bridge routes by provider session id and a
+ * second binding would misroute turns between the two threads.
+ */
+export function findLiveThreadIdByProviderThreadId(
+  db: DbQueryConnection,
+  args: { hostId: string; providerThreadId: string },
```
SawyerHood: 🚨 `slopcop/review` — Include the provider ID in this lookup key.

Provider session IDs belong to a provider namespace. This query uses only the host and session ID. A session named `abc` in `acp-omp` can therefore block a valid `abc` import from `acp-opencode` on the same host. Pass `providerId` into this function and filter on `threads.providerId`. Use the same provider dimension in the atomic reservation.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

This pull request lets BB adopt a conversation that started in an ACP agent. BB copies the old conversation into a thread, then lets the user continue it.

### Findings

I found four issues that need fixes:

1. Two imports can bind the same provider session when they use different environments. The database check is not atomic, and each environment has a separate bridge process.
2. The duplicate lookup omits the provider ID. Equal session IDs from two different ACP providers can block each other.
3. A custom ACP agent can replace the validated import directory with its configured directory. The provider can load a path that the server did not validate.
4. The new SDK import method loses automatic plugin attribution. The plugin SDK wrapper covers spawn and fork, but it does not cover import.

The server should own a persistent provider-session reservation. The reservation key should include the host, provider, and provider session. This design also removes the duplicate event-log lookup policy from environment-local bridge processes.

I found no command injection, path traversal, new secret exposure, or direct authorization bypass. The daemon protocol version correctly moves from 77 to 78. The CLI, SDK, route, guide, and built-in skill surfaces exist.

### Validation

- All selected tests passed: 840 agent-runtime, 361 database, 49 host-daemon-contract, 41 server-contract, 86 SDK, and 1,347 server tests.
- All seven selected package type checks passed.
- The selected package builds passed.
- The built `bb thread import --help` command passed.

I did not use a browser test. This change adds an API, SDK, and CLI creation path, but it adds no browser import action. A full live check also needs a real external ACP session. The adapter and server integration tests cover replay, lifecycle, and route behavior with the fake ACP agent.

I posted this as a comment-only review. I did not approve the pull request or request changes through GitHub.

mgpai22:
All four findings are addressed, and the branch is rebased onto current main (protocol version now 84).

**F1 — atomic reservation.** New `provider_session_reservations` table (migration `0089`), unique on `(host_id, provider_id, provider_session_id)`, claimed inside the existing `createThread` immediate transaction. Release is FK cascade on `threads.id`, which covers both failed creation and permanent deletion; a soft-deleted thread deliberately keeps its claim until hard delete. Tests: a real two-project/two-environment `Promise.all` race asserting exactly one 201 and one 409, plus a deterministic case for the window before any identity event exists.

I kept the event-log check alongside the reservation rather than replacing it: sessions bound by plain start/resume, and imports created before this migration, exist only in the event log. The reservation is checked first and is the race-safe source of truth.

**F2 — plugin attribution.** `wrapSdkForPlugin` gains an `import` wrapper using the same rules as spawn and fork. Test covers default and explicit origins. I also fixed `fake-sdk.ts`, which claimed to mirror the wrapper but applied attribution only to spawn.

**F3 — validated cwd.** `thread/import` now pins `command.cwd`; start and resume still honour `profile.cwd`. Review found the first fix was incomplete: a later `thread/resume` of an imported thread re-loaded at `profile.cwd` after any daemon restart. Resume now carries the validated directory too, and a cwd assertion that disagrees with a pinned agent directory is refused at the boundary instead of silently ignored.

**F4 — provider-scoped key.** `findLiveThreadIdByProviderThreadId` takes `providerId` and filters on it; the reservation carries the same dimension. Tests cover the cross-provider collision at both db and route level.

Protocol version is bumped only for the earlier `thread.start` descriptor. These four fixes changed no wire payload.

I also dropped two commits that were only working around slowness on my machine: the `@bb/server` timeout raise and the command-poll default. Both were out of scope, as flagged. Locally the server suite needs `--testTimeout=20000` to pass here; with that it is 1385/1386, and the one failure is a hardcoded `0644` expectation under my `umask 0002`. Happy to file that separately if it is worth fixing.

SawyerHood:

smomen:
@SawyerHood any chance we can revive and consider merging this? Important for migrating to bb on a bunch of open threads using native resume functionality!

PR #1049, opened 2026-08-06
Scope the workspace path claim to the project

### LINKED ISSUE #1043: A folder used by a Personal thread can never be opened as a project — "Workspace path is already attached to a different project"
# A folder used by a Personal thread can never be opened as a project — "Workspace path is already attached to a different project"

## What I was trying to do

I was working on a repo inside a Personal thread and asked it to switch its working directory to that repo (normal, supported usage). Later I created a proper bb project from that same folder, expecting to open it and work on it there.

## What happened

Creating the project succeeded, but **every attempt to open it failed** with:

> Workspace path is already attached to a different project

The project was effectively unusable — starting any thread in it hit the same error, every time. Nothing in the UI or CLI said *which* project owned the folder, and there was no supported action to resolve it. The only way I could unblock myself was editing bb's database by hand.

## What I expected instead

I understand a folder probably shouldn't be owned by two projects at once — that rule makes sense. The issue is that this state is created silently, reported confusingly, and can't be fixed in the app. It would help if any of these were true:

1. Creating a project (or adding a source) at a folder already claimed by another project warns or rejects up front, ideally with an "adopt this folder" option; **or**
2. The open error says which project owns the folder and how to resolve it; **and**
3. There is a supported way to transfer or release a folder's ownership — today there is none.

## Steps to reproduce

1. In a Personal thread, ask the agent to switch its working directory to an existing repo folder (uses the `update_environment_directory` tool).
2. Create a new project from that same folder (`bb project create --root <path>`, or the UI).
3. Open the new project / start a thread in it.
4. It fails with *"Workspace path is already attached to a different project"* — and keeps failing on every open.

## Why this happens (technical notes for maintainers)

- Switching a thread's directory records the folder as a workspace owned by that thread's project (`thread-environment-directory.ts`). From a Personal thread, the folder therefore becomes owned by the Personal project — permanently, even after the thread is archived.
- That ownership is never released: project deletion skips non-managed workspaces (`project-deletion.ts`), and workspace cleanup only handles managed ones (`environment-cleanup-internal.ts`). Deleting the owning project wouldn't free the folder either.
- Project creation and source-add never check existing workspace ownership (`routes/projects.ts`), so the conflict only surfaces later, at thread open, in `thread-create.ts` (`existingUnmanagedEnvironmentIntentByHostPath` → 409).
- There is no API, CLI, or UI to transfer or release the claim — hence the manual database edit.

## Suggested directions

1. Validate at project-create / add-source time, and offer to adopt the folder if it's owned by another project.
2. Make the open error actionable: name the owning project and point to a supported fix.
3. Avoid permanently claiming a folder when a Personal thread switches into it (or release the claim when the owning thread is archived/deleted).

## Environment

- bb 0.35.1 (desktop app)
- macOS
- Local repo folder on the primary machine

Happy to provide more details, logs, or help test a fix.

---

*Note: this report was drafted with the help of an AI agent. The reporter is actively involved and will respond to any follow-up questions or requests for details.*

SawyerHood:
Fixes #1043.

## Problem

A folder used by a Personal thread could never be opened as a project. Every attempt failed with `Workspace path is already attached to a different project`, and nothing in the UI or CLI said which project owned the folder or offered a way to release it. The reporter had to edit the database by hand.

## Root cause

The `environments` table had a **global** unique index on `(host_id, path)`. One host folder therefore belonged to exactly one environment, and so to exactly one project.

When a Personal thread called `update_environment_directory`, it created an unmanaged environment under the Personal project. That row claimed the folder permanently. A later thread create in a real project looked the environment up by host and path alone, saw a different `projectId`, and threw a 409 in `thread-create.ts`.

The claim was global, but a project is the unit that owns work. Two projects can already point at the same source folder — only the environment index disagreed.

## Fix

Scope the workspace-path claim to the project:

- `packages/db/src/schema.ts` — the unique index becomes `(project_id, host_id, path)`, with migration `0086_project_scoped_environment_path.sql`. The new index is weaker than the old one, so existing rows always satisfy it.
- `packages/db/src/data/environments.ts` — `findEnvironmentByHostPath` becomes `findProjectEnvironmentByHostPath` and takes a `projectId`.
- `apps/server/src/services/threads/thread-create.ts` and `thread-environment-directory.ts` — both lookups are project-scoped, which removes the two cross-project failure branches.

Each project now gets its own environment row for a shared folder. A Personal thread that visits a folder no longer blocks a project from using it, and the reverse also works.

This removes the error class the issue describes, so the report's other suggestions (adopt-on-create, an error that names the owner, a release action) are no longer needed.

## Notes

- The migration DDL is idempotent (`IF EXISTS` / `IF NOT EXISTS`) because `migrate.test.ts` replays the newest migrations after deleting ledger rows.
- No `HOST_DAEMON_PROTOCOL_VERSION` bump: no session payload, WebSocket message, host RPC command, or result changed.

## Tests

- New test in `thread-create-shared-workspace-path.test.ts`: a project thread starts in a folder the Personal project already claims and gets its own environment. It reproduced the reported 409 before the fix.
- New test in `internal-events-tool-calls.test.ts`: `update_environment_directory` switches into a folder another project uses and leaves the other project's claim alone.
- `@bb/db` 359 pass, `@bb/server` 1325 pass. Repo-wide `turbo run typecheck` passes (57 tasks).
- `internal-skill-trees.test.ts` fails locally on a file-mode comparison (420 vs 436). It fails identically without these changes — a local umask difference, unrelated.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am reviewing this pull request against the security, code quality, performance, architecture, and end-to-end rules.

The stack contains PR #1049 only. No pull request was skipped.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The security phase is complete. It found two possible cross-project hazards around cleanup and branch changes on one physical path.

I will verify both hazards against the cleanup and workspace logic before the final review.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The quality and architecture phase is complete. It confirmed the branch hazard and found an index regression plus a stale tool description.

The final review will include only findings that remain valid after verification.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The performance phase is complete. It confirmed a full environment scan in a host-thread query after the index change.

The exact project, host, and path lookup still uses the new unique index.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The end-to-end phase passed on the development application at the pull request SHA.

Two projects created idle threads on one host path. The server returned different project-owned environment IDs.

apps/server/src/services/threads/thread-create.ts
```diff
@@ -387,19 +387,16 @@ function existingUnmanagedEnvironmentIntentByHostPath(
   deps: ThreadCreateDeps,
   args: ExistingUnmanagedEnvironmentIntentByHostPathArgs,
 ): ExistingUnmanagedEnvironmentIntentResult | null {
-  const existing = findEnvironmentByHostPath(deps.db, args.hostId, args.path);
+  const existing = findProjectEnvironmentByHostPath(
+    deps.db,
+    args.request.projectId,
```
SawyerHood, line 424: 🚨 `slopcop/review` — P1: Keep the active-thread guard across projects.

This project-scoped lookup hides an environment from another project. A branch request can then run a checkout on the shared directory.

The current guard checks only the environment that this lookup returns.

Check all host and path environments before checkout. Return 409 when any nonterminal thread uses the directory.

Add a cross-project branch test that verifies no checkout command runs.

apps/server/src/services/threads/thread-environment-directory.ts
```diff
@@ -286,20 +286,16 @@ export async function handleUpdateEnvironmentDirectoryToolCall(
     );
   }
 
-  const existingEnvironment = findEnvironmentByHostPath(
+  const existingEnvironment = findProjectEnvironmentByHostPath(
     deps.db,
+    args.thread.projectId,
```
SawyerHood, line 306: 🚨 `slopcop/review` — P3: Update the tool description for project-scoped reuse.

This lookup now reuses an environment only in the current project. The tool description still promises reuse of any host and path environment.

State the project limit in the description so the agent contract matches the behavior.

packages/db/src/schema.ts
```diff
@@ -454,7 +454,13 @@ export const environments = sqliteTable(
     updatedAt: integer("updated_at").notNull(),
   },
   (table) => [
-    uniqueIndex("environments_host_path_idx").on(table.hostId, table.path),
+    // A workspace path is claimed per project, not globally. Two projects may
+    // point at the same folder; each gets its own environment for it.
+    uniqueIndex("environments_project_host_path_idx").on(
```
SawyerHood, line 459: 🚨 `slopcop/review` — P1: Protect shared physical paths from managed cleanup.

This index lets a managed environment and an unmanaged environment use one path across projects.

Cleanup checks only the managed environment ID. The daemon can remove the directory while another project still uses it.

Reject unmanaged aliases to managed paths. Add a host and path guard before each destroy command.

Add a test where cleanup must preserve an active unmanaged alias.

packages/db/drizzle/0086_project_scoped_environment_path.sql
```diff
@@ -0,0 +1,2 @@
+DROP INDEX IF EXISTS `environments_host_path_idx`;--> statement-breakpoint
+CREATE UNIQUE INDEX IF NOT EXISTS `environments_project_host_path_idx` ON `environments` (`project_id`,`host_id`,`path`);
```
SawyerHood: 🚨 `slopcop/review` — P2: Retain a host-leading environment index.

The replacement index starts with project_id. A query that filters only host_id now scans all environments.

The host session paths call listHostThreadIds, which uses that filter.

Use a host-leading unique tuple or add a host and path index. Verify the plan uses an index search.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

I found two P1 safety defects, one P2 index regression, and one P3 contract mismatch.

- P1: Shared managed and unmanaged aliases can let cleanup remove an active directory.
- P1: A project-scoped lookup lets a branch checkout bypass the active-thread guard from another project.
- P2: The project-leading unique index causes host-only queries to scan all environments.
- P3: The directory tool description still promises global environment reuse.

The focused server tests passed with 24 tests. The server and database type checks passed.

The browser test also passed for two projects with idle threads on one host path. That test does not cover the two unsafe operations.

Fix the physical path safety defects before merge. Also restore a host-leading index and update the tool description.

SawyerHood:
Addressed all four SlopCop findings in b897f3963. Notes on each:

**P1 — cross-project branch checkout (thread-create.ts).** Confirmed, and slightly wider than reported. The guard only saw the requesting project's environment, but a project with *no* environment at the path skipped the check entirely by falling through to `direct-unmanaged` with a branch. Moved the guard to the call site, before the intent is chosen, and made it cross-project via a new `hasLiveThreadAtHostPath`. Test: `refuses a branch checkout while another project works in the directory`, which asserts no environment is created.

**P1 — managed/unmanaged aliasing (schema.ts).** Confirmed as reachable. Managed worktree paths embed the environment ID, so two *managed* environments still cannot collide, but an unmanaged environment in another project could now attach in place to a managed worktree — and its cleanup deletes the directory.

I took the narrower of the two suggested fixes: reject the alias rather than add a host+path guard before each destroy. That keeps the invariant "a managed path has exactly one environment", so the destroy precondition stays valid as-is instead of every future destroy path needing to remember a new check. Enforced in both thread create and `update_environment_directory`, with a test on each.

**P2 — host-leading index.** Confirmed with `EXPLAIN QUERY PLAN`. Before: `SELECT id FROM environments WHERE host_id = ?` → `SCAN environments`. Added a non-unique `(host_id, path)` index in migration 0087. After, all four affected shapes search an index, including the new cross-project host+path guard query. Named it `environments_host_path_lookup_idx` rather than reusing the dropped unique index's name, so replaying 0086 cannot drop it.

**P3 — tool description.** Fixed. It now states that reuse is project-scoped, that another project may hold its own environment for the same directory, and that a managed worktree owned by another project is refused.

Thanks — the two P1s were real regressions introduced by the scoping change.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The stack contains PR #1049 only. The author has write access, and the pull request is not a draft. I skipped no pull requests.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The performance phase is complete. It found a medium N+1 query pattern in shared-path branch checks.

It also found a low-cost duplicate-index concern. I will verify both findings before the final review.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The end-to-end phase passed on the development application at the pull request SHA.

A Personal thread changed to the test directory. A standard-project thread then used the same host path successfully.

The browser showed the completed standard-project thread without the old path-claim error. The database held separate project-owned environment rows.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The security phase is complete. It found three possible physical-path alias or race hazards.

I will compare them with the base branch before the final review. Only pull-request regressions will become findings.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The code-quality and architecture phase is complete. It found two shared-path consistency risks.

It also confirmed the avoidable database query cost. I will compare each risk with the base branch before the final review.

apps/server/src/services/threads/thread-create.ts
```diff
@@ -383,22 +386,61 @@ async function resolveManagedDefaultBaseBranchForCreate(
...
+      "Cannot checkout branch while another thread is using this workspace",
     );
   }
+}
+
+function existingUnmanagedEnvironmentIntentByHostPath(
+  deps: ThreadCreateDeps,
+  args: ExistingUnmanagedEnvironmentIntentByHostPathArgs,
+): ExistingUnmanagedEnvironmentIntentResult | null {
+  const existing = findProjectEnvironmentByHostPath(
```
SawyerHood, line 422: 🚨 `slopcop/review` — [P2] Keep branch metadata consistent across project aliases

This lookup selects only the current project environment. A branch checkout changes one physical Git tree, but settlement updates only this row. Other project aliases retain the old `branchName`, which the sidebar, runtime configuration, and fork logic can later use.

Update the physical branch fields for all live aliases, or refresh an alias before reuse. Add a two-project test that checks out through one alias and verifies both rows.

apps/server/src/services/threads/workspace-path-claims.ts
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Whether any project has a live thread working in this directory. A branch
+ * checkout rewrites the working tree, so it must not run while another
+ * project's agent is using the same folder.
+ */
+export function hasLiveThreadAtHostPath(
+  db: DbConnection,
+  args: HostPathArgs,
+): boolean {
+  return listEnvironmentsByHostPath(db, args.hostId, args.path).some(
```
SawyerHood: 🚨 `slopcop/review` — [P3] Replace the N+1 path check with a targeted query

This code loads all matching environments and runs one thread query for each row. The foreign-managed check also filters loaded rows in JavaScript. This conflicts with the repository data-access rule and makes a branch request perform `1 + N` checks.

Use a filtered `LIMIT 1` query for the managed predicate. Use one environment-to-thread join or `EXISTS` query for live threads.

apps/server/src/services/threads/workspace-path-claims.ts
```diff
@@ -0,0 +1,52 @@
...
+ * A bb-managed workspace (a worktree bb created and will destroy) owned by a
+ * different project. Attaching to it in place is unsafe: cleanup of the owning
+ * environment deletes the directory out from under the attached thread.
+ */
+export function foreignManagedEnvironmentAtHostPath(
+  db: DbConnection,
+  args: HostPathArgs & { projectId: string },
+): Environment | null {
+  return (
+    listEnvironmentsByHostPath(db, args.hostId, args.path).find(
```
SawyerHood: 🚨 `slopcop/review` — [P1] Reserve managed target paths before host completion

This exact-path lookup cannot see a managed environment while its path is still `NULL`. Another project can attach after the host creates the worktree but before the owner stores its path. The old global index rejected the owner update, but the new project-scoped index accepts both rows. Cleanup can then delete a directory used by the alias thread.

Store the managed target path, or an equivalent atomic claim, before host dispatch. Add a test that attempts a foreign attachment while managed provisioning is pending and requires rejection.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

I found three pull-request regressions.

- P1: A managed path remains unclaimed until host completion. The new index can then accept a dangerous cross-project alias.
- P2: A branch checkout updates one environment row but leaves sibling project aliases with stale branch metadata.
- P3: The new safety checks filter rows in JavaScript and perform an N+1 thread query.

The development application passed the shared-path workflow. A Personal thread and a standard-project thread used one host path with distinct ready environments.

The focused server tests passed with 27 tests. The database migration tests passed with 39 tests. Server and database type checks passed.

GitHub CI passed. The final GPT-5.6 gate returned REQUEST CHANGES, but this review uses the required comment-only action.

Please fix the P1 and P2 consistency defects before merge. The P3 change also enforces the repository data-access rule.

SawyerHood:
Round two addressed in 815ef098e. Migrations also squashed into a single 0086 (9bda88f11) — 0086 and 0087 both rebuilt the `environments` indexes and neither had shipped, so upgrading installs now do one rebuild. Regenerated via `drizzle-kit` off `origin/main` rather than hand-edited, so the snapshot chain stays consistent.

**P1 — managed path unclaimed until host completion.** Confirmed. `buildManagedEnvironmentPlan` creates the row with no `path` and computes the target only in `buildRequest`, so the row claims nothing for the whole provisioning window.

I defended the *workspace roots* instead of the row. bb derives every managed path from `<dataDir>/worktrees` or `<dataDir>/personal-workspaces`, so a path under either belongs to a managed environment whether or not it has been stored yet. This closes the window completely and needs no lifecycle change — I avoided writing the path at creation time because a provisioning failure would then leave an `error` row carrying a path to a directory that was never created, which `workspaceCanBeDestroyedNow` treats as destroyable. A project may still attach to a managed path it already owns, so same-project reuse keeps working. Test: `refuses a managed path whose owner has not stored it yet`, using a `provisioning` environment with a NULL path.

**P3 — N+1.** Agreed, and it conflicted with the explicit AGENTS.md data-access rule. `findForeignManagedEnvironmentAtHostPath` now filters `managed`/`projectId`/`status` in SQL with `LIMIT 1`, and `hasLiveThreadAtHostPath` is a single `threads`⋈`environments` join instead of one thread query per environment at the path.

**P2 — branch metadata across aliases.** I'm not fixing this one, and I'd rather say why than silently skip it.

`branchName` on an unmanaged environment has always been a cached *observation* of a directory bb does not control — any user can `git checkout` in that folder from their own terminal at any time, and bb has never held a lock on it. The convergence mechanism for that already exists and is keyed on the physical directory, not the row: `callEnvironmentWorkspaceStatus` reads the real `currentBranch` from the host and calls `recordEnvironmentCurrentBranch` per environment. Each alias polls the same directory and converges independently.

So this is not a new class of staleness, and the proposed fix would make things worse: writing branch metadata into another project's environment row from a checkout that project never requested creates a cross-project write where today there is only a per-project cached read. Happy to reconsider if you see a path where a stale read causes something worse than a briefly wrong branch label.

All checks green; the only local failure is a `internal-skill-trees` file-mode comparison that reproduces on the base branch (umask), and it passes in CI.

PR #1053, opened 2026-08-06
Document visual verification against the desktop shell

brsbl:
Verifying UI work meant reading markup and reasoning about what it renders, because nothing in the repo said to launch the app and look at it. Adds a **Visual Verification** section to `docs/debugging-and-qa.md` with the actual recipe, plus a pointer from `AGENTS.md`.

The substance is two findings that each cost real time in the session that prompted this:

- **A DevTools screenshot can never show traffic lights.** CDP captures the renderer's web contents; macOS draws the window frame outside it. So `screencapture` against the desktop shell is the only way to verify anything about window chrome — and its `could not create image from display` failure means the host process lacks Screen Recording permission, not that the command is malformed. Worth stating plainly, because the message reads like a bug in the invocation.
- **Faking `window.bbDesktop` to coax macOS chrome out of the web build does not work.** The desktop path expects the whole `BbDesktopApi`; a partial stub throws mid-render and silently blanks the page (the specific trap: subscribe methods must *return* an unsubscribe function, not be one). Even a complete stub only proves what the renderer does, never what the native frame does.

The section also records the cheaper path for in-page changes — the dev web app, with a static headless capture or a single long-lived CDP session when the check needs interaction across several states.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -57,7 +57,8 @@
 ## Debugging And QA
 
 - Do not assume. Inspect logs, query the database, call server APIs, or use the CLI to observe real state.
-- See [docs/debugging-and-qa.md](docs/debugging-and-qa.md) for dev ports/data dirs, entity-ID lookups, and the `scripts/bb-dev-app` local dev QA launcher.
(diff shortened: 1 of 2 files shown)


brsbl:
Closing — this is personal workflow guidance, not repo policy. The screencapture / Screen Recording and CDP-can't-show-native-chrome findings already live in ~/.claude/CLAUDE.md and ~/.bb/AGENTS.md, so the repo copy only duplicates them.

PR #1050, opened 2026-08-06
Add Disable and Uninstall to plugin sidebar row menus

brsbl:
## What changed

A plugin's page row in the side panel offered one action — hide it from the sidebar — so turning a plugin off or removing it meant leaving for Extensions. The row now carries the plugin's lifecycle too:

| | Enabled | Disabled | Ships with bb |
|---|---|---|---|
| 1 | Hide from sidebar | Hide from sidebar | Hide from sidebar |
| 2 | **Disable** | **Enable** | Disable |
| 3 | **Uninstall** / *Remove from bb* (destructive) | same | Uninstall, greyed — *"Included with BB; disable this plugin instead."* |

Both call the plugin admin routes the Extensions pages already use — `setPluginEnabled` (POST `…/enable`|`…/disable`) and `removePlugin` (DELETE `/plugins/:id`). No new server or daemon surface.

Removal policy and copy move into `plugin-removal.ts`, shared with the Extensions detail page, so the two menus cannot drift. That is also where the builtin refusal lives, since `DELETE /plugins/:id` answers 409 for plugins that ship with bb.

## Two details worth reviewing

**The list owns the mutations and the confirmation dialog, not the row.** Uninstalling unregisters the plugin's panels, so the row that started the request unmounts before it settles — a row-owned toast, navigation, or dialog would be lost.

**Uninstall navigates away; disable does not.** After a successful removal, a panel sitting on that plugin's route goes to the app root instead of standing on the "panel is not available" placeholder. Disable deliberately stays put so the state change is visible.

Lifecycle items render only once the installed-plugin list resolves — an action needs real `enabled`/provenance state before it can label itself. The query is gated on the sidebar actually having plugin rows, so an Extensions-only sidebar never asks for it.

## Verification

Driven against the dev app with a `path:`-installed test plugin contributing a nav panel:

- Disable → server reported `enabled:false`, the row left the sidebar, the panel fell back to the placeholder, and Extensions → Plugins showed the plugin unchecked. Re-enabled from the detail page → row returned.
- Uninstall → confirmation dialog, confirm, route moved `/plugins/qa-panel/qa` → `/`, row gone; after a reload the plugin was absent from both the sidebar and `GET /api/v1/plugins`.

[...]

PR #1054, opened 2026-08-06
Clear the traffic lights when a panel tab goes full screen

brsbl:
Full-screening a tab leaves the tab strip sitting under the macOS traffic lights — but only on a thread that is **not** in a split. In a split, the same action clears the lights correctly.

## Root cause

Collapsing the conversation produces the same layout on both thread surfaces:

- **Split host** sets its group layout to `[0, MAIN_PANEL_OPEN_SIZE_PERCENT]` (`SplitWorkspaceSecondaryPanelHost.tsx:129`).
- **Inline thread detail** sizes the timeline panel to `COLLAPSED_TIMELINE_PANEL_SIZE_PERCENT = 0`.

Either way the conversation column goes to zero width, and the thread header rides inside that column, so the panel's toolbar is left alone on the title-bar row on both surfaces.

Only the reserve differed. `resolveCollapsedPanelTrafficLightReserveClassName` gated it on `isInSplitHost`, justified in its own doc comment: inline thread detail "keeps a full-width header on the traffic-light row above the panel, so it needs no reserve here." That is not true once the conversation is collapsed — inline renders `{header}` inside the wrapper that collapsing makes `opacity-0` and `inert` (`ThreadDetailSecondaryContent.tsx:400`), and the comment there says as much: "`inert` removes the hidden conversation (header, timeline, composer)".

So the split host cleared the lights because it passed the host gate, and inline thread detail — laid out identically — did not. That asymmetry is exactly the reported symptom.

## Fix

Drop `isInSplitHost`. The reserve is about the collapse, not the host. The parameter is removed rather than left accepted-and-ignored.

## Verification

Measured the working reference — `AppPageHeader`'s reserve — in the running app under real macOS chrome, driving Chrome for Testing over CDP:

```
sidebar hidden, panel docked right:
  header:        left 16,   padding-left 104px,  content at 120   <- reference
  panel toolbar: left 1517, padding-left 0px                      <- correct while docked
```

120px is the shared target: `px-4` (16) + `MACOS_COLLAPSED_TOP_LEFT_RESERVE_CLASS` (104), which clears the lights and the pinned sidebar trigger. After this change the panel lands its leading controls on that same 120px when it owns the top-left, on both surfaces.

`turbo run typecheck --filter=@bb/app` clean; `ThreadSecondaryPanel.test.ts` (13) + `AppPageHeader.test.tsx` + `bb-desktop.test.ts` — 17 passed. The test that encoded the wrong premise now asserts the opposite, with the reason recorded.

[...]

PR #1058, opened 2026-08-06
Let a menu item carry its own aria-disabled

brsbl:
Split out of #1050, where it was buried in a 17-file plugin change. It touches every menu in the app, so it deserves its own review.

## The problem

A menu item that must refuse its action *while explaining why* cannot be natively disabled: both menu primitives pair `disabled` with `pointer-events-none`, so the item stops receiving the hover or focus that would reveal the explanation. Such an item has to supply its own `aria-disabled` and description instead — and `DropdownMenuItem` dropped both. The compact branch renders a plain `<button>` that never spread caller DOM props, so the item was announced as an ordinary **enabled** button that silently did nothing when tapped.

## The fix

Forward `aria-disabled` and `title` explicitly rather than widening the DOM-prop spread, because the compact `<button>`'s prop type rejects the primitive's div-shaped props wholesale.

The fallback to the native disabled state on **both** branches is load-bearing, not defensive:

```ts
aria-disabled={ariaDisabled ?? (disabled || undefined)}
```

Radix emits its own `aria-disabled={disabled || undefined}` and then spreads caller props over it, so forwarding a bare `undefined` *deletes* the attribute React would otherwise have rendered — for every ordinarily-disabled item in every menu. I shipped exactly that bug in the first attempt and it was invisible: `data-disabled` and the roving-focus skip are unaffected, nothing renders differently, no test failed. Only a screen reader could tell. Credit to thread `thr_dj4xb4qppr` for catching it and proving causation.

Both directions are covered by the new test, since an assertion is the only thing that can catch the regression.

## Verification

`turbo run typecheck --filter=@bb/app --filter=@bb/shared-ui` clean; 352 tests across 50 files pass on the full stack.


SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull-request stack for security, code quality, performance, architecture, duplication, and end-to-end behavior.

Stack order:

1. #1058 — Let a menu item carry its own aria-disabled
2. #1050 — Add Disable and Uninstall to plugin sidebar row menus
3. #1054 — Clear the traffic lights when a panel tab goes full screen

I will run the end-to-end check once on the top pull request, #1054.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Performance phase complete for #1058.

No performance issue found. The two forwarded properties add no event listeners, requests, state, or material render cost.

The registry was stale during this phase. The author updated it, and the registry retest now passes.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Security phase complete for #1058.

No security issue found. This change forwards accessibility fields and does not grant access or execute actions.

packages/shared-ui/src/components/ui/dropdown-menu.tsx
```diff
@@ -246,6 +246,13 @@ const DropdownMenuItem = React.forwardRef<
       disabled,
       role = "menuitem",
       "aria-checked": ariaChecked,
+      // Forwarded explicitly rather than via the DOM-prop spread: the compact
```
SawyerHood, line 249: 🚨 `slopcop/review` — Regenerate the dropdown registry before merge.

This source change leaves `packages/plugin-registry/r/dropdown-menu.json` stale. The registry test fails and reports this exact file.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Code-quality phase complete for #1058.

The registry was stale during this phase. The author updated it in the current head.

The plugin-registry retest now passes. No open finding remains on #1058.

packages/shared-ui/src/components/ui/dropdown-menu.tsx
SawyerHood: 🚨 `slopcop/review` — Fixed in the current #1058 head.

Commit `5270895e7d9b7c402cf5e3fdbc0b354886d58d61` now updates the generated dropdown registry.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Final SlopCop result for #1058: no open finding.

The author updated the generated dropdown registry during this review. The local registry retest passed, and all current #1058 GitHub checks passed.

Security: no finding.

Performance: no finding.

Quality: no open finding.

I used only a comment review. I did not approve or request changes.

brsbl:
Closing. This existed to let the plugin sidebar menu's disabled Uninstall item explain why it was refused — #1050, now closed as not worth the change. Without that caller, what remains is forwarding two attributes plus the fallback that keeps ordinarily-disabled items announcing correctly, which was a regression this PR introduced and then fixed. Not worth carrying a shared-ui change every menu depends on for that.

Noted for anyone who hits it later: `DropdownMenuItem` drops caller DOM props on the compact branch, so an item cannot supply its own `aria-disabled`, description, or tooltip handlers on touch. That is unchanged from before this PR. If you add one, forward them — and keep `aria-disabled={ariaDisabled ?? (disabled || undefined)}`, because Radix emits its own and spreads caller props over it, so a bare `undefined` silently strips the attribute from every disabled item in every menu.

PR #1061, opened 2026-08-06
Drop panel tabs of a plugin that is no longer installed

brsbl:
Split out of #1050. Self-contained: it adds the cleanup mechanisms, and the layer above wires the sidebar's uninstall to them.

## The problem

An uninstalled plugin's panels could outlive it in two places, both then rendering the "This plugin panel is not available" placeholder for a page nothing can restore:

- an open **split pane**
- a persisted **panel tab** — the worse of the two, because it came back on every reload

## Two mechanisms, because the two states are stored differently

**`pruneUninstalledPluginPanelTabs`** drops the tabs, applied from an effect that keys on the *installed plugin list* rather than the registered slots. A plugin whose frontend has not loaded yet registers nothing and its tabs must survive that window — as must the tabs of a merely *disabled* plugin, which stays installed.

The effect also depends on the tab list itself. Tabs arrive twice: synchronously from local storage, then again when the server list replaces it wholesale. The plugin list is already warm by then, so keying on it alone would silently miss whatever the server adds afterwards — for the rest of the session, since React Query's structural sharing keeps that identity stable.

**`closePanesForPluginAtom`** drops the panes. It clears the maximized pane when that no longer names a live pane, and clears the layout outright when the only pane that could keep focus is itself one of the removed ones — better an empty layout than a dead pane.

## Verification

This layer was checked out in isolation, not just as part of the stack tip: `tsc --noEmit` clean and 46 tests across 5 files pass with nothing above it applied.

### DIFF
--- a/apps/app/src/components/secondary-panel/useThreadFileTabs.ts
+++ b/apps/app/src/components/secondary-panel/useThreadFileTabs.ts
@@ -20,6 +20,7 @@ import {
   type WorkspaceFilePreviewFixedPanelTab,
 } from "@/lib/fixed-panel-tabs-state";
 import { usePluginSlots } from "@/lib/plugin-slots";
+import { usePluginList } from "@/hooks/queries/plugin-settings-queries";
 import { useFileOpenerPreferenceValue } from "@/lib/file-opener-preference";
 import {
   createFileOpenerTabForRequest,
@@ -47,6 +48,7 @@ import {
   isBrowserTab,
   openSecondaryPanelTabInState,
   pruneStorageTabs,
+  pruneUninstalledPluginPanelTabs,
   removeWorkspaceTabsForOtherEnvironments,
   replaceNewTabWithSecondaryPanelTabInState,
   reorderSecondaryPanelFileTabInState,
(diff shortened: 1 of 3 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am reviewing this stack under the `review` rule.

The stack contains:

- #1061: Clean up after a plugin that is no longer installed
- #1050: Add Disable and Uninstall to plugin sidebar row menus
- #1054: Clear the traffic lights when a panel tab goes full screen

I will review security, code quality, architecture, duplication, performance, and the full user path.

I will post a comment after each phase.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Security phase complete.

I found no new exploitable security issue.

I found one medium lifecycle gap. The new cleanup removes fixed panel tabs, but it does not remove persisted split panes.

Removal through Settings, the CLI, or another browser tab does not call the split-pane cleanup.

The stale plugin pane can return after a reload.

The agent ran the app typecheck, 55 focused tests, and diff checks.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Code quality phase complete.

Medium, blocking: the cleanup covers fixed panel tabs only.

Browse, Settings, CLI, SDK, and another browser tab can bypass the explicit split cleanup.

These paths can persist an unavailable plugin pane.

Move the cleanup to an installed-list owner that prunes fixed tabs and split panes.

The application typecheck passed. All 2,367 application tests passed.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Performance phase complete.

I found no performance defect in PR #1061.

The cleanup scans only the current tab list. React Query shares the plugin-list request across consumers.

The application typecheck passed. All 2,367 application tests passed.

apps/app/src/components/secondary-panel/useThreadFileTabs.ts
```diff
@@ -403,6 +405,54 @@ export function useThreadFileTabs({
...
+  // without this they come back as the "plugin tab is not available"
+  // placeholder on every reload, for a tab nothing can restore. Undefined while
+  // the list has not loaded, so a failed or pending fetch never prunes.
+  useEffect(() => {
+    if (!isPanelStateResolved || installedPluginIds === undefined) return;
+    updateFixedPanelTabsState((state) => {
+      const pruned = setPrunedSecondaryTabs({
+        activeTabId: state.secondary.activeTabId,
+        stateTabs: state.secondary.tabs,
+        tabs: pruneUninstalledPluginPanelTabs(
```
SawyerHood, line 429: 🚨 `slopcop/review` — This cleanup misses persisted split panes.

This effect removes fixed tabs only. Settings, Browse, CLI, and SDK removal paths do not call the split-pane cleanup.

Realtime refreshes this list, but it does not remove the split layout. A dead plugin pane can return after a reload.

Please reconcile split panes from the installed list or another global removal signal.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Final review: I found one medium, blocking defect.

The new installed-list effect removes fixed plugin tabs. It does not remove matching panes from the persisted split layout.

Settings, Browse, CLI, SDK, and other browser removal paths do not call the split cleanup.

A dead plugin pane can remain and return after a reload.

I found no exploitable security issue and no performance defect.

The application typecheck passed. All 2,367 application tests passed.

GitHub checks fail because the lower base branch has a stale generated dropdown registry. This PR does not introduce that failure.

Browser QA could not remove a plugin because the isolated server had no removable plugin.

I posted this as a comment review. I did not approve the PR or request changes.

brsbl:
Closing. This came in as review feedback on #1050, which added Uninstall to the sidebar row menu — cleanup was in scope because that PR created a new way to uninstall. #1050 is now closed as not worth the change, so the reason for this one went with it.

The stale-tab bug is real and predates all of this: uninstalling from the Extensions page has always been able to strand a persisted `plugin-panel` tab that then renders the unavailable-panel placeholder on every reload. But nothing in the remaining stack introduces or depends on the fix, so it stays a pre-existing issue rather than scope carried by an unrelated layer. Worth its own PR if it actually bites someone.

PR #1077, opened 2026-08-06
fix(github-plugin): Github PRs fail to sync when Issues are disabled on a repo

kschrader:
## Problem

The official GitHub plugin's `syncRepo` fetches issues and PRs in a single `Promise.all`. For a repo with GitHub Issues disabled, the two `gh issue list` calls fail (`the '<owner>/<repo>' repository has disabled issues`), which rejects the whole batch — the PR results are discarded, every sync logs `sync failed for <repo>` followed by `synced 0 item(s)`, and the plugin never shows any PRs for that repo.

Easy to reproduce: track any repo with Issues turned off and watch `bb plugin logs github`.

## Fix

Route the two issue-list calls through a small wrapper that treats the disabled-issues error as an empty issue list (`"[]"`, which `toItems` parses to no items) so PR sync proceeds. Any other `gh` failure — auth, network, unknown repo — still rejects and surfaces as a sync failure exactly as before. The PR-list calls are untouched.

Verified locally against a repo with issues disabled: sync went from `synced 0 item(s)` to `synced 80 item(s)`, with open/closed PRs cached and browsable.


SawyerHood:
TY for the fix!

PR #1079, opened 2026-08-06
Merge official-plugins into plugins

SawyerHood:
## What

`plugins/` and `official-plugins/` both held first-party plugins that ship bundled inside the app. The split duplicated what `autoInstall` in the bundled-plugin registry already records, so the two could drift — `t3sidebar` already lived in `plugins/` while the registry listed it as store-only, and it needed a per-entry `repoDirectory` override to work.

This folds `official-plugins/*` into `plugins/*` and makes the registry the only source of install policy.

## Changes

- **Registry** (`apps/server/src/services/plugins/builtin-registry.ts`): delete `BundledPluginDefinition.repoDirectory` and the `"builtin-plugins" | "official-plugins"` union. `resolveBuiltinPluginRootPathForModuleDir` loses its per-plugin lookup and fallback chain, and resolves against one constant. Both registry `.map()` bodies are one line each again, and the `t3sidebar` override is gone.
- **Copy script**: `copyBuiltinPlugins` narrows its `plugins` param to `Pick<…, "name">`.
- **`scripts/build-official-plugins.mjs`**: derives its list from `OFFICIAL_PLUGINS` instead of a hardcoded name→directory map, so adding a store-only plugin no longer needs an edit here.
- **`turbo.json`**: 39 per-plugin input globs collapse into 3 wildcards. This also closes a gap — `t3sidebar` had no input globs at all.
- Docs, guide template (plus regenerated output), `bb-plugin-authoring` skill, and plan-doc relative links updated. The release doc now states explicitly that the directory does not encode policy and `autoInstall` does.

## Tradeoff

Promoting a plugin to store-only is now a one-line flag change rather than a visible directory move in the diff. If that review signal is worth keeping, `official-plugins.test.ts` is the natural place to assert the expected split by name.

## Verification

- `pnpm install`; typecheck + lint on server, app, templates, t3sidebar — pass.
- `turbo run build --filter=bb-app` — all 13 plugins land in `packages/bb-app/server/dist/builtin-plugins`.
- Tests: `@bb/app` 317 files, `@bb/integration-tests` 25, `@bb/cli` 44, and 12 of 13 plugin packages — pass.
- Smoke-tested `build-official-plugins.mjs` for both a successful build and the unknown-name error.

Two failures are pre-existing and unrelated; both reproduce on a clean tree:

- `@bb/server` `internal-skill-trees.test.ts` expects file mode `0644`; my umask `0002` yields `0664`.
- `bb-plugin-tasks` loses the jsdom environment in 46 tests (`window.localStorage` undefined). That package has no changes here, and the same 46 failed before the move.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

The stack contains one pull request: #1079, “Merge official-plugins into plugins.” I found no draft or untrusted stack pull request to skip.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Performance review complete. I found no actionable performance or resource-use regression.

The review built and copied all 13 plugins. It also verified Turbo exclusions, discovery paths, startup registration, runtime copies, and 52 focused tests.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Security review complete. I found no actionable security issue.

The review checked source and target paths, install policy, provenance, file modes, symlinks, manifests, dependencies, and all 13 plugin roots.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

End-to-end review complete. I found no actionable browser or runtime issue.

The live Browse route showed all 13 plugin cards and the five optional install actions. I installed Docs from `plugins/docs` and verified its detail page, app bundle, command, skill, and running service.

packages/bb-app/scripts/build.mjs
```diff
@@ -73,7 +73,7 @@ await copyBuildOutput({
 });
 // Builtin plugins are bundled at packaging time (not in @bb/server's build,
 // which source checkouts don't need — the registry falls back to the repo's
-// plugins/<name> there). Runs in apps/server so tsx + workspace imports
+// builtin-plugins/<name> there). Runs in apps/server so tsx + workspace imports
```
SawyerHood: 🚨 `slopcop/review` — This changed comment describes the wrong source fallback.

The registry resolves source checkouts to `plugins/<name>`. It uses `builtin-plugins/<name>` only for packaged output. Restore `plugins/<name>` here so the packaging note matches the resolver.

packages/templates/src/templates/bb-guide-plugins.md
```diff
@@ -424,7 +424,8 @@ in a checkout). The builtin `inline-vis` plugin renders
 `::inline-vis{file="demo.html" height="480"}` through the sidebar's
 path-shaped, sandboxed worktree HTML iframe preview; `height` is optional.
 Its card header includes an open-in-sidebar action for the source HTML file.
-The `official-plugins/` directory contains the BB Official GitHub, Docs,
-Memory, and Tasks plugins. The remaining `examples/plugins/` reference plugins
+The `plugins/` directory contains every bundled plugin: the auto-installed
+builtins and the store-only BB Official GitHub, Docs, Memory, Tasks, and T3
```
SawyerHood, line 430: 🚨 `slopcop/review` — The guide lists T3 Sidebar here but omits it from its install command list.

This file lists only GitHub, Docs, Memory, and Tasks at line 141. The configuration guide and the bb-cli skill repeat that incomplete list. Add T3 Sidebar to all discoverable lists, then regenerate the template output.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

Code-quality and architecture review complete. I found two low-severity documentation issues and posted both on their changed lines.

The code scan found no stale `official-plugins/` path, no stale `repoDirectory` contract, and no new code duplication. The build, type checks, focused tests, and GitHub checks passed.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

I found two low-severity documentation issues:

- The packaging comment names `builtin-plugins/<name>` as the source-checkout fallback. The resolver uses `plugins/<name>`.
- The updated guide includes T3 Sidebar in one official-plugin list. Install commands and other guides still omit it.

I found no actionable security or performance issue. I also found no stale directory contract or new code duplication.

The live Browse route showed all 13 plugins and five optional install actions. Docs installed from `plugins/docs` and reached a running state.

Focused tests, affected type checks, both plugin builds, the complete `bb-app` build, and all GitHub checks passed. One unchanged local machine-install test failed twice, while GitHub passed its server suite.

These findings are non-blocking. I used a comment-only review, as required.

packages/bb-app/scripts/build.mjs
SawyerHood: Fixed in 212ce1f18.

packages/templates/src/templates/bb-guide-plugins.md
SawyerHood: Fixed in 212ce1f18.

PR #1081, opened 2026-08-06
Fix Git status for newly initialized repositories

SawyerHood:
## Summary

- detect when a plain workspace becomes a Git repository after `git init`
- refresh the daemon, server, and cached runtime metadata for the repository
- support status and diff commands before the first commit exists
- increase the host daemon protocol version to 76

## Root cause

The workspace watcher stopped when `.git` did not exist. Static workspace metadata then kept Git support disabled after `git init`.

Some diff paths also used `HEAD` directly. A new repository has no valid `HEAD` commit.

## User impact

BB now enables Git status after a user adds Git to an existing workspace. BB also reports staged and untracked changes before the first commit.

## Validation

- `pnpm exec turbo run test --filter=@bb/host-workspace --filter=@bb/host-watcher --filter=@bb/host-daemon-contract --filter=@bb/host-daemon --force`
- `pnpm exec turbo run test --filter=@bb/server --force -- test/internal/internal-environment-change.test.ts`
- `pnpm exec turbo run typecheck --filter=@bb/host-workspace --filter=@bb/host-watcher --filter=@bb/host-daemon-contract --filter=@bb/host-daemon --filter=@bb/server`

### DIFF
--- a/apps/host-daemon/src/watch-manager.ts
+++ b/apps/host-daemon/src/watch-manager.ts
@@ -23,7 +23,11 @@ type StopWatching = () => void | Promise<void>;
 
 const STOP_WATCHING: StopWatching = () => undefined;
 const LOCAL_WORKSPACE_WATCH_CHANGE_KINDS: readonly WorkspaceStatusWatchChangeKind[] =
-  ["workspace-content-changed", "workspace-git-changed"];
+  [
+    "workspace-content-changed",
+    "workspace-git-changed",
+    "workspace-git-repository-created",
+  ];
 
 interface WorkspaceWatchState {
   lastLocalFingerprint: string | null;
@@ -39,12 +43,19 @@ interface WorkspaceWatchEntry {
   workspace: HostWorkspace;
 }
 
+interface RefreshWorkspaceArgs {
+  environmentId: string;
+  provision: ProvisionWorkspaceArgs;
+  workspacePath: string;
+}
+
 export interface WatchManagerOptions {
   dataDir?: string;
   hostWatcher?: HostWatcher;
   provisionWorkspace?: (
     options: ProvisionWorkspaceArgs,
   ) => Promise<HostWorkspace>;
+  refreshWorkspace?: (args: RefreshWorkspaceArgs) => Promise<HostWorkspace>;
   threadStorageRootPath?: string | null;
   onThreadStorageChanged?: (args: {
     environmentId: string;
@@ -100,6 +111,7 @@ function sameWorkspaceTarget(
 export class WatchManager {
   private readonly hostWatcher;
   private readonly provisionWorkspace;
+  private readonly refreshWorkspace;
   private readonly threadStorageTargets = new Map<
     string,
     HostDaemonWatchSetThreadStorageTarget
@@ -112,6 +124,10 @@ export class WatchManager {
   constructor(private readonly options: WatchManagerOptions = {}) {
     this.hostWatcher = options.hostWatcher;
     this.provisionWorkspace = options.provisionWorkspace ?? provisionWorkspace;
+    this.refreshWorkspace =
+      options.refreshWorkspace ??
+      ((args: RefreshWorkspaceArgs) =>
+        this.provisionWorkspace(args.provision));
   }
 
   async replaceWatchSet(watchSet: HostDaemonWatchSet): Promise<void> {
@@ -326,7 +342,10 @@ export class WatchManager {
     try {
       const changeKinds: HostDaemonEnvironmentChange[] = [];
       if (workspaceWatchKindsIncludeLocalState(pendingKinds)) {
-        if (!args.entry.workspace.isGitRepo) {
+        if (
+          pendingKinds.includes("workspace-git-repository-created") &&
+          !args.entry.workspace.isGitRepo
+        ) {
           await this.refreshGitWorkspaceMetadata(args.entry);
         }
         // A real workspace file event must reach live content consumers even
@@ -395,19 +414,22 @@ export class WatchManager {
     if (entry.workspace.isGitRepo) {
       return;
     }
-    const workspace = await this.provisionWorkspace(
-      reconnectProvisionArgsFromWorkspaceContext({
-        environmentId: entry.target.environmentId,
-        ...(this.options.dataDir
-          ? {
-              personalWorkspaceRoot: getPersonalWorkspaceRoot(
-                this.options.dataDir,
-              ),
-            }
-          : {}),
-        workspaceContext: entry.target.workspaceContext,
-      }),
-    );
+    const provision = reconnectProvisionArgsFromWorkspaceContext({
+      environmentId: entry.target.environmentId,
+      ...(this.options.dataDir
+        ? {
+            personalWorkspaceRoot: getPersonalWorkspaceRoot(
+              this.options.dataDir,
+            ),
+          }
+        : {}),
+      workspaceContext: entry.target.workspaceContext,
+    });
+    const workspace = await this.refreshWorkspace({
+      environmentId: entry.target.environmentId,
+      provision,
+      workspacePath: entry.target.workspaceContext.workspacePath,
+    });
     if (!workspace.isGitRepo) {
       return;
     }
--- a/apps/host-daemon/src/watch-manager.ts
+++ b/apps/host-daemon/src/watch-manager.ts
@@ -1,5 +1,6 @@
-import { getPersonalWorkspaceRoot } from "@bb/host-workspace";
+import type { DiscoveredWorkspaceProperties } from "@bb/domain";
 import {
+  getPersonalWorkspaceRoot,
   provisionWorkspace,
   type HostWorkspace,
   type ProvisionWorkspaceArgs,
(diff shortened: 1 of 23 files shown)


SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

The stack contains one pull request: #1081. The second scan found no child pull requests. No pull request was skipped.

I will run security, code-quality, architecture, performance, and end-to-end phases. I will post a comment after each phase.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

End-to-end phase: failed.

I started the isolated development app and opened a personal thread in the browser. The environment first reported `isGitRepo: false`. After `git init -b main`, the live watcher updated it to `isGitRepo: true` with branch `main`.

The CLI status without a merge base correctly reported an unborn branch, one staged file, one untracked file, and two insertions. The browser then requested status relative to `main`. The thread panel showed `Git status: Unknown` because `git rev-list main...HEAD` fails before the first commit.

This issue blocks the main browser workflow that this pull request intends to fix.

packages/host-workspace/src/workspace.ts
```diff
@@ -528,7 +529,8 @@ async function readHeadNumstat(
   workspacePath: string,
   timeoutMs?: number,
 ): Promise<string> {
-  const result = await runGit(["diff", "--numstat", "-z", "HEAD", "--"], {
+  const diffBase = await resolveUncommittedDiffBase(workspacePath, timeoutMs);
```
SawyerHood: 🚨 `slopcop/review` — [P1] Handle merge-base status for an unborn branch

`resolveUncommittedDiffBase` fixes the working-tree numstat path, but the browser requests `getStatus({ mergeBaseBranch: "main" })`. A new repository has no `main` ref or `HEAD` commit, so `readMergeBaseStatus` runs `git rev-list main...HEAD` and the complete status request fails. I reproduced this in the development app after `git init`: the environment changed to Git, the CLI status worked, but the thread panel showed `Git status: Unknown`. Please skip merge-base comparison for an unborn checkout, or return an empty merge-base result, and add a test with `mergeBaseBranch: "main"` before the initial commit.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Security phase: one high-severity finding.

A personal workspace can create a `.git` file that points to Git data outside the personal workspace. The new discovery path accepts that repository without a containment check. BB can then read diffs and run Git mutations through the host daemon. A linked-worktree target can also become an external provider write root after reprovision.

This path crosses the personal workspace boundary. The protocol update, message authorization, exact environment path check, and fixed Git arguments otherwise look correct. I found no command injection in the new Git calls.

packages/host-workspace/src/provision.ts
```diff
@@ -718,11 +718,14 @@ async function provisionPersonalWorkspace(
     throw error;
   }
 
+  const isGitRepo = await detectGitRepo(targetPath);
```
SawyerHood: 🚨 `slopcop/review` — [P1] Keep personal Git metadata inside the personal workspace

`detectGitRepo` accepts a `.git` file that points to another local repository. This new reprovision path therefore promotes a personal workspace even when its Git directory or common directory sits outside `targetPath`. BB can then execute daemon-owned status, diff, and commit operations against that external Git data. A linked-worktree target can also expose external Git paths as provider write roots after reprovision. Please validate that the resolved Git directory and common directory stay inside a personal workspace. Keep external metadata support for trusted unmanaged and managed-worktree provision types.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Code-quality and architecture phase: two findings.

First, the watcher compares canonical event paths with the original workspace path. Symbolic path components can prevent repository promotion. The new watcher test fails consistently on macOS for this reason.

Second, repository promotion updates only the `WatchManager` copy of the workspace. `RuntimeManager` keeps `isGitRepo: false`. Command resolution adds a disk probe around this stale value, while other daemon paths still use it directly.

The repository already has a daemon-owned runtime entry. One refresh operation should update that entry and serve both the watcher and command paths. The protocol version increase to 76 is correct. I found no useful duplicate abstraction beyond this split refresh logic.

packages/host-watcher/src/workspace-status-watcher.ts
```diff
@@ -341,6 +355,48 @@ export class WorkspaceStatusWatcher {
       this.changeKinds.add(changeKind);
     }
     this.changeScheduler.schedule();
+    if (
+      spec.kind === "workspace-root" &&
+      spec.options?.ignore?.includes(".git") !== true &&
+      changeEvent.changedPaths.some((changedPath) =>
+        isPathInsideDotGit(this.args.cwd, changedPath),
```
SawyerHood: 🚨 `slopcop/review` — [P1] Compare paths in the same canonical form

`spec.rootPath` is canonical, but this check uses the original `cwd`. A valid event can therefore compare `/private/var/.../.git` against `/var/...`, or hit the same issue through any symbolic workspace path. The promotion does not run, so the environment remains non-Git. The new test fails twice on macOS because it sends the non-canonical form. Use `spec.rootPath` here, and send the canonical event path in the test.

apps/host-daemon/src/watch-manager.ts
```diff
@@ -366,23 +372,66 @@ export class WatchManager {
...
+      return;
+    }
+
+    const [branchName, resolvedDefaultBranch, sharedRefsFingerprint] =
+      await Promise.all([
+        workspace.getCurrentBranch(),
+        workspace.getDefaultBranch(),
+        workspace.getSharedGitRefsFingerprint(),
+      ]);
+    entry.workspace = workspace;
```
SawyerHood, line 443: 🚨 `slopcop/review` — [P2] Refresh the runtime workspace entry

This assignment updates only the watch entry. `RuntimeManager` keeps its original workspace object with `isGitRepo: false`. The new command-resolution disk probe works around that stale cache for Git-gated commands, but repeated provision and other daemon paths can still report non-Git metadata. A later provision can overwrite the correct server metadata. Add one daemon-owned refresh operation that replaces the runtime workspace, then reuse it here and in command resolution.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Performance phase: two material findings and two secondary costs.

A watched plain workspace now runs Git discovery after every file-change burst. It also emits a work-status change for each burst. Build output can therefore cause repeated Git processes, WebSocket messages, and server notifications before any `.git` directory exists.

Normal repositories also pay one extra Git process for each status and uncommitted-diff request. Trace measurements showed `getStatus` increasing from 9 to 10 Git processes, and `getDiff` increasing from 5 to 6.

The stale runtime cache causes another repeated Git probe for each later Git command. A new empty personal workspace also runs one discovery process although creation already proved that the directory did not exist. These last two costs can share the runtime-refresh fix and an existence guard.

packages/host-watcher/src/workspace-status-watcher.ts
```diff
@@ -215,13 +222,20 @@ export class WorkspaceStatusWatcher {
   }
 
   private async startAsync(): Promise<void> {
-    if (!(await pathExists(path.join(this.args.cwd, ".git")))) {
+    const rootPath = await resolveWatchRootPath(this.args.cwd);
+    if (this.disposed) {
       return;
     }
-    if (this.disposed) {
+    if (!(await pathExists(path.join(this.args.cwd, ".git")))) {
```
SawyerHood, line 229: 🚨 `slopcop/review` — [P2] Do not probe Git after every plain-workspace change burst

This recursive watch has no ignore rules. Each ordinary file burst reaches `WatchManager`, which reprovisions the still-plain workspace and starts `git rev-parse --is-inside-work-tree`. It also emits a work-status notification. A busy build tree can repeat this work at the debounce cadence although only creation of `.git` matters. Send a distinct repository-promotion signal when the watcher sees `.git`, and run discovery only for that signal.

packages/host-workspace/src/workspace.ts
SawyerHood: 🚨 `slopcop/review` — [P2] Keep the common committed-repository path to one diff process

This preflight adds `git rev-parse --verify HEAD` to every status and uncommitted-diff request. Trace measurements increased `getStatus` from 9 to 10 Git processes and `getDiff` from 5 to 6. Only unborn repositories need the empty-tree fallback. Run the existing `HEAD` diff first, then retry with the empty tree only when Git reports a missing `HEAD`.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

Final review summary for stack: #1081 only.

Blocking findings:

1. The browser requests status relative to `main`. An unborn repository has no `main` ref or `HEAD` commit, so the thread panel shows `Git status: Unknown`.
2. A personal workspace can use a `.git` target outside its managed root. Host-daemon Git operations can then cross the personal workspace boundary.

Medium finding:

3. Canonical watcher event paths can differ from the original workspace path. Repository promotion can fail for a symbolic workspace path.

Low-severity follow-up:

- `WatchManager` refreshes its workspace copy but leaves the runtime copy stale.
- Plain-workspace bursts can repeat Git discovery and notifications.
- Normal status and diff requests start one additional Git process.

Validation:

- I reviewed the exact head `edc3383e9799b3fbcb1177da02f693401204804b`.
- The five affected package type checks passed.
- The focused tests produced 204 passes and one macOS watcher-test failure. I reproduced that failure twice.
- The browser end-to-end test reproduced the unborn-branch status failure.
- GitHub server, integration, smoke, and static-check jobs passed. Three other jobs failed during GitHub action downloads.
- `git diff --check` passed.

I used a comment-only review. I did not approve this pull request or request changes through GitHub.

PR #1099, opened 2026-08-06
fix(app): submit on Magic Keyboard Enter in the iPadOS PWA composer

salemsayed:
## Problem

In the installed iPad PWA, there is no way to send a prompt from a Magic Keyboard. Pressing Enter in the composer always inserts a newline, and the only way to submit is to tap the send button.

The composer gates Enter-to-submit on pointer coarseness alone:

```ts
const canSubmitWithEnterKey = !isPointerCoarse;
```

An iPad reports a coarse pointer whether or not a hardware keyboard is attached, so the hardware Enter is permanently routed to the newline path. Making this device-wide would break the software keyboard, whose Return must stay a newline.

Two details make the physical key hard to detect on this platform:

- ProseMirror deliberately defers iOS Enter handling and later replays a **synthetic** Enter into `handleKeyDown`, so the software keyboard can finish its own DOM mutation first. That synthetic event no longer carries the physical `code`.
- The iPad software keyboard's Return arrives with an empty `code`, while a Magic Keyboard Enter arrives as `code: "Enter"` (or `"NumpadEnter"`).

## Fix

A custom ProseMirror DOM `keydown` hook now looks at the **original** event before ProseMirror's deferral, and only on iPadOS WebKit. When that original event carries a physical `Enter`/`NumpadEnter` code, the existing key handler is invoked with `isOriginalIPadHardwareEnter`, which enables Enter-to-submit for that one event:

```ts
const canSubmitWithEnterKey = !isPointerCoarse || isOriginalIPadHardwareEnter;
```

Device detection combines an Apple WebKit vendor/UA check with either the legacy `iPad` platform or the modern desktop-like `MacIntel` platform plus `maxTouchPoints > 2`, which is what distinguishes current iPadOS from macOS.

Everything else keeps ProseMirror's normal path, including the software keyboard's Return, which still inserts a newline.

## Steer/queue mapping is unchanged

No submit routing changed. `canSubmitWithEnterKey` simply moved from a component-level constant to a per-event value inside the key handler.

- `steerActiveThreadOnEnter` off (default): Enter queues a follow-up, Command+Enter steers.
- `steerActiveThreadOnEnter` on: Enter steers, Command+Enter queues.
- Shift+Enter inserts a newline; in zen mode unmodified Enter inserts a newline too.
- On iPadOS, a Magic Keyboard Enter/Command+Enter now reaches exactly these existing paths instead of being swallowed.

`FollowUpPromptBox.test.tsx` consolidates the two overlapping shortcut tests into one parameterized case that asserts both settings, including the scroll-to-bottom behavior that differs between queue and steer.

## Android and desktop safeguards

The interception is scoped as narrowly as possible, and both non-iPad coarse-pointer cases are covered by tests:

- **Android** (Pixel Tablet, `Google Inc.` vendor, `Linux armv8l`) reports `code: "Enter"` from its software keyboard. It fails the Apple WebKit check, is never intercepted, and keeps Return as a newline.
- **Windows touch hybrids** (`Win32`, `maxTouchPoints: 10`) likewise fail the vendor check and are untouched.
- **Fine-pointer desktop** is entirely unaffected — the hook is a no-op there, and Enter-to-submit continues to run through the existing `!isPointerCoarse` path.

## IME and typeahead

Composition is excluded on several fronts, so confirming a candidate never submits:

- `view.composing`, `event.isComposing`, and `event.keyCode === 229` all bail out.
- The hook mirrors ProseMirror's own Safari `compositionend` → `keydown` safeguard with a 500 ms window. Because this DOM hook runs *before* ProseMirror's keydown handler, skipping that safeguard would submit an IME candidate confirmation.
- An event suppressed by that window is recorded in a `WeakSet`, so if it later arrives through the normal handler it is still refused rather than double-processed.

Typeahead is unaffected: with an open command or mention menu, a Magic Keyboard Enter applies the highlighted suggestion (`/` → `/review `, with the mention resource attached) and does **not** submit, matching the fine-pointer behavior.

## Performance

- `isIPadOSWebKit()` is computed once per mount via `useMemo`. The value is stable for the lifetime of the page, so no additional media-query or event listener is registered.
- The keydown hook returns after a few boolean checks for every non-iPad platform and every non-Enter key, so no per-keystroke cost is added anywhere else.
- The composition guard is an O(1) timestamp comparison plus a `WeakSet` lookup. Keying the set on the event object means entries are collected along with the event, so it cannot grow unbounded.
- `canSubmitWithEnterKey` left the `useCallback` dependency list and was replaced by `isPointerCoarse`, so handler identity churns no more often than before.
- Assigning `handleEditorKeyDownRef` moved from `useEffect` to `useLayoutEffect`, so the ref is current before any keydown can be dispatched after a commit and a stale handler cannot fire in the same frame.

## Documentation

Per the repository's rule that a user-facing behavior change updates its discoverable surfaces in the same change, the Enter behavior is now described in:

- `docs/configuration.md`
- the built-in `bb-cli` skill (`SKILL.md` and `references/app-settings.md`)
- the bb guide template `bb-guide-customization.md` and its generated counterpart

Each now states the typeahead exception, Shift+Enter and zen-mode newlines, the coarse-pointer software-keyboard path, and the iPadOS Magic Keyboard behavior.

## Validation

- **Tests:** 317 files and 2,384 app tests passed.
- **Typecheck:** app and templates passed.
- **Lint:** app passed, with existing warnings only.
- **Diff check:** passed.
- **Device verification:** confirmed on a physical iPad running the installed PWA — a Magic Keyboard Enter submits, and the software keyboard's Return still inserts a newline.

New coverage in `PromptBoxInternal.test.tsx` spans fine-pointer Enter, Magic Keyboard Enter, software-keyboard Enter, Android, a Windows touch hybrid, Shift+Enter, Command+Enter, IME composition, the post-`compositionend` window, zen mode, and typeahead application.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am now reviewing this pull request under the security, code quality, performance, architecture, and end-to-end guidelines.

apps/app/src/components/promptbox/PromptBoxInternal.test.tsx
```diff
@@ -439,6 +441,59 @@ function mockPointerCoarse(matches: boolean): () => void {
...
+        "maxTouchPoints",
+        maxTouchPointsDescriptor,
+      );
+    } else {
+      Reflect.deleteProperty(navigator, "maxTouchPoints");
+    }
+    platformMock.mockRestore();
+    vendorMock.mockRestore();
+    userAgentMock.mockRestore();
+  };
```
SawyerHood, line 483: 🚨 `slopcop/review` — These tests do not activate ProseMirror's iOS branch.

ProseMirror reads `navigator` once during module load. This helper changes it after the TipTap import. Thus, `browser.ios` and `browser.safari` stay false in this suite. The tests cover this custom hook, but not ProseMirror's 200 ms replay. Please set the iPad identity before module import. Then add a test that confirms one submission after the replay timeout.

apps/app/src/components/promptbox/PromptBoxInternal.tsx
```diff
@@ -1145,8 +1170,12 @@ export function PromptBoxInternal({
   const skipEditorChangeRef = useRef(false);
   const editorValueKeyRef = useRef("");
   const triggerKeyRef = useRef("");
-  const handleEditorKeyDownRef = useRef<(event: KeyboardEvent) => boolean>(
-    () => false,
+  const handleEditorKeyDownRef = useRef<
+    (event: KeyboardEvent, isOriginalIPadHardwareEnter?: boolean) => boolean
+  >(() => false);
+  const compositionEndedAtRef = useRef(Number.NEGATIVE_INFINITY);
+  const postCompositionKeyDownEventsRef = useRef(
```
SawyerHood: 🚨 `slopcop/review` — This expression creates a new WeakSet during every render.

React keeps only the first value that `useRef` receives. The composer renders after each controlled value change. Thus, each edit also creates a discarded WeakSet. Please use a lazy ref initializer, or initialize a nullable ref once.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: This change lets an iPad Magic Keyboard send a prompt with Enter. The screen keyboard Return key still adds a new line.

I found two items.

1. The new tests change the iPad identity after ProseMirror loads. ProseMirror reads that identity only once. Thus, the tests do not cover its iOS replay path. This path is the main reason for the change. I left an inline comment with a test fix.
2. `useRef(new WeakSet())` creates a discarded WeakSet during every render. I left an inline comment with a lazy initialization fix.

The security review found no issue. The change adds no server, network, authentication, storage, or protocol surface.

The architecture search found no other iPad helper or rich composer to reuse. The change does copy a small composition check and ProseMirror's 500 ms guard. I do not see a larger refactor requirement.

Validation passed for head `2481c025cd0c6be1263eefc9ca50a1cc473cd474`:

- The app type check passed through Turbo.
- All 321 app test files passed. All 2,425 app tests passed.
- The browser check used an iPad identity and a coarse pointer. Shift+Enter added a new line. Hardware Enter reached the submit shortcut and prevented the browser default.
- The browser check used Chromium. It did not reproduce the WebKit engine or ProseMirror's true iOS module-load state.

apps/app/src/components/promptbox/PromptBoxInternal.tsx
SawyerHood: Fixed in dfbdbf5e. `useRef` now receives `null`, and a `usePostCompositionKeyDownEvents` hook fills it once with `??=`, so no `WeakSet` is built and discarded per render.

apps/app/src/components/promptbox/PromptBoxInternal.test.tsx
SawyerHood: Fixed in dfbdbf5e, which adds `PromptBoxInternal.ipados.test.tsx`. It claims the iPad identity in a `vi.hoisted` block, before any import, so ProseMirror's `browser.safari` and `browser.ios` are genuinely on. Two tests: a Magic Keyboard Enter submits exactly once with no replayed second submit after the 200 ms window, and a software Return (`code: ""`) becomes a newline only after that replay. Both fail if the hoisted block is removed.

One correction for the record: the synthetic replay does still carry `code: "Enter"` — ProseMirror's `keyEvent` sets `event.key = event.code = key` (`prosemirror-view/dist/index.js:137`). The hook is needed because ProseMirror's iOS branch swallows the original event and never passes it to `handleKeyDown`, not because the replay loses the physical code.

PR #1109, opened 2026-08-06
Move T3 sidebar plugin to examples

SawyerHood:
## Summary

- Move the T3 sidebar plugin from `plugins/` to `examples/plugins/`.
- Remove T3 Sidebar from the bundled official plugin catalog.
- Add local icon, select, and style helpers so the example does not use BB's private shared UI package.
- Update the plugin guides, reference documents, catalog tests, generated templates, and lockfile.

## Why

T3 Sidebar is a reference implementation for the sidebar replacement APIs. It should remain an installable example instead of a bundled product plugin.

## Impact

New BB builds will not include T3 Sidebar in the official plugin catalog. Developers can install it from a checkout with:

```sh
bb plugin install ./examples/plugins/t3sidebar
```

## Validation

- `pnpm exec turbo run typecheck --filter=bb-plugin-t3sidebar --filter=@bb/server --filter=@bb/templates`
- `pnpm exec turbo run test --filter=bb-plugin-t3sidebar --force`
- `pnpm exec vitest run --config vitest.config.ts test/services/plugins/official-plugins.test.ts test/services/plugins/builtin-plugins.test.ts`
- `pnpm run bb -- plugin build examples/plugins/t3sidebar`

PR #1108, opened 2026-08-06
Gate new onboarding behind an experiment

SawyerHood:
## Summary

- add a default-off newOnboarding experiment with persisted database support
- gate the onboarding flow and provider checks on the experiment
- expose the experiment through Settings and the bb CLI
- keep replay-onboarding useful by enabling the experiment before replay
- update generated contracts, guides, and migration rewind coverage

## Tests

- type checks passed for app, CLI, database, desktop, domain, plugin SDK, server, and templates
- 47 focused tests passed across app, CLI, database, domain, and server

PR #1136, opened 2026-08-07
Fix the plugin catalog route test count

SawyerHood:
## Summary

- Derive `pluginCount`, `includedPluginCount`, and `optionalPluginCount` in `plugin-catalog-routes.test.ts` from the bundled plugin registry instead of hardcoding `13 / 8 / 5`.

## Why

CI is red on `main`. Commit f13c2d35f ("Move T3 sidebar plugin to examples", #1109) removed `t3sidebar` from `OFFICIAL_PLUGINS`, so the bundled catalog went from 13 plugins to 12. That PR updated `plugin-catalog-service.test.ts`, which derives its counts from the registry, but it missed the route test, which hardcoded the numbers.

```
AssertionError: expected { catalog: { pluginCount: 12, ... } }
                to match object { catalog: { pluginCount: 13, ... } }
```

#1109 landed without a green run: its own CI run (31186905645) was cancelled when #1108 pushed on top of it.

The route test now uses the same source of truth as the service test, so a future plugin add or remove will not break it again.

## Tests

- `pnpm exec turbo run test --filter=@bb/server -- plugin-catalog` — 7 passed.
- `pnpm exec turbo run test --filter=@bb/server` — 1336 passed. One unrelated local-only failure in `internal-skill-trees.test.ts`, which asserts file mode `0o644` but writes the file with no explicit `chmod`. It follows the umask, so it fails under umask `0002` and passes on CI runners under `0022`.
- `pnpm exec turbo run typecheck --filter=@bb/server` — pass.
- CI run 31187487485 on `main`: `Tests (server)` was the only failing job. All other jobs passed.

## Risk

Test-only change. No product code is touched.

### DIFF
--- a/apps/server/test/services/plugin-catalog/plugin-catalog-routes.test.ts
+++ b/apps/server/test/services/plugin-catalog/plugin-catalog-routes.test.ts
@@ -3,6 +3,11 @@ import { Hono } from "hono";
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import { registerPluginCatalogRoutes } from "../../../src/routes/plugin-catalog.js";
 import { createPluginCatalogService } from "../../../src/services/plugin-catalog/plugin-catalog-service.js";
+import {
+  BUILTIN_PLUGINS,
+  BUNDLED_PLUGINS,
+  OFFICIAL_PLUGINS,
+} from "../../../src/services/plugins/builtin-registry.js";
 
 describe("plugin catalog routes", () => {
   let db: DbConnection;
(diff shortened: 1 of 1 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am now reviewing this pull request for security, code quality, performance, duplication, architecture, and test coverage.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** The test used a fixed plugin count. One plugin left the catalog, so the fixed count became wrong. This change counts the lists directly.

I found no blocking issue and no inline finding.

- **Security:** This change affects test code only. The new registry import adds no network, file-write, authentication, or dependency risk.
- **Code quality:** The three expressions match the registry and the sibling service test. They still detect an incorrect `autoInstall` filter.
- **Performance:** The imported module already exists in the test graph. This change adds no runtime cost or material test cost.
- **Validation:** The forced plugin-catalog test passed at the pull request SHA. Both files and all seven tests passed.
- **CI:** All completed GitHub checks passed. The compatibility job skipped as expected.

The wider scan found old `13 / 8 / 5` values in unrelated mock and schema fixtures. Those values do not control catalog behavior.

The scan also found two pre-existing plugin-management UI trees. That architecture issue does not affect this small test fix.

I did not start a browser test because the pull request changes no runtime route or UI behavior. The focused test sends Hono requests through the real route and database.

PR #1080, opened 2026-08-06
Load user Pi configuration in the bundled runtime

SawyerHood:
## Summary

- Ship a pinned physical Pi 0.84.0 package tree.
- Load global and project Pi settings through Pi's full service path.
- Load configured packages, extensions, skills, prompts, themes, context, authentication, and models.
- Include extension models in the BB model list.
- Preserve Pi shell settings when BB adds thread environment values.
- Test a custom extension provider, extension tool, and BB dynamic tool from an installed package.

## Why

The bridge created in-memory settings and bundled Pi into one file. This bypassed the user's Pi configuration.

The single-file package also prevented extensions from sharing the host Pi modules.

## Impact

Users keep their Pi settings and extensions while BB controls the Pi package version.

Users do not need an installed Pi executable.

## Validation

- `pnpm exec turbo run typecheck --filter=@bb/agent-runtime --filter=@bb/host-daemon --filter=@bb/server --filter=bb-app --force`
- 810 agent runtime tests
- 479 host daemon tests
- 1,335 server tests
- 48 BB application tests
- `pnpm exec turbo run smoke:tarball --filter=bb-app --force`
- A real model request through the packaged bridge with an existing Pi configuration

PR #1148, opened 2026-08-07
Keep the Pi model list alive when an extension is broken

### LINKED ISSUE #1086: Port primeExtensions() to production Pi bridge so commandcode/extension models appear in stock Pi provider
## Problem

The stock `pi` provider in the bb desktop app only shows built-in providers (`openai-codex`, `opencode-go`) — extension-registered providers like `commandcode`, `router`, and `deepseek` never appear. Users must create a custom ACP agent to access CommandCode's 52+ models (DeepSeek V4, Claude Sonnet/Opus 5, GPT-5.6, Gemini, Kimi, GLM, Qwen, etc.).

## Root cause

The production `daemon-bundle.mjs` (bb-app@0.35.1) calls `ModelRuntime.create().getAvailable()` directly when listing models. Extensions register through `createAgentSession()`, but the bridge never creates a session — so pi-commandcode-provider, pi-model-router, and other extensions never fire, and their models stay invisible.

## Working fix (dev bridge)

The dev bridge at `packages/agent-runtime/src/pi/bridge/model-runtime.ts` already has a `primeExtensions()` function that creates a disposable Pi session before any model query:

```ts
async function primeExtensions(modelRuntime: ModelRuntime): Promise<void> {
  if (extensionsPrimed) return;
  const resourceLoader = new DefaultResourceLoader({ cwd, agentDir });
  await resourceLoader.reload();
  const { session } = await createAgentSession({ modelRuntime, cwd, resourceLoader, ... });
  session.dispose();
  extensionsPrimed = true;
}
```

After priming, `model/list` returns ALL models including 52 commandcode models, router profiles, etc.

## Evidence

- Verified: dev bridge shows 81 models (25 built-in + 52 commandcode + 4 router)
- Verified: production bridge shows 25 models (openai-codex + opencode-go only)
- Reproducible on bb-app@0.35.1 with pi@0.83.0 and pi-commandcode-provider installed

## Workaround until shipped

Users can create a custom ACP agent pointing at `pi-acp` which spawns `pi --mode rpc` and inherits extension registrations.

SawyerHood:
## Summary

- Verify that the `primeExtensions()` port asked for in #1086 already landed in #1080, end to end against the built bridge.
- Stop one broken Pi extension from emptying the whole model picker.
- Stop an extension model without `input` types from failing the whole model list.

## Why

#1086 asks to port `primeExtensions()` to the production Pi bridge so extension providers such as `commandcode` appear in the stock Pi provider.

That port already landed in #1080 (`8d7bf6124`). The bridge builds its model runtime through `createConfiguredPiServices()`, and Pi 0.84.0 flushes `pendingProviderRegistrations` inside `createAgentSessionServices()`. A disposable priming session is not necessary.

I confirmed this with the real extension from the issue, not a mock: `pi-commandcode-provider@0.5.0` installed from npm, driven by a real `model/list` JSON-RPC request to the built `dist/bb-pi-bridge.mjs`. It returns **51 CommandCode models**.

That test found two defects that still produce the exact symptom in the issue.

### 1. A broken extension emptied the whole picker

`collectServiceErrors()` throws on any extension load error, and `getPiModelRuntime()` passed that on, so `model/list` answered with a JSON-RPC error. The user lost every model on the machine, built-in providers included, because of one bad third-party extension.

Repro before this change, with a working extension and a broken one installed together:

```
Error: Failed to load Pi extension "/tmp/pi-broken/ext/broken.ts": Cannot find module '@not-installed/nope'
```

The fix splits the service builder. `loadConfiguredPiServices()` reports the problems, and `createConfiguredPiServices()` still throws. Model listing writes the problems to stderr and lists what loaded. Thread start still fails loudly, so a run never uses a partial configuration in silence.

After, through the shipped bundle:

```
pi bridge: Failed to load Pi extension "/tmp/pi-broken/ext/broken.ts": ...
TOTAL 2
  commandcode-fake: 2
```

### 2. An extension model without `input` crashed the list

`input: [...model.input]` threw `TypeError: model.input is not iterable`. Pi types `input` as required, but an extension registers its models from plain JavaScript and Pi does not validate them. One malformed model hid every other provider.

## Impact

A user with a broken or outdated Pi extension keeps a working model picker and gets the reason on stderr, instead of an empty picker with no explanation.

## Note on model counts

Extension models appear only when the provider has credentials. With `COMMANDCODE_API_KEY` unset, `getAvailable()` filters CommandCode out and the count returns to 0. That is correct Pi behavior, not a BB defect, but it may explain part of what the reporter saw.

`HOST_DAEMON_PROTOCOL_VERSION` is unchanged: this is internal to the bridge process and the `model/list` result shape is the same.

## Validation

- 823 `@bb/agent-runtime` tests, including 3 new regression tests
- 482 `@bb/host-daemon` tests
- `pnpm exec turbo run typecheck --filter=@bb/agent-runtime --filter=@bb/host-daemon`
- ESLint on `packages/agent-runtime/src/pi`
- All three scenarios driven over JSON-RPC against the built `bb-pi-bridge.mjs`, not only against source

Fixes #1086

### DIFF
--- a/packages/agent-runtime/src/pi/bridge/model-list.ts
+++ b/packages/agent-runtime/src/pi/bridge/model-list.ts
@@ -1,7 +1,7 @@
 import type { AvailableModel } from "@bb/domain";
 import { getSupportedThinkingLevels } from "@earendil-works/pi-ai";
 import type { ModelRuntime } from "@earendil-works/pi-coding-agent";
-import { buildPiAvailableModels } from "../model-list.js";
+import { buildPiAvailableModels, type PiCatalogModel } from "../model-list.js";
 
 const NETWORK_REFRESH_TIMEOUT_MS = 5_000;
 
@@ -99,6 +99,38 @@ async function ensureNetworkRefresh(modelRuntime: ModelRuntime): Promise<void> {
   }
 }
 
+type PiAvailableModel = Awaited<
+  ReturnType<ModelRuntime["getAvailable"]>
+>[number];
+
+/**
+ * Narrow one Pi model for the list builder, or drop it.
+ *
+ * Pi marks `id`, `name`, `provider`, and `input` as required, but an extension
+ * registers its models from plain JavaScript and Pi does not validate them. A
+ * missing field threw here or inside the builder, which failed the whole call
+ * and hid every other provider's models. Skip the bad model instead.
+ */
+function toPiCatalogModel(model: PiAvailableModel): PiCatalogModel | undefined {
+  if (
+    typeof model.id !== "string" ||
+    model.id.length === 0 ||
+    typeof model.name !== "string" ||
+    typeof model.provider !== "string" ||
+    model.provider.length === 0
+  ) {
+    return undefined;
+  }
+  return {
+    id: model.id,
+    input: Array.isArray(model.input) ? [...model.input] : [],
+    name: model.name,
+    provider: model.provider,
+    reasoning: model.reasoning === true,
+    supportedThinkingLevels: getSupportedThinkingLevels(model),
+  };
+}
+
 export async function listPiBridgeModels(modelRuntime: ModelRuntime): Promise<{
   models: AvailableModel[];
   selectedOnlyModels: AvailableModel[];
@@ -106,19 +138,19 @@ export async function listPiBridgeModels(modelRuntime: ModelRuntime): Promise<{
   await ensureNetworkRefresh(modelRuntime);
   const availableModels = await modelRuntime.getAvailable();
 
-  return buildPiAvailableModels({
-    models: availableModels.map((model) => ({
-      id: model.id,
-      // An extension registers its models from plain JavaScript, so Pi's
-      // required `input` can still be missing at runtime. Reading it blindly
-      // threw and failed the whole list, which hid every other provider too.
-      input: Array.isArray(model.input) ? [...model.input] : [],
-      name: model.name,
-      provider: model.provider,
-      reasoning: model.reasoning,
-      supportedThinkingLevels: getSupportedThinkingLevels(model),
-    })),
-  });
+  const models: PiCatalogModel[] = [];
+  for (const model of availableModels) {
+    const catalogModel = toPiCatalogModel(model);
+    if (catalogModel) {
+      models.push(catalogModel);
+      continue;
+    }
+    process.stderr.write(
+      `pi bridge: skipped an incomplete model from provider "${String(model.provider)}"\n`,
+    );
+  }
+
+  return buildPiAvailableModels({ models });
 }
 
 /** @internal Test seam: reset the per-process refresh latch. */
--- a/packages/agent-runtime/src/pi/bridge/model-list.ts
+++ b/packages/agent-runtime/src/pi/bridge/model-list.ts
@@ -109,7 +109,10 @@ export async function listPiBridgeModels(modelRuntime: ModelRuntime): Promise<{
   return buildPiAvailableModels({
     models: availableModels.map((model) => ({
       id: model.id,
-      input: [...model.input],
+      // An extension registers its models from plain JavaScript, so Pi's
+      // required `input` can still be missing at runtime. Reading it blindly
+      // threw and failed the whole list, which hid every other provider too.
+      input: Array.isArray(model.input) ? [...model.input] : [],
       name: model.name,
       provider: model.provider,
       reasoning: model.reasoning,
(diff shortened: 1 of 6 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am reviewing this pull request for security, code quality, architecture, performance, tests, and product behavior.

packages/agent-runtime/src/pi/bridge/model-runtime.ts
```diff
@@ -14,8 +14,17 @@ export function getPiModelRuntime(cwd = process.cwd()): Promise<ModelRuntime> {
...
+  const modelRuntimePromise = loadConfiguredPiServices({ cwd: resolvedCwd })
+    .then(({ configErrors, services }) => {
+      // One broken extension must not empty the picker. Pi still registered
+      // every provider it did load, so report the problem and list those
+      // models. Thread start keeps failing on the same configuration, so the
+      // user still learns about it before a run uses a partial setup.
+      for (const configError of configErrors) {
+        process.stderr.write(`pi bridge: ${configError}\n`);
+      }
+      return services.modelRuntime;
```
SawyerHood, line 33: 🚨 `slopcop/review` — P2: Do not cache a partial runtime after configuration errors

`loadConfiguredPiServices()` now resolves with `configErrors`. The promise therefore stays in `modelRuntimePromises`.

If the user fixes the extension, later `model/list` calls still use the partial runtime. The repaired provider stays absent until the long-lived bridge restarts.

Delete this cache entry when configuration errors exist. Add a test where the first load has an error and the next load returns the repaired runtime.

packages/agent-runtime/src/pi/bridge/model-list.ts
```diff
@@ -109,7 +109,10 @@ export async function listPiBridgeModels(modelRuntime: ModelRuntime): Promise<{
   return buildPiAvailableModels({
     models: availableModels.map((model) => ({
       id: model.id,
-      input: [...model.input],
+      // An extension registers its models from plain JavaScript, so Pi's
+      // required `input` can still be missing at runtime. Reading it blindly
+      // threw and failed the whole list, which hid every other provider too.
+      input: Array.isArray(model.input) ? [...model.input] : [],
```
SawyerHood: 🚨 `slopcop/review` — P2: Validate each model before the list builder uses it

This guard handles a missing `input` value, but Pi can also return a JavaScript extension model without an `id`. `buildPiAvailableModels()` then calls `id.endsWith()` and drops the complete model list.

Skip each malformed model at this boundary. Add a test with one model without an identifier and one valid model.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

The model list is like a toy shelf. One broken toy no longer hides all the good toys.

Two problems remain. A repaired toy stays hidden until the room restarts. A toy without a name tag can still hide the shelf.

### Findings

- **P2 — A partial runtime stays cached.** When an extension fails, the bridge caches the partial runtime for that workspace. A later extension repair does not restore its models until the long-lived bridge restarts.
- **P2 — One malformed model can still drop the complete list.** The new guard handles a missing `input` value. A JavaScript extension model without an `id` still causes `buildPiAvailableModels()` to throw.

I left inline comments with a test case for each problem.

### Other checks

I found no security problem. Project trust still controls extension load, and thread start still uses the strict configuration path.

I found no performance regression. Clean workspaces keep the same one-time service load, and the new array check has a constant cost.

The tolerant loader and strict wrapper form a clear architecture. I found no duplicate error collector or duplicate fallback logic.

### Validation

- The exact head SHA passed the `@bb/agent-runtime` Turbo type check.
- All 823 `@bb/agent-runtime` tests passed.
- `git diff --check` passed.
- All applicable GitHub checks passed.
- I did not run a browser test. This host-local Pi configuration path has no direct browser route.

This review uses comments only. It does not approve the pull request or request changes.

packages/agent-runtime/src/pi/bridge/model-runtime.ts
SawyerHood: Fixed in 493bb9158. `getPiModelRuntime()` now deletes the memo when `configErrors` is non-empty, so the next call reloads.

Verified end to end against the built `bb-pi-bridge.mjs`, repairing the extension without restarting the bridge process:

```
1. broken extension present: commandcode-fake: 2
2. after repair, same process: commandcode-fake: 2, repaired: 1
```

Added two tests: one where the first load reports an error and the next returns the repaired runtime, and one asserting a clean load still memoizes.

packages/agent-runtime/src/pi/bridge/model-list.ts
SawyerHood: Fixed in 493bb9158. You are right that `id` is equally extension-supplied, and I confirmed the failure: `isModelAlias()` calls `id.endsWith()` and throws, which fails the whole list.

Replaced the single-field guard with boundary validation. `toPiCatalogModel()` now drops any model missing a usable `id`, `name`, or `provider`, defaults `input`, and writes the skip to stderr. Added a test with one model without an identifier alongside a valid one.

PR #1149, opened 2026-08-07
Fix Cmd+W crash in About window

### LINKED ISSUE #1100: A JavaScript error occurred in the main process
open "About bb" and press cmd+w, I got 

A JavaScript error occurred in the main process

Uncaught Exception:
TypeError: Cannot read properties of null (reading 'close')
at Object.closeWindowOrSideTab (/Applications/bb.app/Contents/Resources/app.asar/dist/main.js:43037:23)
at click (/Applications/bb.app/Contents/Resources/app.asar/dist/main.js:38961:18)
at MenuItem.click (node:electron/js2c/browser_init:2:41229)
at a._executeCommand (node:electron/js2c/browser_init:2:47143)

Image

SawyerHood:
Closes #1100.

## Summary
- handle the null window value that Electron sends for native panels
- send the native macOS close action for the About panel
- preserve the existing app-window and detached DevTools close paths
- add a regression test for the null window case

## Tests
- pnpm exec turbo run test typecheck --filter=@bb/desktop --force
- 230 desktop tests passed
- desktop type check passed

### DIFF
--- a/apps/desktop/test/menu.test.ts
+++ b/apps/desktop/test/menu.test.ts
@@ -52,6 +52,7 @@ function findServerSubmenu(
 
 describe("application menu", () => {
   it("closes a native panel when Electron omits its window", () => {
+    vi.mocked(Menu.sendActionToFirstResponder).mockClear();
     const closeWindowOrSideTab = vi.fn();
     const template = buildApplicationMenuTemplate(
       menuArgs(() => {}, { closeWindowOrSideTab }),
@@ -66,6 +67,22 @@ describe("application menu", () => {
       "performClose:",
     );
     expect(closeWindowOrSideTab).not.toHaveBeenCalled();
+  });
+
+  it("forwards an undefined window for detached DevTools", () => {
+    vi.mocked(Menu.sendActionToFirstResponder).mockClear();
+    const closeWindowOrSideTab = vi.fn();
+    const template = buildApplicationMenuTemplate(
+      menuArgs(() => {}, { closeWindowOrSideTab }),
+    );
+    const fileMenu = template.find((item) => item.label === "File");
+    const submenu = fileMenu?.submenu as MenuItemConstructorOptions[];
+    const closeWindow = submenu.find((item) => item.label === "Close Window");
+
+    closeWindow?.click?.({} as never, undefined, {} as never);
+
+    expect(closeWindowOrSideTab).toHaveBeenCalledWith(undefined);
+    expect(Menu.sendActionToFirstResponder).not.toHaveBeenCalled();
   });
 
   it("shows reload shortcuts without globally stealing browser commands", () => {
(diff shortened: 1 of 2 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I will review this pull request now.

I will check security, code quality, performance, architecture, duplication, and practical test coverage. I will follow the stated review rules.

apps/desktop/test/menu.test.ts
```diff
@@ -49,6 +51,23 @@ function findServerSubmenu(
...
+  it("closes a native panel when Electron omits its window", () => {
+    const closeWindowOrSideTab = vi.fn();
+    const template = buildApplicationMenuTemplate(
+      menuArgs(() => {}, { closeWindowOrSideTab }),
+    );
+    const fileMenu = template.find((item) => item.label === "File");
+    const submenu = fileMenu?.submenu as MenuItemConstructorOptions[];
+    const closeWindow = submenu.find((item) => item.label === "Close Window");
+
+    closeWindow?.click?.({} as never, null as never, {} as never);
```
SawyerHood, line 64: 🚨 `slopcop/review` — Add a test for the existing `undefined` route.

This new branch must keep `null` separate from `undefined`. The `undefined` value reaches `closeWindowOrSideTab`, which closes detached DevTools. This test covers only `null`. Add a second call with `undefined`. Assert that the callback receives `undefined`. Also assert that the native action does not run.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

The About box received the wrong close message. This change sends the normal Mac close message. It is like using the correct key.

### Review result

I found one low-risk test gap. I found no blocking problem.

- **Security:** The fixed native selector accepts no user input. The change adds no permission, process, storage, network, or IPC boundary.
- **Code quality:** The test covers `null`, but it does not protect the existing `undefined` route for detached DevTools. I left an inline comment.
- **Performance:** The normal route adds one constant-time check. The change adds no listener, timer, loop, retained object, or repeated allocation.
- **Architecture:** A project search found no duplicate native-close action. The direct native action fits this small Mac-only boundary.
- **Validation:** The desktop suite passed all 32 files and 230 tests. The forced desktop type check passed. `git diff --check` passed.
- **CI:** All active GitHub checks passed, including the macOS package smoke check.

Dev-browser cannot control the native macOS About panel. A web dev server cannot exercise this route, so I did not claim browser E2E coverage.

A manual native check should open About, press Cmd+W, and confirm the app stays open.

apps/desktop/test/menu.test.ts
SawyerHood: Fixed in b7467442f. The new regression test verifies that undefined reaches the detached DevTools callback and does not use the native panel action.

PR #1151, opened 2026-08-07
Fix connected provider defaults for new threads

ymichael:
## Summary

- mark the automatically selected connected provider as a client preference so thread creation uses the provider shown in the composer
- derive onboarding agent order from the same built-in provider catalog used by the model picker
- add regression coverage for client provenance, server resolution, and connected-agent ordering

## Root cause

For projects without saved execution defaults, the composer displayed the first connected provider but sent empty execution-input provenance. The server intentionally discarded displayed-only provider/model values, fell back to Codex, and persisted that resolved default.

## Testing

- `pnpm exec turbo run test --filter=@bb/app -- --run src/hooks/useThreadCreationOptions.test.tsx src/views/RootComposeView.test.ts`
- `pnpm exec turbo run test --filter=@bb/server -- --run test/system/onboarding.test.ts test/services/threads/thread-execution-plan.test.ts`
- `pnpm exec turbo run lint --filter=@bb/app --filter=@bb/server`


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am reviewing this PR now. I will check security, code quality, architecture, performance, and the new-thread flow in the app.

apps/app/src/views/RootComposeView.tsx
```diff
@@ -1017,6 +1045,10 @@ export function RootComposeView() {
   const connectedProviderId = agentOverviewQuery.data?.agents.find(
     (agent) => agent.status === "connected",
   )?.providerId;
+  const initialProvider = resolveRootComposeInitialProvider({
```
SawyerHood: 🚨 `slopcop/review` — Unknown project defaults can become a saved client override

An optimistic sidebar item uses null for unknown defaults. This helper treats the same null as a confirmed absence.

Cached agent data can arrive before the default query. A fast submit can override the stored default and save the temporary provider.

Keep separate unknown and absent states. Alternatively, block submit until the default query ends. Add a delayed-query test.

apps/app/src/views/RootComposeView.tsx
```diff
@@ -1025,8 +1057,8 @@ export function RootComposeView() {
     // actually signed in to. The raw provider catalog is a fixed list, so
     // `providers[0]` would always be Codex — wrong for anyone who only has,
     // say, Claude Code connected.
-    initialProviderId:
-      projectDefaultExecutionOptions?.providerId ?? connectedProviderId,
+    initialProviderId: initialProvider.providerId,
+    initialProviderSource: initialProvider.source,
```
SawyerHood: 🚨 `slopcop/review` — This fallback can use the wrong machine's provider

The agent query omits a host, so the server checks the primary machine. This composer can target another machine or a reused environment.

This line marks the primary machine's provider as a client preference. The server then honors it on the target machine.

A new thread can use a provider without a valid sign-in. Resolve the target host before the query.

Add the host to the query key and the SDK call. Please add a two-machine test with different connected providers.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

The form now puts a name tag on the helper it shows. Another computer or a fast click can still choose the wrong helper.

### Findings

1. **High — The fallback can use the wrong machine's provider.** The agent query checks the primary machine. The composer can target another machine.

   The new source marks the primary provider as a client preference. The server then honors it on the target machine.

2. **Medium — Unknown project defaults can become a saved client override.** An optimistic project uses null for unknown defaults.

   Cached agent data can arrive first. A fast submit can override the real default and save the temporary provider.

3. **Low — Provider priority has three sources.** Catalog order and separate client and server constants all define the policy.

   Move this product policy to one server-owned source.

I found no security or performance issue.

### Validation

- All 67 focused application tests passed.
- All 11 focused server tests passed.
- The application and server type checks passed.
- The diff check passed.
- The local app created a thread with the displayed Codex provider.
- I stopped the local test thread and the development server.
- The single-machine browser check cannot cover the two-machine defect.

I posted this as a comment-only review, as required.

apps/app/src/views/RootComposeView.tsx
ymichael: Addressed in 1e3840d45. Root compose now keeps project defaults in explicit pending/error/resolved states. Pending or placeholder query data cannot enable the connected-provider fallback, and normal submission stays disabled until defaults are authoritative. Added delayed-query and cross-project-placeholder regression coverage.

apps/app/src/views/RootComposeView.tsx
ymichael: Addressed in 1e3840d45. The connected-provider lookup now receives the same host/environment routing as model discovery, with host/environment included in the query key and SDK request. Reused environments resolve to their owning host server-side, and submission waits for the routed lookup. Added distinct two-host cache/result tests plus reused-environment SDK/server routing tests.

ymichael:
Also addressed the provider-policy cleanup in 1e3840d45: the app no longer hard-codes Codex for an unset selection, and the server derives its product fallback from the first provider in the canonical catalog. Model-picker order, connected-provider priority, and the no-selection fallback now share that catalog source.

PR #1152, opened 2026-08-07
Add manual context compaction across providers

### LINKED ISSUE #1103: pi provider: no way to manually compact a thread's context
> **⚠ Agent-generated:** This issue was drafted by an AI agent, not a human. Facts were gathered from the installed bb app, provider bridges, and CLI; please verify details before acting on them.

## Related

[#448 “Please add a way to compact threads”](https://github.com/get-bb/bb/issues/448) (closed as completed) covered `/compact` working on Claude and not on Codex; the Codex case was fixed. **The `pi` provider was not covered there** and the maintainer’s working assumption — “You should be able to `/compact`” — does not hold for pi threads. This issue is the pi-specific gap.

## Summary

BB exposes no way to manually compact a thread's context when the thread runs on the `pi` provider. Typing `/compact` in the BB composer does nothing on a pi thread — the text is forwarded to pi's headless agent as an ordinary user message, and pi's RPC mode explicitly does not execute built-in TUI commands sent as prompts. On a `claude-code` thread, the exact same `/compact` message *does* trigger real compaction, because the Claude CLI parses slash commands from any user input. The asymmetry is confusing and undocumented.

## Environment

- bb version: 0.35.1
- provider / harness: pi (headless via `bb-pi-bridge`, `pi --mode rpc`), version 0.83.0; claude-code 2.1.221 for the comparison
- model (relevant): deepseek v4 flash (large context window — auto-compaction fires rarely)

## Repro

1. Create/open a BB thread with provider `pi`.
2. Type `/compact` in the composer and send it.
3. Observed: the model receives "/compact" as a prompt and answers conversationally; no "Context compacted" entry appears in the timeline.

Same steps on a `claude-code` thread: the Claude CLI executes its built-in `/compact`, and BB records a "Context compacted" timeline entry.

## Expected vs actual

- Expected: a way to manually compact a pi thread (a `/compact` composer action, a thread action, or a `bb` CLI command), or at least consistent behavior across harnesses — including `/compact` autocomplete in the composer, which #448 noted is missing.
- Actual: `/compact` is silently inert on pi threads (not even autocompleted) and works on claude-code threads; the only manual workaround is forking/starting a new thread.

## Evidence / root cause

- **BB itself has no native compaction control.** There is no composer action named compact for any provider (pi: `{kind:"skills"}` only; claude-code/codex: skills + `/plan` + `/goal`), no thread action, no `bb` CLI command, and no server endpoint. The server only *records* provider-initiated compactions — `thread/compacted` events and `contextCompaction` items — and renders them as "Compacting context" → "Context compacted" timeline entries (`onCompactionBegin`/`onCompactionEnd` in `server/dist/start-server.js`).
- **pi has the capability but BB never calls it.** pi ships a first-class RPC `compact` command (`{"type": "compact"}` in `pi --mode rpc`) plus auto-compaction (threshold/overflow). But pi's docs state built-in TUI commands like `/compact` are "handled only in interactive mode and would not execute if sent via `prompt`" — which is exactly what happens when the text is typed in BB's composer. The `bb-pi-bridge` never sends the RPC `compact` command.
- **claude-code works by accident of forwarding.** `bb-claude-code-bridge` passes user messages verbatim to the Claude CLI (`pushInput({type:"user", message:{role:"user", content}})`); the CLI processes slash commands from any user input (`disableSlashCommands` is false), so `/compact` reaches Claude's own compaction, surfaced to BB via its `PreCompact`/`PostCompact` hooks.

## Suggested fix

Add a manual compact surface for pi threads — e.g. a `/compact` composer action or a thread action that sends pi's RPC `compact` command through the bridge. The capability exists end-to-end; only the UI trigger is missing. Ideally make it generic per provider: pi → RPC `compact`; claude-code → send `/compact` (already works) or use an explicit SDK path; codex → verify what exists before implementing.

## Impact

Users of pi-powered threads (e.g. deepseek models with large context windows) cannot manually compact long sessions; they must fork or start new threads or wait for automatic threshold/overflow compaction. Low severity, but a real confusion risk given the claude-code behavior works.

ymichael:
## Summary

- reuse the existing structured `/compact` turn path instead of adding a second host-daemon compaction command
- preserve the existing Codex and Claude Code behavior; add native Pi compaction and OpenCode ACP `/compact` handling at the provider adapter boundary
- add provider-aware typeahead plus `POST /threads/:id/compact`, `sdk.threads.compact()`, and `bb thread compact`
- leave Cursor ACP and arbitrary custom ACP agents unsupported because ACP does not expose a standard compaction method and Cursor does not interpret its interactive `/compress` command over ACP
- report Pi and OpenCode completion, failure, and interruption through normal visible timeline events

## Provider behavior

| Provider | Behavior |
| --- | --- |
| Codex | Existing native `thread/compact/start` mapping |
| Claude Code | Existing selected `/compact` turn behavior |
| Pi | Selected `/compact` maps to the Pi SDK `compact()` operation |
| OpenCode ACP | Selected `/compact` maps to an ACP `session/prompt` containing the exact provider control |
| Cursor ACP | Unsupported; Cursor does not expose compatible compaction through ACP |
| Other ACP agents | Unsupported unless a concrete provider integration is added |

## Simplified request path

The composer submits `/compact` exactly like any other selected built-in command. The SDK/CLI endpoint synthesizes that same structured input and delegates to the existing `sendThreadMessage` path. This means request persistence, composer clearing, lifecycle settlement, and failure display all use the ordinary turn machinery.

Only the Pi and OpenCode adapters need new dispatch behavior. They recognize the exact selected built-in command, start provider-local compaction, acknowledge the turn promptly, and translate the provider outcome back into the existing `contextCompaction`, `thread/compacted`, and `turn/completed` events.

This branch does not add a host-daemon RPC, change the daemon protocol, configure generic ACP prompt compaction, intercept composer submission, or summarize/reseed sessions.

## Validation

- relevant package typechecks passed
- relevant package lint passed
- focused provider/runtime tests: 339 passed
- focused server compaction/typeahead tests: 6 passed
- app command-suggestion tests: 5 passed
- CLI suite: 383 passed
- the full agent-runtime suite passed 859 tests but still has one unrelated existing failure in `runtime.process-lifecycle.test.ts` (`bounds provider stderr while data arrives without a newline`), reproduced on retry; all changed provider tests pass

## Diff shape after cleanup

The branch is now 1,022 additions / 126 deletions across 39 files. It removes the branch-only daemon command, runtime API, protocol bump, generic ACP configuration, Claude/Codex duplicate plumbing, composer interception, and unrelated `dev:stop` work.

Fixes #1103


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

packages/agent-runtime/src/acp/bridge/bridge.ts
```diff
@@ -1693,6 +1692,53 @@ function runTurn(session: AcpThreadSession, firstInput: PromptInput[]): void {
...
+async function compactSession(
+  session: AcpThreadSession,
+  compaction: AcpManualCompaction,
+): Promise<void> {
+  if (session.promptActive) {
+    throw new Error("Cannot compact context while an ACP turn is active");
+  }
+
+  session.promptActive = true;
+  session.compactionActive = true;
```
SawyerHood: 🚨 `slopcop/review` — [P1] ACP compaction still permits file operations.

This flag hides session updates, but the request handler still permits file reads and writes. The bridge drops the normal tool lifecycle.

A maintenance prompt can therefore read or change files unexpectedly. Reject file and tool requests during prompt-based compaction.

Use a provider-native maintenance method when the provider supplies one.

apps/server/src/routes/projects.ts
```diff
@@ -704,6 +705,10 @@ export function registerProjectRoutes(app: Hono, deps: AppDeps): void {
     return context.json(
       buildCommandListResponse({
         commands: result.commands,
+        includeBuiltinCompact: providerSupportsManualCompaction(
```
SawyerHood: 🚨 `slopcop/review` — [P2] New-thread composers offer `/compact` before a thread exists.

The project command route serves new-thread and follow-up composers. This flag adds `compact` to both surfaces.

The new-thread path creates a thread with `/compact` as its first input. Add a command scope and exclude this command there.

apps/host-daemon/src/command-dispatch.ts
```diff
@@ -273,6 +273,18 @@ const commandHandlers: CommandHandlerMap = {
     await options.eventSink.flush();
     return {};
   },
+  "thread.compact": async (command, options) => {
+    const entry = await ensureThreadRuntime(command, options);
```
SawyerHood: 🚨 `slopcop/review` — [P2] Compaction does not retain the environment runtime.

Start and submit commands retain the runtime. This handler does not retain it during compaction.

An environment refresh can evict an idle ACP runtime before compaction finishes. Retain the runtime and release it in a `finally` block.

packages/agent-runtime/src/claude-code/bridge/bridge.ts
```diff
@@ -1478,6 +1481,28 @@ async function handleThreadStop(
...
+  if (!threadSession) {
+    sendError(id, -32000, "No active session");
+    return;
+  }
+  if (threadSession.session.getIsProcessing()) {
+    sendError(id, -32000, "Cannot compact context while a turn is active");
+    return;
+  }
+  try {
+    await threadSession.session.pushInput("/compact");
```
SawyerHood: 🚨 `slopcop/review` — [P2] Claude reports success before compaction finishes.

`pushInput()` resolves when the SDK consumes `/compact`. It does not wait for a terminal result or compact boundary.

The API can report success before a later provider failure. Wait for the matching terminal result and return its error.

packages/domain/src/shared-types.ts
```diff
@@ -356,6 +356,53 @@ function isSelectedPromptCommandMention(
...
+        isSelectedBuiltinPromptCommandMention(mention, selector),
+      )
+      .map((mention) => ({ start: mention.start, end: mention.end }))
+      .sort((left, right) => left.start - right.start || left.end - right.end);
+    mentionCount += ranges.length;
+
+    let remainingText = "";
+    let cursor = 0;
+    for (const range of ranges) {
+      remainingText += item.text.slice(cursor, range.start);
```
SawyerHood: 🚨 `slopcop/review` — [P2] Invalid mention ranges can replace normal text with compaction.

The schema permits an end offset outside the text. This loop also does not verify the selected substring.

A malformed compact mention from `0` to `999` can remove a normal prompt. Validate offsets and the selected command text.

apps/app/src/hooks/mutations/thread-runtime-mutations.ts
```diff
@@ -491,6 +491,20 @@ export function useStopThread() {
...
 
+export function useCompactThread() {
+  const queryClient = useQueryClient();
+
+  return useMutation({
+    meta: { errorMessage: "Failed to compact context." },
+    mutationFn: async (threadId: string) => {
+      await sdk.threads.compact({ threadId });
+    },
+    onSettled: (_data, _error, threadId) => {
```
SawyerHood: 🚨 `slopcop/review` — [P3] Failed compactions start broad query refreshes.

`onSettled` runs after success and failure. It refreshes detail, timeline, list, sidebar, and search query groups.

A known provider rejection needs no refresh. Move this work to `onSuccess` and use focused recovery for timeouts.

apps/host-daemon/src/command-router.ts
```diff
@@ -666,6 +666,7 @@ export class CommandRouter {
           ? this.getInFlightThreadStopProviderLane(command)
           : null;
       }
+      case "thread.compact":
```
SawyerHood: 🚨 `slopcop/review` — [P1] Compaction blocks unrelated provider sessions.

This write lane covers all non-Codex sessions for the same provider and environment. A stalled compaction can block them for 20 minutes.

ACP remains idle, so the stop route cannot cancel this work. Use a read process lane and a serial session lane.

Reject duplicate compaction requests. Add cancellation and a shorter timeout.

packages/agent-runtime/src/pi/adapter.ts
```diff
@@ -967,6 +976,23 @@ export function createPiProviderAdapter(
           providerThreadId: "",
           scope: turnScope(turnId),
         });
+        if (parsed.data.reason === "manual" && state.currentTurnId === turnId) {
```
SawyerHood, line 999: 🚨 `slopcop/review` — [P2] Pi reports failed compactions as successful.

This code emits `thread/compacted` before it checks `aborted` and `errorMessage`. The timeline can show a false success.

Emit `thread/compacted` only after a successful result. Add tests for errors and aborted compactions.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: This change adds a context clean-up command for several providers. Some providers finish incorrectly, block other work, or expose the command too early.

I reviewed the requested SHA `9a63de9699e8a5beea001a336df9fb6fedcb87eb`. The pull request later moved to `041b652e`, but these code paths remain unchanged.

I found these defects:

- [P1] Non-Codex compaction takes a shared provider write lane. A stalled request can block unrelated sessions for 20 minutes.
- [P1] ACP prompt compaction hides tool activity but still permits file reads and writes.
- [P2] The daemon does not retain the runtime during compaction.
- [P2] Claude returns success when the SDK consumes `/compact`, before compaction completes.
- [P2] Pi emits `thread/compacted` for aborted and failed compactions.
- [P2] New-thread composers offer `/compact` before a thread exists.
- [P2] Invalid mention ranges can replace a normal prompt with compaction.
- [P3] Failed UI compactions refresh broad, unrelated query groups.

The implementation also repeats compact detection and draft recovery in four UI paths. One shared action can reduce this duplication.

The provider adapters use different completion rules. A shared terminal contract and the existing compaction lifecycle can prevent these differences.

I ran the branch dev app and used the browser command surface. Codex returned 200, showed compaction, and reduced context use from 7% to 2%.

The source CLI also compacted the same thread. Ten changed packages passed type checks, and 129 focused tests passed.

The full local run passed nine package tasks. One unrelated machine-install test failed, while GitHub checks passed.

apps/server/src/routes/projects.ts
ymichael: Fixed in 78396d41e. Command suggestions now receive an explicit new-thread versus thread scope. Root compose, plugin new-thread compose, and an uncommitted embedded chat all exclude the built-in compact command; existing-thread composers retain it. The project command endpoint remains a context-neutral catalog, while availability is enforced at the composer boundary that knows whether a thread exists.

packages/agent-runtime/src/acp/bridge/bridge.ts
ymichael: Fixed in 78396d41e. The ACP bridge now rejects both fs/read_text_file and fs/write_text_file while a compaction prompt is active; permission requests were already cancelled outside a normal turn. The regression test drives a fake ACP agent that attempts both operations and verifies that neither succeeds.

packages/agent-runtime/src/claude-code/bridge/bridge.ts
ymichael: This is intentional and I am not changing the completion boundary. Host-daemon protocol version 82 defines thread.compact as settling once provider work is accepted. For Claude, SDK input consumption is the available acceptance boundary; terminal success or failure continues through provider lifecycle events. Waiting for the compact boundary here would recreate an HTTP request that can fail or time out after compaction already started, and would differ from Pi and ACP. I added an explicit code comment in 78396d41e to preserve that contract.

packages/agent-runtime/src/pi/adapter.ts
ymichael: Already fixed on the current tip before this review was submitted. Pi now emits thread/compacted only when both aborted is false and errorMessage is absent, and the adapter suite has explicit failed and aborted manual-compaction cases. No additional change was needed for this thread.

apps/host-daemon/src/command-dispatch.ts
ymichael: Fixed in 78396d41e. thread.compact now retains the environment runtime before ensuring/resuming the provider session and releases it in a finally block. The dispatch regression asserts both retention and release.

packages/domain/src/shared-types.ts
ymichael: Fixed in 78396d41e. Standalone built-in command detection now rejects empty/reversed or out-of-bounds ranges and verifies that the selected substring is exactly the trigger plus command name. The regression covers a normal prompt paired with a forged 0..999 compact mention and a mismatched selected substring.

apps/app/src/hooks/mutations/thread-runtime-mutations.ts
ymichael: Fixed in 78396d41e. Compaction invalidation now runs in onSuccess rather than onSettled. A rejected-compaction regression verifies that no query group is invalidated.

apps/host-daemon/src/command-router.ts
ymichael: Fixed the valid lane issue in 78396d41e. Compaction now takes a shared-process read lane and keeps the existing serial provider-session lane, so unrelated sessions for Pi/Claude/ACP are not blocked. A regression holds Pi compaction open and verifies that a turn on another Pi session completes. The 20-minute detail referred to the earlier implementation: the custom terminal wait was removed in c24661953, and provider commands now acknowledge when work starts while terminal outcome remains event-driven.

ymichael:
Review triage complete at 78396d41e.

- Fixed: ACP file-operation isolation, new-thread command scoping, runtime retention, invalid mention ranges, failure-only query churn, and shared-provider lane contention.
- Already fixed on the current tip: Pi aborted/failed compactions no longer emit thread/compacted, with explicit regressions.
- Intentional contract: Claude acknowledges when the SDK consumes the maintenance input; terminal outcome remains event-driven. HOST_DAEMON_PROTOCOL_VERSION 82 documents that acceptance boundary.

For the two general observations: compact detection is now centralized in isStandaloneBuiltinCompactCommand and shared by both submission controllers. Draft ownership remains controller-specific because the full thread composer and plugin/embedded composer have different persistence and callback lifecycles; combining those controllers is outside this feature. Provider-specific signals are already normalized into the shared compaction lifecycle, while command settlement consistently means accepted rather than terminally completed.

Verification on the new tip: all affected typechecks pass; full suites pass for domain (129), agent-runtime (840), host-daemon (493), and app (2,419).

PR #1155, opened 2026-08-07
Fix enrolled host daemon port collisions

### LINKED ISSUE #1121: Enrolled host-daemon has no collision handling for the default port (38887) when the desktop app is already running on the same machine
## Workflow

Machine already running the packaged desktop app (`bb.app`), which runs its own local server + host-daemon for "This Mac" as its primary host. Used **Settings → Machines → Add machine** on a *different*, remote bb server to enroll this same Mac as an additional execution host (installed via `/install.sh ... --join-code ... --host-id ... --server ...`, which sets up a separate `launchd` service for the new host-daemon, pointed at the remote server).

## What happened

The installer printed a failure at the very end, after the npm install and the join API call both succeeded:

```
Joining http://<remote-server>:38886 as host_xxxxxxxxx...
Joined successfully.
Could not find service "app.getbb.host-daemon.<label>" in domain for user gui: 501
```

(exit code 113). That "Could not find service" text is `launchctl`'s own native error text, not a bb-authored message — the installer shells out to `launchctl` to verify the service came up and surfaces whatever it says.

The actual root cause: the desktop app's own host-daemon was **already bound to 127.0.0.1:38887** (the fixed default host-daemon port, hardcoded regardless of which server a given daemon is enrolled to). The newly-created `launchd` service for the remote-server enrollment tried to bind the same port, hit `EADDRINUSE`, and crashed:

```
Error: listen EADDRINUSE: address already in use 127.0.0.1:38887
    at Server.setupListenHandle [as _listen2] (node:net:2167:16)
```

`launchd`'s `KeepAlive` then restarted it in a loop. The installer's own post-bootstrap `launchctl` check happened to run during one of the crashed windows, producing the confusing "Could not find service" text — but even after that, the daemon kept silently crash-looping forever: it can never win the race for port 38887 while the desktop app's own daemon holds it permanently. The remote server's `hosts` table showed the new host as enrolled (the join API call had already succeeded), but its `host_daemon_sessions` table never got a row for it, and `last_seen_at` stayed null indefinitely — the remote server's UI correctly showed the machine as offline, with no error surfaced anywhere pointing at the actual port conflict.

Fixed manually by editing the generated plist to add `--host-daemon-port <alternate-port>` (a flag `bb-app host-daemon` already supports) and reloading via `launchctl bootout`/`bootstrap`. After that the session appeared immediately.

## Expected

- The installer/`host-daemon` process should detect `EADDRINUSE` on its own default port and either auto-select a free port or fail with a clear, actionable message ("port 38887 already in use by another bb host-daemon — pass --host-daemon-port to use a different one"), instead of crash-looping silently under `launchd`.
- The "Add machine" installer shouldn't report a bare `launchctl` error as if it were bb's own diagnostic; it should distinguish "service didn't register" from "service is crash-looping" and say which.

This will affect anyone who enrolls a machine that's already running the desktop app locally on the same box — plausibly a common setup, not an edge case.

## Environment

- bb-app 0.35.1
- macOS 26.3, Apple Silicon, launchd

ymichael:
## Summary

- allocate and persist a distinct local API port for each enrolled host daemon, and pass it through foreground joins plus generated launchd/systemd services
- validate readiness through `/status`, requiring the expected `hostId`, normalized `serverUrl`, and an active server connection
- turn remaining `EADDRINUSE` and service startup failures into actionable bb-authored diagnostics
- document the installer override and port-selection behavior

## Testing

- `sh -n apps/server/src/assets/install-machine.sh`
- `pnpm exec turbo run typecheck --filter=@bb/host-daemon --filter=bb-app --filter=@bb/server`
- `pnpm exec turbo run test --filter=@bb/host-daemon --force -- --run src/local-api.test.ts`
- `pnpm exec turbo run test --filter=bb-app --force -- --run test/index.test.ts`
- `pnpm exec turbo run test --filter=@bb/server --force -- --run test/app/install-machine-script.test.ts test/app/skeleton.test.ts`

Fixes #1121


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull request for security, code quality, performance, architecture, duplication, and end-to-end behavior.

apps/server/src/assets/install-machine.sh
```diff
@@ -91,6 +94,139 @@ data_dir=${BB_DATA_DIR:-"$HOME/.bb-machines/$server_host"}
...
+          if (error) {
+            process.stderr.write(`Could not release host-daemon port probe ${port}: ${error.message}\n`);
+            process.exit(1);
+          }
+          process.stdout.write(String(port));
+        });
+      });
+    };
+    tryPort(38888);
+  ' "$HOME/.bb-machines"
```
SawyerHood: 🚨 `slopcop/review` — **Medium: include custom data directories in port reservations**

This scan reads only `$HOME/.bb-machines`, but the installer supports a custom `BB_DATA_DIR`.

A stopped custom daemon releases its port. A later installation can then select that stored port and break the first daemon's next start.

Please use one global reservation registry that records every canonical data directory.

apps/server/src/assets/install-machine.sh
```diff
@@ -91,6 +94,139 @@ data_dir=${BB_DATA_DIR:-"$HOME/.bb-machines/$server_host"}
...
+      const server = net.createServer();
+      server.once("error", (error) => {
+        if (error && (error.code === "EADDRINUSE" || error.code === "EACCES")) {
+          tryPort(port + 1);
+          return;
+        }
+        process.stderr.write(`Could not check host-daemon port ${port}: ${error.message}\n`);
+        process.exit(1);
+      });
+      server.listen({ host: "127.0.0.1", port, exclusive: true }, () => {
```
SawyerHood: 🚨 `slopcop/review` — **Medium: make port allocation atomic**

The probe closes before the installer records the selected port. Two concurrent installers can therefore select the same port.

One daemon then fails with `EADDRINUSE`, which recreates the collision this change must prevent.

Please hold one global allocation lock while the script selects and records the port.

apps/server/src/assets/install-machine.sh
```diff
@@ -351,8 +506,19 @@ RestartSec=2
 WantedBy=default.target
 EOF
   systemctl --user daemon-reload
-  systemctl --user enable --now "$service_name.service"
+  if ! systemctl_error=$(systemctl --user enable --now "$service_name.service" 2>&1); then
```
SawyerHood: 🚨 `slopcop/review` — **High: restart the active Linux service**

`systemctl enable --now` does not restart an active unit after `daemon-reload`.

An existing Linux daemon continues on its old port. The new readiness check then polls the new port and times out.

Please enable the unit and explicitly restart it before the readiness check. Add a test for an already active service.

apps/server/test/app/install-machine-script.test.ts
```diff
@@ -289,6 +350,40 @@ describe("machine install script", () => {
     expect(result.stdout).not.toContain("Joined successfully");
   });
 
+  it("assigns a different port when the first enrolled-daemon port is occupied", async () => {
+    const occupied = createNetServer();
+    await new Promise<void>((resolve, reject) => {
+      occupied.once("error", reject);
+      occupied.listen(38888, "127.0.0.1", resolve);
```
SawyerHood: 🚨 `slopcop/review` — **Medium: do not require port 38888 to be free**

A valid enrolled daemon commonly uses port `38888`. This test then fails before it tests allocation behavior.

It failed in this review environment for that reason. The service tests also use fixed ports.

Please inject an isolated allocation base or tolerate an existing listener. Assert only that the selected port differs from the occupied port.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

Each daemon needs a different port.

This change chooses one, but some Linux and multi-install cases can still choose or use the wrong port.

### Findings

I found four issues:

1. **High:** An active Linux service does not restart after its unit changes.
2. **Medium:** Custom `BB_DATA_DIR` daemons do not reserve their ports globally.
3. **Medium:** Concurrent installers can select the same port.
4. **Medium:** The tests require fixed ports that real enrolled daemons can already use.

I left an inline comment for each issue. I would fix these issues before merge.

### Security and performance

I found no new critical security boundary.

The allocation race can still cause `EADDRINUSE`. A failed connection loop can also take about 105 seconds.

### Architecture

The shell script now repeats port validation, allocation, URL normalization, and status parsing.

A typed `bb-app` allocation command could own this policy. A global registry and lock could make reservations complete and atomic.

The daemon protocol version does not need a change because this pull request does not change the server-daemon wire contract.

### Validation

Three GPT-5.6-Sol agents reviewed security, quality, performance, architecture, and end-to-end behavior. A final GPT-5.6-Sol gate confirmed the four findings.

- All affected Turbo type checks passed.
- The host-daemon tests passed.
- The bb-app launcher tests passed.
- The server suite passed 1,344 tests and failed two installer tests.
- One failure confirmed the fixed-port defect.
- The other failure exposed incomplete `PATH` isolation in the npm fallback fixture.
- The live launcher check accepted the correct enrolled daemon in 21 milliseconds.
- The live launcher check rejected a wrong host identity in 207 milliseconds.
- No browser test applied because this change has no browser route.

I used a comment-only review. I did not approve or request changes through GitHub.

apps/server/test/app/install-machine-script.test.ts
ymichael: Fixed in 90807af20. The collision test now tolerates an existing 38888 listener and asserts only that allocation skips it. The launchd/systemd fixtures read the dynamically selected port instead of using 42111/42112, and the fixture PATH is isolated so a globally installed bb-app cannot change npm-fallback behavior.

apps/server/src/assets/install-machine.sh
ymichael: Fixed in 90807af20. Each port is now claimed with an atomic mkdir before its availability probe; that per-port reservation remains held after the probe closes, so concurrent installers cannot select the same port. A concurrent two-installer regression test verifies distinct assignments.

apps/server/src/assets/install-machine.sh
ymichael: Fixed in 90807af20. The installer now canonicalizes every selected data directory, including custom BB_DATA_DIR values, and records it in the global ~/.bb-machines/host-daemon-ports/<port>/data-dir registry. The new regression test runs two custom data-directory installs and verifies both canonical owners are registered.

apps/server/src/assets/install-machine.sh
ymichael: Fixed in 90807af20. The Linux path now runs systemctl --user enable followed by an explicit systemctl --user restart after daemon-reload. The service test models an already-active unit and only launches the replacement daemon on restart.

PR #1158, opened 2026-08-07
Add configurable thread rename shortcut

### LINKED ISSUE #1157: Add a configurable shortcut for renaming the focused thread
## Problem

BB lets users rename a thread from its actions menu, but renaming is not available as a keyboard command. A keyboard-focused workflow must therefore switch to the pointer for this action.

This is also relevant in split view, where a rename command should target the focused thread rather than another visible pane.

## Proposed behavior

- Add **Rename thread** to Settings → Keyboard.
- Leave the command unassigned by default to avoid introducing a shortcut conflict.
- Let users assign, clear, and reset the shortcut like other keyboard commands.
- Open the existing rename dialog for the focused thread.
- In split view, target only the focused pane.
- Keep the current thread-title persistence behavior unchanged.

## Prototype

I implemented and tested this behavior on a fork:

- Branch: https://github.com/ryanbbrown/bb/tree/feature/thread-rename-shortcut
- Commit: https://github.com/ryanbbrown/bb/commit/8ef3e1c34

The prototype adds focused app and server tests, including split-pane routing through the real pane context.

Does this behavior and scope align with the project before I open a pull request?

ryanbbrown:
Closes #1157.

## Summary

- expose **Rename thread** in Settings → Keyboard without assigning a default shortcut
- support assigning, clearing, and resetting keyboard commands that have no default shortcut
- route the command to the focused thread and reuse the existing rename dialog
- target only the focused pane when multiple threads are visible

## User impact

Users can choose a keyboard shortcut for renaming the focused thread. The existing thread-title persistence behavior remains unchanged.

## Testing

- `pnpm exec turbo run test --filter=@bb/app -- --run src/components/settings/KeyboardSettingsSection.test.tsx src/lib/keyboard-shortcut-settings.test.ts src/views/thread-detail/ThreadRenameCommandHandler.test.tsx`
- `pnpm exec turbo run test --filter=@bb/server -- --run test/system/app-keybindings.test.ts`
- `pnpm exec turbo run typecheck --filter=@bb/app --filter=@bb/server --filter=@bb/domain --filter=@bb/server-contract --filter=@bb/plugin-sdk --filter=@bb/templates`
- manual QA in the isolated development app by assigning the shortcut and renaming a thread
- `git diff --check`


ymichael:
Superseded by #1172, which contains this rename implementation rebased onto current `main` together with the closely related configurable archive shortcut. Commit authorship is preserved there, and #1172 now closes #1157.

PR #1165, opened 2026-08-07
Preserve background tasks across daemon reconnects

### LINKED ISSUE #1127: bb kills background tasks that span an idle gap between turns — bb 0.35.1, provider claude-code.
Based on `bb` self-debugging:

Steps:
- Start a run_in_background job longer than one turn (~10 min).
- Let the turn end.
- Send any follow-up message.
- Expected: still running. 
- Actual: taskStatus: "stopped", process gone, output file 0 bytes.

The replacement claude --resume <same-session-id> process starts at the same second as the stopped event, in two independent threads. Environment-wide, so sibling threads lose their runs too.

`bb does have a guard for exactly this (entryHasActiveRuntimeWork, with a comment saying background tasks must count as active work), and I verified its inputs are correct — bb emits item/started / backgroundTask / pending as the tracker expects. So the teardown is bypassing that guard, not failing it.`

ymichael:
## Summary

- distinguish a reconnect from a genuinely new host-daemon instance when replacing a server session
- close only the superseded WebSocket for same-instance reconnects, preserving resident provider runtimes and their background processes
- retain the fatal `session-close` and task reconciliation behavior when a different daemon instance takes over
- bump the host-daemon protocol version and add unit plus real-provider regression coverage

## Root cause

Opening a replacement server session always sent `session-close:replaced` to the previous socket. On an ordinary reconnect, that socket belonged to the same live daemon instance. The daemon treated the message as terminal and shut down every environment runtime, which killed background commands before its supervisor restarted it.

The server already has both session instance IDs, so it can tell these cases apart. A same-instance replacement now closes only the stale transport. A different-instance replacement still shuts down the superseded daemon and settles its work.

## Reproduction

I reproduced this with standalone dev servers and the `bb` CLI against both the v0.35.1 tag and this branch:

- v0.35.1: replacing the live session with the same daemon instance immediately killed the background shell and sleep PIDs, recorded the task as stopped/interrupted, and logged `session-close:replaced` daemon shutdown
- patched branch: the daemon reconnected with the same instance ID, the same background PIDs remained alive across a CLI follow-up, and the task completed and wrote its marker normally
- control: an unchanged CLI follow-up without a reconnect preserved the background PIDs, confirming the reported failure is specifically the reconnect path

## Tests

- `@bb/server` lifecycle and hub regression tests: 32 passed
- `@bb/host-daemon-contract` tests: 49 passed
- sanitized install-machine regression suite: 10 passed
- typechecks for `@bb/server`, `@bb/host-daemon-contract`, and `@bb/integration-tests`
- real Claude end-to-end reconnect/background-process regression: passed
- full server suite: 1,357 passed; the one environment-dependent install-script failure passed when rerun with the expected sanitized `PATH`

Closes #1127

### DIFF
--- a/tests/integration/real/provider-background-survival.test.ts
+++ b/tests/integration/real/provider-background-survival.test.ts
@@ -1,0 +1,106 @@
+import fs from "node:fs/promises";
+import path from "node:path";
+import { describe, expect, it } from "vitest";
+import {
+  createRealThread,
+  pathExists,
+  REAL_POLL_INTERVAL_MS,
+  sendAndWaitForIdle,
+  TEST_TIMEOUT_MS,
+} from "./provider-smoke-harness.js";
+
+const PROVIDER_ID = "claude-code";
+
+async function waitForCondition(
+  predicate: () => boolean | Promise<boolean>,
+  timeoutMs: number,
+  message: string,
+): Promise<void> {
+  const startedAt = Date.now();
+  while (Date.now() - startedAt <= timeoutMs) {
+    if (await predicate()) {
+      return;
+    }
+    await new Promise((resolve) => setTimeout(resolve, REAL_POLL_INTERVAL_MS));
+  }
+  throw new Error(message);
+}
+
+describe("real Claude background process integration", () => {
+  it(
+    "survives a same-daemon session reconnect and follow-up",
+    async () => {
+      const { environment, harness, thread } = await createRealThread({
+        providerId: PROVIDER_ID,
+        workspace: { path: null, type: "unmanaged" },
+      });
+
+      try {
+        if (!environment.path) {
+          throw new Error("Expected an unmanaged workspace path");
+        }
+        const markerPath = path.join(
+          environment.path,
+          "background-survived-reconnect.txt",
+        );
+
+        await sendAndWaitForIdle({
+          harness,
+          providerId: PROVIDER_ID,
+          threadId: thread.id,
+          text: `Use the Bash tool exactly once to execute this exact command: sleep 40; printf survived > ${JSON.stringify(markerPath)}. Set Bash's run_in_background parameter to true. As soon as Bash accepts it, reply exactly STARTED. Do not poll it, wait for it, or use another tool.`,
+        });
+        expect(await pathExists(markerPath)).toBe(false);
+
+        // Cross the host daemon's 10-second shell-environment refresh TTL so
+        // the follow-up exercises the same idle-gap maintenance path as the
+        // report, while the background command remains live.
+        await new Promise((resolve) => setTimeout(resolve, 12_000));
+        expect(await pathExists(markerPath)).toBe(false);
+
+        const runtimeBeforeReconnect = harness.daemonApp.runtimeManager.get(
+          environment.id,
+        )?.runtime;
+        const previousSessionId = harness.daemonApp.connection.sessionId;
+        if (!runtimeBeforeReconnect || !previousSessionId) {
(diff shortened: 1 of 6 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** A daemon reconnect is like a phone call that drops. This change keeps its long task alive after the new call starts.

I am SlopCop. I am reviewing this pull request for security, code quality, performance, tests, and architectural duplication. Three BB child threads will inspect these areas in parallel. I will post one final review after I verify their findings.

tests/integration/real/provider-background-survival.test.ts
```diff
@@ -0,0 +1,106 @@
...
+
+        harness.daemonApp.connection.handleSessionInvalidated({
+          code: "inactive_session",
+          observedSessionId: previousSessionId,
+          source: "postEvents",
+        });
+        await waitForCondition(
+          () => {
+            const sessionId = harness.daemonApp.connection.sessionId;
+            return sessionId !== null && sessionId !== previousSessionId;
```
SawyerHood: 🚨 `slopcop/review` — Wait for the replacement socket, not only a new session ID.

`openSession()` assigns the new ID before the socket opens. This test can send the follow-up during that gap. The server can then return `host_unavailable`. Please wait for daemon socket registration before the follow-up.

tests/integration/real/provider-background-survival.test.ts
```diff
@@ -0,0 +1,106 @@
...
+          45_000,
+          "Background command did not write its marker after reconnect",
+        );
+
+        expect(await fs.readFile(markerPath, "utf8")).toBe("survived");
+      } finally {
+        await harness.cleanup();
+      }
+    },
+    TEST_TIMEOUT_MS,
```
SawyerHood: 🚨 `slopcop/review` — Give this real-provider test a separate larger timeout.

The shared limit is 120 seconds. This test includes provider setup, two turns, a 12-second delay, and a 40-second process. A valid slow run can exceed the limit. Please use a dedicated timeout or reduce the fixed waits.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** A daemon reconnect is like a phone call that drops. This change keeps a long job alive after the new call starts.

I found no blocking product defect. The change keeps server policy in the server. It also adds a small socket-only hub method. I found no useful duplicate code or larger refactor.

I left two low-risk test comments:

- Wait for the replacement WebSocket before the follow-up turn.
- Give the real-provider test a larger timeout.

Security, code-quality, and performance threads reviewed the change. A separate GPT-5.6 gate confirmed the two test risks and found no product defect.

Targeted server tests passed with 11 tests. Contract tests passed with 49 tests. Server, contract, and integration type checks passed. The dev app loaded in a browser. All GitHub checks are green.

The real Claude test stopped before the reconnect scenario. The local Claude OAuth session had expired. This failure does not show a pull-request defect.

tests/integration/real/provider-background-survival.test.ts
ymichael: Addressed in 62282a3d9. This scenario now has a dedicated scaled 240-second timeout. The updated live test completed in 63 seconds locally.

tests/integration/real/provider-background-survival.test.ts
ymichael: Addressed in 62282a3d9. The reconnect wait now requires the daemon connection session ID to match the hub's registered session ID, so the follow-up cannot run before the replacement WebSocket is available. The updated live test passes.

PR #1166, opened 2026-08-07
Fix app shortcuts and Escape in the chat input

### LINKED ISSUE #1162: Sidebar shortcut does not work while chat input is focused
## Problem

The keyboard shortcut that toggles the sidebar does not work while the chat input has focus. The Escape key also does not remove focus from the input, so the user must use the mouse.

## Steps to reproduce

1. Focus the chat input.
2. Use the keyboard shortcut that toggles the sidebar.
3. Confirm that the sidebar does not toggle.
4. Press Escape.
5. Confirm that the input keeps focus.

## Expected behavior

The sidebar shortcut toggles the sidebar while the chat input has focus. Escape should also remove focus from the chat input.

SawyerHood:
Fixes #1162

## Root cause

The prompt editor consumes key events before the app command layer sees them.

1. `AppCommandProvider` listens on `window` in the bubble phase and skips any event whose default is already prevented (`AppCommandProvider.tsx:270`). TipTap's keymap cancels every chord it owns. Live check in the dev app, with the chat input focused: `Ctrl+Shift+B` (blockquote) and `Ctrl+B` reached the `window` listener with `defaultPrevented === true`. A sidebar shortcut bound to such a chord therefore did nothing while the chat input had focus.
2. ProseMirror cancels Escape unconditionally, but nothing released the editor. Live check: after Escape, `document.activeElement` was still the editor.

## User impact

- The sidebar shortcut, and any other app shortcut, silently did nothing while the chat input had focus, whenever its chord collided with the editor keymap. Users had to click the toggle.
- Escape did not release the chat input, so the keyboard could not reach the rest of the app.

## Implementation

- `AppCommandProvider.tsx`: the keybinding match and dispatch loop moves into `handleKeyboardEvent`, which reports whether a command ran. A new `useAppCommandKeyDispatch()` hook exposes it. The window listener keeps its current behavior.
- `PromptBoxInternal.tsx`: the editor keydown handler asks the app command layer first, so app keybindings win over the editor keymap. Escape blurs the editor DOM and drops the caret.

Escape priority stays intact. The typeahead menu claims Escape earlier in the same handler, so the first Escape dismisses the menu and the second releases focus. Voice recording cancels from a window capture listener that stops the event before the editor.

No server or daemon payload changed, so `HOST_DAEMON_PROTOCOL_VERSION` stays as it is.

## Validation

- New `PromptBoxAppShortcuts.test.tsx`: shortcut from a focused composer, shortcut on a chord the editor keymap also claims, Escape blur, and plain typing unaffected.
- `pnpm exec turbo run typecheck --filter=@bb/app` — pass.
- `pnpm exec vitest run --project @bb/app` — 322 files, 2418 tests, all pass.
- `pnpm exec turbo run lint --filter=@bb/app` — 0 errors, no warnings in the changed files.
- Live QA in the dev app, thread view and new-thread composer: the sidebar toggles from the focused chat input, the first Escape dismisses the mention menu and keeps focus, the second Escape blurs, and the draft text survives.

### DIFF
--- a/apps/app/src/components/commands/AppCommandProvider.tsx
+++ b/apps/app/src/components/commands/AppCommandProvider.tsx
@@ -112,6 +112,10 @@ export function AppCommandProvider({ children }: { children: ReactNode }) {
     new Map<AppCommandContextKey, Set<symbol>>(),
   );
   const sequenceRef = useRef(0);
+  // Key events already offered to the handlers, so a second delivery of the
+  // same event is a no-op. Weak so entries drop with the event.
+  const attemptedEventsRef = useRef(new WeakSet<KeyboardEvent>());
+  const clearShortcutHintHoldRef = useRef<() => void>(() => {});
 
   useEffect(() => {
     if (!showKeyboardHints) return;
@@ -126,6 +130,9 @@ export function AppCommandProvider({ children }: { children: ReactNode }) {
       shortcutHintModifierHeldRef.current = false;
       setIsShortcutHintModifierHeld(false);
     };
+    // A widget that dispatches a chord itself stops the event before this
+    // listener sees it, so the dispatcher clears the hint through this ref.
+    clearShortcutHintHoldRef.current = clearModifierHold;
     const handleKeyDown = (event: KeyboardEvent) => {
       if (!isShortcutHintModifier(event.key)) {
         if (
@@ -166,6 +173,7 @@ export function AppCommandProvider({ children }: { children: ReactNode }) {
     window.addEventListener("blur", handleBlur);
     return () => {
       clearModifierHold();
+      clearShortcutHintHoldRef.current = () => {};
       window.removeEventListener("keydown", handleKeyDown);
       window.removeEventListener("keyup", handleKeyUp);
       window.removeEventListener("blur", handleBlur);
@@ -277,6 +285,11 @@ export function AppCommandProvider({ children }: { children: ReactNode }) {
       if (event.defaultPrevented || event.isComposing || event.repeat) {
         return false;
       }
+      // A widget that dispatched first leaves the event alone when every
+      // handler declines, so the same event still reaches the window listener.
+      // Without this, those handlers would run a second time.
+      if (attemptedEventsRef.current.has(event)) return false;
+      attemptedEventsRef.current.add(event);
       const bindings = keybindingsRef.current;
       let context: AppCommandContext | null = null;
       const isMac = isMacKeyboardPlatform(browserPlatform());
@@ -292,6 +305,7 @@ export function AppCommandProvider({ children }: { children: ReactNode }) {
         context ??= currentContext(event.target);
         if (!matchesAppCommandContext(binding, context)) continue;
         if (!dispatch(binding.command, event.target)) return false;
+        clearShortcutHintHoldRef.current();
         event.preventDefault();
         event.stopPropagation();
         return true;
--- a/apps/app/src/components/commands/AppCommandProvider.tsx
+++ b/apps/app/src/components/commands/AppCommandProvider.tsx
@@ -45,6 +45,7 @@ interface AppCommandHandlerRegistration {
 interface AppCommandProviderValue {
   dispatch: (command: AppCommandId, target: EventTarget | null) => boolean;
   getShortcut: (command: AppCommandId) => AppShortcut | null;
+  handleKeyboardEvent: (event: KeyboardEvent) => boolean;
   registerContext: (
     key: AppCommandContextKey,
     source: symbol,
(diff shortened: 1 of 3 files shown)


SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am now reviewing this change for security, quality, performance, architecture, and user behavior.
I will post the complete findings after the parallel checks finish.

apps/app/src/components/commands/AppCommandProvider.tsx
```diff
@@ -283,15 +291,23 @@ export function AppCommandProvider({ children }: { children: ReactNode }) {
         if (!matchesAppShortcut(event, binding.shortcut, isMac)) continue;
         context ??= currentContext(event.target);
         if (!matchesAppCommandContext(binding, context)) continue;
-        if (!dispatch(binding.command, event.target)) return;
+        if (!dispatch(binding.command, event.target)) return false;
```
SawyerHood, line 307: 🚨 `slopcop/review` — Prevent a second command dispatch after all handlers return false.

The editor calls `dispatch` here. If all handlers return `false`, this function leaves the event unconsumed.

The same event then reaches the window listener. The provider calls the same handlers again.

Please mark each keyboard event after its first dispatch attempt. Add a test where all handlers return `false`.

apps/app/src/components/promptbox/PromptBoxInternal.tsx
```diff
@@ -2579,6 +2590,18 @@ export function PromptBoxInternal({
         }
       }
 
+      // Escape releases the composer so the keyboard can reach the rest of the
+      // app. Higher-priority Escape behavior still runs first: the typeahead
+      // menu above dismisses itself, and voice recording cancels from a window
+      // capture listener that stops the event before the editor sees it.
+      // TipTap's `blur` command defers to the next animation frame, so blur the
+      // editor DOM directly and drop the caret with it.
+      if (event.key === "Escape") {
```
SawyerHood, line 2605: 🚨 `slopcop/review` — Make Escape release a focused editor after the input becomes locked.

`setEditable(false)` prevents ProseMirror from calling this key handler. A focused editor can remain focused after the lock.

Escape cannot run this blur path in that state. Please handle Escape outside the editable-only handler.

Add a test for a locked editor that already has focus.

apps/app/src/components/commands/AppCommandProvider.tsx
```diff
@@ -283,15 +291,23 @@ export function AppCommandProvider({ children }: { children: ReactNode }) {
...
         event.preventDefault();
```
SawyerHood, line 309: 🚨 `slopcop/review` — Clear keyboard hints before this editor path stops event propagation.

The window hint listener normally clears hints after a non-modifier key. This direct editor path stops the event before that listener runs.

A displayed hint stays active until modifier release or window blur.

Please clear the hint state in the shared dispatcher. Add a test for a consumed editor shortcut.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

The chat box now lets app key shortcuts pass through. Escape moves focus out without deleting text.
It works like a door that lets app keys leave the chat box.

### Findings

I found two correctness issues and one small UI issue.

1. When all handlers say no, the editor path and window path call them twice.
2. A locked editor skips ProseMirror key handlers, so Escape cannot release existing focus.
3. An editor shortcut can stop the event before the window clears keyboard hints.

I found no security defect. The performance review found negligible extra work and no listener or render loop.

### Architecture and duplicate code

The shared dispatcher removes duplicate shortcut matching logic. Its two event paths still need per-event dispatch state.

An older `useEscapeToHide` hook is inactive and appears stale. The modal-menu gap and alternate editable checks predate this pull request.

### Validation

- GitHub CI passed all required jobs.
- The Turbo app type check passed.
- The two focused files passed all 17 tests.
- Browser QA confirmed that Escape keeps text and removes focus.
- Browser QA confirmed that Meta+Backslash toggles the sidebar from the composer.
- The full local suite passed 2,417 tests. One unrelated scroll test failed, while the GitHub app job passed.

I posted a comment review. I did not approve the pull request or request changes.

apps/app/src/components/commands/AppCommandProvider.tsx
SawyerHood: Fixed in 426ab772f. The provider now keeps a `WeakSet` of key events it has already offered to the handlers, so the window listener skips an event the editor path already tried. Test added: `offers a declined chord to the handlers only once`.

apps/app/src/components/promptbox/PromptBoxInternal.tsx
SawyerHood: Fixed in 426ab772f. Escape now also runs from an `onKeyDown` on the editor container, which fires while the editor is locked and ProseMirror skips its own handlers. It acts only when `editor.isEditable` is false, so the editable path keeps its typeahead-first priority. Test added: `releases focus on Escape while a plugin locks the composer`.

apps/app/src/components/commands/AppCommandProvider.tsx
SawyerHood: Fixed in 426ab772f. The shared dispatcher clears the hint hold through a ref before it consumes the event, so a chord dispatched from the composer releases the hint exactly as the window path did. Test added: `clears the keyboard hint when the composer runs a shortcut`.

PR #1167, opened 2026-08-07
Reset sidebar search after opening a thread

### LINKED ISSUE #1163: Opening a thread from search leaves the sidebar in search state
## Problem

When a user opens sidebar search with Command-K and selects a thread with the keyboard, the sidebar stays in search mode. The search input and query remain visible after navigation.

## Steps to reproduce

1. Press Command-K to open sidebar search.
2. Enter a search query.
3. Use the keyboard to select a thread.
4. Open the selected thread.
5. Confirm that the sidebar remains in the search state.

## Expected behavior

After the selected thread opens, clear the search input and return the sidebar to its state before the search started.

SawyerHood:
Fixes #1163

## Root cause

`AppSidebar` opened the thread in `handleThreadSearchSelectItem` (navigate + `closeOnMobile`) but never cleared the search state. `isThreadSearchActive`, the query, the cursor index, and the result items all stayed set. `ProjectList` renders the search panel in place of the thread list while `threadSearch.isActive` is true, so the sidebar stayed in search mode after navigation.

## User impact

After Command-K search and a keyboard selection, the thread opened but the sidebar kept the search field, the typed query, and the result list. The user had to press Escape twice, or click the X, to get the thread list back. Pointer selection had the same problem.

## Implementation

- New hook `useSidebarThreadSearch` owns the query, the result items, the keyboard cursor, the input ref, and the sidebar key handler. `AppSidebar` supplies only the two effects it owns: open the sidebar, and open a thread.
- `onSelectItem` opens the thread and then clears all search state. Keyboard Enter and pointer clicks share this path, so both behave the same.
- A plugin thread list reads the host search field, so its `onNavigate` gets the same reset.
- `isThreadSearchKeyboardEventTarget` moved to `sidebarThreadSearch.ts`, next to the other pure helpers.
- Escape keeps its two-step behavior: the first press clears the query, the second closes search. Search focus, section labels, and sidebar visibility do not change. Search still opens the sidebar and does not close it after navigation.

## Automated validation

- New `useSidebarThreadSearch.test.tsx`: 4 tests covering keyboard selection, pointer selection, the Escape steps, and activation.
- Mutation check: with the reset call removed, 3 of the 4 new tests fail. With it restored, all pass.
- Sidebar suite: 25 files, 200 tests, all pass.
- `pnpm exec turbo run typecheck --filter=@bb/app` — pass.
- `pnpm exec turbo run lint --filter=@bb/app` — 0 errors. The 141 warnings match the state before this change; these files add none.

## Dev Browser validation

I ran the real dev app from `scripts/bb-dev-app current` and drove it with headless Chromium through dev-browser.

Keyboard path — `Control+K`, query `needle`, `ArrowDown`, `Enter`:

| Point | URL | Search input | Options |
|---|---|---|---|
| Before search | `/` | absent | 0 |
| Query typed | `/` | `"needle"` | 3 |
| After Enter | `/projects/.../threads/thr_8b9g6fvnkd` | absent | 0 |

A string comparison of the sidebar text before search and after navigation returned `true`, so the sidebar returns to its earlier state. The opened thread row is selected and the sidebar width is unchanged.

Pointer path: `Control+K`, query `gamma`, mouse click on the row. The thread opened and the sidebar list came back the same way.

Bug reproduction in the live app: with the reset call removed, the same keyboard flow opened the thread but left `"needle"` in the input and 3 result rows on screen — the state the issue describes. Restoring the file removed that state again.


SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** The sidebar search is like a small find box. This change clears it after you open a thread.

I am SlopCop, and I am reviewing this pull request now. I will check security, code quality, performance, architecture, and the user flow.

apps/app/src/components/sidebar/AppSidebar.tsx
```diff
@@ -203,6 +140,20 @@ export function AppSidebar({
...
+    onOpenSidebar: openSidebarForThreadSearch,
+    onOpenThread: openSearchedThread,
+  });
+  const closeThreadSearch = threadSearch.onClose;
+
+  // A plugin thread list reads the host search field, so opening a thread from
+  // it ends search the same way a built-in result row does.
+  const handlePluginThreadListNavigate = useCallback(() => {
+    closeOnMobile();
+    closeThreadSearch();
```
SawyerHood: 🚨 `slopcop/review` — Update the public plugin navigation contract

This callback now clears the host search on desktop. However, `PluginThreadListProps.onNavigate` still says it is a desktop no-op in `packages/plugin-sdk/src/app-contract.ts`. The feature guide, generated declarations, and plugin-authoring skill repeat that statement.

Please update those contract surfaces, regenerate the declarations, and add a desktop plugin-path test that verifies a complete search reset.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** The search box now cleans itself after you open a thread. One instruction sheet still describes the old behavior.

I found one P2 issue. The plugin thread-list callback now clears desktop search, but the public SDK contract still calls it a desktop no-op. The feature guide, generated declarations, and plugin-authoring skill repeat that stale contract. I left an inline comment with the required updates and test.

Security and performance reviews found no issues. The architecture scan found no additional duplication that needs action. The extracted hook keeps the search state and keyboard behavior in one place.

Validation results:

- The 16 focused sidebar tests passed.
- The `@bb/app` Turbo typecheck passed.
- `git diff --check` passed.
- The browser flow passed. Search returned the target thread, navigation opened it, search closed, and the normal list returned.
- All GitHub checks passed.
- The full local app run completed all 2,418 tests, but an unrelated delayed cleanup in `splitDragSession` caused one unhandled-error exit.

I used a comment review. I did not approve the pull request or request changes.

apps/app/src/components/sidebar/AppSidebar.tsx
SawyerHood: Fixed in e150af6f4. The finding was correct: `onNavigate` no longer is a desktop no-op.

- `packages/plugin-sdk/src/app-contract.ts`, `docs/plugin-sidebar-thread-list.md`, and the `bb-plugin-authoring` skill now say it closes the mobile drawer and clears the host search field on every viewport.
- Regenerated `packages/plugin-sdk/bundled-types/*.d.ts` and `packages/templates/src/generated/plugin-sdk-dts.generated.ts`.
- The hook now owns both reset paths through `onExternalThreadOpen`, so `AppSidebar` passes one member straight to `PluginThreadList`. That removed the duplicate composition.
- New test: "clears the search state when a plugin thread list opens a thread" asserts the full reset on the desktop plugin path.

PR #1168, opened 2026-08-07
Add a usable local Cloud development workflow

ymichael:
## Summary

- add `pnpm cloud:dev` as the one-command local workflow for the Cloud dashboard and Connect worker
- run both services against one worktree-local D1 database and local Durable Objects
- use real Better Auth email/password accounts and sessions in local Cloud instead of a seeded-user authentication bypass
- route `bb.localhost` to the dashboard and `<handle>.bb.localhost` through Connect, including WebSocket upgrades
- use the existing Vite app as the tunneled origin when a source-development bb pairs with local Cloud
- make `pnpm dev` point unpaired Connect settings and code-only pairing at the matching local Cloud; explicit `--server` and `--base-url` targets still override it
- support local machine creation, revocation, `ws:` machine tunnels, and `http:` machine share URLs while keeping non-local machine enrollment HTTPS-only
- preflight all local ports, wait for the routed services to become ready, await child shutdown with SIGKILL escalation, handle SIGHUP, and tolerate clients closing rejected WebSocket upgrades

## Auth behavior

Email/password endpoints and controls are enabled only when the launcher supplies the fail-closed local auth flag on an HTTP `*.localhost` origin. Production remains GitHub-only and keeps the secure production cookie names. Local sessions use non-secure cookie names scoped to `.bb.localhost`, allowing the same real session to authenticate the dashboard and handle subdomains.

The existing GitHub control remains in the UI, but this workflow does not configure a local GitHub OAuth application; the launcher continues to supply placeholder OAuth credentials.

Local D1 state is stored under `.wrangler/cloud-dev`.

## Review follow-up

- readiness uses Node HTTP so the routing `Host` header is preserved, response bodies are drained, and failures back off
- local Vite/auth overrides apply only to `vite serve`; a production build ignores them even if they are present in the shell
- local Connect sign-in reconstructs the public handle URL, preserving `returnTo`
- HTTP versus HTTPS is derived centrally from the trusted base domain and stays off the daemon wire, so no protocol bump is needed
- gateway ports cannot overlap worker ports, and detached children are cleaned up on SIGHUP
- removed dead account payload, duplicated auth configuration/header constants, redundant cookie defaults/re-exports, and formatting-only churn
- added coverage for forwarded-header stripping, URL-template host validation, HTTP-to-WebSocket conversion, public machine origins, and the launcher regressions

## Verification

- Turbo typechecks passed for all 11 affected packages (12 tasks including the plugin SDK build)
- Turbo tests passed across 13 tasks: Connect Cloud 95, Web 53, config 96, scripts 85, Connect client 8, daemon contract 49, host daemon 505, plugin SDK 74, server 1,379, templates 23, and Connect plugin 77
- production Web build passed with local Cloud variables deliberately present; the emitted config remained on `https://getbb.app`, production D1, with no dev-auth flag or local secret
- local auth smoke passed: dashboard 200, email/password sign-up, and authenticated session lookup
- launcher smoke passed: routed readiness, bold ready URL, SIGHUP cleanup, and release of all three worktree ports


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull request for security, code quality, performance, duplication, architecture, and end-to-end behavior.

scripts/bb-cloud-dev.mjs
```diff
@@ -0,0 +1,621 @@
...
+
+const web = spawnService(
+  "web",
+  [
+    "--filter",
+    "@bb/web",
+    "exec",
+    "vite",
+    "dev",
+    "--host",
```
SawyerHood, line 239: 🚨 `slopcop/review` — P1: The seeded account is open on every network interface.

This listener uses `0.0.0.0`, while line 502 enables `DEV_AUTH_USER_ID`. The auth boundary checks the configured URL, not the request source. A LAN or Tailscale client can change account state without authentication.

Bind seeded mode to `127.0.0.1`. Disable seeded auth when remote access is necessary. Add a request test from a non-loopback host.

scripts/bb-cloud-dev.mjs
```diff
@@ -0,0 +1,621 @@
...
+    "0.0.0.0",
+    "--port",
+    String(WEB_PORT),
+  ],
+  {
+    BB_CLOUD_DEV_APP_URL: `http://127.0.0.1:${WEB_PORT}`,
+    BB_CLOUD_DEV_BASE_DOMAIN: BASE_DOMAIN,
+    BB_CLOUD_DEV_CONNECT_SERVER_URL_TEMPLATE: `http://{label}.${BASE_DOMAIN}:${CONNECT_PORT}`,
```
SawyerHood: 🚨 `slopcop/review` — P1: The GitHub-auth mode cannot authorize this local gate.

Better Auth creates a host-only cookie for `127.0.0.1`. This gate uses `<label>.localhost`, and the return URL check rejects that different base host. The desktop fallback also installs an `__Secure-` cookie with `secure: false` for this HTTP URL, which Chromium rejects.

Use a shared development parent domain or mint a gate-scoped signed session. Add a browser test for the complete `--github-auth` return flow.

plugins/connect/src/server.ts
```diff
@@ -31,7 +37,8 @@ export default async function plugin(bb: BbPluginApi) {
   tunnel = new ConnectTunnel({
     store,
     shares,
-    getLoopbackBaseUrl: () => bb.server.loopbackBaseUrl,
+    getLoopbackBaseUrl,
```
SawyerHood, line 42: 🚨 `slopcop/review` — P1: The account-origin override does not reach machine creation or revocation.

Pairing and the dashboard use this override. However, both machine operations still derive `http://localhost:<connect-port>` from the gate credential. Their routes run on the separate web port, so both calls fail.

Pass the account origin to both machine helpers. Add tests with different account and gate ports.

apps/web/src/server/api.ts
```diff
@@ -559,7 +613,7 @@ export async function createMachineCode(
   return {
     code,
     expiresInMs: CONNECT_CODE_TTL_MS,
-    serverUrl: `https://${srv.subdomain}.${baseDomain}`,
+    serverUrl: serverUrlForLabel(srv.subdomain, serverUrlTemplate),
```
SawyerHood, line 600: 🚨 `slopcop/review` — P1: The returned HTTP gate URL cannot complete the local machine lifecycle.

The installer and host daemon still require HTTPS. The daemon always creates a `wss:` tunnel. Machine share URLs also use HTTPS, and the machine page omits the local port.

Permit HTTP only for validated loopback domains. Derive the HTTP and WebSocket schemes from the trusted gate URL. Add an end-to-end machine test.

scripts/bb-cloud-dev.mjs
```diff
@@ -0,0 +1,621 @@
...
+  const deadline = Date.now() + 30_000;
+  while (Date.now() < deadline) {
+    if (child.exitCode !== null) {
+      throw new Error(`${label} exited before becoming ready`);
+    }
+    try {
+      const response = await fetch(url, {
+        signal: AbortSignal.timeout(1_000),
+      });
+      if (response.status < 500) return;
```
SawyerHood: 🚨 `slopcop/review` — P2: An HTTP 500 response causes a tight readiness loop.

The delay only occurs when `fetch` throws. A fast 500 response causes immediate retries for 30 seconds, and the response body remains unread. This can create thousands of requests and consume sockets.

Consume or cancel each body. Add the 250 ms delay after every result that is not ready. Test a server that always returns 500.

scripts/bb-cloud-dev.mjs
```diff
@@ -0,0 +1,621 @@
...
+
+function stopAll(exitCode = 0) {
+  if (stopping) return;
+  stopping = true;
+  restoreManagedDevEnv();
+  connectProxy.close();
+  connectProxy.closeAllConnections();
+  stopService(connect);
+  stopService(web);
+  setTimeout(() => process.exit(exitCode), 100);
```
SawyerHood: 🚨 `slopcop/review` — P2: Startup and shutdown can leave detached services active.

A proxy bind failure occurs after Connect starts but before cleanup exists. Normal shutdown sends `SIGTERM` and exits the parent after only 100 ms. Wrangler or Vite can keep a port and the shared database open.

Register cleanup before the first spawn. Await both child exits, then use `SIGKILL` after a deadline. Test bind failure and slow shutdown.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: This pull request adds one command for local Cloud work. The main server flow starts, but several security and machine paths still fail.

I found six issues:

- P1: Seeded authentication is reachable from LAN and Tailscale clients because Vite listens on every interface.
- P1: The `--github-auth` mode cannot authorize an HTTP `<label>.localhost` gate.
- P1: Machine creation and revocation call the Connect port instead of the separate account port.
- P1: Machine enrollment, tunnels, shares, and the machine page still require the production HTTPS topology.
- P2: HTTP 500 readiness responses cause a tight retry loop and leave response bodies unread.
- P2: Bind failures and the 100 ms shutdown path can leave detached services active.

The architecture still infers an account origin from a gate origin in several places. Carry both origins explicitly through each contract. Centralize loopback validation so IPv4, IPv6, protocol, and port rules cannot drift.

Validation passed for all affected Turbo tests and type checks. The app ran 2,415 tests, and the desktop ran 231 tests. The Cloud launcher started successfully in a real browser smoke test. The dashboard claimed a handle, generated a pairing code, and rendered the offline gate.

A LAN browser also received the seeded account state without authentication. Chromium rejected the insecure `__Secure-` desktop cookie for the HTTP gate.

I left a comment-only review, as required.

scripts/bb-cloud-dev.mjs
ymichael: Addressed by the current rewrite: seeded authentication no longer exists. Local Cloud uses real Better Auth email/password, and the Connect worker, Vite dashboard, and gateway all bind to 127.0.0.1. The local auth flag is also fail-closed to an HTTP *.localhost origin.

scripts/bb-cloud-dev.mjs
ymichael: The reviewed --github-auth mode was removed. Local development now uses email/password with cookies scoped to .bb.localhost; production remains GitHub-only. Local Cloud no longer attempts to install an insecure __Secure- cookie.

scripts/bb-cloud-dev.mjs
ymichael: Updated after the follow-up audit: readiness now lives in `scripts/lib` and uses `node:http` instead of `fetch`, so the routing `Host` header is actually preserved. It drains each response and backs off after HTTP 500 or transport failures. A real HTTP-server test verifies the received Host header, and a fake-timer test verifies the 250 ms backoff.

apps/web/src/server/api.ts
ymichael: Updated after the follow-up audit: local HTTP is derived centrally from a validated `.localhost` base domain in `@bb/connect-client`; non-local gates remain HTTPS-only. The protocol intentionally stays out of `HostDaemonConnectTunnelIdentity`, so `HOST_DAEMON_PROTOCOL_VERSION` remains 90 and no daemon rollout is triggered. Daemon tests cover HTTP→WebSocket conversion, local and production share origins, and rejection of protocol/domain mismatches. The local machine page also preserves HTTP and the gateway port.

plugins/connect/src/server.ts
ymichael: The split account/gate topology was removed. bb.localhost and every handle host now share one gateway port, so deriving the apex routes machine-code and revoke-machine to the correct account service. A regression test now covers both local machine operations.

scripts/bb-cloud-dev.mjs
ymichael: Fixed. The launcher preflights all three ports and binds the gateway before spawning either child. Signal cleanup is installed before the first spawn, awaits both process groups after SIGTERM, and escalates to SIGKILL after five seconds. A live smoke verified duplicate-port rejection, Ctrl-C exit, and release of all three worktree ports.

ymichael:
Addressed the valid findings from the follow-up review in `000b12a61`:

- fixed routed readiness checks, response draining/backoff, SIGHUP cleanup, and the high-offset port collision
- made local Vite/auth overrides serve-only and removed the duplicated local auth secret
- preserved the public local URL for sign-in `returnTo` and fixed local machine-page links
- removed the unnecessary protocol field/version bump; protocol is derived centrally from the validated base domain
- removed dead/duplicated fields, constants, defaults, re-exports, and formatting-only churn
- added regression coverage for all noted test gaps

Verification: all 12 affected-package typecheck tasks and all 13 affected-package test tasks pass. A production Web build with local variables present stayed on production config. Live local Cloud sign-up/session, routed readiness, and SIGHUP cleanup also passed.

ymichael:
Fixed the paired-status issues in a18a1ed99:

- bb connect off now revokes the presenting server credential in Cloud and closes that server tunnel; local disconnect still succeeds if Cloud is unreachable.
- The dashboard continues polling after pairing, so local disconnects and later liveness changes are reflected.
- A completed pairing now hides stale setup UI immediately, avoiding Online alongside Waiting for it to connect.

Verified with affected Turbo typechecks/tests (230 tests), a production web build, and a live local Cloud credential-revocation smoke test.

PR #1169, opened 2026-08-07
Resolve bare Pi models through sole authenticated provider

### LINKED ISSUE #1123: Pi bridge: bare model ids from bb's own spawns fail at thread.start ("Ambiguous Pi model")
## Summary

A Pi thread spawned with a bare model id (e.g. `gpt-5.6-terra`) fails at `thread.start` with:

```
Ambiguous Pi model "gpt-5.6-terra": served by azure-openai-responses, cloudflare-ai-gateway, github-copilot, openai, openai-codex, opencode. Prefix it with the provider you want.
```

(or `Failed to resolve Pi model "..."` on older Pi catalogs). Two aspects of this feel like bugs rather than the documented don't-guess policy:

### 1. bb itself produces the bare ids

The error is easy to hit without the user ever typing a model name: in our case a Codex parent thread spawned Pi worker threads, and the model id carried over to the spawn was the bare `gpt-5.6-terra`. Selections stored before #1037 have the same shape. So the "prefix it" advice targets a string the user never wrote.

### 2. The failure is deferred from spawn to start

`bb thread spawn` accepts the unresolvable model without complaint; the thread only errors when it is started — in our case days later, inside automation, where the error surfaced as silently dead worker threads. Validating the model against the provider catalog at spawn time (or at least warning) would catch this where the caller can react.

### Suggestion: resolve ambiguity via the sole authenticated provider

`resolveConfiguredModel` in `packages/agent-runtime/src/pi/bridge/sdk-session.ts` refuses to guess between providers serving the same id, on the (sound) rationale that substituting a vendor would send workspace content and billing somewhere the user never chose. But when exactly one of the matching providers has credentials configured (`modelRuntime.hasConfiguredAuth`), the others could never serve the request at all — picking the sole authenticated provider sends nothing anywhere the user hasn't already signed in to. In practice this makes bare ids from stored selections and cross-provider spawns Just Work on single-account machines, while genuinely ambiguous cases (two authed providers) still error.

We're running this as a small patch on our fork — happy to send it as a PR if you want it:
https://github.com/amadad/bb/commit/c26b69604348c161cb43f8d753997a585c844955

```ts
if (bare.length > 1) {
  const authed = bare.filter((candidate) =>
    modelRuntime.hasConfiguredAuth(candidate.provider),
  );
  if (authed.length === 1) {
    return authed[0];
  }
  // ... existing ambiguity error
}
```

### Environment

- bb-app 0.35.1 (self-hosted, built from `origin/main` @ `cce08058`)
- Pi `@earendil-works/pi-coding-agent` 0.84.0, auth: `openai-codex` OAuth only

SawyerHood:
Fixes #1123.

## Summary

- resolve an ambiguous bare Pi model through its sole authenticated provider
- keep explicit provider selections authoritative
- retain the ambiguity error when zero or multiple matching providers have authentication
- bump the host daemon protocol to update enrolled daemons before the model meaning changes

## Tests

- `pnpm exec turbo run test typecheck --filter=@bb/host-daemon-contract --filter=@bb/agent-runtime --force`
- 879 tests passed after rebasing on #1148
- both package typechecks passed

## Risk

The behavior changes only for bare model IDs with multiple catalog matches and exactly one authenticated provider. Explicit provider-prefixed IDs remain unchanged.

### DIFF
--- a/packages/agent-runtime/src/pi/bridge/__tests__/sdk-session.test.ts
+++ b/packages/agent-runtime/src/pi/bridge/__tests__/sdk-session.test.ts
@@ -449,6 +449,34 @@ describe("PiSdkSession", () => {
     );
   });
 
+  it("uses the sole authenticated provider for an ambiguous bare model id", async () => {
+    mockGetModel.mockReturnValue(undefined);
+    mockGetModels.mockReturnValue([
+      { id: "gpt-5.6-terra", provider: "azure-openai-responses" },
+      { id: "gpt-5.6-terra", provider: "openai" },
+      { id: "gpt-5.6-terra", provider: "openai-codex" },
+    ]);
+    mockHasConfiguredAuth.mockImplementation(
+      (provider: string) => provider === "openai-codex",
+    );
+    const session = new PiSdkSession(
+      {
+        cwd: "/tmp/project",
+        model: "gpt-5.6-terra",
+      },
+      vi.fn(),
+      vi.fn(),
+    );
+
+    await session.start();
+
+    expect(mockCreateAgentSession).toHaveBeenCalledWith(
+      expect.objectContaining({
+        model: { id: "gpt-5.6-terra", provider: "openai-codex" },
+      }),
+    );
+  });
+
   it("refuses to guess when two providers serve one bare model id", async () => {
     mockGetModel.mockReturnValue(undefined);
     mockGetModels.mockReturnValue([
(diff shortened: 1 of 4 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am reviewing this pull request for security, code quality, performance, architecture, and test coverage. I will also test the complete path when this change has a suitable route.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

Many doors can have the same name. This change picks the only door with a working key. If two keys work, it asks you to name the door.

### Review

I found no actionable issue.

- Security: An explicit provider still wins. A bare model selects only one authenticated match. The code does not expose credentials.
- Code quality: The strict resolver keeps the current error behavior. The new test covers the new route. Existing tests cover explicit and multiple-match routes.
- Architecture: The host daemon owns the local Pi authentication state. Protocol version 82 correctly updates old daemons before the meaning changes.
- Duplicate code: The Pi command-line resolver has a similar rule. It also accepts fuzzy and custom models. Reuse would change this strict bridge behavior.
- Performance: The change adds one small in-memory scan during session start. It adds no network, disk, or retry work.
- Tests: All 872 package tests passed. Both package type checks passed. `git diff --check` also passed.
- Browser test: This change has no user interface route, so browser testing does not apply.

PR #1171, opened 2026-08-08
Add official System Monitor plugin

charpeni:
This is a proposal to add `System Monitor` as an official plugin, please close this pull request if you believe it shouldn't be part of the official/built-in plugins and I will move this to my repository.

I've been running bb on a VM and I've been enjoying this plugin architecture, I figured it could be a good idea to show stats of the host machine, especially the remaining space.

| Desktop | Mobile |
| - | - |
| bettershot_1786149251793 | bettershot_1786149239526 |

## Summary

- Add System Monitor as an install-on-demand official plugin under the Host access category
- Provide a responsive sidebar panel for CPU usage and speed, memory, disk space, load average, uptime, and host details
- Refresh metrics every five seconds and handle mobile sidebar-button spacing
- Add `bb system-monitor [show] [--json]` for human-readable and structured output
- Document the plugin across the CLI guide, agent skill, configuration, and official plugin release process

System Monitor reports metrics for the host running the bb server, not separately enrolled execution machines. CPU speed uses `os.cpus()` with a Linux `/proc/cpuinfo` fallback.


MGrin:
Nice to see this proposed as an official plugin — host visibility is genuinely useful when agents are the ones loading the machine.

I built something very similar independently ([bb-plugin-system](https://github.com/MGrin/bb-plugin-system)) and, in auditing it, measured a few things on macOS that may be worth checking against this implementation before it becomes official. Sharing them because they bit me, not as criticism of the PR:

**1. `vm_stat`-derived "memory used" is easy to get wrong.** Summing `Pages active + wired + compressor` looks right but isn't: the active *queue* includes file-backed cache (counted as used when it shouldn't be) and excludes dirty anonymous pages sitting on the inactive queue (excluded when they shouldn't be). Activity Monitor's formula is `anonymous - purgeable + wired + compressed`. On my machine (M4, 24 GiB, 16 KiB pages) the difference was **16.0 GiB vs 18.1 GiB — understated by 2.2 GiB / ~9 points**, and because it moves with the cache/anon split it distorts the shape of a history graph, not just the level.

**2. "Memory pressure" isn't `used / total`.** macOS exposes the real signal at `kern.memorystatus_vm_pressure_level` (1 = normal). I measured it reporting green while my own UI painted a 67% "pressure" meter. A Mac at 95% used with no swap activity is genuinely fine, so the label matters.

**3. Load average is not CPU%.** Dividing `vm.loadavg` by `hw.ncpu` gave me a 39% CPU bar on an otherwise idle machine (load counts blocked threads). Tick-delta utilization is the honest number. Related: on Apple Silicon `hw.ncpu` mixes P- and E-cores (`hw.perflevel0/1.logicalcpu`), so per-core normalization is a bit of a lie either way.

[`systeminformation`](https://www.npmjs.com/package/systeminformation) (MIT, zero deps, ~858 KB) implements 1 and 3 correctly and gets Linux/Windows for free; I'm migrating to it. Its process table is just `ps` with more columns though, so hand-rolling that part is fine.

**On refresh cadence:** a fixed short interval spawning several child processes per tick fights macOS timer coalescing, and App Nap silently stretches it when the app is backgrounded anyway. What I'm moving to is an adaptive interval — frequent while a thread is running a turn or the panel is open, sparse otherwise — which bb can do better than a generic monitor because it knows when work is happening. Worth considering for an official plugin that will run on every user's machine.

The one thing mine has that this doesn't is **retained history** (24h ring buffer + sparklines + `bb system history`), which is what lets you answer "what was the machine doing while that fleet ran?" after the fact. If that's interesting for the official plugin I'm happy to contribute it here rather than maintain a parallel implementation — just say the word and I'll open a PR against this branch. Otherwise I'll keep mine as a community plugin and point people here.

SawyerHood:
> \<Automated Message\>
>
> We will make sure to consider this to our extension marketplace when we have it up and running. We will contact you when it is ready. We are closing this out for now.

PR #1174, opened 2026-08-08
Show thread numbers in sidebar

### LINKED ISSUE #1243: Optionally show thread numbers in the sidebar
## Problem

BB supports direct navigation to the first nine visible threads, but the sidebar does not show which number maps to each thread. Users must remember or infer the mapping.

## Proposed behavior

- Add a **Show thread numbers** toggle under Settings → General.
- Keep the toggle off by default.
- Show subtle fixed-width `1` through `9` prefixes beside the first nine built-in sidebar threads when enabled.
- Derive each number from the rendered thread order so it matches the existing navigation commands.
- Persist the setting across clients and restarts.
- Expose the same setting through the `bb` CLI.

## Prototype

I implemented and tested this behavior in PR #1174.

Does this behavior and scope align with the project?

ryanbbrown:
Closes #1243.

## Summary

- add a browser-local `Thread numbers` option to the sidebar display menu
- keep the option off by default and remember it in local storage
- show subtle, fixed-width `1`–`9` prefixes beside the first nine rendered thread rows
- update the numbers when rows mount, move, unmount, or change viewport so they match the existing thread jump shortcuts
- observe sidebar changes only while numbers or keyboard shortcut hints are visible, and clear assignments when observation stops

## User impact

Thread numbers stay hidden by default. A user can open the sidebar display menu and select `Thread numbers`. The choice applies to that browser instead of every bb client.

## Testing

- `pnpm exec turbo run test --filter=@bb/app -- --run src/components/sidebar/AppSidebar.test.tsx src/components/sidebar/ThreadRow.test.tsx src/components/sidebar/sidebarThreadShortcuts.test.ts src/components/sidebar/SidebarDisplayOptionsMenu.test.tsx`
  - 4 files, 80 tests passed


SawyerHood:

PR #1177, opened 2026-08-08
Find Claude CLI in well-known locations and clarify the missing-CLI error

SawyerHood:
## Summary

A bb 0.35.1 user hit this error on every Claude turn:

> Native CLI binary for darwin-arm64 not found. Reinstall @anthropic-ai/claude-agent-sdk without --omit=optional, or set options.pathToClaudeCodeExecutable.

Packaged bb builds inline the Claude Agent SDK into the bridge bundle and do not ship the SDK's native platform package (217 MB per platform), so the SDK's bundled-binary fallback can never work in a shipped install. When the login-shell PATH probe also misses the user's claude install (theirs was at `~/.local/bin/claude`), the session fails with the raw SDK error above.

## Changes

- `resolveClaudeCodeExecutable` now checks well-known Claude install locations (`~/.local/bin/claude`, `~/.claude/local/claude`, `/opt/homebrew/bin/claude`, `/usr/local/bin/claude`) when PATH discovery finds nothing. This also benefits model listing, which shares the resolver.
- The bridge translates the SDK's "Native CLI binary ... not found" error into bb guidance: install Claude Code, or set `BB_CLAUDE_CODE_EXECUTABLE`, then restart bb. The old message told users to reinstall an npm package they do not own.

## Tests

- New test: falls back to well-known install locations when PATH discovery fails.
- `pnpm exec turbo run typecheck test --filter=@bb/agent-runtime`: 43 files, 831 tests pass.

## Risks

- The well-known-location scan runs only after PATH discovery fails, so existing setups keep their current resolution.
- A machine with a stale claude binary at a well-known location could now pick it up where the session previously failed outright; that is strictly better than failing.

### DIFF
--- a/packages/agent-runtime/src/claude-code/bridge/sdk-session.ts
+++ b/packages/agent-runtime/src/claude-code/bridge/sdk-session.ts
@@ -8,6 +8,11 @@ import {
   type SDKUserMessage,
 } from "@anthropic-ai/claude-agent-sdk";
 import type { ClaudePermissionMode } from "../interactive-contract.js";
+import {
+  isMissingClaudeCliMessage,
+  missingClaudeCliGuidance,
+  translateMissingClaudeCliError,
+} from "./missing-cli-error.js";
 
 export interface SdkSessionOptions {
   cwd: string;
@@ -74,25 +79,10 @@ function appendBoundedText(args: AppendBoundedTextArgs): string {
   return next.slice(next.length - SDK_STDERR_TAIL_MAX_CHARS);
 }
 
-// The Agent SDK throws this when it falls back to its bundled CLI binary and
-// cannot resolve the platform package. Packaged bb builds never ship that
-// package (the bridge bundle inlines the SDK), so the SDK's "reinstall"
-// guidance is wrong for bb users; point them at the real remedies instead.
-const MISSING_NATIVE_CLI_PATTERN = /Native CLI binary for .+ not found/;
-const MISSING_CLAUDE_CLI_GUIDANCE =
-  "bb could not find the Claude Code CLI on this machine. Install Claude Code (https://claude.com/claude-code), or set BB_CLAUDE_CODE_EXECUTABLE to the full path of the claude binary, then restart bb.";
-
-function translateMissingClaudeCliError(error: unknown): unknown {
-  if (error instanceof Error && MISSING_NATIVE_CLI_PATTERN.test(error.message)) {
-    return new Error(MISSING_CLAUDE_CLI_GUIDANCE, { cause: error });
-  }
-  return error;
-}
-
 function getErrorMessage(error: unknown): string {
   const message = error instanceof Error ? error.message : String(error);
-  return MISSING_NATIVE_CLI_PATTERN.test(message)
-    ? MISSING_CLAUDE_CLI_GUIDANCE
+  return isMissingClaudeCliMessage(message)
+    ? missingClaudeCliGuidance()
     : message;
 }
 
--- a/packages/agent-runtime/src/claude-code/bridge/sdk-session.ts
+++ b/packages/agent-runtime/src/claude-code/bridge/sdk-session.ts
@@ -74,8 +74,26 @@ function appendBoundedText(args: AppendBoundedTextArgs): string {
   return next.slice(next.length - SDK_STDERR_TAIL_MAX_CHARS);
 }
 
+// The Agent SDK throws this when it falls back to its bundled CLI binary and
+// cannot resolve the platform package. Packaged bb builds never ship that
+// package (the bridge bundle inlines the SDK), so the SDK's "reinstall"
+// guidance is wrong for bb users; point them at the real remedies instead.
+const MISSING_NATIVE_CLI_PATTERN = /Native CLI binary for .+ not found/;
+const MISSING_CLAUDE_CLI_GUIDANCE =
+  "bb could not find the Claude Code CLI on this machine. Install Claude Code (https://claude.com/claude-code), or set BB_CLAUDE_CODE_EXECUTABLE to the full path of the claude binary, then restart bb.";
+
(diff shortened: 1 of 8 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the Slop Cop. I am reviewing this pull request for security, code quality, architecture, performance, and test coverage.

I will also test the affected behavior where practical. I will post one final review after the parallel checks finish.

packages/agent-runtime/src/claude-code/bridge/sdk-session.ts
```diff
@@ -74,8 +74,26 @@ function appendBoundedText(args: AppendBoundedTextArgs): string {
   return next.slice(next.length - SDK_STDERR_TAIL_MAX_CHARS);
 }
 
+// The Agent SDK throws this when it falls back to its bundled CLI binary and
+// cannot resolve the platform package. Packaged bb builds never ship that
+// package (the bridge bundle inlines the SDK), so the SDK's "reinstall"
+// guidance is wrong for bb users; point them at the real remedies instead.
+const MISSING_NATIVE_CLI_PATTERN = /Native CLI binary for .+ not found/;
+const MISSING_CLAUDE_CLI_GUIDANCE =
+  "bb could not find the Claude Code CLI on this machine. Install Claude Code (https://claude.com/claude-code), or set BB_CLAUDE_CODE_EXECUTABLE to the full path of the claude binary, then restart bb.";
```
SawyerHood: 🚨 `slopcop/review` — The suggested override cannot reach this process

`RuntimeProviderProcessManager` removes inherited `BB_*` variables before the bridge starts. The host then adds only `PATH`. Thus, `BB_CLAUDE_CODE_EXECUTABLE=...` plus a restart still leaves this value absent.

Pass this supported override into the bridge process. A process-level test should verify the complete path.

packages/agent-runtime/src/claude-code/bridge/sdk-session.ts
```diff
@@ -74,8 +74,26 @@ function appendBoundedText(args: AppendBoundedTextArgs): string {
...
+
+function translateMissingClaudeCliError(error: unknown): unknown {
```
SawyerHood: 🚨 `slopcop/review` — The error translator does not cover model discovery

`model-list.ts` calls the SDK `query()` function directly. It then sends the raw failure to the caller. A missing CLI can fail during `model/list`, before `SdkSession` starts.

Share this translator with the model probe. Test the model probe and session paths.

packages/agent-runtime/src/claude-code/bridge/session-options.ts
```diff
@@ -223,10 +240,24 @@ export function resolveClaudeCodeExecutable(
...
     executableName: "claude",
     pathEnv: args.env.PATH,
   });
+  if (executableOnPath) {
+    return executableOnPath;
+  }
+
+  for (const candidate of wellKnownClaudeExecutablePaths(args.env)) {
+    try {
+      accessSync(candidate, constants.X_OK);
```
SawyerHood: 🚨 `slopcop/review` — Implicit paths can cross a privilege boundary

When bb runs with elevated privileges, this code can select a user-controlled `HOME` binary or a writable global path. The `X_OK` check also accepts searchable directories. It does not verify the file owner or a symlink target.

Do not use implicit paths during elevated execution. Otherwise, verify a regular file, its resolved target, its owner, and its writable bits.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: bb now looks for Claude in common folders. It also tries to replace a confusing missing-file message with useful instructions.

I found three issues:

- Medium: The suggested `BB_CLAUDE_CODE_EXECUTABLE` override does not reach the bridge process. The runtime removes inherited `BB_*` variables and restores only `PATH`.
- Medium: Model discovery calls the SDK directly. It can still show the incorrect SDK reinstall message before a session starts.
- Medium: The new implicit paths can execute an untrusted file when bb runs with elevated privileges. The `X_OK` check also accepts directories.

The architecture scan found one related mismatch. Host provider status checks only `PATH`, while the runtime now checks extra paths. One shared, platform-aware resolver would keep these results consistent.

The local probe cost is small. Four failed fixed-path checks took about 12 microseconds. A browser test does not apply because this change only affects the backend bridge.

Validation completed on the exact pull request SHA:

- The `@bb/agent-runtime` Turbo type check passed.
- All 831 package tests passed across 43 files.
- A focused run passed 77 bridge and session tests.
- `git diff --check` passed.

I added inline comments for the three confirmed issues. I did not use the approve or request-changes review states.

packages/agent-runtime/src/claude-code/bridge/sdk-session.ts
SawyerHood: Fixed in ea90e4fa57: the daemon now forwards BB_CLAUDE_CODE_EXECUTABLE into the runtime shell env, and providerProcessEnvFromShellEnv passes it through the BB_* spawn sanitization to the bridge process.

packages/agent-runtime/src/claude-code/bridge/sdk-session.ts
SawyerHood: Fixed in ea90e4fa57: the translator moved to missing-cli-error.ts and model-list.ts now wraps both the query() call and initializationResult() with it.

packages/agent-runtime/src/claude-code/bridge/session-options.ts
SawyerHood: Fixed in ea90e4fa57: well-known locations are skipped when the process runs as root, and all candidates (PATH and well-known) must now be regular files after symlink resolution, not just X_OK. Ownership checks are intentionally omitted: the daemon runs as the user, and the explicit BB_CLAUDE_CODE_EXECUTABLE override remains available for elevated setups.

PR #1195, opened 2026-08-08
Preserve host-daemon startup error causes

### LINKED ISSUE #1119: host-daemon startup fetch failure swallows underlying error cause
## Workflow

Launched the packaged desktop app (`bb.app`) on a fresh macOS install (Apple Silicon). The app was closed shortly before via the normal app-quit flow, then reopened.

## What happened

The app failed to start with:

> Could not start bb — Port http://127.0.0.1:38886 is responding, but it does not look like bb: /api/v1/system/config returned "This operation was aborted."

`~/.bb/logs/` contains a `process-host-daemon-startupFailure-*.json` diagnostic:

```json
{
  "diagnosticVersion": 1,
  "kind": "startupFailure",
  "processName": "host-daemon",
  "error": {
    "name": "TypeError",
    "message": "fetch failed",
    "stack": "TypeError: fetch failed"
  }
}
```

`server.1.log` shows the server bound its listener (`"msg":"Server listening","port":38886`) at essentially the same moment the host-daemon crashed. No `cause` is attached to the logged error, so the underlying reason the `fetch` failed (connection refused because the server wasn't listening yet? DNS? something else?) is not recoverable from the logs.

## Expected

Either:
- the host-daemon's startup fetch should retry with backoff instead of treating a single early failure as fatal (plausible race: it fires before the server's listener is fully bound), and/or
- the error handler should log the fetch failure's `cause` (Node's `fetch` attaches the underlying `Error` — e.g. `ECONNREFUSED` — to `TypeError: fetch failed` via `.cause`), so the failure is diagnosable from `~/.bb/logs/` without needing to reproduce it under a debugger.

Right now the diagnostic JSON captures `error.stack` but not `error.cause`, so the most useful piece of information is dropped.

## Environment

- bb 0.35.1, packaged desktop app
- macOS 26.3, Apple Silicon
- Node (bundled Electron) — desktop app's own bundled runtime

SawyerHood:
## Reproduction

Starting the host daemon against a closed server port produced TypeError: fetch failed. The startup report omitted the nested ECONNREFUSED cause.

## Root cause

The safe diagnostic writer copied only the outer error name, message, and stack.

## Fix

Serialize nested error causes and system error codes. Stop cyclic or deep cause chains with explicit truncation markers. Add regression tests for each case.

Fixes #1119

### DIFF
--- a/packages/process-utils/src/index.ts
+++ b/packages/process-utils/src/index.ts
@@ -63,6 +63,7 @@ export interface WriteSafeProcessDiagnosticReportArgs
 }
 
 const MAX_DIAGNOSTIC_ERROR_CAUSE_DEPTH = 8;
+const MAX_DIAGNOSTIC_AGGREGATE_ERRORS = 8;
 
 interface SafeProcessDiagnosticError {
   name: string;
@@ -70,6 +71,8 @@ interface SafeProcessDiagnosticError {
   stack?: string;
   code?: string;
   cause?: SafeProcessDiagnosticError;
+  errors?: SafeProcessDiagnosticError[];
+  errorsTruncated?: number;
   truncationReason?: "cycle" | "depth";
 }
 
@@ -241,6 +244,20 @@ function serializeDiagnosticError(
         depth + 1,
       );
     }
+    if (error instanceof AggregateError) {
+      const aggregateErrors = error.errors.slice(
+        0,
+        MAX_DIAGNOSTIC_AGGREGATE_ERRORS,
+      );
+      serialized.errors = aggregateErrors.map((aggregateError) =>
+        serializeDiagnosticError(aggregateError, seenErrors, depth + 1),
+      );
+      const errorsTruncated = error.errors.length - aggregateErrors.length;
+      if (errorsTruncated > 0) {
+        serialized.errorsTruncated = errorsTruncated;
+      }
+    }
+    seenErrors.delete(error);
     return serialized;
   }
 
--- a/packages/process-utils/src/index.ts
+++ b/packages/process-utils/src/index.ts
@@ -62,10 +62,15 @@ export interface WriteSafeProcessDiagnosticReportArgs
   createReportId?: () => string;
 }
 
+const MAX_DIAGNOSTIC_ERROR_CAUSE_DEPTH = 8;
+
 interface SafeProcessDiagnosticError {
   name: string;
   message: string;
   stack?: string;
+  code?: string;
+  cause?: SafeProcessDiagnosticError;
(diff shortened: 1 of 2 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the Slop Cop. I am reviewing this pull request for security, code quality, architecture, performance, and test behavior.

packages/process-utils/src/index.ts
```diff
@@ -192,17 +197,50 @@ function formatDiagnosticTimestamp(date: Date): string {
...
       name: error.name,
       message: error.message,
     };
     if (error.stack !== undefined) {
       serialized.stack = error.stack;
     }
+    if ("code" in error && typeof error.code === "string") {
+      serialized.code = error.code;
+    }
+    if (error.cause !== undefined) {
```
SawyerHood, line 240: 🚨 `slopcop/review` — `AggregateError` details are lost

Node 22 uses `AggregateError` when `localhost` resolves to IPv4 and IPv6. The report records an empty message and `ECONNREFUSED`, but drops the `errors` array. That array contains each failed address and can identify an IPv4 or IPv6 connection fault. Please serialize a bounded `AggregateError.errors` list with the same cycle and depth guards.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

## ELI5

The program now opens a small error box inside a large error box. This helps people find why it stopped.

## Review

I found one non-blocking diagnostic gap. Node 22 can put IPv4 and IPv6 failures in `AggregateError.errors`. The new serializer drops that array, so the report loses each failed address. I left an inline comment with the verified case.

The security review found no issue. Reports stay local, filenames remain safe, and this change adds no wire contract.

The performance review found no issue. The cause walk has one path, a depth limit of eight, and cycle detection.

The codebase has no equivalent safe JSON serializer. Some message-only formatters exist, but they do not duplicate this code. The host-daemon and server have similar stderr helpers, but this change does not require that refactor.

## Validation

- `@bb/process-utils` type check passed at the exact head SHA.
- All 11 `@bb/process-utils` tests passed at the exact head SHA.
- A Node 22 runtime check reproduced the lost `AggregateError.errors` details.
- All active GitHub checks passed.
- Browser testing does not apply because this pull request adds no browser route or UI behavior.

Outcome: comment only. I found no blocking issue.

packages/process-utils/src/index.ts
SawyerHood: Fixed in 2b4057874. The report now serializes up to eight AggregateError entries with the same cycle and depth guards, and records the omitted count. I added a regression test for the bounded list.

PR #1196, opened 2026-08-08
Protect pending turns from runtime eviction

### LINKED ISSUE #1156: claude-code: new thread's first turn silently killed by idle-runtime eviction on cosmetic shell-env drift (fnm multishells) — stuck on "Working…" forever
## Summary

On this machine, **14 of the last 16 claude-code threads created from the UI hung on "Working…" forever at their first message**. The claude CLI spawns and receives the prompt, then is silently shut down a few seconds later; the server never learns the turn died, shows "Working…" indefinitely, and the Stop button fails internally with `No runtime exists for environment …`. Stopping and re-sending ("kicking") the thread always recovers it.

Root cause (traced end-to-end with strace on the host daemon, session files, and the events table): **the daemon's shell-env refresh treats a cosmetic PATH difference as a real change and mass-evicts "idle" runtimes — and a claude-code thread whose first turn is still starting up counts as idle.**

## Environment

- bb-app 0.35.1, headless Linux server install (Ubuntu, Node v22.22.2 via **fnm**), accessed remotely via bb connect
- Provider: claude-code (Claude Code CLI under `~/.local/share/fnm/...`)
- zsh with `eval "$(fnm env)"` in `.zshenv`/`.zprofile`/`.zshrc`

## The failure chain

1. The daemon resolves the user's login-shell environment (`zsh -ilc 'printf __BB_SHELL_ENV_START__; env; …'`) with a ~10s TTL cache (`DK=1e4` in the daemon bundle).
2. Every probe evals `fnm env`, which mints a **unique** `/run/user/<uid>/fnm_multishells/<pid>_<timestamp>` directory and prepends it to PATH. So **every resolution differs from the last one** (this box has accumulated >100k multishell dirs).
3. `replaceBaseShellEnv` compares envs by strict equality, sees a "change", and calls `shutdownProviderMaintenanceRuntime()` + **`evictIdleRuntimeEntries()`**.
4. `evictIdleRuntimeEntries()` shuts down every runtime entry with no active thread IDs, terminals, or in-flight thread commands. A claude-code thread's first turn is only **queued** during CLI startup — `thread.start` settles immediately (prompt queued via queue-operations) and `turn/started` doesn't fire until ~5–7s later — so the freshly spawned runtime **looks idle** and is evicted mid-startup.
5. The CLI dies gracefully (session jsonl records `[Request interrupted by user]` with `"interruptedByShutdown": true` ~2.4s after the prompt was committed). Nothing is logged daemon- or server-side.
6. Server-side, the turn was dispatched but `turn/started` never arrives, and there is **no watchdog/timeout** for this state → the thread shows "Working…" forever.
7. Because eviction deletes the entry, a subsequent user Stop fails with `CommandDispatchError: No runtime exists for environment env_…` (`unknown_environment`) — visible in host-daemon/server logs for every hung thread.
8. Stop + re-send recovers because `turn.submit` recreates the runtime and resumes the session.

The UI flow reliably loses the race because opening the new thread view triggers extra daemon work (composer/model warm-up → maintenance runtime creation → another TTL-expired env probe) right inside the vulnerable window. `bb thread spawn` threads usually survive — same `thread/start` payload byte-for-byte, just no follow-on probe in the window.

## Captured evidence (one instance, strace attached to the daemon)

- `14:31:12.084` server: thread created, `thread/start` + `turn/start` dispatched to the (pre-existing) env bridge — both visible in the trace as daemon pipe writes
- `14:31:12.349` daemon: `execve zsh -ilc 'env'` probe → `fnm env --shell zsh` → new multishell dir
- `14:31:13.081` bridge spawns `claude --output-format stream-json … --session-id 82f29bc8-…`
- `14:31:15.361` second probe `zsh -lc 'env'` → another new multishell dir (`…_1786131075413`)
- `14:31:20.3` probe result lands → env "differs" → eviction: **the env bridge exits, the 7-second-old claude process is orphaned (reparented to init) and dies**. A new maintenance bridge spawns at `14:31:20.28` using the fresh multishell path.
- Thread shows "Working…" until the user stops it at `14:31:37` (event `system/thread/interrupted`), with no `turn/started` ever recorded.

Multishell-dir creation timestamps line up within seconds of the CLI death in **every** hung thread examined across two days.

## Suggested fixes (any one of these breaks the chain; all three are worth having)

1. **Don't treat cosmetic env drift as a change.** Ignore volatile components when diffing the resolved shell env (fnm multishell paths are the obvious case), or only evict when something the runtimes actually depend on changed — or don't evict running-state environments on env refresh at all.
2. **Don't classify a starting thread as idle.** A runtime whose `thread.start`/`turn/start` was accepted but whose turn hasn't started yet should count as active work for `evictIdleRuntimeEntries()`/`replaceEntryForSkillCatalog()` (e.g., keep the in-flight retain until `turn/started`, not until the RPC settles).
3. **Add a turn-start watchdog.** If a dispatched turn produces no `turn/started` within N seconds, fail the turn visibly instead of showing "Working…" forever. Relatedly, when a runtime is shut down/exits while a turn is pending, surface that as a thread error — today it is completely silent, and the only artifact is the later `No runtime exists for environment` error when the user clicks Stop.

---

*Investigated with Claude Code on the affected machine; happy to provide the full strace capture, session files, or run a patched build to verify a fix.*

SawyerHood:
## Reproduction

A focused provider test accepts `turn/start` without emitting `turn/started`. An FNM-shaped `PATH` refresh then exercises idle-runtime eviction during that gap.

## Root cause

The command retain ended after the provider accepted the queued prompt. The runtime had pending turn work, but neither the active-turn predicate nor the in-flight-command predicate reported it.

## Fix

Expose pending turn-start work from the agent runtime and count it as active environment work. Regression tests cover the event gap and FNM-shaped shell environment changes.

Fixes #1156

### DIFF
--- a/apps/host-daemon/src/runtime-manager.ts
+++ b/apps/host-daemon/src/runtime-manager.ts
@@ -397,7 +397,7 @@ export class RuntimeManager {
   listActiveThreads(): HostDaemonActiveThread[] {
     const activeThreads: HostDaemonActiveThread[] = [];
     for (const entry of this.entries.values()) {
-      for (const threadId of entry.runtime.getActiveThreadIds()) {
+      for (const threadId of entry.runtime.getLiveThreadIds()) {
         activeThreads.push({
           threadId,
         });
@@ -485,8 +485,7 @@ export class RuntimeManager {
   private entryHasActiveRuntimeWork(entry: RuntimeEntry): boolean {
     return (
       entry.terminals.size > 0 ||
-      entry.runtime.getActiveThreadIds().length > 0 ||
-      entry.runtime.hasPendingTurnStart() ||
+      entry.runtime.getLiveThreadIds().length > 0 ||
       entry.runtime.hasOpenBackgroundWork()
     );
   }
--- a/apps/host-daemon/src/runtime-manager.ts
+++ b/apps/host-daemon/src/runtime-manager.ts
@@ -486,6 +486,7 @@ export class RuntimeManager {
     return (
       entry.terminals.size > 0 ||
       entry.runtime.getActiveThreadIds().length > 0 ||
+      entry.runtime.hasPendingTurnStart() ||
       entry.runtime.hasOpenBackgroundWork()
     );
   }
--- a/apps/host-daemon/src/app.test.ts
+++ b/apps/host-daemon/src/app.test.ts
@@ -278,11 +278,8 @@ function createFakeRuntime(): AgentRuntime {
     hasThread() {
       return false;
     },
-    getActiveThreadIds() {
+    getLiveThreadIds() {
       return [];
-    },
-    hasPendingTurnStart() {
-      return false;
     },
     hasOpenBackgroundWork() {
       return false;
--- a/apps/host-daemon/src/app.test.ts
+++ b/apps/host-daemon/src/app.test.ts
@@ -281,6 +281,9 @@ function createFakeRuntime(): AgentRuntime {
     getActiveThreadIds() {
       return [];
     },
+    hasPendingTurnStart() {
+      return false;
+    },
     hasOpenBackgroundWork() {
       return false;
     },
--- a/apps/host-daemon/src/command-dispatch-support.test.ts
+++ b/apps/host-daemon/src/command-dispatch-support.test.ts
@@ -84,6 +84,9 @@ function makeRuntime(args: MakeRuntimeArgs): AgentRuntime {
     getActiveThreadIds() {
       return [];
     },
+    hasPendingTurnStart() {
+      return false;
+    },
     hasOpenBackgroundWork() {
       return false;
     },
--- a/apps/host-daemon/src/command-dispatch-support.test.ts
+++ b/apps/host-daemon/src/command-dispatch-support.test.ts
@@ -81,11 +81,8 @@ function makeRuntime(args: MakeRuntimeArgs): AgentRuntime {
     hasThread() {
       return false;
     },
-    getActiveThreadIds() {
+    getLiveThreadIds() {
       return [];
-    },
-    hasPendingTurnStart() {
-      return false;
     },
     hasOpenBackgroundWork() {
       return false;
--- a/apps/host-daemon/src/command-dispatch.test.ts
+++ b/apps/host-daemon/src/command-dispatch.test.ts
@@ -205,8 +205,7 @@ function createRuntime(): FakeDispatchRuntime {
         : null,
     reapIdleProviderSessions: vi.fn(async () => ({ reapedSessions: [] })),
     hasThread: (threadId) => hostedThreadIds.has(threadId),
-    getActiveThreadIds: () => [...activeTurnsByThreadId.keys()],
-    hasPendingTurnStart: () => false,
+    getLiveThreadIds: () => [...activeTurnsByThreadId.keys()],
     hasOpenBackgroundWork: () => false,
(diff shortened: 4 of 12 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I started the review. I will check security, code quality, performance, architecture, tests, and end-to-end scope.

apps/host-daemon/src/runtime-manager.ts
```diff
@@ -477,6 +477,7 @@ export class RuntimeManager {
     return (
       entry.terminals.size > 0 ||
       entry.runtime.getActiveThreadIds().length > 0 ||
+      entry.runtime.hasPendingTurnStart() ||
```
SawyerHood: 🚨 `slopcop/review` — A daemon reconnect can interrupt the pending turn.

This check protects a pending turn from local runtime eviction. However, `listActiveThreads()` still reports only started turns.

A session open sends that incomplete list. The server then interrupts each active thread that the daemon omits.

A same-instance reconnect before `turn/started` can therefore interrupt an accepted live turn.

Expose pending thread identifiers through a shared live-thread query. Use that query for eviction and session reports.

Add a daemon reconnect test for the period before the first turn event. A wire-content change also requires a daemon protocol version increase.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: The change keeps a new turn alive while its provider prepares the first event. A daemon reconnect can still tell the server that this turn does not exist.

I found one high-severity correctness issue. A same-instance daemon reconnect can interrupt the accepted turn before its first lifecycle event.

The inline comment gives the full failure path and a specific fix. Use one shared live-thread query for eviction and reconnect reports.

I found no new security issue. The possible long retention matches the existing behavior for a started active turn.

The architecture scan found no other duplicate implementation. The new separate boolean does expose split definitions of live work across eviction and reconnect code.

Validation passed:

- `git diff --check`
- 831 agent-runtime tests
- 493 host-daemon tests
- Turbo type checks for both packages
- All GitHub checks, including both package smoke checks

A browser test does not apply because this pull request has no UI path. Add a daemon and server reconnect integration test instead.

I posted a comment review only. I did not approve this pull request or request changes.

apps/host-daemon/src/runtime-manager.ts
SawyerHood: Fixed in 357b27191 and updated after the main merge in be8d41eb1. AgentRuntime now exposes one live-thread list that includes active and pending turns. Runtime eviction and session reconciliation use that list. The RuntimeManager test now covers a pending turn in the reconnect report. I also increased HOST_DAEMON_PROTOCOL_VERSION to 84 because current main already used version 83 and the report meaning changed.

SawyerHood:
Reconnect check after the rebase:

I confirmed that `listActiveThreads` must include a thread after the provider accepts `turn/start` and before `turn/started` arrives.

`reconcileDaemonReportedThreads` uses each reported id with the saved thread status. It does not inspect provider event rows.

- An `active` row stays active and avoids a false restart interruption.
- A `starting` or `idle` row receives `run.started`. This is correct because the provider accepted the turn.
- A `stopping` or deleted row still requests a stop.

A different daemon instance cannot retain this in-memory turn state. The new report only changes the same-instance reconnect path that preserves the runtime.

The branch now uses protocol version 85, after version 84 from PR #1197.

PR #1198, opened 2026-08-08
Move thread list reads off the event loop

### LINKED ISSUE #1131: Prevent synchronous SQLite reads and retention sweeps from blocking the server
## Current umbrella scope

This is the parent tracker for synchronous SQLite work blocking the server event loop across heavy reads and retention sweeps. The original 11-second thread-list cause was addressed by #1204; later reports identify residual reads, timeline stalls, and periodic sweep costs. Do not treat the original measurement as proof that its exact query is still unfixed.

- [ ] #1207 — Residual thread-list/read isolation and shared machinery for other heavy reads, including the timeline path described in follow-up comments.
- [ ] #3196 — Investigate truncation-sweep stalls, bounded/yielding work, and accurate blocking-work attribution on the reported 0.42.1 path.
- [ ] Preserve query-level efficiency as well as event-loop isolation; use measured offending operations to choose the fix.
- [ ] Validate cold/warm large-history reads and retention activity while unrelated requests remain responsive, recording query, decoding, projection, and sweep costs separately.
- [ ] Review retention threshold and batch throughput against history generation rates; document any resulting configuration consistently across SDK/CLI and UI where applicable.

#3196's truncation mechanism is a suspected cause, not a newly confirmed reproduction. Child issues retain their independent evidence and acceptance criteria. Remaining umbrella retention questions, including automation-run retention, are not completed merely by closing a read-isolation child.

## Historical report

The following measurements and environment describe the original 0.35.1 incident; follow-up evidence is in the comments and linked children.

## Summary

On a self-hosted instance with a grown `bb.db` (1.7GB), a single cold thread-list query stalled the server event loop for **11.2 seconds** — during which even static asset requests hung (measured 7.3s TTFB on `/`). On mobile Safari over Tailscale this presents as the whole app freezing for seconds at a time; the client is fine (boot payload is well-split and cached — nice work on #1071).

## Measurements

- `Event loop stalled {"maxDelayMs":11232.3, ...}` logged at the exact moment of a first (cold-cache) `bb thread list --json` after a restart; subsequent identical calls ~0.5s server-side.
- While stalled, `curl` TTFB on `/` (static index.html) went from 2ms → 7.3s. One synchronous query freezes every client and every route.
- We had previously logged stalls up to 22s under memory pressure (leak to 4.2G before MemoryHigh, since capped).

## Why the DB got big (contributing factor)

better-sqlite3 is synchronous on the serving loop, so stall duration scales with DB size/page-cache misses. Ours grew because the completed-event output truncation sweep effectively never fires:

- `events` = 1,447MB of 1.7GB; `item/completed` payloads = 913MB, 815MB of it `commandExecution.aggregatedOutput`.
- `COMPLETED_EVENT_OUTPUT_TRUNCATION_THRESHOLD_CHARS` is 32KB, but the real-world median payload is ~5-10KB — almost nothing qualifies, so the sweep reclaims ~0.
- `DEFAULT_COMPLETED_EVENT_OUTPUT_TRUNCATION_BATCH_SIZE` of 250/hour is below the generation rate of a busy multi-agent instance (~240 qualifying rows/hour for us) — even a threshold fix can never work down a backlog.

We're running both tuned on our fork (4KB threshold, 1000 batch): https://github.com/amadad/bb/commits/release/atum-pi-model-resolve — happy to PR either.

## Suggestions (in rough order of value)

1. Move heavy read queries (thread list, timeline) off the event loop — worker thread pool for better-sqlite3, or async driver for the hot read paths. This is the structural fix; retention only shrinks the constant.
2. Lower the truncation threshold default (32KB → 4-8KB) and/or make it configurable; size the batch relative to generation rate.
3. Consider retention for `automation_runs` (currently unbounded, only cascade-deleted).

## Environment

bb-app 0.35.1 self-hosted (Hetzner 4 vCPU/8GB, Node 24), bb.db 1.7GB / ~595K events, heavy automation usage (~1,000+ bb-originated provider sessions/month).

> AGENT GENERATED

SawyerHood:
## Reproduction

A cold thread-list query on a 2.0 GB test database took 1.150 seconds. A concurrent health request took 1.098 seconds, and the server logged a 1.166-second event-loop stall.

## Root cause

The thread-list handler ran synchronous better-sqlite3 reads on the server event loop. SQLite used temporary B-trees for the list query, so every client waited for the read to finish.

## Fix

Move the complete thread-list read phase to a worker thread with a read-only database connection. Each worker operation uses one deferred read transaction.

The worker derives database-backed disconnected state. After the worker returns, the main thread applies current live daemon session state from the notification hub. This prevents stale host status during a slow read.

The same cold query now leaves the health route responsive at 1.2 ms. Slow-query logs return to the server logger, worker messages use validated contracts, and shutdown terminates the worker.

Fixes #1131

### DIFF
--- a/apps/server/src/routes/projects.ts
+++ b/apps/server/src/routes/projects.ts
@@ -194,21 +194,25 @@ function parseProjectListIncludes(
 async function buildProjectsWithThreadsResponse(
   deps: AppDeps,
   options: ProjectListOptions,
+  signal: AbortSignal,
 ): Promise<ProjectWithThreadsResponse[]> {
   return buildProjectsWithThreadsResponseFromRows(
     deps,
     listDiscoverableProjects(deps, options),
+    signal,
   );
 }
 
 async function buildProjectsWithThreadsResponseFromRows(
   deps: AppDeps,
   projectRows: ProjectResponseRow[],
+  signal: AbortSignal,
 ): Promise<ProjectWithThreadsResponse[]> {
   const projects = buildProjectResponsesFromRows(deps, projectRows);
   const projectIds = projects.map((project) => project.id);
   const threadResponses = await deps.databaseReads.listThreadEntriesForProjects(
     { archived: false, projectIds },
+    { signal },
   );
   const threadsByProjectId = new Map<
     string,
@@ -236,7 +240,10 @@ async function buildProjectsWithThreadsResponseFromRows(
   }));
 }
 
-async function buildSidebarBootstrapResponse(deps: AppDeps) {
+async function buildSidebarBootstrapResponse(
+  deps: AppDeps,
+  signal: AbortSignal,
+) {
   const personalProject = getPersonalProject(deps.db);
   if (!personalProject) {
     throw new ApiError(
@@ -246,8 +253,12 @@ async function buildSidebarBootstrapResponse(deps: AppDeps) {
     );
   }
   const [personalProjectResponses, projects] = await Promise.all([
-    buildProjectsWithThreadsResponseFromRows(deps, [personalProject]),
-    buildProjectsWithThreadsResponseFromRows(deps, listPublicProjects(deps.db)),
+    buildProjectsWithThreadsResponseFromRows(deps, [personalProject], signal),
+    buildProjectsWithThreadsResponseFromRows(
+      deps,
+      listPublicProjects(deps.db),
+      signal,
+    ),
   ]);
   const personalProjectResponse = personalProjectResponses[0];
   if (!personalProjectResponse) {
@@ -328,7 +339,11 @@ export function registerProjectRoutes(app: Hono, deps: AppDeps): void {
     };
     if (includes.has("threads")) {
       return context.json(
-        await buildProjectsWithThreadsResponse(deps, options),
+        await buildProjectsWithThreadsResponse(
+          deps,
+          options,
+          context.req.raw.signal,
+        ),
       );
     }
     return context.json(
@@ -340,7 +355,9 @@ export function registerProjectRoutes(app: Hono, deps: AppDeps): void {
   });
 
   get(routes.sidebarBootstrap, async (context) =>
-    context.json(await buildSidebarBootstrapResponse(deps)),
+    context.json(
+      await buildSidebarBootstrapResponse(deps, context.req.raw.signal),
+    ),
   );
 
   post(routes.create, async (context, payload) => {
--- a/apps/server/src/routes/projects.ts
+++ b/apps/server/src/routes/projects.ts
@@ -7,10 +7,8 @@ import {
   deleteProjectSource,
   getProjectSourceByHost,
   getProjectSourceForProject,
-  listProjectExecutionDefaultsByProjectIds,
   listPublicProjects,
   listProjectSourcesByProjectIds,
-  listThreadSections,
   reorderProject,
   updateProject,
   updateProjectSource,
@@ -44,7 +42,6 @@ import {
   requirePublicStandardProject,
 } from "../services/lib/entity-lookup.js";
 import { PROMPT_HISTORY_ENTRY_LIMIT } from "@bb/domain";
-import { resolveCreateThreadExecutionDefaults } from "../services/threads/thread-default-policy.js";
 import { resolveProjectCreateDefaultExecutionPlan } from "../services/threads/thread-execution-plan.js";
 import { callHostRetryableOnlineRpc } from "../services/hosts/online-rpc.js";
 import { runLiveHostCommand } from "../services/hosts/live-command.js";
@@ -196,80 +193,17 @@ async function buildProjectsWithThreadsResponse(
   options: ProjectListOptions,
   signal: AbortSignal,
 ): Promise<ProjectWithThreadsResponse[]> {
-  return buildProjectsWithThreadsResponseFromRows(
-    deps,
-    listDiscoverableProjects(deps, options),
-    signal,
-  );
-}
-
-async function buildProjectsWithThreadsResponseFromRows(
-  deps: AppDeps,
-  projectRows: ProjectResponseRow[],
-  signal: AbortSignal,
-): Promise<ProjectWithThreadsResponse[]> {
-  const projects = buildProjectResponsesFromRows(deps, projectRows);
-  const projectIds = projects.map((project) => project.id);
-  const threadResponses = await deps.databaseReads.listThreadEntriesForProjects(
-    { archived: false, projectIds },
+  return deps.databaseReads.listProjectsWithThreads(
+    { includePersonal: options.includePersonal },
     { signal },
   );
-  const threadsByProjectId = new Map<
-    string,
-    ProjectWithThreadsResponse["threads"]
-  >();
-  for (const thread of threadResponses) {
-    const projectThreads = threadsByProjectId.get(thread.projectId);
-    if (projectThreads) {
-      projectThreads.push(thread);
-      continue;
-    }
-    threadsByProjectId.set(thread.projectId, [thread]);
-  }
-  const defaultsByProjectId = listProjectExecutionDefaultsByProjectIds(
-    deps.db,
-    { projectIds },
-  );
-
-  return projects.map((project) => ({
-    ...project,
-    threads: threadsByProjectId.get(project.id) ?? [],
-    defaultExecutionOptions: resolveCreateThreadExecutionDefaults({
-      storedDefaults: defaultsByProjectId.get(project.id) ?? null,
-    }).executionDefaults,
-  }));
 }
 
 async function buildSidebarBootstrapResponse(
   deps: AppDeps,
   signal: AbortSignal,
 ) {
-  const personalProject = getPersonalProject(deps.db);
-  if (!personalProject) {
-    throw new ApiError(
-      500,
-      "internal_error",
-      "Personal project is not initialized",
-    );
-  }
-  const [personalProjectResponse, ...projects] =
-    await buildProjectsWithThreadsResponseFromRows(
-      deps,
-      [personalProject, ...listPublicProjects(deps.db)],
-      signal,
-    );
-  if (!personalProjectResponse) {
-    throw new ApiError(
-      500,
-      "internal_error",
-      "Personal project response was not built",
-    );
-  }
-  return {
-    sections: listThreadSections(deps.db),
-    projects,
-    personalProject: personalProjectResponse,
-  };
+  return deps.databaseReads.getSidebarBootstrap({ signal });
 }
 
 interface RequireProjectSourceArgs {
--- a/apps/server/src/routes/projects.ts
+++ b/apps/server/src/routes/projects.ts
@@ -11,7 +11,6 @@ import {
   listPublicProjects,
   listProjectSourcesByProjectIds,
   listThreadSections,
-  listThreadsWithPendingInteractionStateForProjects,
   reorderProject,
   updateProject,
   updateProjectSource,
@@ -47,7 +46,6 @@ import {
 import { PROMPT_HISTORY_ENTRY_LIMIT } from "@bb/domain";
 import { resolveCreateThreadExecutionDefaults } from "../services/threads/thread-default-policy.js";
 import { resolveProjectCreateDefaultExecutionPlan } from "../services/threads/thread-execution-plan.js";
-import { toThreadListEntryResponses } from "../services/threads/thread-runtime-display.js";
 import { callHostRetryableOnlineRpc } from "../services/hosts/online-rpc.js";
 import { runLiveHostCommand } from "../services/hosts/live-command.js";
 import {
@@ -193,29 +191,25 @@ function parseProjectListIncludes(
   return includes;
 }
 
-function buildProjectsWithThreadsResponse(
+async function buildProjectsWithThreadsResponse(
   deps: AppDeps,
   options: ProjectListOptions,
-): ProjectWithThreadsResponse[] {
+): Promise<ProjectWithThreadsResponse[]> {
   return buildProjectsWithThreadsResponseFromRows(
     deps,
     listDiscoverableProjects(deps, options),
   );
 }
 
-function buildProjectsWithThreadsResponseFromRows(
+async function buildProjectsWithThreadsResponseFromRows(
   deps: AppDeps,
   projectRows: ProjectResponseRow[],
-): ProjectWithThreadsResponse[] {
+): Promise<ProjectWithThreadsResponse[]> {
   const projects = buildProjectResponsesFromRows(deps, projectRows);
   const projectIds = projects.map((project) => project.id);
-  const threadRows = listThreadsWithPendingInteractionStateForProjects(
-    deps.db,
+  const threadResponses = await deps.databaseReads.listThreadEntriesForProjects(
     { archived: false, projectIds },
   );
-  const threadResponses = toThreadListEntryResponses(deps, {
-    threads: threadRows,
-  });
   const threadsByProjectId = new Map<
     string,
     ProjectWithThreadsResponse["threads"]
@@ -242,7 +236,7 @@ function buildProjectsWithThreadsResponseFromRows(
   }));
 }
 
-function buildSidebarBootstrapResponse(deps: AppDeps) {
+async function buildSidebarBootstrapResponse(deps: AppDeps) {
   const personalProject = getPersonalProject(deps.db);
   if (!personalProject) {
     throw new ApiError(
@@ -251,10 +245,11 @@ function buildSidebarBootstrapResponse(deps: AppDeps) {
       "Personal project is not initialized",
     );
   }
-  const personalProjectResponse = buildProjectsWithThreadsResponseFromRows(
-    deps,
-    [personalProject],
-  )[0];
+  const [personalProjectResponses, projects] = await Promise.all([
+    buildProjectsWithThreadsResponseFromRows(deps, [personalProject]),
+    buildProjectsWithThreadsResponseFromRows(deps, listPublicProjects(deps.db)),
+  ]);
+  const personalProjectResponse = personalProjectResponses[0];
   if (!personalProjectResponse) {
     throw new ApiError(
       500,
@@ -264,10 +259,7 @@ function buildSidebarBootstrapResponse(deps: AppDeps) {
   }
   return {
     sections: listThreadSections(deps.db),
-    projects: buildProjectsWithThreadsResponseFromRows(
-      deps,
-      listPublicProjects(deps.db),
-    ),
+    projects,
     personalProject: personalProjectResponse,
   };
 }
@@ -327,7 +319,7 @@ export function registerProjectRoutes(app: Hono, deps: AppDeps): void {
   });
   const routes = publicApiRoutes.projects;
 
-  get(routes.list, (context, query) => {
+  get(routes.list, async (context, query) => {
     const includes = parseProjectListIncludes(query);
     // Compatibility is resolved once at the HTTP boundary: ordinary projects
     // remain the default, and all internal list paths receive an explicit flag.
@@ -335,7 +327,9 @@ export function registerProjectRoutes(app: Hono, deps: AppDeps): void {
       includePersonal: query.includePersonal === "true",
     };
     if (includes.has("threads")) {
-      return context.json(buildProjectsWithThreadsResponse(deps, options));
+      return context.json(
+        await buildProjectsWithThreadsResponse(deps, options),
+      );
     }
     return context.json(
       buildProjectResponsesFromRows(
@@ -345,8 +339,8 @@ export function registerProjectRoutes(app: Hono, deps: AppDeps): void {
     );
   });
 
-  get(routes.sidebarBootstrap, (context) =>
-    context.json(buildSidebarBootstrapResponse(deps)),
+  get(routes.sidebarBootstrap, async (context) =>
+    context.json(await buildSidebarBootstrapResponse(deps)),
   );
 
   post(routes.create, async (context, payload) => {
--- a/apps/server/src/routes/projects.ts
+++ b/apps/server/src/routes/projects.ts
@@ -252,15 +252,12 @@ async function buildSidebarBootstrapResponse(
       "Personal project is not initialized",
     );
   }
-  const [personalProjectResponses, projects] = await Promise.all([
-    buildProjectsWithThreadsResponseFromRows(deps, [personalProject], signal),
-    buildProjectsWithThreadsResponseFromRows(
+  const [personalProjectResponse, ...projects] =
+    await buildProjectsWithThreadsResponseFromRows(
       deps,
-      listPublicProjects(deps.db),
+      [personalProject, ...listPublicProjects(deps.db)],
       signal,
-    ),
-  ]);
-  const personalProjectResponse = personalProjectResponses[0];
+    );
   if (!personalProjectResponse) {
     throw new ApiError(
       500,
--- a/apps/server/src/routes/threads/base.ts
+++ b/apps/server/src/routes/threads/base.ts
@@ -4,7 +4,6 @@ import {
   countNonDeletedAssignedChildThreads,
   getEnvironment,
   getThreadSectionById,
-  listThreadsWithPendingInteractionState,
   markThreadDeleted,
   searchThreadsWithPendingInteractionState,
   updateThread,
@@ -195,7 +194,7 @@ export function registerThreadBaseRoutes(app: Hono, deps: AppDeps): void {
   });
   const routes = publicApiRoutes.threads;
 
-  get(routes.list, (context, query) => {
+  get(routes.list, async (context, query) => {
     const limit = parseOptionalInteger(query.limit, "limit");
     if (limit !== undefined && limit <= 0) {
       throw new ApiError(400, "invalid_request", "limit must be positive");
@@ -217,7 +216,7 @@ export function registerThreadBaseRoutes(app: Hono, deps: AppDeps): void {
     if (query.sectionId) {
       requireThreadSection(deps, query.sectionId);
     }
-    const threads = listThreadsWithPendingInteractionState(deps.db, {
+    const threadEntries = await deps.databaseReads.listThreadEntries({
       ...(query.projectId ? { projectId: query.projectId } : {}),
       ...(query.parentThreadId ? { parentThreadId: query.parentThreadId } : {}),
       ...(query.sourceThreadId ? { sourceThreadId: query.sourceThreadId } : {}),
@@ -234,9 +233,7 @@ export function registerThreadBaseRoutes(app: Hono, deps: AppDeps): void {
       ...(limit !== undefined ? { limit } : {}),
       ...(offset !== undefined ? { offset } : {}),
     });
-    return context.json(
-      toThreadListEntryResponses(deps, { threads }) satisfies ThreadListEntry[],
-    );
+    return context.json(threadEntries satisfies ThreadListEntry[]);
   });
 
   get(routes.search, (context, query) => {
--- a/apps/server/src/routes/threads/base.ts
+++ b/apps/server/src/routes/threads/base.ts
@@ -216,23 +216,34 @@ export function registerThreadBaseRoutes(app: Hono, deps: AppDeps): void {
     if (query.sectionId) {
       requireThreadSection(deps, query.sectionId);
     }
-    const threadEntries = await deps.databaseReads.listThreadEntries({
-      ...(query.projectId ? { projectId: query.projectId } : {}),
-      ...(query.parentThreadId ? { parentThreadId: query.parentThreadId } : {}),
-      ...(query.sourceThreadId ? { sourceThreadId: query.sourceThreadId } : {}),
-      ...(query.sectionId ? { sectionId: query.sectionId } : {}),
-      ...(query.unsectioned === "true" ? { unsectioned: true } : {}),
-      ...(query.originKind ? { originKind: query.originKind } : {}),
-      ...(query.originPluginId ? { originPluginId: query.originPluginId } : {}),
-      ...(query.childOrigin ? { childOrigin: query.childOrigin } : {}),
-      includeHidden: query.includeHidden === "true",
-      archived:
-        query.archived === undefined ? undefined : query.archived === "true",
-      hasParent:
-        query.hasParent === undefined ? undefined : query.hasParent === "true",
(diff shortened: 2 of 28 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am now reviewing this change for security, code quality, performance, architecture, and end-to-end behavior.

apps/server/src/services/database/database-read-service.ts
```diff
@@ -0,0 +1,160 @@
...
+}
+
+export function createWorkerDatabaseReadService(
+  args: CreateWorkerDatabaseReadServiceArgs,
+): DatabaseReadService {
+  const workerData: DatabaseReadWorkerData = {
+    databasePath: args.databasePath,
+  };
+  const sourceFile = import.meta.url.endsWith(".ts");
+  const worker = new Worker(resolveWorkerEntryUrl(), {
```
SawyerHood: 🚨 `slopcop/review` — The server does not verify worker readiness.

The factory returns as soon as it starts the worker. It does not wait until the worker opens the database. An initialization error stores a permanent failure, while the server can still report readiness. All later list requests then fail until restart. Add a ready message before listener startup. Restart the worker or stop the server after an unexpected exit. Test both paths.

apps/server/src/start-server.ts
```diff
@@ -51,6 +52,11 @@ export async function runServer(serverConfig: ServerConfig): Promise<void> {
     logger,
   });
   const hub = new NotificationHub();
+  const databaseReads = createWorkerDatabaseReadService({
```
SawyerHood: 🚨 `slopcop/review` — The worker can keep a failed startup process alive.

`runServer` creates the worker here, before later operations that can fail. The top-level error handler only sets `process.exitCode`. A process test confirmed that the worker then keeps Node active. Close the worker on every startup error. Add a child-process test that requires exit code 1 quickly.

apps/server/src/routes/threads/base.ts
```diff
@@ -217,7 +216,7 @@ export function registerThreadBaseRoutes(app: Hono, deps: AppDeps): void {
     if (query.sectionId) {
       requireThreadSection(deps, query.sectionId);
     }
-    const threads = listThreadsWithPendingInteractionState(deps.db, {
+    const threadEntries = await deps.databaseReads.listThreadEntries({
```
SawyerHood: 🚨 `slopcop/review` — The main app still runs thread-list reads on the server event loop.

This call only moves `GET /threads` to the worker. The app loads `/sidebar-bootstrap`, not this route, for its normal sidebar. That route still calls `listThreadsWithPendingInteractionStateForProjects` and `toThreadListEntryResponses` on the server event loop. Move the sidebar and `projects?include=threads` read paths into the same worker service. Add route tests that check a health request during each slow read.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

This change gives a slow database job to a separate worker. The main server can then answer other requests. The idea works for `GET /threads`, but the normal app uses a different route. Startup and worker failures can also leave the server in a bad state.

I recommend fixes before merge. I posted a comment-only review, as required.

### Findings

- **P1 — The normal app still uses the server event loop for thread-list reads.** The app loads `/sidebar-bootstrap`. That route and `projects?include=threads` still run the thread queries and display queries on the server event loop. Move these complete read paths into the worker service.

- **P1 — A later startup error can leave the process alive.** `runServer` creates the worker before later operations that can fail. The top-level error handler only sets `process.exitCode`. Two reviews reproduced a process that did not exit. Close the worker on every startup error.

- **P1 — A worker fault can permanently break thread lists.** The service does not wait for a ready message. It stores worker errors and rejects all later list requests. The server can still report that it is ready. Require worker readiness before listener startup. Restart the worker or stop the server after an unexpected exit.

### Architecture and security

I found no reusable server worker pool in this repository. Extend this database service for the sidebar and project read paths. The browser worker pool does not fit this server use.

The worker validates its messages and opens the database with read-only and query-only controls. I found no direct path-trust or data-exposure fault.

### Validation

- Three GPT-5.6 workers reviewed security, code quality, architecture, and performance. One GPT-5.6 gate checked the combined result.
- Turbo type checks passed for `@bb/server` and `@bb/db`.
- Turbo tests passed for both packages. The server suite passed 1,368 tests in 159 files.
- The production server build passed and included the worker bundle.
- The source app started at the pull request SHA. A browser call to `GET /api/v1/threads` returned status 200.
- The browser received one seeded thread entry. The normal app displayed the same thread.
- `git diff --check origin/main...HEAD` passed. The worktree stayed clean.

apps/server/src/services/database/database-read-service.ts
SawyerHood: Fixed in 3f24fc89a. The service now waits for the worker ready message. It also starts a replacement after an unexpected exit. Tests cover initialization failure and restart.

apps/server/src/start-server.ts
SawyerHood: Fixed in 3f24fc89a. Startup now closes the database read service when later startup work fails. A child-process test confirms a prompt exit with code 1.

apps/server/src/routes/threads/base.ts
SawyerHood: Fixed in 3f24fc89a. The sidebar and projects-with-threads routes now use the worker service. Both route tests confirm that health requests remain available during a delayed read.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am reviewing this pull request for security, code quality, performance, architecture, and product behavior.

apps/server/test/app/startup-diagnostics.test.ts
```diff
@@ -36,6 +68,64 @@ describe("server startup diagnostics", () => {
...
+          cwd: resolve(testDir, "../../../.."),
+          env: {
+            ...process.env,
+            BB_STARTUP_FAILURE_TEST_DATABASE_PATH: databasePath,
+          },
+          stdio: "ignore",
+        },
+      );
+
+      await expect(waitForChildExit(child, 10_000)).resolves.toBe(1);
```
SawyerHood: 🚨 `slopcop/review` — The outer test can expire before this ten-second wait.

This test uses the default five-second Vitest timeout. The helper can wait ten seconds before it kills the child and reports the useful error. The full exact-SHA server suite failed here locally after five seconds. The focused test passed. Set the test timeout above ten seconds, or reduce this helper timeout below five seconds.

apps/server/src/services/database/database-read-service.ts
```diff
@@ -0,0 +1,279 @@
...
+  ): Promise<ThreadListEntry[]> {
+    const state = await requireReadyWorker();
+    const id = nextRequestId;
+    nextRequestId += 1;
+    return new Promise((resolve, reject) => {
+      if (state.failed) {
+        reject(new Error("The database read worker stopped"));
+        return;
+      }
+      pending.set(id, { reject, resolve });
```
SawyerHood: 🚨 `slopcop/review` — The single worker has no load control.

Every request enters the `pending` map and the worker message queue. The service has no queue limit, deadline, or HTTP abort support. A client can submit slow list requests faster than the worker completes them. This can exhaust memory and delay current reads. Add a bounded queue, a deadline, and abort support. Test overload rejection and client disconnects.

apps/server/src/services/database/database-read-service.ts
```diff
@@ -0,0 +1,279 @@
...
+      rejectReady,
+      resolveReady,
+      wasReady: false,
+      worker,
+    };
+    workers.add(worker);
+    args.onWorkerCreated?.(worker);
+
+    function fail(error: Error): void {
+      if (state.failed || closed) {
```
SawyerHood, line 385: 🚨 `slopcop/review` — A shutdown can leave a read promise pending.

`close()` sets `closed` before it terminates the worker. The exit event calls `fail()`, but this return skips `rejectReady()`. A request that waits for a replacement worker never resolves or rejects. I reproduced this at the required SHA. The request stayed pending after `close()` completed. Reject each unresolved ready promise during close. Add a test that closes during replacement startup.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

The server used to stop and read a large list by itself. This change gives that job to a helper. It is like asking a friend to count blocks while you answer the door. The idea works, but the helper can get stuck and accept too much work.

### Findings

- High: A shutdown can leave a request pending forever during replacement worker startup. I reproduced this race at the required SHA. Reject every unresolved readiness promise during `close()`. Add a test for this shutdown point.
- Medium: The single worker queue has no limit, deadline, or client abort support. Slow repeated requests can consume memory and delay current reads. Add load control and tests for overload and disconnects.
- Low: The startup test allows a ten-second child wait inside a five-second Vitest test. The full local suite failed at five seconds. Give the test a longer timeout, or reduce the child wait.

### Architecture

The worker boundary otherwise looks sound. It uses a read-only connection, strict messages, and a fixed worker entry path. I found no SQL injection or access-control regression.

The direct service and worker entry duplicate operation dispatch. The contract also repeats the daemon-session schema. A shared operation definition would reduce future drift.

Thread search still uses synchronous database work. I did not count that existing path as a defect in this pull request.

### Validation

- The exact-SHA server and database type checks passed.
- The full server and database tests completed 1,372 of 1,373 tests. The new startup test hit its five-second timeout.
- The focused startup test passed. GitHub checks for the required SHA also passed during this review.
- Browser checks returned HTTP 200 for health, thread list, and sidebar routes.
- With 250,000 temporary rows, health returned in 2 ms while a 278 ms list request still ran.
- A 10,000-row response kept the maximum observed health latency near 30 ms.

The main event-loop goal works. The shutdown race needs a fix before merge. I left three inline comments with specific changes.

apps/server/src/services/database/database-read-service.ts
SawyerHood: Fixed in fb250b719. The parent now posts one worker request at a time, permits 32 pending reads, and applies a 30-second deadline. Routes pass request abort signals. Tests cover overload, abort removal, deadlines, and retryable 503 responses.

apps/server/test/app/startup-diagnostics.test.ts
SawyerHood: Fixed after the rebase in 8f084cbb2. The child now reports worker readiness first. The test then requires exit within five seconds and allows 35 seconds for loaded CI startup.

apps/server/src/services/database/database-read-service.ts
SawyerHood: Fixed in fb250b719. Close now rejects the active worker readiness promise before termination. A test closes during replacement startup and confirms that the waiting read rejects.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am now reviewing this pull request under the configured security, code quality, performance, architecture, and end-to-end guidelines.

apps/server/src/services/database/database-read-service.ts
```diff
@@ -0,0 +1,480 @@
...
+      }
+      activeWorker = null;
+      if (state.wasReady) {
+        activeWorker = startWorker();
+        void activeWorker.ready.catch(() => {});
+      }
+    }
+
+    worker.on("message", (value: unknown) => {
+      const parsedMessage = databaseReadWorkerMessageSchema.safeParse(value);
```
SawyerHood, line 414: 🚨 `slopcop/review` — Full result validation still stops the server event loop.

`safeParse` walks every `ThreadListEntry` after the worker returns. The worker transfer already copies the whole array. A local test measured about 90 ms to parse 100,000 rows. The transfer decode took about 225 ms. This creates another long pause on the target path.

Please validate the full payload inside the worker. Check only the small message envelope here.

apps/server/src/services/database/database-read-worker-entry.ts
```diff
@@ -0,0 +1,90 @@
...
+      { threads },
+    );
+    port.postMessage({
+      entries,
+      id: request.id,
+      kind: "result",
+    } satisfies DatabaseReadWorkerMessage);
+  } catch (error) {
+    const requestId = request?.id;
+    if (requestId === undefined) {
```
SawyerHood: 🚨 `slopcop/review` — A malformed list request can stop the shared worker.

The HTTP route accepts integers larger than `Number.MAX_SAFE_INTEGER`. This schema rejects them before `request` gets a value. The catch then throws, so the worker exits. `fail()` rejects every active and queued read. An unauthenticated caller can repeat this request and break thread lists for all clients.

Please validate bounds at the HTTP boundary. Also parse the request ID separately and return one request error. Keep the worker active.

apps/server/src/services/database/database-read-service.ts
```diff
@@ -0,0 +1,480 @@
...
+    const id = nextRequestId;
+    nextRequestId += 1;
+    return new Promise((resolve, reject) => {
+      const read: PendingDatabaseRead = {
+        createRequest,
+        id,
+        reject,
+        resolve,
+        settled: false,
+        timeout: setTimeout(() => {
```
SawyerHood, line 608: 🚨 `slopcop/review` — An active timeout does not release or replace the worker.

This handler settles an active read but leaves `activeRead` set. `scheduleDispatch()` then exits, and the settled read still uses one queue slot. The timer also starts when the request enters the queue. If a query exceeds 30 seconds, later reads wait or time out until that query returns.

Please replace the worker after an active timeout. Use a separate queue wait limit if needed.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

The app gives a slow list job to a helper. It works like asking a friend to count toys while you keep playing.

### Review

The main idea works. The database query now runs outside the server event loop. I found these issues:

1. **High:** A large `limit` value fails worker validation before the request gets an ID. The worker exits and rejects unrelated reads.
2. **High:** An active timeout settles the promise but leaves `activeRead` set. Later reads wait or time out until the query returns.
3. **Medium:** The main thread validates every returned entry with Zod. Tests measured about 315 ms for transfer decode and validation of 100,000 rows.
4. **Medium:** The service marks queue and timeout errors as retryable 503 responses. The app retry policy rejects all `HttpError` values.
5. **Medium:** A client abort creates a plain `Error`. The server then records an error log and maps the error to HTTP 500.

The read-only SQLite connection has the correct write controls. Existing project and thread access checks still run before worker reads.

The architecture scan found no reusable server worker pool, so this service does not duplicate an existing worker layer. However, the Zod schema repeats `ListThreadsOptions`. Add a key coverage guard to prevent silent contract drift.

The route and integration harnesses use the direct service. Thus, route tests do not test the production worker boundary. Add one route test with the worker and a full thread entry.

Validation passed:

- Server typecheck
- Server production build
- 14 focused server tests
- 5 database tests
- Local UI load with the real worker
- HTTP 200 from `/api/v1/sidebar-bootstrap` and `/api/v1/threads?limit=10`
- No database worker errors in the local server log

I left this as a comment review. I did not approve the pull request or request changes.

apps/server/src/services/database/database-read-service.ts
SawyerHood: Fixed in 3bdc475e3. An active timeout now clears the active read, starts a replacement worker, and terminates the slow worker. Queued reads can continue on the replacement. A regression test verifies that the next read succeeds.

SawyerHood:
Review follow-up for 3bdc475e3:

- Moved full result validation into the worker.
- Added safe HTTP bounds and request-local worker validation errors.
- Replaced the worker after an active read timeout.
- Added compile-time option-key coverage and a production-worker route test.
- Kept the 503 retry behavior unchanged. The maintainer deferred that product decision.
- The abort and response-byte findings were fixed in 14573f087.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

apps/server/src/services/database/database-read-service.ts
```diff
@@ -0,0 +1,499 @@
...
+          const timedOutWhileActive = activeRead === read;
+          removeQueuedRead(read);
+          if (timedOutWhileActive) {
+            activeRead = null;
+          }
+          const error = new DatabaseReadUnavailableError(
```
SawyerHood: 🚨 `slopcop/review` — Medium: One slow read can cause five 30-second attempts.

This timeout uses retryable `DatabaseReadUnavailableError`. The global query policy permits five attempts.

A repeatable slow query can delay the final error for about 150 seconds. Each attempt starts another worker.

Please use a non-retryable timeout error or a smaller timeout retry limit.

apps/server/src/routes/projects.ts
```diff
@@ -251,10 +252,15 @@ function buildSidebarBootstrapResponse(deps: AppDeps) {
       "Personal project is not initialized",
     );
   }
-  const personalProjectResponse = buildProjectsWithThreadsResponseFromRows(
-    deps,
-    [personalProject],
-  )[0];
+  const [personalProjectResponses, projects] = await Promise.all([
```
SawyerHood: 🚨 `slopcop/review` — Medium: Sidebar bootstrap splits one logical read into two serial worker requests.

Both calls share one worker, but `Promise.all()` submits them separately. Near the queue limit, one call can remain queued after its sibling fails.

That work uses capacity and can delay a client retry.

Please send one request with all project identifiers. Then split the results by project type.

apps/server/src/services/database/database-read-service.ts
```diff
@@ -0,0 +1,499 @@
...
+      if (state.failed || closed) {
+        return;
+      }
+      state.failed = true;
+      args.logger.warn(
+        { err: error, workerWasReady: state.wasReady },
+        "Database read worker failed",
+      );
+      state.rejectReady(error);
+      rejectAllReads(error);
```
SawyerHood: 🚨 `slopcop/review` — High: A worker failure returns a non-retryable error.

`fail()` rejects all reads with a plain `Error`, although it starts a replacement worker. The server then returns a non-retryable 500 response.

The new retry policy cannot use the ready replacement worker.

Please reject affected reads with `DatabaseReadUnavailableError`. Add a test for a pending read during an unexpected worker exit.

apps/server/src/services/database/database-read-service.ts
```diff
@@ -0,0 +1,499 @@
...
+            activeWorker = startWorker();
+            void activeWorker.ready.catch(() => {});
+            void timedOutWorker.worker.terminate();
+          }
+          scheduleDispatch();
+        }, requestTimeoutMs),
+      };
+      if (context?.signal !== undefined) {
+        read.signal = context.signal;
+        read.abortHandler = (): void => {
```
SawyerHood, line 624: 🚨 `slopcop/review` — High: An active abort removes the worker deadline.

When this signal stops an active request, `rejectRead()` clears the timer. It does not clear `activeRead` or stop the worker.

Later request timeouts see only queued work. They cannot replace the blocked worker.

Please replace the worker when an active request stops. Add a test that stops an active slow read.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

This change gives slow list work to a helper. It is like asking a helper to sort cards. The main app can keep moving.

Two stop rules can fail. The helper can then block new work or return an error that the app will not retry.

### Findings

I found four issues:

- High: An active request abort clears its timer but leaves the worker busy.
- High: An unexpected worker failure returns a non-retryable error, although the service starts a replacement.
- Medium: A slow read can trigger five 30-second attempts and five worker replacements.
- Medium: Sidebar bootstrap sends two related reads into one serial queue.

The inline comments contain the impact and a concrete fix for each issue.

The sidebar should send one request for all project identifiers. This refactor removes duplicate queue work and gives one result boundary.

I found no authorization, data exposure, injection, or unsafe path issue. The active-abort issue can cause an availability failure.

### Validation

- I reviewed all 27 changed files at the stated commit.
- `git diff --check` passed.
- Turbo type checks passed for the app, server, SDK, and database packages.
- The focused worker test file passed all 12 tests.
- The local full run passed 1,383 of 1,385 tests.
- The relevant timeout test passed alone. The other local failure came from an unrelated plugin test.
- All GitHub checks passed.
- Browser QA created an isolated project and thread.
- The sidebar showed both records.
- The thread page opened successfully.
- All observed thread and list routes returned status 200.

I would fix the two high findings before merge. I left a comment review only. I did not approve or request changes.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** This change gives slow thread-list database work to a background worker. The main server can then answer other requests.

I reviewed `a0af660ae9`. The pull request advanced to `938458ad24` during the review. That later commit only changed test synchronization.

I found three medium-severity issues:

1. The project routes read projects before the worker await and defaults after it. A concurrent deletion can produce an impossible sidebar response.
2. Each worker operation runs several SQLite reads without a transaction. A write can make one response combine data from different WAL snapshots.
3. Each active client cancellation replaces the worker. Repeated request cancellations can cause worker churn and deny thread-list service.

The first two issues need one consistent read transaction. The project and sidebar operations must move all response reads into that transaction.

For cancellation, reject the caller but keep the worker request active. Discard its result, and replace a worker only after an internal timeout.

The reviewed SHA also had an intermittent server test failure. Commit `938458ad24` added a replacement-ready wait, and the new server CI run passed.

I found no reusable server worker framework. The dedicated service fits the current architecture.

The direct and worker adapters repeat the operation dispatch. A shared operation function can remove that duplication.

Validation passed locally:

- The full server test run passed with 1,388 tests.
- The server Turbo typecheck passed.
- The server Turbo build passed and included the worker artifact.
- The browser loaded the sidebar without console errors.
- The thread list, project list with threads, and sidebar bootstrap routes returned valid responses.
- `git diff --check` passed.

I left this as a comment review. I did not approve the pull request or request changes.

SawyerHood:
🚨 `slopcop/review` — Run the complete worker operation in one read transaction.

[Reviewed line](https://github.com/get-bb/bb/blob/a0af660ae91716506d9f5549da1e8c9d0de9384e/apps/server/src/services/database/database-read-worker-entry.ts#L58)

This query and `toThreadListEntryResponses` run several database reads. A main-thread write can commit between those reads.

The result can combine thread rows and activity rows from different WAL snapshots.

Use one deferred read transaction for each worker request. This transaction will pin one consistent SQLite snapshot.

SawyerHood:
🚨 `slopcop/review` — This await breaks the sidebar response's database snapshot.

[Reviewed line](https://github.com/get-bb/bb/blob/a0af660ae91716506d9f5549da1e8c9d0de9384e/apps/server/src/routes/projects.ts#L213)

The route builds projects and sources before this await. It reads defaults after the await.

I reproduced a concurrent deletion during this wait. The response included the deleted project without its former thread.

That response matched neither the state before deletion nor the state after deletion.

Move the complete sidebar read into one worker-side read transaction. Include projects, sources, sections, defaults, and threads in that transaction.

SawyerHood:
🚨 `slopcop/review` — Do not replace the worker when a client cancels an active read.

[Reviewed line](https://github.com/get-bb/bb/blob/a0af660ae91716506d9f5549da1e8c9d0de9384e/apps/server/src/services/database/database-read-service.ts#L463)

Each active cancellation starts a new worker and stops the current worker. The queue limit does not bound this restart path.

A client can repeat a list request and close its socket after dispatch. This can deny service to other list requests.

Reject the caller, but keep the active slot until the worker returns. Discard that result and reserve worker replacement for internal timeouts.

SawyerHood:
Addressed all three findings from SlopCop run `run_6vklc90omskvazsv` in ba8543a7b.

- Each worker operation now uses one deferred SQLite read transaction.
- The project and sidebar worker operations now read projects, sources, defaults, sections, and threads in that transaction.
- An active client cancellation now rejects the caller but lets the worker finish. Only an internal timeout replaces the worker.
- A concurrent-delete test confirms that a sidebar response keeps one database snapshot.
- A worker-count test confirms that an active cancellation does not replace the worker.
- A route test confirms that the worker and direct sidebar paths return identical bytes.

All app, database, SDK, and server typechecks pass. The app, database, and SDK suites pass. The server suite passes 1,389 tests and has only the known 0664-versus-0644 skill-file failure.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

apps/server/src/services/database/database-read-operations.ts
```diff
@@ -0,0 +1,213 @@
...
+): DatabaseReadWorkerResult {
+  return deps.db.$client
+    .transaction(() => {
+      switch (request.operation) {
+        case "listThreadEntries":
+          return {
+            ...toThreadEntries(
+              deps,
+              request,
+              listThreadsWithPendingInteractionState(deps.db, request.options),
```
SawyerHood, line 153: 🚨 `slopcop/review` — Large thread lists still fail at the SQLite variable limit.

This path later passes all thread IDs to `listActiveBackgroundTaskCountsByThreadIds`. That query binds each ID twice.

An exact-SHA test with 16,377 visible threads failed with `too many SQL variables`.

Please split the ID queries into safe groups. Please also add pagination to limit the total work.

apps/server/src/services/database/database-read-service.ts
```diff
@@ -0,0 +1,651 @@
...
+
+  return {
+    close,
+    listThreadEntries(
+      options: ListThreadsOptions,
+      context?: DatabaseReadRequestContext,
+    ): Promise<ThreadListEntry[]> {
+      return sendRequest(
+        (id) => ({
+          daemonSessions: args.hub.listDaemonSessions(),
```
SawyerHood: 🚨 `slopcop/review` — The worker can return old daemon state.

This call records connected daemon sessions before the worker completes the database read. The hub can change during that read.

An exact-SHA race test returned `waiting-for-host` after the host connected. The direct path returned `active`.

Please apply current hub state after the worker result. A hub version check with a retry can also prevent stale results.

apps/server/src/services/database/database-read-service.ts
```diff
@@ -0,0 +1,651 @@
...
+          )
+        : error;
+      state.rejectReady(requestError);
+      rejectAllReads(requestError);
+      if (activeWorker !== state) {
+        return;
+      }
+      activeWorker = null;
+      if (state.wasReady) {
+        activeWorker = startWorker();
```
SawyerHood: 🚨 `slopcop/review` — A recovery restart can throw outside the service error path.

`new Worker()` can throw immediately. This event callback calls `startWorker()` without a guard, so the error can stop the server.

The replacement also has no ready deadline. A worker that never sends `ready` can stop all later reads.

Please use one guarded start function. Give worker startup a deadline and report a controlled unavailable error.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: The server now gives large thread-list database work to a helper. The main server can answer other work during that read.

I found no security defects. I found three concrete reliability and performance defects.

1. **High:** A list with 16,377 visible threads still fails with SQLite's variable limit. The background-task query binds every thread ID twice.
2. **Medium:** The worker can return old host status. It records daemon sessions before the database read finishes.
3. **Medium:** Worker recovery calls `startWorker()` without a guard. A direct constructor error can stop the server, and readiness has no deadline.

The change improves event-loop availability, but it does not reduce the database cost. One slow read still blocks the single worker queue.

The active read also continues after a caller aborts. This keeps the worker busy until the query finishes or reaches its deadline.

The thread search route still runs its database read on the main event loop. This route remains outside the new service.

The project response code now repeats source grouping in two files. A shared pure helper can prevent future response drift.

Validation completed at exact SHA `ba8543a7b61136154d2f0bcb75a7a4dee4828db5`:

- The server, app, database, and SDK type checks passed.
- The production server build passed and included the worker entry.
- All 15 worker-service tests passed in an isolated rerun.
- The route, database, SDK, and retry test groups passed.
- The full parallel run passed 1,389 of 1,390 tests. One worker test reached its time limit under heavy host load.
- The same worker test passed alone in 3.6 seconds.
- The exact dev server returned valid sidebar and thread-list responses.
- A headless browser loaded the thread-list route without page errors.

I left inline notes on the three concrete defects. I did not approve or request changes.

apps/server/src/services/database/database-read-operations.ts
SawyerHood: Confirmed as a real pre-existing defect. Issue #1205 contains the exact 16,376/16,377 boundary and reproduces it on current main. This worker change does not create the variable limit; the direct path fails at the same query. I will keep its query chunking and partial-index fix in #1205. Adding response pagination here would silently omit sidebar threads and change the API contract. No change in this pull request.

apps/server/src/services/database/database-read-service.ts
SawyerHood: Fixed in daaa7481e. The worker now derives database-backed disconnected state. The main thread applies current notification-hub state after each result. A regression test connects a daemon after dispatch and confirms that the returned active thread has active runtime state.

apps/server/src/services/database/database-read-service.ts
SawyerHood: Fixed in daaa7481e. Recovery starts now use one guarded helper, synchronous constructor failures become controlled retryable unavailable errors, and worker readiness has a bounded deadline. A regression test confirms that an expired startup deadline returns DatabaseReadUnavailableError.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am reviewing this pull request for security, quality, performance, architecture, and user behavior.

I will send the checks to three parallel reviewers. I will post one final review after the checks finish.

SawyerHood:
Closing this in favor of #1204.

This PR answered the 11.2 second event-loop stall in #1131. A measurement pass
(#1204) then showed the stall came from `listLatestGoalEventRowsByThreadIds`,
not from the thread-list query. A partial index cut that query from 10.5s to
0.10s cold. A worker thread moves a slow read off the loop; it does not make the
read fast. #1204 makes it fast, so the reason for this machinery is gone.

The work here was good and the review found real problems, which is exactly why
it is worth being deliberate about adopting it. The residual cost is recorded
with its measurements in #1207 (Task, Low), so this option stays revivable. The
branch is kept.

Two pieces of this PR outlive it:

- The `retryable` finding is a pre-existing bug on `main`, independent of this
  change. `retryable: true` is produced today (for example
  `apps/server/src/services/threads/thread-create.ts:146` and several
  host-daemon contract errors) and no client code reads it. That deserves its
  own PR.
- The `listActiveBackgroundTaskCountsByThreadIds` variable-limit crash found
  during this work is filed as #1205.

Thanks for the thorough work on this one.

PR #1199, opened 2026-08-08
Page large finished turns by stored byte size

### LINKED ISSUE #1129: Large finished turns can OOM bb-server during timeline projection and cause a restart loop
> Agent-generated report. I asked an agent to investigate and file this from production logs. I reviewed the facts below; private hostnames and thread IDs have been removed.

## Summary

`bb-server` 0.35.1 can exhaust the V8 heap while building the latest timeline for a large thread. One timeline request can materialize and decode tens of megabytes of event JSON before returning a small projected response. In my case this killed the process repeatedly and put the service into a systemd restart loop.

The service recovered once the timeline requests stopped, but the same workload can trigger it again.

## Environment

- bb: 0.35.1, npm release
- Node.js: 22.23.2
- OS: Linux arm64
- Storage: SQLite with `better-sqlite3`
- Service manager: systemd with `Restart=on-failure`
- `BB_FF_TIMELINE_WINDOW_EVENT_BUDGET`: unset, so the default applies
- cgroup limits: `MemoryHigh=640M`, `MemoryMax=768M`
- Host: 1.8 GiB RAM, 1 GiB swap

## What happened

Before the first crash, the server logged repeated event loop stalls during timeline projection. One request had this shape:

```json
{
  "selectionStrategy": "standard-window",
  "pageKind": "latest",
  "segmentLimit": 20,
  "eventRowCount": 1306,
  "eventDataBytes": 29135518,
  "decodedEventCount": 1306,
  "projectedRowCount": 45,
  "responseRowCount": 45,
  "stageTimings": [
    { "stage": "event-query", "durationMs": 713.4 },
    { "stage": "event-json-decode", "durationMs": 597.7 },
    { "stage": "thread-view-projection", "durationMs": 76.6 }
  ]
}
```

Another request read 5,594 event rows totaling 6.2 MB and returned 15 projected rows:

```json
{
  "selectionStrategy": "standard-window",
  "pageKind": "latest",
  "segmentLimit": 20,
  "eventRowCount": 5594,
  "eventDataBytes": 6228463,
  "decodedEventCount": 5594,
  "projectedRowCount": 15,
  "responseRowCount": 15
}
```

The process then failed around a 300 MB V8 heap:

```text
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```

The native stacks varied between failures. Some ended in `better_sqlite3.node` while converting rows from `Statement::JS_all`; others ended in V8 `JsonParser` during `JSON.parse`.

systemd restarted the service after five seconds. Connected clients reconnected and requested timelines again, and the new process often crashed within 15 to 45 seconds. This repeated 11 times. Observed process memory peaks ranged from roughly 426 MB to 520 MB.

## Reproduction shape

I do not have a small deterministic fixture yet, but the production trigger is consistent:

1. Create a thread with a large finished turn containing many tool events or large event payloads.
2. Open that thread so the client requests the latest timeline.
3. Restart the server, or let it restart while clients remain connected.
4. The server rebuilds the timeline, materializes a large event set, stalls the event loop, and can exhaust the V8 heap.
5. Reconnecting clients can turn the failure into a restart loop.

The logs showing 5,594 selected events are also surprising with the default timeline event budget. If a finished turn intentionally bypasses that budget, a single large turn still needs a byte or memory ceiling.

## Expected behavior

A single thread or timeline request should not terminate the server. Oversized turns should be paginated, truncated with an explicit continuation, or rejected with a bounded per-request error that leaves the process healthy.

## Suggested investigation

- Apply event count and byte limits before `Statement.all()` materializes all selected rows.
- Bound JSON decoding and projection by bytes, not only event count.
- Make the limit cover a single oversized finished turn.
- Avoid re-running the same fatal timeline projection immediately after clients reconnect.
- Add a regression fixture with a finished turn containing large tool payloads and assert that the latest timeline request stays within a fixed memory envelope.

I can provide more anonymized journal output or inspect the SQLite event distribution if that would help narrow down the query path.

SawyerHood:
## Reproduction

A finished turn with 1,305 events and 32,686,106 stored bytes exhausted a 180 MiB V8 heap. The timeline request returned an empty reply, and the server restarted.

## Root cause

The 1,500-event budget did not limit stored bytes. Finished turns also kept one summary segment, so the server loaded and decoded the complete 32 MiB turn.

## Fix

Measure stored UTF-8 bytes before row allocation. Move the sequence floor until the newest suffix fits. Return a truncated page with an older cursor. This changes timeline pagination semantics because one finished summary can now span several sequence pages.

The limit is 4 MiB. A 16 MiB latest page failed the heap test. A later 8 MiB nested page also failed. At 4 MiB, eight HTTP 200 pages retained all 650 tool calls under a 180 MiB heap.

A single event over 4 MiB cannot fit in a smaller window. The server now returns HTTP 200 with a safe placeholder. It gives the client an older cursor when earlier events exist. This replaces the previous permanent 413 response.

Tests cover byte floors, delegated children, finished-turn page identity, client page merge, and the single-event placeholder. All touched package type checks passed. All relevant tests passed. The full server suite has only the known local file-mode skill-tree failure.

Fixes #1129


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am reviewing this pull request for security, code quality, architecture, performance, and practical behavior.

apps/server/src/services/threads/timeline.ts
```diff
@@ -880,20 +888,75 @@ interface ResolveTimelineSegmentWindowArgs {
...
+        },
+        retryable: false,
+      },
+    );
+  }
+  if (floor.kind === "fits") {
+    return args.window;
+  }
+
+  return {
```
SawyerHood, line 1024: 🚨 `slopcop/review` — This cut can restore an unbounded parent payload.

The count-budget path checks `hasParentedEventCrossingSequence` before a cut. This byte path does not check it. `ensureTimelineWindowParentedRows` then fetches all descendants without sequence bounds.

I reproduced a 4 MiB limit that returned a 16.5 MiB response from a finished delegation turn. Please keep cuts outside parent spans, or add a bounded parent-pagination contract.

apps/server/src/services/threads/timeline.ts
```diff
@@ -880,20 +888,75 @@ interface ResolveTimelineSegmentWindowArgs {
...
+    ...args.window,
+    requiresWholeItemClosure: true,
```
SawyerHood, line 1027: 🚨 `slopcop/review` — This cut hides the newest part of a finished turn after the client merges pages.

Only the newest page contains `turn/completed`. Older pages project the same turn as ordinary work.

With 650 commands, the API returned eight complete pages. The browser then showed 567 commands, and expansion ended at command 566. Commands 567 through 649 were inaccessible.

Please preserve completion and use distinct row identities for page slices. Add a client merge test for a byte-split finished turn.

apps/server/src/services/threads/timeline.ts
```diff
@@ -880,20 +888,75 @@ interface ResolveTimelineSegmentWindowArgs {
...
+  const eventDataBytes = getStoredTimelineWindowEventDataBytes(db, windowArgs);
+  if (eventDataBytes <= THREAD_TIMELINE_EVENT_DATA_BYTE_LIMIT) {
+    return args.window;
+  }
+
+  const floor = findStoredTimelineWindowByteBudgetFloor(db, {
+    ...windowArgs,
+    maxDataBytes: THREAD_TIMELINE_EVENT_DATA_BYTE_LIMIT,
+  });
+  if (floor.kind === "single-event-too-large") {
```
SawyerHood, line 989: 🚨 `slopcop/review` — This 413 can make a thread unopenable.

The latest page uses this path. The SQL limit does not shorten tool arguments, file data, or user text.

I stored a newest 4.19 MiB tool input. The timeline route returned a non-retryable 413 before it rendered any row.

Please return a safe placeholder page for one oversized event. Add a latest-page test because the current test covers only an older page.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: This change splits one large finished turn into smaller pages. That reduces server memory use during a timeline load.

I found three important problems.

1. **High: delegation closure defeats the byte limit.** The byte cut skips the existing parent-crossing guard. Parent closure then fetches all descendants without sequence bounds. A delegated fixture loaded 32.7 MiB of event data with a 4 MiB limit. The live route returned a 16.5 MiB response. Please page nested rows or keep the cut outside parent spans.

2. **High: one newest oversized event makes the thread unopenable.** I stored a newest 4.19 MiB tool input. The latest timeline route returned a non-retryable 413 before any row rendered. The read limit does not shorten tool arguments, file data, or user text. Please return a bounded placeholder page for one oversized event.

3. **Medium: the client loses the newest slice of a finished turn.** The API returned all 650 commands across eight pages. After browser auto-load, the UI showed 567 commands. Expansion ended at command 566, so commands 567 through 649 were inaccessible. Please preserve completion state and use distinct row identities for page slices.

The architecture scan found one useful refactor. `getStoredTimelineWindowEventDataBytes` scans the same window before the floor query scans it again. The floor query already returns `fits`. One query can handle all three results.

The security review found no SQL injection or cross-thread cursor access. Drizzle binds the raw prepared-query parameters.

Validation passed:

- Server timeline tests: 17 passed.
- Database event tests: 70 passed.
- Turbo type checks passed for `@bb/server` and `@bb/db`.
- The development app and timeline route started successfully.
- Browser automation loaded the eight-page finished-turn fixture and reproduced the missing newest slice.

I left inline comments on the three important findings. I did not approve or request changes.

apps/server/src/services/threads/timeline.ts
SawyerHood: Fixed in 48daf8b10. A single oversized event now returns HTTP 200 with a safe error placeholder. It returns an older cursor when earlier rows exist. The latest-page regression confirms the server does not allocate the event row.

SawyerHood:
Addressed all SlopCop findings in 48daf8b10.

- Parent and delegation closure now respects byte-page sequence bounds.
- Finished-turn slices retain completion and use stable page identities.
- One oversized event returns a safe HTTP 200 placeholder instead of 413.
- The floor query now reports fits and byte totals, so the server no longer scans the window twice.

The original 1,305-event fixture returned eight HTTP 200 pages under a 180 MiB heap. All 650 tool calls remained available. App and database suites passed. The server suite had only the known local file-mode skill-tree failure.

PR #1204, opened 2026-08-08
Fix the #1131 cold stall at its source: index the latest-goal lookup, drop the thread-list GROUP BY

SawyerHood:
## Summary

This PR investigates issue #1131 with measurements. The result changes the diagnosis.

The 11-second cold stall did not come from `listThreadsWithPendingInteractionState`. It came from `listLatestGoalEventRowsByThreadIds`, which the same route calls for every listed thread. Its `ORDER BY` made the planner walk `events_thread_sequence_idx` and fetch **every event row of every listed thread**. On a grown database, that read is close to the whole file.

This PR fixes that query with one small partial index and a pinned rewrite. Cold reads for that query drop from **444,010 to 357** at 26K threads. The PR also replaces the thread-list `GROUP BY` + join with a correlated `EXISTS`. That removes both temp B-trees, but measurement shows the temp B-trees were never the real cost.

## How I measured

- I built a 1.45GB `bb.db` at the reported scale with the real `migrate()`: 30,000 threads, 594,971 events (issue reports ~595K events, 1447MB events table), 29,842 interactions, 25 pending. I inserted rows in global timestamp order across concurrent threads, so `threads` pages interleave with `events` pages like a real grown file. No `ANALYZE`, because production never runs it.
- I ran the real Drizzle query functions from `@bb/db` and captured the exact generated SQL and `EXPLAIN QUERY PLAN`.
- Cold means evicted page cache: `posix_fadvise(DONTNEED)` on the db/wal/shm files, fresh process, verified with `/proc/self/io` read counters.
- The host has a fast NVMe disk. Absolute cold times transfer badly, so I also report read-syscall counts, and I re-ran cold tests inside a cgroup capped at 1000 read IOPS to model a stressed cloud disk. All cold runs are latency-bound: time ≈ reads ÷ IOPS.

## What the 11.2s actually was

`bb thread list --json` at 26,459 visible threads, cold:

| query in the route | cold reads | cold (NVMe) | cold (1000 IOPS) | warm |
|---|---|---|---|---|
| `listThreadsWithPendingInteractionState` | 2,044 | 0.30s | 2.2s | 0.18s |
| `listLatestGoalEventRowsByThreadIds` (before) | **444,010** | **10.5s** | ~7 min (projected) | 0.77s |
| `listLatestGoalEventRowsByThreadIds` (after) | **357** | **0.10s** | 0.29s | 0.10s |

The before-numbers reproduce the issue precisely: 10.5s cold matches the reported 11.2s stall, and 0.77s warm matches the reported ~0.5s warm repeats.

Why it happened: the query had `ORDER BY thread_id, sequence`. The stats-less planner served that order from `events_thread_sequence_idx` to avoid a sort, then filtered by type per row. That plan walks and fetches every event row (multi-KB payloads included) of every listed thread. The fix has three parts:

1. A partial index `events_goal_thread_sequence_idx ON events (thread_id, sequence) WHERE type IN ('thread/goal/updated', 'thread/goal/cleared')`. Goal events are rare, so the index stays tiny.
2. The query spells the goal types as SQL literals and pins the index with `INDEXED BY`. Both are required: SQLite cannot match a partial index against a bound parameter, and the stats-less planner assumes every index is table-sized, so it never picks the small one on its own (verified: it kept the big index until pinned).
3. No `ORDER BY`. `sequence` is unique per thread, so the query returns at most one row per thread, and both callers sort or ignore order.

I verified the new query returns row-identical results against the old SQL on a dataset with 10,579 goal events across 3,000 threads (2,750 result rows, zero differences).

## The thread-list query itself, and the temp B-tree theory

The prior observation was correct: the query used temp B-trees for GROUP BY and ORDER BY, and `bb thread list` walked all threads in id-index order (random page order). The `EXISTS` rewrite in this PR removes the join, the GROUP BY, and one temp B-tree, with row-identical results (verified across 31,309 rows in four query shapes). Measured effect, before → after:

| shape | rows | cold reads | cold (1000 IOPS) | warm |
|---|---|---|---|---|
| `bb thread list` (all visible) | 26,459 | 2,044 → 1,959 | 2.16s → 2.05s | 0.18s → 0.17s |
| sidebar bootstrap (active, all projects) | 4,950 | 1,771 → 1,767 | 1.76s → 1.75s | 51ms → 47ms |
| one project, active | 169 | 352 → 352 | 0.40s → 0.40s | 6ms |
| one project, archived | 731 | 973 → 973 | 1.01s → 1.01s | 19ms |

The temp B-trees were a signature, not the cost. The cold cost is the row fetches: the query returns every visible thread row, and those rows sit on ~2,000 pages scattered through the file. No index can remove that. I tested the extremes:

- A 24-column covering partial index (the entire row duplicated): helps only the single-project shape (352 → 195 reads), the planner does not pick it for the IN-list sidebar shape, and it abandons it once stats exist. Not shipped.
- Full `VACUUM` (contiguous pages) plus the ideal sequential scan: cli-list cold still needs 0.85s at 1000 IOPS. bb never full-vacuums an incremental-mode database in steady state, so real files stay fragmented.

So for the thread-list query proper, the honest answer to "can an index make cold fast?" is **no**. About 1,700–2,000 scattered cold reads are inherent at this scale (roughly 2s at 1000 IOPS), and the ~50–170ms warm cost is per-row CPU (26K rows: ~63ms inside SQLite, the rest row mapping), which no index changes.

## Relationship to #1198 (now closed) and #1207

I checked these claims against the #1198 branch (`database-read-service.ts`, `database-read-worker-entry.ts`, `start-server.ts`), not from memory.

- #1198 ran the whole route read inside its worker: the worker entry called `listThreadsWithPendingInteractionState` and `toThreadListEntryResponses`, and the latter reaches `listLatestGoalEventRowsByThreadIds`. So #1198 **would** have moved the 10.5s read off the event loop. The global freeze would stop: static assets and every other route would stay responsive.
- #1198 would **not** have made the thread list fast. The request itself would still have taken the same ~10.5s cold inside the worker. It would not have 503'd: the request timeout default was 30 seconds (`DEFAULT_DATABASE_READ_TIMEOUT_MS = 30_000`, and the only construction site passed no override), which a 10.5s read does not reach.
- The accurate division of labor: **#1198 contained the blast radius of a slow read; this PR removes the dominant slow read.** One masks the symptom, the other fixes the cause.
- #1198 is now closed. The residual synchronous cost that remains after this PR — the inherent thread-list row fetches, roughly 2s cold at 1000 IOPS at 26K threads, tens of milliseconds on a healthy disk with a warm cache — is tracked in #1207.

## Related findings (not addressed here)

- `listActiveBackgroundTaskCountsByThreadIds` (same route) has the same access-path class: ~4,150 cold reads at 5K threads, ~6,888 at 15K. Worse, it binds the thread-id list twice, so at 16,377 or more visible threads (measured boundary; the double-bound list plus fixed parameters exceeds SQLite's 32,766-variable limit) it throws `too many SQL variables` and the thread-list route fails outright. Tracked separately with a reproduction in #1205.
- The route's per-thread queries all scale with visible-thread count. Pagination for `bb thread list` would cap every term.

## Test plan

- `pnpm exec turbo run test --filter=@bb/db`: 359/359 pass. The migration-replay suite forced `CREATE INDEX IF NOT EXISTS` (same pattern as 0084/0086), and the index-inventory expectation in `migrate.test.ts` now includes the new index.
- `pnpm exec turbo run test --filter=@bb/server`: 1366/1367 pass; the one failure is `internal-skill-trees.test.ts` expecting file mode 0644 under my machine's umask 0002, which also fails on `main` and is unrelated.
- Row-identical result verification for both rewritten queries, described above.

Refs #1131 (measurement basis; deliberately not "Fixes" — #1131 stays open for the maintainer to close once follow-up #1205 also lands). Residual event-loop work is tracked in #1207.

### DIFF
--- a/packages/db/src/data/events.ts
+++ b/packages/db/src/data/events.ts
@@ -988,27 +988,49 @@ export function listLatestGoalEventRowsByThreadIds(
     return [];
   }
 
+  // This runs over every listed thread on each sidebar bootstrap, so it must
+  // stay proportional to the number of goal events, not the number of events.
+  // Three deliberate choices make that hold on a stats-less database:
+  //   - The goal-type list is spelled as SQL literals matching the WHERE
+  //     clause of events_goal_thread_sequence_idx exactly; SQLite only uses a
+  //     partial index when the query predicate provably implies the index
+  //     predicate at prepare time, which a bound parameter never does.
+  //   - INDEXED BY pins the tiny partial index. Without ANALYZE data the
+  //     planner assumes every index is table-sized: it previously served an
+  //     ORDER BY from events_thread_sequence_idx and walked every event row
+  //     of every listed thread — a full-file read on a grown events table
+  //     (issue #1131).
+  //   - No ORDER BY: sequence is unique per thread, so this returns at most
+  //     one row per thread and callers do not depend on row order.
   const goalTypes = [
     "thread/goal/updated",
     "thread/goal/cleared",
-  ] satisfies ThreadEventType[];
+  ] as const satisfies readonly ThreadEventType[];
+  const goalTypesPredicate = sql.raw(
+    `IN (${goalTypes.map((type) => `'${type}'`).join(", ")})`,
+  );
+  const threadIdList = sql.join(
+    args.threadIds.map((threadId) => sql`${threadId}`),
+    sql`, `,
+  );
 
   return db
     .select(storedEventRowFields)
     .from(events)
     .where(
-      and(
-        inArray(events.threadId, [...args.threadIds]),
-        inArray(events.type, goalTypes),
-        sql`${events.sequence} = (
-          SELECT MAX(latest.sequence)
-          FROM events latest
-          WHERE latest.thread_id = ${events.threadId}
-            AND latest.type IN (${goalTypes[0]}, ${goalTypes[1]})
-        )`,
-      ),
+      sql`${events}.rowid IN (
+        SELECT latest_goal.rowid
+        FROM ${events} AS latest_goal INDEXED BY events_goal_thread_sequence_idx
+        WHERE latest_goal.thread_id IN (${threadIdList})
+          AND latest_goal.type ${goalTypesPredicate}
+          AND latest_goal.sequence = (
+            SELECT MAX(candidate.sequence)
+            FROM ${events} AS candidate INDEXED BY events_goal_thread_sequence_idx
+            WHERE candidate.thread_id = latest_goal.thread_id
+              AND candidate.type ${goalTypesPredicate}
+          )
+      )`,
     )
-    .orderBy(events.threadId, events.sequence)
     .all();
 }
 
--- a/packages/db/drizzle/0088_narrow_kronos.sql
+++ b/packages/db/drizzle/0088_narrow_kronos.sql
@@ -1,0 +1,1 @@
+CREATE INDEX IF NOT EXISTS `events_goal_thread_sequence_idx` ON `events` (`thread_id`,`sequence`) WHERE "events"."type" IN ('thread/goal/updated', 'thread/goal/cleared');
--- a/packages/db/drizzle/meta/0088_snapshot.json
+++ b/packages/db/drizzle/meta/0088_snapshot.json
@@ -1,0 +1,3437 @@
+{
+  "version": "6",
+  "dialect": "sqlite",
+  "id": "e3c1c069-39fa-4091-8b7c-a202d370f8ba",
+  "prevId": "3a348e16-0592-4d7b-8f90-b6afcafffe48",
+  "tables": {
+    "app_settings": {
+      "name": "app_settings",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "text",
+          "primaryKey": true,
+          "notNull": true,
+          "autoincrement": false
+        },
+        "caffeinate": {
+          "name": "caffeinate",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true,
+          "autoincrement": false,
+          "default": false
+        },
+        "show_keyboard_hints": {
+          "name": "show_keyboard_hints",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true,
+          "autoincrement": false,
+          "default": true
+        },
+        "steer_active_thread_on_enter": {
+          "name": "steer_active_thread_on_enter",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true,
+          "autoincrement": false,
+          "default": false
+        },
+        "show_unhandled_provider_events": {
+          "name": "show_unhandled_provider_events",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true,
+          "autoincrement": false,
+          "default": false
+        },
+        "codex_memory_enabled": {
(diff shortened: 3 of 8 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull request for security, code quality, performance, architecture, and test coverage.

I will also test the changed route when the local product supports that test.

packages/db/src/data/events.ts
```diff
@@ -988,27 +988,49 @@ export function listLatestGoalEventRowsByThreadIds(
...
-        sql`${events.sequence} = (
-          SELECT MAX(latest.sequence)
-          FROM events latest
-          WHERE latest.thread_id = ${events.threadId}
-            AND latest.type IN (${goalTypes[0]}, ${goalTypes[1]})
-        )`,
-      ),
+      sql`${events}.rowid IN (
+        SELECT latest_goal.rowid
+        FROM ${events} AS latest_goal INDEXED BY events_goal_thread_sequence_idx
```
SawyerHood, line 1023: 🚨 `slopcop/review` — Add a query-plan regression test for both rewritten hot paths.

The behavior tests prove result equivalence, but they do not protect the performance contract. Assert that this query uses `events_goal_thread_sequence_idx` twice. Reject `events_thread_sequence_idx` and a temporary sort. Also confirm `pending_interactions_thread_status_created_idx` for the correlated `EXISTS` query.

This comment is non-blocking. The test will protect the exact failure mode that this pull request fixes.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** SQLite now uses a small goal-only index instead of reading every event. The thread list now asks a simple yes-or-no question for pending input.

**Findings**

- Low, non-blocking: The new performance contract has no query-plan regression test. I left one inline comment with the required checks.
- I found no security, correctness, migration, architecture, duplication, or current performance defect.

**Architecture**

The repeated goal-type literals are necessary because SQLite must match the partial-index condition during query preparation. A shared helper would add complexity without a clear benefit.

**Validation**

- All 359 database tests passed at the exact head SHA.
- The Turbo database type check passed.
- SQLite used `events_goal_thread_sequence_idx` for both goal scans.
- SQLite used `pending_interactions_thread_status_created_idx` for the pending-input check.
- A local server and browser test returned status 200 for both thread-list routes.
- The row-backed response reported one active goal and one pending interaction.
- All required GitHub CI jobs passed.
- The final GPT-5.6 review gate found no blocking defect.

I found no blocking issue. This GitHub review uses comment status only.

packages/db/src/data/events.ts
SawyerHood: Added in f3c6eb3e1 (`packages/db/test/query-plans.test.ts`): the latest-goal plan must probe `events_goal_thread_sequence_idx` exactly twice, must not use `events_thread_sequence_idx`, and must not use a temp B-tree; the thread-list plan must serve the pending probe from `pending_interactions_thread_status_created_idx` with no GROUP BY temp B-tree. The tests capture the exact prepared SQL rather than the slow-query log, because the log redacts literals and truncates — either would change the plan under EXPLAIN (a redacted literal no longer implies the partial-index predicate).

PR #1208, opened 2026-08-08
fix(pi): protect bridge JSON-RPC stdout

### LINKED ISSUE #1180: pi: thread stuck on Working...
And nothing appears to be happening. No relevant logs.

bb-app@0.36.0

raincodes64:
Reserve the Pi bridge stdout pipe for protocol frames and redirect extension output to stderr. Add regression coverage for OSC notification output.

Fixes #1180

### DIFF
--- a/packages/agent-runtime/src/pi/bridge/output-guard.ts
+++ b/packages/agent-runtime/src/pi/bridge/output-guard.ts
@@ -1,0 +1,70 @@
+type ProcessWriteCallback = (error?: Error | null) => void;
+type RedirectedProcessWrite = (
+  chunk: string | Uint8Array,
+  encodingOrCallback?: BufferEncoding | ProcessWriteCallback,
+  callback?: ProcessWriteCallback,
+) => boolean;
+
+interface PiBridgeStdoutTakeoverState {
+  originalStdoutWrite: typeof process.stdout.write;
+  protocolStdoutWrite: (
+    chunk: string,
+    callback?: ProcessWriteCallback,
+  ) => boolean;
+}
+
+let stdoutTakeoverState: PiBridgeStdoutTakeoverState | undefined;
+
+export function takeOverPiBridgeStdout(): void {
+  if (stdoutTakeoverState) {
+    return;
+  }
+
+  const protocolStdoutWrite = process.stdout.write.bind(
+    process.stdout,
+  ) as PiBridgeStdoutTakeoverState["protocolStdoutWrite"];
+  const stderrWrite = process.stderr.write.bind(
+    process.stderr,
+  ) as RedirectedProcessWrite;
+  const originalStdoutWrite = process.stdout.write;
+
+  process.stdout.write = ((
+    chunk: string | Uint8Array,
+    encodingOrCallback?: BufferEncoding | ProcessWriteCallback,
+    callback?: ProcessWriteCallback,
+  ): boolean => {
+    if (typeof encodingOrCallback === "function") {
+      return stderrWrite(chunk, encodingOrCallback);
+    }
+    if (encodingOrCallback !== undefined) {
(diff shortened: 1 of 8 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull request for security, code quality, performance, architecture, and test coverage.

packages/agent-runtime/src/pi/bridge/output-guard.ts
```diff
@@ -0,0 +1,70 @@
...
+    if (typeof encodingOrCallback === "function") {
+      return stderrWrite(chunk, encodingOrCallback);
+    }
+    if (encodingOrCallback !== undefined) {
+      return stderrWrite(chunk, encodingOrCallback, callback);
+    }
+    if (callback) {
+      return stderrWrite(chunk, callback);
+    }
+    return stderrWrite(chunk);
```
SawyerHood: 🚨 `slopcop/review` — The redirect can increase memory use without a limit

The redirect sends all extension output to provider stderr. The parent runtime uses `readline` and stores every completed line for the process lifetime. Newline-free OSC output stays in the `readline` buffer. At exit, the runtime joins all stored lines before it keeps only 4,000 characters. A verbose extension can therefore increase memory use without a limit. Limit bytes as data arrives, including incomplete lines. The existing ACP and Claude bounded-tail code shows suitable patterns.

packages/agent-runtime/src/pi/bridge/output-guard.ts
```diff
@@ -0,0 +1,70 @@
...
+  ) as RedirectedProcessWrite;
+  const originalStdoutWrite = process.stdout.write;
+
+  process.stdout.write = ((
+    chunk: string | Uint8Array,
+    encodingOrCallback?: BufferEncoding | ProcessWriteCallback,
+    callback?: ProcessWriteCallback,
+  ): boolean => {
+    if (typeof encodingOrCallback === "function") {
+      return stderrWrite(chunk, encodingOrCallback);
```
SawyerHood: 🚨 `slopcop/review` — The stdout redirect can stop a writer forever

`process.stdout.write()` returns the result from `stderr.write()`. If stderr returns `false`, the caller waits for `drain` on stdout. Only stderr emits that event. For example, `source.pipe(process.stdout)` can pause forever after enough output. Keep the write result and its `drain` event on the same stream. Add a test where stderr returns `false`.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: The bridge needs a clean JSON message line. This change moves extension text away, but the new path can stop and fill memory.

I found two merge blockers:

1. **High — The redirect breaks stdout flow control.** It reports stderr backpressure through stdout. A caller can then wait for `drain` on the wrong stream and stop forever.

2. **High — Redirected output has no memory limit.** The parent stores every stderr line for the process lifetime. Newline-free OSC text also stays in the `readline` buffer. The exit formatter joins all data before it keeps the last 4,000 characters.

Security and architecture:

- The Pi-specific guard has the correct scope because Pi loads extensions in its bridge process.
- The guard does not isolate hostile code. Direct file-descriptor writes, saved writers, and inherited stdout can bypass it.
- I treat those bypasses as a documented limit, not a blocker for this transport fix.
- The shared provider-process boundary should use the existing ACP or Claude bounded-tail pattern.

Validation:

- Three independent GPT-5.6 reviews covered security, code quality, architecture, and performance.
- The final GPT-5.6 gate confirmed both findings.
- GitHub CI passed all required jobs.
- `git diff --check` passed.
- I did not execute code from this untrusted fork.
- This backend bridge change has no browser route, so browser end-to-end testing does not apply.

I submitted this as a comment-only review. Fix both stream issues and add tests for backpressure and bounded capture.

PR #1209, opened 2026-08-08
Fix thread list failure above SQLite variable limit

### LINKED ISSUE #1205: listActiveBackgroundTaskCountsByThreadIds throws 'too many SQL variables' at ≥16,377 visible threads, failing the thread-list route
## Summary

`listActiveBackgroundTaskCountsByThreadIds` (`packages/db/src/data/events.ts`) binds the caller's thread-id list twice — once in each CTE. SQLite's variable limit is 32,766. At **16,377 or more thread ids the query throws `SqliteError: too many SQL variables`** (measured boundary: 16,376 works, 16,377 throws; the double-bound list plus the fixed parameters crosses the limit).

`toThreadListEntryResponses` calls this function with every listed thread id. So on an instance with ≥16,377 visible threads, `GET /api/threads` (and `bb thread list`) does not get slow — it **fails outright**. The projects-with-threads bootstrap fails the same way once the active thread count crosses the boundary.

Found while measuring #1131 at 26K visible threads; the thread-list route could not run at all until I capped the id list.

## Reproduction

Verified against `main` (d807f1364):

```ts
import {
  createConnection,
  migrate,
  listActiveBackgroundTaskCountsByThreadIds,
} from "@bb/db";

const db = createConnection(":memory:");
migrate(db);
const ids = (n: number) => Array.from({ length: n }, (_, i) => `thr_${i}`);

listActiveBackgroundTaskCountsByThreadIds(db, { threadIds: ids(16_376) }); // ok
listActiveBackgroundTaskCountsByThreadIds(db, { threadIds: ids(16_377) });
// SqliteError: too many SQL variables
```

No thread rows need to exist; the prepare fails before execution.

## Also: the same query has the pre-fix #1131 access pattern

Cold-cache cost measured on a 1.45GB bb.db (30K threads / 595K events, page cache evicted with `posix_fadvise(DONTNEED)`, reads counted via `/proc/self/io`):

| thread ids | cold read syscalls | cold time @1000 IOPS |
|---|---|---|
| 5,000 | ~4,150 | ~4s |
| 15,000 | ~6,888 | ~7s |

The probes run over the full-size `events_thread_type_item_kind_sequence_idx`, so cold cost scales with the events index, not with the number of background-task events. This is the same class of problem #1204 fixed for `listLatestGoalEventRowsByThreadIds`.

## Suggested fix

1. Chunk the thread-id list (or use a temp table / JSON parameter) so the route survives any thread count.
2. Give it the #1204 treatment: a partial index over `item_kind = 'backgroundTask'` events with literal predicates (a partial index never matches a bound parameter) and an `INDEXED BY` pin, since the stats-less planner will not pick a small partial index on its own.

Refs #1131, #1204.

SawyerHood:
## Summary

- Chunk active background-task count reads at 16,377 thread IDs.
- Derive that size from the 32,766-variable limit, 11 fixed values, and two uses per ID.
- Batch every thread-list event lookup through one variable-budget helper.
- Add and pin a covering partial index for background-task events.
- Add boundary, route, result-equivalence, migration, and query-plan tests.

## Reproduction

I ran the issue reproduction before any change. A list of 16,376 IDs completed. A list of 16,377 IDs failed with `SqliteError: too many SQL variables`.

After the fix, lists of 16,377, 16,378, and 32,754 IDs completed. The thread-list response also completed with 32,767 active threads.

## Measurements

I used the 1.45 GB synthetic database from #1204. It has 30,000 threads and about 595,000 events. It has no background-task events. Thus, these results measure rejection of unrelated events. Correctness tests use real background-task rows.

Each cold sample used a fresh process. I evicted the database page cache with `posix_fadvise(DONTNEED)`. I measured reads with `/proc/self/io`. I capped the NVMe device at 1,000 read IOPS. Each value below is the median of three samples.

| Thread IDs | Cold reads before | Cold reads after | Cold time before | Cold time after | Warm time before | Warm time after |
|---:|---:|---:|---:|---:|---:|---:|
| 5,000 | 4,149 | 3 | 3.96 s | 18.7 ms | 29.1 ms | 18.7 ms |
| 15,000 | 6,888 | 3 | 6.87 s | 52.7 ms | 75.5 ms | 53.5 ms |

The old plan used `events_thread_type_item_kind_sequence_idx` for both task scans. The fixed plan uses `events_background_task_thread_type_item_sequence_idx` as a covering index for both scans. The query has a literal partial-index predicate and two `INDEXED BY` pins.

## Verification

- `pnpm exec turbo run test --filter=@bb/db --force`: 370 tests passed.
- The focused server route suite passed all 9 tests.
- `pnpm exec turbo run typecheck --filter=@bb/db --filter=@bb/server`: passed.
- A regression test crosses the old active-task variable boundary.
- A regression test compares chunked counts with single-query counts.
- A route regression builds 32,767 active thread-list entries.
- Query tests cross the goal and request-key limits.
- A query-plan test confirms both partial-index scans.
- Drizzle generated migration 0089 and its snapshot.

Fixes #1205.

### DIFF
--- a/packages/db/src/data/events.ts
+++ b/packages/db/src/data/events.ts
@@ -53,16 +53,40 @@ import {
 } from "./threads.js";
 
 const STORED_EVENT_SEQUENCE_LOOKUP_CHUNK_SIZE = 250;
-// SQLite accepts 32,766 variables. This query binds each thread ID twice and
-// also binds 11 fixed values.
 const SQLITE_MAX_VARIABLE_NUMBER = 32_766;
-const ACTIVE_BACKGROUND_TASK_FIXED_PARAMETER_COUNT = 11;
-const ACTIVE_BACKGROUND_TASK_THREAD_ID_PARAMETER_COUNT = 2;
-const ACTIVE_BACKGROUND_TASK_THREAD_ID_CHUNK_SIZE = Math.floor(
-  (SQLITE_MAX_VARIABLE_NUMBER -
-    ACTIVE_BACKGROUND_TASK_FIXED_PARAMETER_COUNT) /
-    ACTIVE_BACKGROUND_TASK_THREAD_ID_PARAMETER_COUNT,
-);
+// This OR query prepares with 995 keys. A 996th key reaches the configured
+// SQLite expression-depth limit of 1,000.
+const CLIENT_TURN_REQUEST_KEY_BATCH_SIZE = 995;
+
+interface QueryInSqliteVariableBatchesArgs<TValue, TRow> {
+  fixedVariableCount: number;
+  maximumValueCount?: number;
+  queryBatch: (values: readonly TValue[]) => readonly TRow[];
+  values: readonly TValue[];
+  variableCountPerValue: number;
+}
+
+function queryInSqliteVariableBatches<TValue, TRow>(
+  args: QueryInSqliteVariableBatchesArgs<TValue, TRow>,
+): TRow[] {
+  const variableBatchSize = Math.floor(
+    (SQLITE_MAX_VARIABLE_NUMBER - args.fixedVariableCount) /
+      args.variableCountPerValue,
+  );
+  const batchSize = Math.min(
+    variableBatchSize,
+    args.maximumValueCount ?? variableBatchSize,
+  );
+  if (batchSize < 1) {
+    throw new Error("The fixed SQL variables exceed the SQLite limit");
+  }
+
+  const rows: TRow[] = [];
+  for (let offset = 0; offset < args.values.length; offset += batchSize) {
+    rows.push(...args.queryBatch(args.values.slice(offset, offset + batchSize)));
+  }
+  return rows;
+}
 
 const isRootTurnStartedEventData = sql`COALESCE(json_extract(${events.data}, '$.parentToolCallId'), '') = ''`;
 const isNotNestedTurnUsageEvent = sql`NOT EXISTS (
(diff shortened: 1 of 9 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

packages/db/src/data/events.ts
```diff
@@ -1859,21 +1869,59 @@ export function listActiveBackgroundTaskCountsByThreadIds(
     return [];
   }
 
+  const threadIds = [...new Set(args.threadIds)];
```
SawyerHood: 🚨 `slopcop/review` — The thread-list route still fails at 32,767 visible threads.

This loop fixes the 16,378-ID failure. The same route then passes every ID to `listLatestGoalEventRowsByThreadIds` without chunks. I tested the exact commit with 32,767 visible threads. `/api/v1/sidebar-bootstrap` returned 500. The server logged `too many SQL variables` from `events.ts:1074`. Please use a shared SQLite batch helper for both queries. Add a route test above 32,766 IDs.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5 summary

Think of the thread names as blocks that must fit in a box.

This change uses two boxes, so 16,378 names fit. A later step still uses one box. At 32,767 names, the list breaks again.

### Finding

- Medium — The thread-list route still exceeds the SQLite variable limit at 32,767 visible threads. The inline comment has the full trace. Please batch `listLatestGoalEventRowsByThreadIds`. The active-thread query has the same one-list pattern.

### Review details

- Security: I found no security issues. The raw SQL predicate uses a fixed typed value.
- Code quality: The migration, snapshot, schema, query, and tests agree. Duplicate IDs and chunk results work correctly.
- Performance: The changed query stays below 32,766 variables. Both scans use the new partial index.
- Architecture: `pending-interactions.ts` has another local batch helper. A shared database helper can protect all bulk-ID queries.

### Validation

- The database suite passed with 368 tests in 28 files.
- The database Turbo type check passed.
- The browser route returned 200 with 16,378 visible threads.
- The same route returned 500 with 32,767 visible threads. The server logged `too many SQL variables`.
- All GitHub checks passed.

packages/db/src/data/events.ts
SawyerHood: Fixed in b87b480fd. One variable-budget helper now batches the active-task, goal, open-input, and request-key queries. A server regression builds 32,767 active thread-list entries. The database and focused server suites pass.

PR #1210, opened 2026-08-08
Fix flaky client request batching test

SawyerHood:
## Summary

- replace the 16,383-key timeout test with the real 996-key SQLite expression-depth boundary
- verify that a matching row from the second query batch is returned
- remove the custom 15-second timeout and its unnecessary query work

## Tests

- `pnpm exec turbo run test --filter=@bb/db --force` (six runs; 370 tests passed each run)
- `pnpm exec turbo run typecheck --filter=@bb/db`

## Risk

Low. This changes only regression test data and assertions.

### DIFF
--- a/packages/db/test/data/events.test.ts
+++ b/packages/db/test/data/events.test.ts
@@ -1390,19 +1390,30 @@ describe("events", () => {
     expect(rowsByThreadId.size).toBe(2);
   });
 
-  it(
-    "batches client turn request keys above the SQLite variable limit",
-    () => {
-      const { db } = setup();
-      const keys = Array.from({ length: 16_383 }, (_, index) => ({
-        requestId: "creq_23456789ab",
+  it("batches client turn request keys above the expression-depth limit", () => {
+    const { db, thread } = setup();
+    const requestId = "creq_23456789ab";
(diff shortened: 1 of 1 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am reviewing this pull request for security, code quality, performance, test behavior, and design duplication.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** The code splits a large request list into smaller groups. This test proves that the code also checks the second group.

I found no problems.

- **Security:** This pull request changes test data only. It does not change a trust boundary or product behavior.
- **Code quality:** Key 996 follows the 995-key batch limit. The stored match proves that the second batch returns its row.
- **Performance:** The test now uses two batches instead of 17. It removes the custom 15-second timeout and much unnecessary work.
- **Design:** I found no duplicate helper or useful architecture change. A shared helper would hide the important 995-key boundary.
- **Validation:** The exact pull-request events test passed all 76 tests. The `@bb/db` type check and `git diff --check` also passed.

This pull request changes no route or user interface. Therefore, a dev-server browser test does not apply.

PR #1212, opened 2026-08-09
fix: preserve provider sessions across environment switches

patleeman:
I hit this when the agent switched from my personal working directory to another directory with `update_environment_directory`.

The switch worked, but the session did not migrate with it. The transcript was still there; the next message started a new provider session.

The same state also broke forks: side chats returned HTTP 409, and Fork into new thread returned the same personal-workspace error.

## Repro

1. Start a thread in a personal workspace.
2. Switch it to another directory with `update_environment_directory`.
3. Ask a follow-up question.
4. Try Reply in side chat or Fork into new thread.
5. Before this change, the transcript context was lost and both fork actions failed.

## Fix

- Keep the provider thread ID when the environment changes.
- Release the old runtime before resuming in the new environment.
- Serialize starts and submits per thread across environment lanes so a move cannot deadlock the handoff.
- Let a source-derived fork reuse only its exact switched unmanaged environment.
- Give isolated forks from personal projects a personal workspace.
- Keep the existing project and host boundary checks.

## Verification

I ran the flow in the desktop app against both builds. The pre-fix build lost the transcript context and showed the 409 on both fork actions. The fixed build preserved the original opening message, the side chat replied `SIDE_CHAT_OK`, and the forked thread replied `FORK_OK`.

59 focused host-daemon tests, 37 focused server tests, and 76 DB event tests pass. The affected packages typecheck and build; formatting and diff checks pass. The PR contains no screenshot artifacts.

### DIFF
--- a/apps/host-daemon/src/command-dispatch.ts
+++ b/apps/host-daemon/src/command-dispatch.ts
@@ -230,10 +230,11 @@ function shouldInvalidateProviderMaintenanceRuntimeAfterProviderCliInstall(args:
 
 const commandHandlers: CommandHandlerMap = {
   "thread.start": async (command, options) => {
-    const release = options.runtimeManager.retainEnvironmentForThreadCommand(
-      command.environmentId,
-      command.threadId,
-    );
+    const release =
+      await options.runtimeManager.retainEnvironmentForThreadCommand(
+        command.environmentId,
+        command.threadId,
+      );
     try {
       return await startThread(command, options);
     } finally {
@@ -241,10 +242,11 @@ const commandHandlers: CommandHandlerMap = {
     }
   },
   "turn.submit": async (command, options) => {
-    const release = options.runtimeManager.retainEnvironmentForThreadCommand(
-      command.environmentId,
-      command.threadId,
-    );
+    const release =
+      await options.runtimeManager.retainEnvironmentForThreadCommand(
+        command.environmentId,
+        command.threadId,
+      );
     try {
       const entry = await ensureThreadRuntime(command, options);
       return await submitTurn(command, entry, options);
@@ -257,6 +259,10 @@ const commandHandlers: CommandHandlerMap = {
       command.environmentId,
       options.runtimeManager,
     );
+    await options.runtimeManager.releaseThreadFromOtherEnvironments({
+      environmentId: command.environmentId,
+      threadId: command.threadId,
+    });
     if (entry.runtime.hasThread(command.threadId)) {
       // Stop can be dispatched while the start/submit RPC is still in flight
       // and the turn/started event has not been observed yet. Wait for the
@@ -286,6 +292,10 @@ const commandHandlers: CommandHandlerMap = {
       command.environmentId,
       options.runtimeManager,
     );
+    await options.runtimeManager.releaseThreadFromOtherEnvironments({
+      environmentId: command.environmentId,
+      threadId: command.threadId,
+    });
     if (!entry.runtime.hasThread(command.threadId)) {
       throw new ExpectedCommandDispatchError(
         "unknown_thread_runtime",
@@ -308,6 +318,10 @@ const commandHandlers: CommandHandlerMap = {
     if (!entry) {
       return {};
     }
+    await options.runtimeManager.releaseThreadFromOtherEnvironments({
+      environmentId: command.environmentId,
+      threadId: command.threadId,
+    });
     await entry.runtime.renameThread({
       threadId: command.threadId,
       title: command.title,
@@ -320,6 +334,10 @@ const commandHandlers: CommandHandlerMap = {
       environmentId: command.environmentId,
       runtimeManager: options.runtimeManager,
       workspaceContext: command.workspaceContext,
+    });
+    await options.runtimeManager.releaseThreadFromOtherEnvironments({
+      environmentId: command.environmentId,
+      threadId: command.threadId,
     });
     await entry.runtime.archiveThread({
       threadId: command.threadId,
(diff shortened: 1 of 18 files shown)


patleeman:
I went back through the diff and pulled the non-obvious rationale into this PR discussion.

- The DB change keeps the last provider thread ID across `environment_directory_update` because that event changes the runtime/workspace, not the provider conversation. The old runtime is released before the new workspace resumes that ID.
- The host-daemon handoff uses an async retain/release barrier because a start or submit can be in the gap before runtime turn state becomes visible. Starts and submits also share a thread-level lane across environment IDs; directory switches make environment lanes distinct, and without this ordering two lanes can wait on each other.
- Stop, cancel, rename, and archive use the same ownership handoff because they can otherwise operate on a provider session still resident in the old runtime.
- The fork exception is deliberately narrow: a personal-project fork may reuse the exact unmanaged environment left by its source directory switch. An isolated personal fork gets a personal workspace, and a different unmanaged environment is still rejected.

The regression tests cover the real directory-update path, transcript/provider continuity, cross-environment handoff ordering, exact-source fork reuse, isolated personal forks, and rejection of a different unmanaged environment. Live QA was run against the pre-fix and fixed builds. The PR contains no screenshot artifacts.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I started the security, code quality, and performance reviews. I will also check the test path.

apps/host-daemon/src/runtime-manager.ts
```diff
@@ -340,6 +345,65 @@ export class RuntimeManager {
...
+  }): Promise<void> {
+    await this.enqueueThreadControl(args.threadId, async () => {
+      const inFlightOldCommands = [
+        ...this.inFlightThreadCommandCompletionsByEnvironmentId.entries(),
+      ].flatMap(([environmentId, commandsByThreadId]) =>
+        environmentId === args.environmentId
+          ? []
+          : [...(commandsByThreadId.get(args.threadId) ?? [])],
+      );
+      await Promise.all(inFlightOldCommands);
```
SawyerHood, line 422: 🚨 `slopcop/review` — This wait can cause a deadlock.

The control queue waits for an old command. The old command can then place its release behind this wait. Neither request can finish.

Wait outside the queue. Then enter the queue again to release the old owner. Add a cross-environment control test.

apps/host-daemon/src/command-dispatch.ts
```diff
@@ -257,6 +259,10 @@ const commandHandlers: CommandHandlerMap = {
       command.environmentId,
       options.runtimeManager,
     );
+    await options.runtimeManager.releaseThreadFromOtherEnvironments({
```
SawyerHood: 🚨 `slopcop/review` — These handlers require the target runtime before they release the old owner.

An existing database environment can be absent from the daemon. The command then fails, and the old provider turn continues.

Find and stop the old owner first. Apply the same fix to `thread.plan.cancel`. Add tests for an unloaded target runtime.

packages/db/src/data/events.ts
```diff
@@ -2835,41 +2834,10 @@ export function getLastStoredProviderThreadId(
...
-    latestEnvironmentDirectoryUpdateRow &&
-    (!latestIdentityRow ||
-      latestIdentityRow.sequence < latestEnvironmentDirectoryUpdateRow.sequence)
-  ) {
-    return null;
-  }
-
+  // A directory switch changes the runtime cwd, not the provider thread's
+  // conversation identity. The host daemon releases the old runtime owner
+  // before the next turn is dispatched through `thread/resume` with the new
```
SawyerHood, line 2839: 🚨 `slopcop/review` — This server behavior needs the new daemon ownership controls.

An old daemon accepts the same command but does not release the old provider owner. Two runtimes can then own one provider session.

Increase `HOST_DAEMON_PROTOCOL_VERSION` and update its contract test.

apps/host-daemon/src/runtime-manager.ts
```diff
@@ -340,6 +345,65 @@ export class RuntimeManager {
...
+  }): Promise<void> {
+    const staleEntries = [...this.entries.values()].filter(
+      (entry) =>
+        entry.environmentId !== args.environmentId &&
+        entry.runtime.hasThread(args.threadId),
+    );
+
+    await Promise.all(
+      staleEntries.map((entry) =>
+        entry.runtime.stopThread({ threadId: args.threadId }),
```
SawyerHood, line 448: 🚨 `slopcop/review` — This handoff always calls `stopThread`.

A rename or goal clear after a directory switch can interrupt the active old turn.

Route non-turn controls to the current owner. Transfer ownership only when a new turn starts. Add an active-turn control test.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: The patch keeps one chat when a thread moves to another folder. It is like moving a game without losing your place.

I found four high-severity defects:

- The server needs a host daemon protocol increase for the new ownership rule.
- Stop and Plan cancel can miss the old owner when the target runtime is absent.
- Rename and goal clear can stop an active old turn.
- The new control queue can deadlock across two environments.

I left one inline comment for each defect.

Architecture: The router and runtime manager both keep per-thread promise queues. One runtime-owner service would reduce lock-order errors and duplicate state.

Three parallel GPT-5.6 workers reviewed security, quality, architecture, and performance. One GPT-5.6 reviewer verified the combined findings.

`git diff --check` passed. GitHub reports no pull request checks.

I did not execute the fork code or start a browser test. The environment has no execution sandbox for untrusted code.

PR #1213, opened 2026-08-09
Fix flaky git-source plugin install test timeout on CI

SawyerHood:
## Why

Main CI failed on run [31284464089](https://github.com/get-bb/bb/actions/runs/31284464089). The test `plugin-install.test.ts > git sources > clones a pinned tag into its exact immutable cache dir and loads it` timed out at the 5s default.

Each test in the `git sources` block clones a repo and runs a full install. The install runs git subprocesses, an esbuild build, and a plugin load. The first test in the block also pays the warm-up cost. On a loaded CI runner, this work can take more than 5 seconds.

## Change

Raise the timeout for the `git sources` describe block to 30 seconds. The `npm sources` test in the same file already uses this pattern with a per-test timeout.

## Test

All 29 tests in the file pass locally in 8 seconds.

### DIFF
--- a/apps/server/test/services/plugins/plugin-install.test.ts
+++ b/apps/server/test/services/plugins/plugin-install.test.ts
@@ -279,7 +279,9 @@ describe("plugin install flows", () => {
     await rm(workDir, { recursive: true, force: true });
   });
 
-  describe.skipIf(!hasGit)("git sources", () => {
+  // Each test clones a repo and runs a full install (git subprocesses +
+  // esbuild build + plugin load); the 5s default flakes on loaded CI runners.
+  describe.skipIf(!hasGit)("git sources", { timeout: 30_000 }, () => {
     it("clones a pinned tag into its exact immutable cache dir and loads it", async () => {
       const repoDir = join(workDir, "repo");
       await writePluginFixture(repoDir, { name: "bb-plugin-gitty" });


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull request for security, code quality, performance, test coverage, duplication, and architecture. I will post one final review after the parallel checks finish.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** Some git tests sometimes need more time on a busy CI machine. This change gives them a larger clock, like extra time for a slow race. It does not change the product.

I found no blocking issue.

- Security: The change touches only a test timeout. It does not change git input, plugin trust, production code, or cleanup.
- Code quality: Vitest 4.1.1 supports this suite option. The comment explains the reason, and the scope matches all git-source tests.
- Duplication and architecture: The file has one similar npm timeout. A shared helper would add more code without a clear benefit.
- Performance: The timeout is a limit, not a delay. Successful tests keep their normal speed. A full block stall could add about five minutes to a failed run. I accept that small diagnostic cost.
- End-to-end test: A browser or dev server test does not apply. This pull request changes no route, interface, or production path.

I ran the exact test file at `da5e5b2651290e3c5f804a68eca4b0ddc81ffbcb`. All 29 tests passed in 11.68 seconds. The first git-source test took 1.11 seconds. The server CI test also passed. The macOS package smoke check was still pending when I posted this review.

PR #1214, opened 2026-08-09
Fix mobile remote session lag caused by hidden thread outline refreshes

wjin17:
## Summary

- Only request the full-history conversation outline when the thread TOC is visible.
- Stop compact thread panes from refetching a hidden outline after every appended-event batch.
- Preserve delayed TOC visibility measurement and add compact-pane regression coverage.

## Root cause

The thread TOC is hidden below its 56rem container breakpoint, but its outline query was enabled whenever timeline rows existed. Because realtime `events-appended` notifications invalidate that query, mobile clients repeatedly fetched and rebuilt the full conversation outline even though they could not display it. Remote browser sessions paid both the server projection cost and tunnel latency for every request.

## Performance

I ran a controlled A/B test through the same phone and bb Connect tunnel using two fresh threads per variant. Every run used Claude Code's default model and the same prompt: 20 sequential tool calls followed by a fixed-length response. Both variants stored 110 events across the two runs.

| Metric | Upstream `main` | Patched | Change |
| --- | ---: | ---: | ---: |
| Stored events | 110 | 110 | Equal workload |
| Conversation-outline requests | 208 | 0 | **-100%** |
| Compressed outline traffic | 79,712 bytes | 0 | **-100%** |
| Aggregate outline request time | 3,073.7 ms | 0 | **-100%** |
| Outline request p95 | 24.2 ms | n/a | Eliminated |
| Normal timeline requests | 3 | 3 | Unchanged |

The individual upstream runs made 108 and 100 outline requests; both patched runs made zero. Run duration remained effectively equal because model generation dominates wall time.

During the original remote-session diagnosis, a 39.1-hour traffic capture contained 1,184 outline requests, 4.61 MB of compressed responses, and 62.26 seconds of aggregate origin request time, with a 198.9 ms p95. This change eliminates that request path whenever the TOC is hidden by the compact-pane breakpoint.

## Validation

- `pnpm exec turbo run test --filter=@bb/app --force` — 324 files and 2,447 tests passed
- `pnpm exec turbo run test --filter=@bb/app --force -- src/components/thread/toc/ThreadTableOfContents.test.tsx` — 20 tests passed
- `pnpm exec turbo run lint --filter=@bb/app` — 0 errors
- `git diff --check`

The new regression fails against the previous enablement condition (`enabled: true` at 400px) and passes with this change (`enabled: false`).

### DIFF
--- a/apps/app/src/components/thread/toc/ThreadTableOfContents.tsx
+++ b/apps/app/src/components/thread/toc/ThreadTableOfContents.tsx
@@ -272,7 +272,7 @@ function useConversationTocItems({
   return outlineTocItems ?? timelineTocItems;
 }
 
-function useThreadTocVisible(rootElement: HTMLDivElement | null): boolean {
+function useThreadTocVisible(rootElement: HTMLElement | null): boolean {
   const [visible, setVisible] = useState(
     () => typeof ResizeObserver === "undefined",
   );
@@ -493,20 +493,19 @@ export function ThreadTableOfContents({
   loadOlderTimelineRows,
 }: ThreadTableOfContentsProps) {
   const bottomAnchor = useBottomAnchoredScroll();
-  // The outline is secondary to the latest timeline. Waiting for that first
-  // window prevents a long outline projection from occupying the synchronous
-  // server before the conversation can paint. Do not gate on `tocVisible`: the
-  // minimum-message early return can unmount the root before it is measured.
+  const [rootElement, setRootElement] = useState<HTMLElement | null>(null);
+  const tocVisible = useThreadTocVisible(rootElement);
+  // The full-thread outline exists only for the wide-layout TOC. Waiting for
+  // both its visible container and the first timeline window keeps compact
+  // clients from rebuilding an invisible outline on every appended-event batch.
   const outlineQuery = useThreadConversationOutline(threadId, {
-    enabled: timelineRows.length > 0,
+    enabled: tocVisible && timelineRows.length > 0,
   });
   const senderThreadMetadataById = useSenderThreadMetadataById();
   const { agentItems, userItems } = useConversationTocItems({
     outlineItems: outlineQuery.data?.items,
     timelineRows,
   });
-  const [rootElement, setRootElement] = useState<HTMLDivElement | null>(null);
-  const tocVisible = useThreadTocVisible(rootElement);
   const [open, setOpen] = useState(false);
   const [tab, setTab] = useState<TocTab>("user");
   const [activeUserId, setActiveUserId] = useState<string | null>(null);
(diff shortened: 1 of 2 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

apps/app/src/components/thread/toc/ThreadTableOfContents.tsx
```diff
@@ -493,20 +493,19 @@ export function ThreadTableOfContents({
...
-  // server before the conversation can paint. Do not gate on `tocVisible`: the
-  // minimum-message early return can unmount the root before it is measured.
+  const [rootElement, setRootElement] = useState<HTMLElement | null>(null);
+  const tocVisible = useThreadTocVisible(rootElement);
+  // The full-thread outline exists only for the wide-layout TOC. Waiting for
+  // both its visible container and the first timeline window keeps compact
+  // clients from rebuilding an invisible outline on every appended-event batch.
   const outlineQuery = useThreadConversationOutline(threadId, {
-    enabled: timelineRows.length > 0,
+    enabled: tocVisible && timelineRows.length > 0,
```
SawyerHood, line 518: 🚨 `slopcop/review` — The query gate uses a different width box than CSS.

`host.clientWidth` includes 24 pixels of horizontal padding. The `@container` rule checks the content-box width.

This difference enables the outline query at host widths from 896 through 919 pixels. CSS still hides the table of contents.

I reproduced the issue with a 900-pixel host. Its content width was 876 pixels, and CSS hid the table of contents. However, JavaScript mounted the table-of-contents button.

Realtime events can therefore rebuild a hidden outline in this range. Please measure the content-box width and add boundary tests.

A 900-pixel padded host must disable the query. A 920-pixel padded host must enable the query.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

This patch stops phones from requesting a menu that they cannot show. It works on phones, but it misses a narrow split-pane range.

### Finding

I found one P2 issue. The JavaScript gate and the CSS breakpoint use different width boxes.

JavaScript includes 24 pixels of host padding. CSS checks only the content width.

Host widths from 896 through 919 pixels still request the outline. CSS hides the table of contents in that range.

I reproduced this issue with a 900-pixel host and a 876-pixel content box. CSS hid the table of contents, but JavaScript mounted its button.

Please use the content-box width for the query gate. Add boundary tests for 900 and 920-pixel padded hosts.

### Other checks

The security review found no issue. The codebase search found no duplicate table-of-contents visibility helper.

The focused test passed all 20 tests. The app type check and `git diff --check` also passed.

The browser test made no outline request with a 390-pixel pane. A 1,120-pixel pane made the expected request.

I posted this result as a comment review. I did not approve the pull request or request changes.

SawyerHood:
Addressed the SlopCop finding in 1344b2e45.

The gate now measures the container-query width. `containerInlineSize()` subtracts the computed horizontal padding from `clientWidth`, so the JavaScript check and the `@container scroll-overlay (min-width: 56rem)` rule agree on the same box. The 896–919px host band no longer requests a hidden outline.

Added two boundary tests with a padded host: a 900px host stays disabled, and a 920px host enables the query. The 900px test fails against the previous `clientWidth` comparison.

Validation: `@bb/app` test, typecheck, and lint all pass (324 files, 2,449 tests, 0 lint errors).

PR #1221, opened 2026-08-09
Allow loopback network access inside the workspace sandbox

### LINKED ISSUE #1220: Sandboxed threads lose all bb CLI access during system-initiated turns (child-completed), blocking orchestration
# Sandboxed threads lose all bb CLI access during system-initiated turns (child-completed), blocking orchestration

## Summary

In `auto`/`accept-edits` modes, the workspace sandbox denies all network including loopback. `bb` CLI calls only work through the unsandboxed-retry escape hatch, which the server enables per turn: `allowUnsandboxedCommands` is true only when the turn initiator is the user and the thread has no parent (`resolvePermissionEscalation`, `apps/server/src/services/threads/thread-runtime-config.ts`).

So a parent thread orchestrating children can use `bb` freely in turns the user starts, but in the system turn injected when a child completes — exactly when it needs to dispatch the next child — every `bb` call fails with `EPERM` (connect to `127.0.0.1:<port>` blocked). Child threads in sandboxed modes can never use `bb` at all, which is why the guide recommends `--permission-mode full` for subagents, giving up the sandbox entirely.

The escalation policy itself is sound (agent-triggered turns should not self-escalate out of containment). The problem is that bb's own control plane rides on the same escape hatch: the sandbox has no allowance for the bb server address.

## Reproduce (one thread, two messages)

1. Start a thread in `auto` permission mode and send:
   "Run `bb status` and tell me if it worked. Then run `bb thread spawn --parent-self --project <this project id> --prompt 'Reply done and finish.'`. When you get the completion notification for that child, immediately run `bb status` again and report the result."
2. Both `bb` calls in this turn succeed (user-initiated turn, `allowUnsandboxedCommands: true`).
3. The child finishes; bb injects a child-completed system turn into the parent. The `bb status` run in that turn fails: `fetch failed` / `Cannot reach bb at http://127.0.0.1:<port>: EPERM — the connection was blocked.`

Verification: `ps aux | grep claude` shows the parent's session process is replaced at the system turn with `"allowUnsandboxedCommands":false` in its `--settings` (was `true` during the user turn — the escalation flip is an execution-settings change, so the session is rebuilt).

## Expected

`bb` CLI reaches the bb server in every turn of every thread, regardless of turn initiator or sandbox mode. Sandboxing should contain the workspace and the open network, not the agent's own control plane.

## Suggested direction

Add a sandbox network allowance for loopback in `buildWorkspaceWriteSandbox` (`packages/agent-runtime/src/claude-code/bridge/session-options.ts`): the SDK sandbox supports `network.allowLocalBinding`, which on macOS emits `(allow network-outbound (remote ip "localhost:*"))` — sandboxed commands can then reach the bb server directly, with no unsandboxed escalation involved. This keeps escalation `deny` semantics for everything else and makes sandboxed child threads first-class `bb` citizens.

sholub-dev:
## Summary

Fixes #1220.

The workspace sandbox (`auto`/`accept-edits` modes) denies all network including loopback, so `bb` CLI calls from sandboxed shells only ever worked through the unsandboxed-command escape hatch (`allowUnsandboxedCommands`, enabled only when permission escalation is `ask`). Turns with escalation `deny` — system-initiated turns such as child-completed notifications, and all child threads — have no retry path, so every `bb` call fails with `EPERM`. A parent thread orchestrating children loses bb access in exactly the turns that exist so it can dispatch the next child.

This passes `network.allowLocalBinding: true` in the sandbox config, which permits direct loopback connections (on macOS the profile emits `(allow network-outbound (remote ip "localhost:*"))`). Sandboxed commands reach the bb server in every turn without involving unsandboxed escalation; escalation semantics for everything else are unchanged.

## Notes

- `allowLocalBinding` also permits binding local ports (its primary documented purpose, e.g. dev servers), so this loosens the sandbox from "no network" to "loopback + local binding, nothing external". The SDK exposes no loopback allowance scoped to a single port; this is the narrowest available knob and is far tighter than the previous workaround of spawning subagents with `--permission-mode full` to keep bb reachable.
- On Linux the sandbox uses a network namespace, where host loopback is unreachable regardless; this change is a no-op there.
- No `HOST_DAEMON_PROTOCOL_VERSION` bump: the change is local to the claude-code bridge session options; nothing on the server/daemon wire changes.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull request for security, code quality, performance, architecture, and practical test coverage.

packages/agent-runtime/src/claude-code/bridge/session-options.ts
```diff
@@ -176,6 +176,10 @@ function buildWorkspaceWriteSandbox(
     failIfUnavailable: false,
     autoAllowBashIfSandboxed: true,
     allowUnsandboxedCommands: params.permissionEscalation === "ask",
+    // Loopback must stay reachable from inside the sandbox: the bb CLI talks
+    // to the local bb server, and escalation-denied turns (system-initiated
+    // turns, child threads) have no unsandboxed-retry path around a block.
+    network: { allowLocalBinding: true },
```
SawyerHood, line 183: 🚨 `slopcop/review` — **High: loopback access lets sandboxed code start an unsandboxed terminal through BB's unauthenticated API.**

The `/api/v1/*` routes have no authentication check. `POST /api/v1/terminals` accepts a command and asks the host daemon to start it outside this Claude sandbox. After this line, untrusted workspace content can call `bb terminal create` or this route during an escalation-denied turn. This bypasses both `allowUnsandboxedCommands: false` and the workspace write boundary. Please authenticate the local API with a session token before you enable loopback access, or use a narrower authenticated transport.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** This PR lets sandboxed Claude commands reach local BB. The fix works on macOS, but it also gives those commands an unchecked path to start unsandboxed terminals.

I found one blocking security issue.

1. **High — the BB API becomes a sandbox escape path.** The local `/api/v1/*` routes have no authentication check. `POST /api/v1/terminals` starts a host terminal outside the Claude sandbox. A command influenced by untrusted repository content can use this path during an escalation-denied turn. The request bypasses both `allowUnsandboxedCommands: false` and the workspace write boundary. Add per-session API authentication before this network grant, or use a narrower authenticated transport.

Other findings:

- **Medium — the grant is wider than the PR description.** Claude 2.1.226 emits rules for outbound access to every localhost port. It also permits bind and inbound access on every interface. A live macOS sandbox probe reached the BB server. It also reached a sandboxed listener through the machine's LAN address. External outbound access stayed blocked.
- **Medium — the agent guidance stays stale.** The BB CLI skill and thread guide still tell subagents to use Full Access. That instruction bypasses the sandbox which this PR aims to make useful. Related memory, secrets, and plugin-authoring documents still say loopback needs Full Access.
- **Medium — Linux keeps the original problem.** The current Claude Linux sandbox path ignores `allowLocalBinding`. The PR states this limit, but the code comment reads as platform-neutral.
- **Low — the flag changes Java behavior.** Claude injects `JAVA_TOOL_OPTIONS=-Djava.net.preferIPv4Stack=true` into macOS sandboxed commands when this flag applies.

The implementation has no meaningful performance cost. It adds a constant session option and starts no new proxy or process. The sandbox construction has one owner, so I found no duplicate code to refactor. A path-scoped Unix socket could provide a narrower future transport.

Validation results:

- `@bb/agent-runtime` typecheck passed.
- The focused bridge suite passed all 62 tests.
- The full package suite passed 837 of 838 tests. One unchanged process lifecycle test failed again when run alone.
- A live macOS `sandbox-exec` probe confirmed IPv4 and IPv6 loopback access after the new rule.
- A browser test did not apply because this change has no UI route. The sandbox probe tested the changed boundary directly.
- No daemon protocol version change is necessary because this option stays inside the Claude bridge.

I would not merge this until the local API cannot turn loopback access into unsandboxed command execution.

sholub-dev:
Addressed the Mediums: the sandbox comment now states the grant is macOS-only and coarse (Linux --unshare-net ignores the flag), the stale loopback/Full-Access caveats in the plugin-authoring, memory, and secrets docs are updated, and the subagent guidance now defaults to inheriting the parent's permission mode.

On the High: /api/v1 being unauthenticated is pre-existing and deliberate (trusted local machine, loopback bind by default). A session token cannot create the boundary described here — the bb CLI itself runs inside the sandbox in escalation-denied turns, so any credential it can read is equally readable by every other sandboxed command. The real fix is a capability-scoped transport for sandboxed callers (e.g. a Unix socket listener mounting only agent-safe routes via allowUnixSockets), which deserves its own design and PR.

SawyerHood:
Yeah I agree on the high one! Slopcop is being picky

SawyerHood:
I'll merge when ci passes! TY for the contribution

PR #1224, opened 2026-08-09
fix(agent-runtime,thread-view): stop timeline 500s from session-scoped ACP fs-write ids

vburojevic:
## Problem

Threads on ACP providers (reproduced with acp-kimi) intermittently fail `GET /api/v1/threads/:id/timeline` with a 500, taking down the whole chat view (`Cannot merge file-edit messages with different scopes`).

Root cause: the ACP adapter minted `fileChange` item ids from a per-session counter (`acp-fs-write-N`, `packages/agent-runtime/src/acp/adapter.ts`). Resumed ACP sessions restart the counter, so a later turn reuses an item id already persisted in an earlier turn. The timeline projection merges file-edit rows by call id (`upsertFileEdit` in `packages/thread-view/src/operation-projection.ts`) and threw on the cross-turn scope mismatch, 500ing every timeline request for the thread from then on. Every resumed session that wrote files created a fresh collision, so affected threads kept re-breaking.

## Fix (two layers)

- **Adapter:** mint `acp-fs-write-<turnId>-<counter>`. Turn ids carry a per-adapter-instance random prefix, so ids are unique across sessions for good.
- **Projection:** `upsertFileEdit` now partitions existing rows by scope — it merges only compatible-scope rows, preserves foreign-scope rows as their own messages, and scope-qualifies message keys on collision. Any reused call id (including rows already persisted by older builds) degrades to separate file cards instead of failing the projection. The strict throw in `updateFileEditMessage` stays as an invariant guard, now unreachable from `upsertFileEdit`.

## Tests

- `acp/adapter.test.ts`: two adapter instances (simulating a resumed session) mint distinct fs-write ids.
- `build-thread-timeline.test.ts`: two turns reusing the same `fileChange` item id produce two distinct file-change rows, each keeping its own diff.

Verified: full `@bb/agent-runtime` and `@bb/thread-view` vitest suites green, `tsc --noEmit` clean in both packages. Also repaired the affected production threads in a live `bb.db` with this recipe (kept earliest turn, suffixed later-turn duplicates in `item_id` and `data->'$.item.id'`); all previously-500ing threads load again.

## Release note

After merge, cut 0.36.1 per the usual flow: `node scripts/bump-version.mjs --patch` → "Prepare bb-app 0.36.1" PR → dispatch `publish-bb-app.yml` with `npm_tag=latest`, `dry_run=false`.


agustif (approved):
test

agustif (commented):
Note: earlier empty/test APPROVE was accidental from a scout dry-run of the gh CLI APIs — please ignore. Not a real review of this PR.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull request for security, code quality, performance, tests, and architecture.

I will post the combined findings after the parallel checks finish.

packages/thread-view/src/operation-projection.ts
```diff
@@ -615,6 +643,20 @@ export function upsertFileEdit(
     ? eventProjectionMessageTurnScopeFields(turnId)
     : eventProjectionMessageThreadScopeFields();
   const existingRows = state.fileEditsByCallId.get(partial.callId) ?? [];
+  // Providers can reuse call ids across scopes (e.g. resumed ACP sessions
+  // restart their synthetic id counters). Merge only rows from a compatible
+  // scope and leave foreign-scope rows untouched, so each scope keeps its own
+  // file-edit message instead of failing the whole projection.
+  const compatibleRows: EventProjectionFileEditMessage[] = [];
```
SawyerHood, line 683: 🚨 `slopcop/review` — High: scope whole-item windowing by turn

This fallback handles reused IDs only after the server selects a timeline window. The server still groups item spans and lifecycle rows by raw `itemId`. A legacy ID reused across turns becomes one thread-wide item.

With 39 events and a budget of 10, the latest page backfilled both turns. The next older page returned neither file row. Larger histories can exceed the 1,500-event budget and restore the expensive projection path.

Please use `(scopeKind, turnId, itemId)` through whole-item closure and its database queries. Add a server test with the reused ID on opposite sides of the event cut.

packages/agent-runtime/src/acp/adapter.ts
```diff
@@ -1168,7 +1168,9 @@ export function createAcpProviderAdapter(
           scope: turnScope(turnId),
           item: {
             type: "fileChange",
-            id: `acp-fs-write-${state.fsWriteCounter}`,
+            // Include the turn id: resumed sessions restart the counter, so a
+            // bare counter would reuse ids already persisted in earlier turns.
+            id: `acp-fs-write-${turnId}-${state.fsWriteCounter}`,
```
SawyerHood, line 1173: 🚨 `slopcop/review` — Medium: update the daemon protocol for this wire change

This changes an event item ID that the host daemon sends to the server. `HOST_DAEMON_PROTOCOL_VERSION` remains 86, so enrolled daemons will not update. They will continue to emit the legacy IDs that this PR tries to stop.

Please increment the protocol version and update its contract expectation.

packages/thread-view/src/operation-projection.ts
```diff
@@ -615,6 +643,20 @@ export function upsertFileEdit(
...
+  for (const row of existingRows) {
+    if (haveCompatibleEventProjectionMessageScope(row, scopeFields)) {
+      compatibleRows.push(row);
+    } else {
+      foreignRows.push(row);
+    }
+  }
+  const foreignMessageIds = new Set(foreignRows.map((row) => row.id));
   const stdoutBuffer =
     state.fileEditStdoutBuffersByCallId.get(partial.callId) ??
```
SawyerHood: 🚨 `slopcop/review` — Medium: scope the file-edit output buffer too

The new row logic separates reused call IDs by turn, but this buffer still uses only `callId`. Two pending file-edit output streams with the same ID therefore share text across turns. The flush path then writes that combined text into every row for the ID.

A focused reproduction made both rows contain `one-outputtwo-output`. Please key buffers by the scoped call identity. Add an output-delta regression test that keeps each turn’s output separate.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: The app gave two file changes the same name. This patch separates them. Some shelves still use the old name, so files can mix or disappear.

I found three actionable problems.

1. High: whole-item timeline closure ignores the turn scope. Legacy IDs can exceed the event budget and make older rows disappear.
2. Medium: the host-daemon wire behavior changed without a protocol update. Enrolled daemons will continue to emit legacy IDs.
3. Medium: file-edit output buffers still use only the call ID. Reused IDs can combine output from different turns.

The security review found no issue. The architecture review recommends one scoped item identity for closure queries, projection maps, and output buffers.

The focused suite passed all 123 tests. Both affected package type checks passed. The complete thread-view suite passed all 349 tests.

The agent-runtime suite passed 838 of 839 tests. The unrelated stderr-tail test failed twice, while GitHub package checks passed.

A browser test was not practical for this historical ACP session path. An in-memory server test used the real database and timeline service.

The GPT-5.6 review gate confirmed all three findings. I posted this review as comment-only, as required.

SawyerHood:
Sorry about that @vburojevic I just asked the review bot to fix and test my b!

PR #1225, opened 2026-08-09
Fix icon appearance in dark mode: macOS Dock, iOS home screen, notification badges

vburojevic:
## What happened

With macOS in dark mode, the Dock icon renders correctly while bb is closed, then flips to the light rendering (black glyph on a white tile) the moment the app launches.

The same family of bug exists on the web-app side: the iOS home-screen icon (installed PWA) is a glyph on transparency, which iOS renders on black and can't derive dark/tinted home-screen treatments from, and the manifests ship no `purpose: "monochrome"` icons, so platform tinting and Web Push notification badges have no alpha mask to work with.

## What I expected

The Dock icon should keep the system's dark-mode rendering while the app is running, and installed-app icons / notification badges should participate in each platform's appearance treatment.

## Root cause (Dock)

`apps/desktop/src/main.ts` calls `app.dock.setIcon(iconPath)` on every launch. That replaces the bundle's `.icns` with a raw `NSImage`, and programmatically set dock images bypass macOS's appearance pipeline entirely — the system gives them the default light treatment.

The call is redundant for packaged builds: `run-electron-builder.mjs` already bakes the channel-correct icon (`icon.icns` / `icon-nightly.icns`) into the bundle. Its only real purpose is dev runs, where the unpackaged Electron binary needs `icon-dev.png` to be distinguishable.

## Changes

- **Desktop:** only call `dock.setIcon` when `!isPackaged`. Packaged builds keep the bundle icon and the system's dark rendering.
- **Web app:** `apple-touch-icon*.png` are now full-bleed opaque (glyph on white tile, matching the maskable design); the generator tints them in tile mode so color variants keep the glyph gradient. Added `purpose: "monochrome"` white-on-transparent alpha-mask icons (derived from glyph darkness) to all manifests — these enable Android themed-icon tinting and give Web Push service workers a correct `badge` asset (a full-color icon there renders as a solid silhouette blob).

## Verification

- A/B build: identical source packaged with and without the guard, both running simultaneously in a dark-mode Dock — the unguarded build reproduces the white tile, the guarded one keeps the system dark rendering.
- `@bb/desktop`: typecheck + 231 tests pass. `@bb/app`: typecheck + 2446 tests pass, including the `generate-pwa-icons --check` gate (all color variants and manifests regenerated via the script).
- Monochrome asset verified as a clean alpha mask (white bb glyph) composited on a dark background.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull request for security, code quality, performance, architecture, and end-to-end behavior. I will post one final review after the parallel checks finish.

apps/app/scripts/generate-pwa-icons.mjs
```diff
@@ -24,7 +24,19 @@ const icons = [
   { file: "icon-512.png", mode: "tile" },
   { file: "icon-192-maskable.png", mode: "tile" },
   { file: "icon-512-maskable.png", mode: "tile" },
-  { file: "apple-touch-icon.png", mode: "glyph" },
+  // Opaque white tile: iOS renders transparency in touch icons as black,
+  // and the system's dark/tinted home-screen treatments need a full-bleed
+  // opaque source.
+  { file: "apple-touch-icon.png", mode: "tile" },
```
SawyerHood: 🚨 `slopcop/review` — The `@bb/web` touch icon remains transparent

`apps/web/src/routes/__root.tsx` still serves `apps/web/public/apple-touch-icon.png`. This generator updates only `apps/app/public`.

The old web icon still has transparent pixels, so its iOS home-screen fault remains. Generate both icons from one source.

Add a check that confirms that each touch-icon pixel is opaque.

SawyerHood: 🚨 `slopcop/review` — The new mode makes the glyph path unreachable

Every icon now uses `tile` mode. Thus, `tintGlyphIcon` and the `glyph` branch cannot run.

Remove the unused function and the unused mode branch.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: This PR gives the mobile icons a solid background. It also lets macOS keep the correct packaged Dock icon after launch.

I found one P2 issue and one non-blocking cleanup issue.

- P2: `@bb/web` still serves its old transparent Apple touch icon. This leaves one iOS home-screen surface unfixed.

- P3: All generated icons now use `tile` mode. The old `glyph` function and branch can no longer run.

I found no security or privacy issue. I found no important performance regression.

The P2 should receive a fix before merge. The generator should use one source for both Apple touch icons.

Checks completed:

- The exact head SHA and base SHA matched the pull request.

- `git diff --check` and the generated-asset check passed.

- Turbo typechecks passed for `@bb/app` and `@bb/desktop`.

- All 2,446 app tests and all 231 desktop tests passed.

- Chromium loaded all nine manifests and all icon URLs successfully.

- The default and blue Apple touch icons were 180×180 pixels and fully opaque.

- The monochrome icons had valid alpha masks.

I did not repeat the packaged macOS Dock visual test. That check needs a packaged app and a direct Dock comparison.

I posted this result as a comment review. I did not approve the pull request or request changes.

PR #1230, opened 2026-08-09
fix(agent-runtime): recover when ACP steer has no active turn

agustif:
## Summary
- Treat ACP bridge `No active turn to steer` / `No active ACP session` rejections as **stale** so host-daemon falls back to a new turn instead of failing `turn.submit` (seen on Grok/ACP harness threads).
- Emit `acp/turn/completed` when `session/prompt` fails so the server turn does not stay open after the bridge prompt ends.

## Test plan
- [x] `vitest` `@bb/agent-runtime` lifecycle + ACP bridge tests (79 passed)
- [ ] Smoke: on an `acp-grok` thread mid-turn, send a follow-up after the prompt has ended; message should start a new turn rather than `Command turn.submit failed`


ymichael:
superseded by #1427

PR #1232, opened 2026-08-09
fix(cursor): use namespaced agent executable

### LINKED ISSUE #1231: Cursor model discovery can launch the wrong agent executable
## Workflow

1. Install and authenticate Cursor's agent CLI.
2. Have another tool provide an `agent` executable earlier on `PATH` (for example, Grok's agent CLI).
3. Open BB's new-thread composer and select Cursor.
4. Open the model picker.

## What happened

BB launches the generic `agent` executable for Cursor model discovery and ACP sessions. If another provider owns that name earlier on `PATH`, BB launches the wrong CLI. Cursor model discovery then fails and the picker falls back to only `Agent default`.

Provider health and update actions use the same ambiguous executable, so they can also inspect or update the wrong tool.

## What I expected

BB should use Cursor's namespaced `cursor-agent` executable consistently for model discovery, ACP sessions, health checks, updates, and login guidance. The model picker should then show the catalog returned by `cursor-agent --list-models`.

## Environment evidence

On the affected machine:

- `agent` resolved to Grok's CLI and rejected Cursor's `--list-models` behavior.
- `cursor-agent --list-models` returned Cursor's full model catalog.
- Pointing BB's built-in Cursor profile at `cursor-agent` restored the model list in the dev app.

This is distinct from #933: the models are not merely ordered under “More models”; the catalog command is sent to the wrong executable and no Cursor catalog is loaded.

DevVig:
## Summary

- use Cursor's namespaced `cursor-agent` executable for ACP launch, model discovery, provider health/update actions, and login guidance
- bump the host-daemon protocol to 88 so enrolled daemons refresh their built-in Cursor profile
- update focused tests and the packaged-app provider documentation

## Why

BB currently resolves the generic `agent` command for Cursor. When another provider owns that name earlier on `PATH`, BB launches the wrong CLI, model discovery fails, and the Cursor picker collapses to `Agent default`.

Cursor's installer creates both `agent` and `cursor-agent` symlinks to the same executable. Using the namespaced alias keeps Cursor behavior intact while preventing collisions with other agent CLIs.

Closes #1231.

## Validation

- `pnpm exec turbo run build typecheck lint --filter=@bb/agent-runtime --filter=@bb/host-daemon --filter=@bb/server --filter=@bb/host-daemon-contract`
- agent-runtime focused tests: 54 passed
- host-daemon provider health tests: 16 passed
- host-daemon contract tests: 35 passed
- server onboarding tests: 2 passed
- `git diff --check`
- dev app: opened the Cursor model picker and verified Auto, Cursor Grok 4.5, GPT-5.6 Sol, Opus 5, Fable 5, Composer 2.5, and More models load from the live Cursor catalog

## Risk and rollout

No UI or persisted-settings schema changes. Existing threads are untouched; new Cursor runtimes use the updated built-in profile after the daemon protocol upgrade. If `cursor-agent` is missing, the existing provider-health error and install flow remain available instead of silently launching an unrelated `agent` binary.

### DIFF
--- a/apps/host-daemon/src/provider-cli-health.test.ts
+++ b/apps/host-daemon/src/provider-cli-health.test.ts
@@ -216,7 +216,7 @@ const CLAUDE_CODE_DEFINITION: ProviderCliDefinition = {
 const CURSOR_DEFINITION: ProviderCliDefinition = {
   key: "cursor",
   displayName: "Cursor",
-  executableName: "agent",
+  executableName: "cursor-agent",
   npmPackageName: null,
   minimumSupportedVersion: null,
   installCommand: {
@@ -225,8 +225,8 @@ const CURSOR_DEFINITION: ProviderCliDefinition = {
   },
   updateCommand: {
     commandKind: "exec",
-    displayCommand: "agent update",
-    command: "agent",
+    displayCommand: "cursor-agent update",
+    command: "cursor-agent",
     args: ["update"],
   },
 };
@@ -280,8 +280,12 @@ function installMissingClaudeCommands(
 function installMissingCursorCommands(
   runner: FakeProviderCliCommandRunner,
 ): void {
-  runner.setExit("which", ["agent"], 1, "agent not found");
-  runner.setSpawnError("agent", ["--version"], "spawn agent ENOENT");
+  runner.setExit("which", ["cursor-agent"], 1, "cursor-agent not found");
+  runner.setSpawnError(
+    "cursor-agent",
+    ["--version"],
+    "spawn cursor-agent ENOENT",
+  );
   runner.setSuccess("npm", ["prefix", "-g"], "/usr/local\n");
 }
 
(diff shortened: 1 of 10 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I started this review under the `review` rule.

I will check security, code quality, performance, architecture, and practical behavior.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** BB used a generic `agent` command that could start the wrong tool. This change uses Cursor's specific `cursor-agent` command.

I found no blocking issues.

The rename covers session launch, model discovery, health checks, updates, onboarding, and documentation. The protocol increase to version 88 correctly forces old daemons to update.

Security review found no new unsafe command construction or trust-boundary change. The namespaced command reduces the risk that BB starts another provider's generic `agent` executable.

Performance review found no new process, network, disk, startup, or cache work. The daemon update cost is necessary for the changed local profile.

Validation results:

- All required GitHub checks pass.
- The local `agent` link points to Grok, while `cursor-agent` points to Cursor. This result confirms the reported collision.
- The installed `cursor-agent` returns a version and a valid model list.
- Cursor documents [`cursor-agent login` and `cursor-agent update`](https://docs.cursor.com/en/cli/reference/parameters).
- Cursor keeps `cursor-agent` as a supported namespaced alias in its [2026 CLI change](https://cursor.com/changelog/cli-jan-08-2026).

I found two low-severity test gaps. Direct tests do not assert `cursor-agent update` or `cursor-agent login`. Several test and story fixtures also keep the old `agent` name for Cursor.

I also checked for duplicate metadata and a useful refactor. The current copies serve separate daemon, runtime, and server contracts. A shared constant would cross those boundaries and would expand this small fix.

I did not run the fork code or a browser end-to-end test. This change has no browser route, and a valid end-to-end test needs an authenticated Cursor session.

I left no inline comments because I found no concrete defect on a changed line.

SawyerHood:
TY for fixing this!

PR #1233, opened 2026-08-09
Reduce the workspace checkout preload below 500 KB

### LINKED ISSUE #1072: Thread route still statically imports the 2.15 MB shared chunk (+ katex): FilePreview → @pierre/diffs, markdown-preview → rehype-katex, PromptBox → tiptap should be lazy
## Problem

After #1070 removed Shiki from the critical path, `workspace-checkout-display` is still **1.7MB** and preloaded via `<link rel="modulepreload">` in the HTML shell. This means every thread page load pulls in the full code editor before rendering.

### What's in it

```
workspace-checkout-display: 1.7MB
  ├── Monaco/Eddie editor (~800KB+)
  ├── Cytoscape (graph visualization) (~425KB)
  ├── Katex (math rendering) (~253KB)
  └── 30+ utility chunks bundled together
```

### Import chain

`index.js` eagerly includes modules that import `formatWorkspaceCheckoutDisplay` from `@/lib/workspace-checkout-display`. Vite sees this as a static dependency and bundles the entire editor tree into the preload set.

Files that pull it in:
- `ThreadDetailView.tsx` — imports `formatWorkspaceCheckoutDisplay`
- `ThreadMetadataContent.tsx` — imports `formatWorkspaceCheckoutDisplay`
- `PluginThreadChat.tsx` — imports `formatWorkspaceCheckoutDisplay`
- `ThreadEnvironmentSummary.tsx` — imports the type

### Fix

1. **Split `@/lib/workspace-checkout-display`** into a light type-only module and a heavy implementation module. The type is used everywhere; the implementation (which pulls in the editor) should be lazy-loaded.

2. **Lazy-load the actual editor** — `ThreadDetailView`, `ThreadMetadataContent`, and `PluginThreadChat` only need `formatWorkspaceCheckoutDisplay` for rendering a small environment label. That function shouldn't pull in the full editor.

3. **Audit why Cytoscape and Katex are bundled here** — they should be in their own lazy chunks.

### Target

- `workspace-checkout-display` under 500KB (from current 1.7MB)
- No editor code in HTML modulepreload
- Thread view first paint: no editor parsing on critical path

Related: #1063, #1070

SawyerHood:
## Results

The raw `workspace-checkout-display` chunk fell from 2,147,954 bytes to 323,524 bytes. After the rebase, the production build reports 316.4 KB raw, below the issue's 500 KB target.

The browser transfer fell from 649,358 bytes to 102,973 bytes. First paint occurred at 128 ms. The checkout chunk started at 603 ms. The browser requested no prompt, diff, or file-editor chunk during the measured initial load.

## Root cause and fix

The issue hypothesis was partly wrong. `workspace-checkout-display.ts` did not import an editor. Rolldown named a shared 713-module chunk after that small helper. Static route imports pulled in the worker pool, timeline rows, file preview, and diff renderer.

This change adds lazy boundaries for the prompt editor, timeline rows, file preview, diff renderer, diff panel, and worker-pool provider. It also moves shared constants into small leaf modules so route code does not cross heavy component seams.

## Guardrails

The bundle check now measures the exact checkout chunk and the workspace route's static chunk closure. `bundle-budget.json` adds a 500 KB per-chunk cap and `forbiddenWorkspaceRoutePackages` for editor and renderer packages.

I rewrote the file's rationale comment because it previously said there was deliberately no per-chunk limit. The checkout chunk now needs a focused limit because its shared ownership hid this regression.

## Rebase

The rebase resolved a conflict with the iPadOS Magic Keyboard work in `PromptBoxInternal.tsx`. This change renames that implementation to `PromptBoxInternalImpl.tsx`, and the merged implementation preserves the iPadOS behavior and tests.

## Validation

- Production app build
- Bundle budget check
- App typecheck
- App lint
- 324 app test files and 2,444 tests
- Cold production browser measurement

Fixes #1072


SawyerHood:
Closing this for now. We cannot take this change yet.

Nothing here is wrong. The branch is kept, CI was green, and the measurements
hold. This is a scheduling decision, not a rejection.

Recorded so the next attempt does not repeat the investigation:

- The result: the checkout chunk went from 2,147,954 bytes to 323,524 bytes raw
  (316.4 KB after the rebase), against the 500 KB target in #1072. Browser
  transfer dropped from 649,358 to 102,973 bytes.
- The issue hypothesis was partly wrong. `workspace-checkout-display.ts` never
  imported an editor. Rolldown named a shared 713-module chunk after that small
  helper. The real weight came from static route imports pulling in the worker
  pool, timeline rows, file preview, and diff renderer.
- The fix adds lazy boundaries at those four points, plus a per-chunk cap and
  `forbiddenWorkspaceRoutePackages` in `bundle-budget.json` to stop a
  regression.
- The rebase resolved a conflict with the iPadOS Magic Keyboard work in
  `PromptBoxInternal.tsx`, which this branch renames to
  `PromptBoxInternalImpl.tsx`. The iPadOS behavior is preserved; the only
  difference from `main` is the extraction of two helpers into
  `prompt-box-runtime.ts`.

#1072 stays open. Branch: `bb/fix-1072-workspace-checkout-display-1-7mb-preloa-thr_k86mrargnh`.

PR #1236, opened 2026-08-09
Apply Claude turn settings without replacing live sessions

### LINKED ISSUE #1235: Background tasks are killed when user and system turns alternate
**Scope:** Claude Code threads only. Workflows, backgrounded commands, and background agents live inside the Claude Agent SDK session; no other provider hosts background tasks, so no other provider is affected.

`permissionEscalation` follows the turn initiator: turns started by the user resolve to "ask", turns started by the system (e.g. a child thread's completion report) resolve to "deny". The runtime treats escalation as a session-level setting, so when consecutive turns have different initiators it sees a settings change and replaces the provider session. Replacing the Claude session settles every background task inside it as interrupted.

**Impact if not fixed:** background work cannot survive orchestration. A manager thread loses all running background work the moment any child reports back, so the threads most likely to use background tasks are the ones guaranteed to lose them.

**Steps to reproduce (Claude Code provider):**
1. Create a thread in Approve for me (auto) permission mode.
2. Ask the agent to start a long-running background task (a workflow, or a backgrounded command like `sleep 300`).
3. Ask the agent to spawn a child thread for any small task.
4. Wait for the child thread to finish. Its completion report starts a system-initiated turn on the parent.
5. **Expected:** the parent's background task keeps running. **Actual:** as soon as the report arrives, the task is marked interrupted and its process is gone.

Note: the repro needs a top-level (non-child) thread — child threads run every turn with "deny" and never flip; the flip happens on a parent that mixes user turns ("ask") with system-initiated turns ("deny").

sholub-dev:
## Summary

Fixes #1235.

Claude background commands, workflows, and agents all live inside one Agent SDK session. bb previously treated execution-option drift as construction-time configuration, dispatched `thread/resume`, and replaced that session. Alternating user/system turns therefore interrupted background work, and several other setting changes had the same destructive behavior.

This PR makes reconfiguration provider-specific. Claude settings that can safely change in place now travel with each turn; only actual session-construction changes rebuild the provider session.

## What changes

- Permission escalation is live per input. A steer adopts its new escalation only after the input is accepted, and prompt/subagent/tool tracking preserves the originating turn's policy for background work.
- Claude background subagents inherit approval provenance from their parent `Agent`/legacy `Task` tool chain, including SDK callbacks that omit agent metadata. Permission-request hooks provide an additional tool-level provenance boundary.
- Approval-policy ownership is explicit on provider adapters. Claude filters allow/ask/deny using the originating work before forwarding an approval, so the runtime no longer reclassifies that request against newer thread settings. Codex, ACP, and Pi retain runtime-owned approval filtering.
- Model changes use the SDK's live `setModel` control.
- Memory, workflow, and reasoning changes use `applyFlagSettings`, including transitions to and from `max` reasoning.
- Claude's native `Agent`/legacy `Task` and `Workflow` toggles are enforced live. They no longer depend on stale construction-time `disallowedTools`, and disabling them rejects new launches without replacing the session or killing existing work.
- Unsupported Claude `serviceTier: "fast"` is normalized to `"default"` before validation, comparison, persistence, and provider command construction, so it cannot cause a rebuild or reach the Claude bridge.
- Provider adapters now classify setting changes as unchanged, live, or session-scoped. Codex, ACP, and Pi retain their existing session-rebuild behavior.

The permission path remains conservative: readonly enforcement resolves escalation at tool-call time, unsandboxed retries are automatically denied on escalation-denied turns, and that denial runs before cached session grants so an approval from an earlier user turn cannot leak into later system work.

The following remain session-scoped and still rebuild when changed:

- Claude Plan mode override
- Permission mode, workspace scope, or approval reviewer
- Mock Claude CLI traffic configuration

`HOST_DAEMON_PROTOCOL_VERSION` is bumped to 99 because the server no longer sends construction-time Claude `disallowedTools` values for workflow/subagent settings. Version 98 is already used by qualified upstream PR lookup on the updated base.

## Validation

- Standalone QA started a real Claude background process, then changed memory, workflow, and subagent settings while switching Haiku/low to Opus 1M/max. The same bridge PID, Claude CLI PID, provider thread ID, and background process survived; the next turn completed with the new model.
- A second real Claude run used `serviceTier: "fast"` on both start and agent-origin steer. It retained one provider thread, preserved the background child's original `ask` policy after the steer changed the current thread to `deny`, surfaced the dangerous write as a pending approval, and completed it after `allow_once`.
- Affected package typechecks pass for `@bb/agent-runtime`, `@bb/server`, `@bb/host-daemon`, and `@bb/host-daemon-contract`.
- Focused execution-option, runtime lifecycle, interactive-request, and Claude bridge coverage: 115 tests passed.
- Full server suite: 1,395 tests passed.
- Host-daemon contract suite: 49 tests passed.
- Full agent-runtime suite: 871/872 passed locally; the only failure is the unchanged timing-sensitive `bounds provider stderr while data arrives without a newline` lifecycle test, which is outside this diff.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull request for security, code quality, architecture, performance, and end-to-end behavior.

packages/agent-runtime/src/claude-code/bridge/bridge.ts
```diff
@@ -1432,6 +1473,7 @@ async function handleTurnStart(
     return;
   }
 
+  threadSession.permissionEscalation = params.permissionEscalation;
```
SawyerHood, line 1978: 🚨 `slopcop/review` — A shared session value applies the newest turn policy to all background work.

Background tasks outlive their start turn. A later turn overwrites this shared value. An older system task can then read `ask` from a user turn and enter the approval path. The reverse change can also deny older user work. Please attach the policy to the work that created each tool request. Add overlap tests for both policy directions.

packages/agent-runtime/src/claude-code/bridge/bridge.ts
```diff
@@ -1452,6 +1494,7 @@ async function handleTurnSteer(
...
```
SawyerHood: 🚨 `slopcop/review` — A failed steer leaves the new permission policy active.

This assignment occurs before `pushInput` confirms the steer. The error path returns an error without restoring the old value. A rejected `ask` steer can therefore loosen the policy for existing `deny` work. Update the value after acceptance, or restore it in the error path. Add a failed-steer test.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: This change keeps old Claude jobs alive. But it gives all jobs one permission switch. It is like one light switch for every room.

I found two defects. I recommend fixes before merge.

1. **High — Background work crosses the per-turn permission boundary.**

   Background tasks outlive their start turn. Each new turn overwrites one session-wide `permissionEscalation` value. An older system task can later read `ask` from a user turn and enter the approval path. The reverse change can deny older user work. Attach the policy to the work that created each tool request. Add overlap tests for both policy directions.

2. **Medium — A failed steer keeps its permission change.**

   The bridge changes the value before `pushInput` succeeds. The error path does not restore the old value. A rejected `ask` steer can loosen existing `deny` work. Update the value after acceptance, or restore it after failure.

**Architecture and performance**

The adapter capability is an acceptable abstraction. The two Claude permission checks serve different SDK enforcement points. I found no useful refactor requirement. The change removes session replacement and adds constant-time state access.

**Validation**

I verified the exact head SHA and reviewed all 14 changed files. Three newer base commits do not touch these files. `git diff --check` passed. All reported GitHub CI checks passed. The change does not require a host-daemon protocol increase.

The pull request comes from an untrusted fork, and this backend change has no browser route. I did not start a server or execute the pull-request code locally.

packages/agent-runtime/src/claude-code/bridge/bridge.ts
sholub-dev: The SDK gives canUseTool no link to the originating turn or task, so per-work policy isn't possible at this layer. The old behavior rebuilt the session on every escalation flip and killed background work — the bug this PR fixes. Both crossovers fail safe: `ask` only happens while a user turn is active, `deny` auto-denies.

packages/agent-runtime/src/claude-code/bridge/bridge.ts
sholub-dev: Fixed in d4bf20e81: escalation is set only after pushInput succeeds. Added a failed-steer test.

PR #1240, opened 2026-08-10
fix: don't show missing provider CLIs as updates in the sidebar badge

patleeman:
## What's wrong

The small chip in the sidebar footer's bottom-right corner advertised an **update available** for Claude Code even on machines where Claude Code was never installed. Clicking it just opened the Updates page, where the row actually said "Not installed" with an Install button — there was no update to apply at all. Codex behaved the same way.

## Why

The daemon correctly reports `needsUpdate: false` when a CLI isn't installed — the confusion was purely in the UI. The badge builds its "stale agent" marks from every provider issue the Updates page surfaces, and "X CLI not installed" is one of those issues. So a missing CLI (an install prompt) was rendered as a brand chip labeled "X update available".

## The fix

The sidebar badge now only counts provider issues where the CLI is actually installed and stale (`needsUpdate` or `versionUnsupported`). Missing CLIs no longer produce an update chip.

Nothing else changes: Settings → Updates still shows the "Not installed" row with its Install action, which is the right place for it. The bb-app/desktop update chip is unaffected.

## Tests

Added two cases to `SidebarUpdatesBadge.test.tsx`:
- a missing CLI alone renders no badge chips at all
- a missing CLI alongside a real bb update still shows the bb chip, but no provider chip

Full `@bb/app` suite passes (325 files, 2453 tests), typecheck clean.

### DIFF
--- a/apps/app/src/components/sidebar/SidebarUpdatesBadge.test.tsx
+++ b/apps/app/src/components/sidebar/SidebarUpdatesBadge.test.tsx
@@ -52,6 +52,34 @@ function providerIssue(
   };
 }
 
+function missingInstallIssue(
+  provider: ProviderCliKey,
+  displayName: string,
+): ProviderCliIssue {
+  return {
+    provider,
+    status: {
+      displayName,
+      executableName: provider,
+      executablePath: null,
+      installed: false,
+      installSource: "notInstalled",
+      currentVersion: null,
+      latestVersion: "1.1.0",
+      minimumSupportedVersion: null,
+      npmPackageName: `@example/${provider}`,
+      npmGlobalPackageVersion: null,
+      installAction: null,
+      needsUpdate: false,
+      versionUnsupported: false,
+    },
+    action: null,
+    title: `${displayName} CLI not installed`,
+    description: `Install ${displayName} so bb can start ${displayName} sessions.`,
+    fingerprint: `${provider}:missing:1.1.0`,
+  };
+}
+
 function host(id: string): Host {
   return {
     id,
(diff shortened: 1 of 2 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

ELI5: The sidebar badge should count real tool updates. It should not count tools that the machine does not have.

I’m SlopCop. I’m reviewing this pull request for security, code quality, performance, architecture, and product behavior.

apps/app/src/components/sidebar/SidebarUpdatesBadge.test.tsx
```diff
@@ -137,6 +165,37 @@ describe("SidebarUpdatesBadge", () => {
...
+      machines: [
+        machine({
+          issues: [missingInstallIssue("claudeCode", "Claude Code")],
+        }),
+      ],
+    });
+
+    expect(
+      screen.queryByTestId("sidebar-updates-badge-providers"),
+    ).toBeNull();
```
SawyerHood: 🚨 `slopcop/review` — The changed test file does not pass the repository Prettier check.

`pnpm exec prettier --check` reports this file. Prettier puts this assertion on one line. This note does not block the review.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: The badge now shows real tool updates. A missing tool remains on the Updates page, but it no longer adds a badge.

I found no blocking issue. The change is correct and small.

One small format note remains. The changed test file does not pass Prettier. I left an inline comment with the exact location.

Security review found no security, privacy, authorization, or unsafe execution issue.

Code review confirmed that the filter has the correct position inside the issue loop. A missing entry cannot hide a stale entry.

The architecture scan found related classification logic in the issue builder, Updates page, and CLI. A typed issue kind could centralize this logic later.

`actionableCount` and `hasAttention` still count missing CLIs. Production code does not use these fields, so this PR does not cause a visible conflict.

Performance review found no regression. The loop keeps its linear cost and adds no request, hook, or large allocation.

Validation results:

- The focused test file passed all seven tests.
- The app type check passed.
- The app lint task passed with no errors.
- `git diff --check` passed.
- Current GitHub checks passed.
- Prettier found only the changed test file.

I started the exact SHA in the dev app. The app loaded correctly. The daemon restored the host CLI path, so I could not reproduce a missing CLI. Browser response mocks did not settle. The focused tests cover the changed behavior directly.

This is a comment-only review. I did not approve or request changes.

Issue #1089, opened 2026-08-06
Browser pane: page content layer renders offset, overlapping the adjacent thread pane


strahinja-rs:
**Version:** bb-app 0.35.1 (npm, packaged desktop app) · macOS 26 (Darwin 25.4.0) · Apple Silicon

**What happens:** With a thread open in the left pane and the built-in browser in a right split, the web page's content layer draws at the wrong position — offset down and left, floating over the thread pane — while the browser pane's own chrome (URL bar, tab strip, toolbar) sits correctly in the right pane. The page remains fully interactive, just in the wrong place.

**Repro:**
1. Open a thread in the left pane
2. Open the built-in browser in a right split
3. Load any page (google.com in the screenshot)

**Expected:** Page content clipped to and positioned within the browser pane, matching its chrome.

**Guess at cause (untested):** looks like stale native-webview bounds after split layout — the chrome is laid out by the web UI while the content layer keeps pre-split coordinates.

**Screenshot:**

![browser pane content overlapping thread pane](https://raw.githubusercontent.com/strahinja-rs/assets/main/2026-08-06-bb-browser-pane-offset.png)

SawyerHood:
ty for the report @strahinja-rs this should be fixed in the nightly and should be out today in the main release

PR #1244, opened 2026-08-10
Fix browser view alignment at window zoom

smsunarto:
## Summary

- Convert renderer-measured browser bounds from zoomed CSS pixels to native window points before desktop IPC.
- Apply the conversion to both initial browser attachment and later bounds updates.
- Add a 125% zoom regression test for both paths.

## Root cause

The browser panel chrome is part of the renderer and uses CSS pixel coordinates. The page is a native Electron `WebContentsView`, whose bounds use window coordinates. These coordinates match at 100% window zoom but diverge after the BB shell is zoomed.

The existing resize synchronization sent fresh bounds, but it repeatedly sent values in the wrong coordinate space. The native page therefore shifted left and overlapped the thread while the renderer-owned toolbar stayed in the correct right panel.

## Manual bug repro

1. Launch the BB desktop app.
2. Open any thread.
3. Open the right panel and select **Open browser**.
4. Load `https://example.com`.
5. Click the thread pane so the BB shell, not the embedded page, owns keyboard focus.
6. Use **View → Zoom In** to increase the BB window zoom above 100%.
7. Observe the browser page layer.

**Before:** The page layer shifts left and overlaps the thread pane. The tab strip, address bar, and other browser chrome remain correctly positioned in the right panel. Additional zoom steps increase the offset.

**Expected:** The page layer stays clipped below the browser toolbar and aligned with the right panel at every BB window zoom level.

### 100% window zoom — control

The native page aligns with the right panel.

CleanShot 2026-08-09 at 20 20 29@2x

### Increased window zoom — failure before this fix

The native page shifts left across the panel divider and covers the thread.

CleanShot 2026-08-09 at 20 22 56@2x

## Verification

- `pnpm exec turbo run test --filter=@bb/desktop --force` — 232 tests passed.
- `pnpm exec turbo run typecheck --filter=@bb/desktop --filter=@bb/desktop-contract` — passed.
- Red/green mutation check: removing the conversion makes the new regression test fail with raw, unscaled bounds; restoring it makes all five preload browser API tests pass.
- Live isolated Electron check: `example.com` stayed aligned with the right panel after increasing the BB shell zoom.

Follow-up for #1089.

### DIFF
--- a/apps/desktop/src/preload.ts
+++ b/apps/desktop/src/preload.ts
@@ -15,6 +15,7 @@ import {
   type BbDesktopBrowserSnapshotHandler,
   type BbDesktopBrowserStateHandler,
   type BbDesktopBrowserUnsubscribe,
+  type BbDesktopBrowserViewBounds,
   type BbDesktopCloseWindowRequestHandler,
   type BbDesktopInfo,
   type BbDesktopInfoChangeHandler,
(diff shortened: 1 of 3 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I started the review under the security, code quality, performance, architecture, and test guidelines.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** The app measures the browser panel in zoomed page pixels. This change scales those measurements for Electron. The native browser view then stays aligned with the panel.

I found no blocking or actionable issue.

- Security: No new trust boundary, IPC channel, listener, or unvalidated event source appears.
- Code quality: The conversion belongs in the preload because that layer can read the current page zoom.
- Architecture: I found no duplicate zoom conversion or another renderer-to-main bounds protocol.
- Performance: The change adds no event, observer, layout read, listener, or IPC message.
- Correctness: Endpoint rounding keeps adjacent edges aligned and avoids one-point seams.

Three optional follow-ups do not block this pull request:

- Add a 110% zoom test where endpoint rounding differs from direct width rounding.
- Keep the zoom factor in the renderer dedupe key if a future fixed-size layout can keep identical CSS bounds.
- Consider separate CSS-bound and native-bound types if more coordinate conversions appear.

Validation passed at `073975048fda138a5b55c72d44affa951d979c08`:

- `pnpm exec turbo run test --filter=@bb/desktop --force`: 232 tests passed.
- `pnpm exec turbo run typecheck --filter=@bb/desktop`: passed.
- `git diff --check`: passed.

I did not run browser automation. It cannot inspect the Electron native `WebContentsView`, so it would not test this alignment path.

SawyerHood:
@smsunarto ty again for the high quality contribution!

PR #1247, opened 2026-08-10
feat: add bidirectional model, provider, and reasoning cycles

smsunarto:
## Summary

Adds fast, bidirectional cycling for the active composer's model, provider, and reasoning effort without opening the picker.

| Area | Forward | Backward |
| --- | --- | --- |
| Model | `Alt+M` | `Alt+Shift+M` |
| Provider | `Alt+P` | `Alt+Shift+P` |
| Reasoning effort | `Alt+T` | `Alt+Shift+T` |

All three dimensions wrap in both directions. Reasoning cycles only through the current model's supported efforts in canonical `reasoningLevelValues` order, so provider response ordering cannot reverse the meaning of forward and backward.

The commands work with the picker open or closed and resolve to the active composer across split panes and portaled picker content. An in-scope cycle chord claims the key even when there is nowhere to rotate, preventing macOS Option-composed characters from landing in the prompt. The same physical-key matching handles Option composition for both the forward and Shift+Option variants. Unrelated editable controls do not yield the chord.

This deliberately avoids `Mod+Arrow` defaults, which collide with native caret and document navigation in editable controls. Native navigation and deletion keys continue to win through the central editable-control arbitration established for #1397.

The keyboard settings metadata, configuration documentation, plugin SDK declarations, and generated template copy are updated with the final command set.

## Validation

- `pnpm exec turbo run typecheck test lint --filter=@bb/app --filter=@bb/server --filter=@bb/domain`
  - 8/8 tasks passed
  - app: 340 files, 2,650 tests
  - server: 162 files, 1,465 tests
  - domain: 23 files, 130 tests
  - lint: 0 errors; 149 existing React Compiler warnings
- Bundled plugin SDK declaration check passed
- Generated template check passed
- `git diff --check` passed

Original implementation by @smsunarto. Their commits and authorship are preserved; the branch was rebased and the review fixes were added by the maintainers.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I started the review for PR #1247.

I will check security, code quality, architecture, duplicate code, performance, and the end-to-end behavior.

apps/server/src/services/system/app-keybindings.ts
```diff
@@ -164,45 +176,34 @@ export const DEFAULT_APP_KEYBINDINGS: AppKeybindings = [
...
+  // match on the physical key — see `normalizeAppShortcutInputKey` in @bb/domain.
+  binding("modelPicker.cycleModel", "m", { alt: true }, composerWithoutModal),
+  binding("modelPicker.cycleProvider", "p", { alt: true }, composerWithoutModal),
+  binding("modelPicker.cycleReasoning", "t", { alt: true }, composerWithoutModal),
+  // Navigation-style chords apply while the editor or any control in that
+  // composer is focused. Their handlers enforce that ownership at the
+  // individual picker level; `editableFocus` would reject Tab-focused buttons
+  // before those handlers can identify their composer. They only act on a
+  // closed picker, so they intentionally have no `modelPickerOpen` copies below.
+  binding("modelPicker.previousModel", "ArrowLeft", { mod: true }, composerWithoutModal),
```
SawyerHood: 🚨 `slopcop/review` — These defaults replace native prompt editing keys.

`Mod+ArrowLeft/Right` moves to a line boundary on macOS. `Control+ArrowLeft/Right` moves by word on Windows and Linux. The app dispatch runs before the editor handles the key. When multiple models exist, these bindings change the model instead of moving the caret. The browser smoke test reproduced that model change. Please use non-editor defaults or leave these commands unbound by default. Add an editor test that protects native caret movement.

apps/app/src/components/pickers/modelPickerCycle.ts
```diff
@@ -17,3 +17,38 @@ export function nextCycleValue<T extends string>(
...
+ * enters from the edge implied by the direction.
+ */
+export function adjacentReasoningValue<T extends string>(
+  options: readonly PickerOption<T>[],
+  current: T,
+  direction: "decrease" | "increase",
+): T | null {
+  if (options.length < 2) return null;
+
+  const currentIndex = options.findIndex((option) => option.value === current);
```
SawyerHood: 🚨 `slopcop/review` — Direction uses provider list order instead of the canonical reasoning rank.

`activeModel.supportedReasoningEfforts` keeps the provider response order. This helper then uses the adjacent array item. A valid unsorted response can make Increase select a lower effort, or make Decrease select a higher effort. `reasoningLevelValues` defines the load-bearing rank. Please traverse supported values by that rank, or sort once at the boundary. Add a test with unsorted options.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** This change adds keyboard controls for providers, models, and reasoning levels. The controls work, but two direction rules need correction.

I found two issues:

1. High: The new `Mod+Arrow` defaults replace standard prompt editing keys. They can change execution options when a user wants to move the caret.
2. Medium: Reasoning direction follows provider response order. An unsorted response can make Increase move down or make Decrease move up.

I left an inline comment for each issue.

The security review found no actionable issue. The performance review found no actionable issue.

The architecture scan found no duplicate model or provider cycle implementation. It found one duplicate policy: reasoning rank already exists in `reasoningLevelValues`.

Validation results:

- The app and server type checks passed.
- The 46 focused app tests passed.
- The five focused server tests passed.
- The browser check passed for model, provider, and reasoning controls.
- The browser check also reproduced the native caret-key conflict.
- A broad server run had 1,369 passes and ten unrelated timeout failures.

PR #1253, opened 2026-08-10
Fix React #185 in split thread views

### LINKED ISSUE #1246: Opening a thread can trigger React error #185 and leave a blank screen
## Summary

Opening an existing thread can trigger React error #185 and leave the entire bb web app as a blank dark screen.

I observed this failure directly in Chrome and captured the console error. The exact thread-data condition is not yet isolated, so there is no shareable reproduction recipe yet.

## Environment

- bb: 0.36.0
- Client: bb web app in Chrome on macOS
- Route shape: `/projects/<project-id>/threads/<thread-id>`
- The served asset hashes match the packaged 0.36.0 build

## Expected

The thread page renders normally. A problem in the composer should not take down the whole route.

## Actual

The route renders as a completely blank dark page with no recovery UI.

## Console error

```text
Error: Minified React error #185; visit https://react.dev/errors/185
    at _i (dist-CiVt1nMI.js:8:27500)
    at mi (dist-CiVt1nMI.js:8:27025)
    at zs (dist-CiVt1nMI.js:8:58555)
    at Rs (dist-CiVt1nMI.js:8:58177)
    at Ue (workspace-checkout-display-Bf-7ns1Z.js:507:123482)
    at Kc (dist-CiVt1nMI.js:8:91672)
    at ll (dist-CiVt1nMI.js:8:96131)
    at Cl (dist-CiVt1nMI.js:8:104917)
    at ll (dist-CiVt1nMI.js:8:96118)
    at Cl (dist-CiVt1nMI.js:8:104917)
```

React error #185 decodes to “Maximum update depth exceeded.”

## Source investigation

The minified `Ue` frame at the reported byte offset maps to the stack-height measurement in:

`apps/app/src/components/promptbox/FollowUpPromptBox.tsx`

```tsx
useLayoutEffect(() => {
  const element = stackRef.current;
  if (!element) return;
  const measured = element.offsetHeight;
  setStackHeight((prev) => (prev === measured ? prev : measured));
}, [stack]);
```

The same component also updates `stackHeight` from a `ResizeObserver`, and then derives the prompt textarea minimum height from `stackHeight`.

Likely cause: a layout/state feedback loop where the measured stack height and elastic textarea height keep affecting each other. This is an investigation lead, not yet a proven root cause.

The code is unchanged on current `main` at the time of filing.

This appears separate from #767, which fixed a ResizeObserver/render loop in `GeneratedConversationMessage`.

## Workaround

The affected thread can still be read and continued through `bb thread` CLI commands against the same server. Avoiding the thread route is the only confirmed workaround so far.

### LINKED ISSUE #1250: Closing and reopening a split thread can trigger React #185
## Summary

After the first split-thread open succeeds, closing the right pane and opening that thread in split again can trigger React error #185 and blank the entire app.

This was discovered while QAing #1246 and its fix in #1249. It is a separate failure: the repeating stack does not enter `FollowUpPromptBox`.

## Version

- bb 0.36.0
- Reproduced in a packaged build from `cb3054d97f0646829a906900a84ca744badc64df` with the #1246 prompt-box fix applied

## Reproduction

1. Open a thread in a single pane.
2. From a sidebar thread's actions, choose **Open in split**.
3. Close the right pane.
4. Open the same sidebar thread in split again.

## Expected

The split reopens with both threads rendered.

## Actual

The second split open blanks the route and logs minified React error #185 (maximum update depth exceeded).

## Evidence

The failing stack maps to:

```text
usePaneSecondaryPanelRegistration layout effect
  -> registration.publish(model)
  -> createPaneSecondaryPanelRegistry.publish
  -> notify(paneId)
  -> focused-pane useSyncExternalStore listener
  -> React #185
```

On the second open, the instrumented build produced 52 fresh model publications for each pane before React stopped the loop. The focused pane had one active listener; the other pane had none. Observable panel fields stayed unchanged, but the hosted panel and model references were new on every pass.

At least three independent identity paths recreate the hosted model:

- a render-created environment open context changes `openPathInFileTarget`, then `renderHostedPanel`
- `useThreadStorageBrowser` returns a fresh controller object, recreating metadata content
- the secondary-panel `fileTabs` array is recreated for the reopened thread

The resulting cycle is:

```text
fresh owner values
  -> fresh hosted panel model
  -> layout-effect publish
  -> registry synchronously notifies focused host
  -> host renders the panel
  -> pane renders again
  -> fresh owner values
```

Suppressing that registry notification moved #185 to a Radix Popper `ResizeObserver` setter during the same repeat transition. The registry loop is proven, but fixing only its first unstable reference may reveal another mount-time writer.

## Fix direction

Avoid publishing an opaque `ReactNode` from a child layout effect into a synchronously subscribed ancestor. The registry should carry semantic layout state, while panel content remains React-owned through a stable host/slot boundary.

A fix should preserve panel focus, live content updates, resize behavior, and composer context without relying on memoizing every current callback or object producer.

## Acceptance criteria

- Open split -> close pane -> reopen split keeps both panes rendered.
- No React #185 or blank route.
- Registry publication/notification remains bounded.
- Hosted panel content still updates correctly.
- A regression test exercises the full open -> close -> reopen transition.

ratulsarna:
Fixes #1246.
Fixes #1250.

## What changed

- Keep the selected merge-base branch local to each thread view instead of sharing it across split panes.
- Keep pending file and commit diff actions with the pane that requested them.
- Deduplicate composer stack-height measurements before scheduling React state updates.

## Why

Split panes could have different merge-base branches while reading and writing the same global selection. Each pane then restored its own value, causing the values to alternate and the panes to rerender until React threw error #185 and blanked the route.

## Verification

- Full @bb/app suite: 326 files, 2,456 tests passed.
- Turbo typecheck passed for @bb/app.
- Regression tests cover two simultaneous merge-base owners, environment changes, delayed pane-local diff actions, and stale-action cleanup.
- Installable bb-app tarball smoke test passed.
- Packaged browser QA passed on the affected thread: initial split, two close/reopen cycles, and persisted split reload. Both panes and composers remained visible, with no React #185 or merge-base request loop.

### DIFF
--- a/apps/app/src/components/secondary-panel/git-diff/useGitDiffPanel.test.tsx
+++ b/apps/app/src/components/secondary-panel/git-diff/useGitDiffPanel.test.tsx
@@ -1,0 +1,108 @@
+// @vitest-environment jsdom
+
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+import { Provider as JotaiProvider } from "jotai";
+import { useEffect, useRef, useState } from "react";
+import { expect, it, vi } from "vitest";
+import { shouldSyncSelectedMergeBaseBranch } from "./useEnvironmentMergeBase";
+import { useGitDiffPanel } from "./useGitDiffPanel";
+
+const requestedBranches = vi.hoisted(
+  () => new Map<string, string | undefined>(),
+);
+
+vi.mock("../../../hooks/queries/environment-queries", () => ({
+  useEnvironmentMergeBaseBranches: (
+    environmentId: string,
+    options?: { selectedBranch?: string },
+  ) => {
+    requestedBranches.set(environmentId, options?.selectedBranch);
+    return { data: undefined, isFetching: false };
+  },
+}));
+
+const noop = () => undefined;
+
+function MergeBaseOwner({
+  environmentId,
+  persistedBranch,
+}: {
+  environmentId: string;
+  persistedBranch: string;
+}) {
+  const stateKeyRef = useRef<string | undefined>(undefined);
+  const syncCountRef = useRef(0);
+  const { selectedMergeBaseBranch, setSelectedMergeBaseBranch } =
+    useGitDiffPanel({
+      activeSecondaryTab: null,
+      clearActiveFileTabs: noop,
+      environmentId,
+      mergeBaseBranchOptionsEnabled: true,
+      setThreadSecondaryPanel: noop,
+    });
+
+  useEffect(() => {
+    if (
+      !shouldSyncSelectedMergeBaseBranch({
+        previousStateKey: stateKeyRef.current,
+        nextStateKey: environmentId,
+        persistedMergeBaseBranch: persistedBranch,
+        selectedMergeBaseBranch,
+        updatePending: false,
+      })
+    ) {
+      return;
+    }
+    syncCountRef.current += 1;
+    if (syncCountRef.current > 10) {
+      throw new Error("Merge-base owner did not settle independently");
+    }
+    stateKeyRef.current = environmentId;
+    setSelectedMergeBaseBranch(persistedBranch);
+  }, [
+    environmentId,
+    persistedBranch,
+    selectedMergeBaseBranch,
+    setSelectedMergeBaseBranch,
+  ]);
+
+  return (
+    <output data-testid={environmentId}>
+      {selectedMergeBaseBranch ?? "unset"}
+    </output>
+  );
+}
+
+function TwoOwnerHarness() {
+  const [leftBranch, setLeftBranch] = useState("main");
+  return (
+    <JotaiProvider>
+      <MergeBaseOwner environmentId="env-left" persistedBranch={leftBranch} />
+      <MergeBaseOwner environmentId="env-right" persistedBranch="release" />
+      <button type="button" onClick={() => setLeftBranch("feature-left")}>
+        Change left branch
+      </button>
+    </JotaiProvider>
+  );
+}
+
+it("keeps simultaneous merge-base owners and queries independent", async () => {
+  requestedBranches.clear();
+  render(<TwoOwnerHarness />);
+
+  await waitFor(() => {
+    expect(screen.getByTestId("env-left").textContent).toBe("main");
+    expect(screen.getByTestId("env-right").textContent).toBe("release");
+  });
+  expect(requestedBranches.get("env-left")).toBe("main");
+  expect(requestedBranches.get("env-right")).toBe("release");
+
+  fireEvent.click(screen.getByRole("button", { name: "Change left branch" }));
+
+  await waitFor(() => {
+    expect(screen.getByTestId("env-left").textContent).toBe("feature-left");
+    expect(requestedBranches.get("env-left")).toBe("feature-left");
+  });
+  expect(screen.getByTestId("env-right").textContent).toBe("release");
+  expect(requestedBranches.get("env-right")).toBe("release");
+});
--- a/apps/app/src/components/secondary-panel/git-diff/useGitDiffPanel.test.tsx
+++ b/apps/app/src/components/secondary-panel/git-diff/useGitDiffPanel.test.tsx
@@ -107,12 +107,12 @@ function useMergeBaseOwner(environment: Environment, threadId: string) {
     updateEnvironment,
   });
   const diffState = useGitDiffPanelState({
-    defaultMergeBaseBranch: effectiveMergeBaseBranch,
     environmentId: environment.id,
     isDiffPanelActive: true,
     onClearPendingGitDiffIntent: panel.clearPendingGitDiffIntent,
     pendingGitDiffCommitSha: panel.pendingGitDiffCommitSha,
     pendingGitDiffScrollPath: panel.pendingGitDiffScrollPath,
+    requestedMergeBaseBranch: effectiveMergeBaseBranch,
   });
   return { ...diffState, ...panel, effectiveMergeBaseBranch };
 }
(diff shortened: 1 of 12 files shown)


SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

I will check security, code quality, performance, architecture, and tests. I will post one final review after these checks.

apps/app/src/components/secondary-panel/git-diff/useGitDiffPanel.ts
```diff
@@ -30,8 +29,8 @@ export function useGitDiffPanel({
   mergeBaseBranchOptionsEnabled = false,
   setThreadSecondaryPanel,
 }: UseGitDiffPanelParams) {
-  const selectedMergeBaseBranch = useAtomValue(selectedMergeBaseBranchAtom);
-  const setSelectedMergeBaseBranch = useSetAtom(selectedMergeBaseBranchAtom);
+  const [selectedMergeBaseBranch, setSelectedMergeBaseBranch] =
```
SawyerHood: 🚨 `slopcop/review` — The branch state has no environment identity.

A pane can change from environment A to B without a remount. The first B render still uses branch A.

The effect at line 82 clears the branch later. This can start B queries with the branch from A.

Store the environment ID with the branch. Ignore the branch when its environment does not match.

Add a test that renders the view again with environment B. Confirm that no B query uses branch A.

apps/app/src/components/secondary-panel/git-diff/useGitDiffPanel.ts
```diff
@@ -30,8 +29,8 @@ export function useGitDiffPanel({
...
+    useState<string>();
   const setPendingGitDiffScrollPath = useSetAtom(pendingGitDiffScrollPathAtom);
```
SawyerHood: 🚨 `slopcop/review` — The file and commit intents still use global atoms.

Every split pane reads these atoms. Pane B can consume or clear a request from pane A.

This event can reset the selection in pane B. It can also cancel the requested scroll in pane A.

Move both intents into the pane owner. An atom family with a pane and thread key also works.

Add a delayed scroll test. Confirm that only pane A consumes its request.

apps/app/src/components/secondary-panel/git-diff/useGitDiffPanel.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const { selectedMergeBaseBranch, setSelectedMergeBaseBranch } =
+    useGitDiffPanel({
+      activeSecondaryTab: null,
+      clearActiveFileTabs: noop,
+      environmentId,
+      mergeBaseBranchOptionsEnabled: true,
+      setThreadSecondaryPanel: noop,
+    });
+
+  useEffect(() => {
```
SawyerHood: 🚨 `slopcop/review` — This test copies the production synchronization effect.

The test does not use `useEnvironmentMergeBase`. It does not change an environment ID.

It also omits the split close and reopen sequence from issue #1250. Therefore, both medium findings can remain while this test passes.

Use the production hooks. Add the complete close and reopen regression test.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: Each split pane needs its own small box for branch and diff requests.

This change gives each pane a branch box. However, two request notes remain in one shared box.

Review result: I found two medium correctness issues and one low test gap.

First, the selected branch has no environment key. A pane can query the new environment with the old branch.

Bind the branch to its environment. Add a test that changes the environment without a remount.

Second, file-open and commit-open requests still use global atoms. Another pane can consume or clear these requests.

Move these requests into the pane owner, or key them by pane and thread.

The new ownership test copies the production synchronization effect. It omits the production hook composition and the split close and reopen sequence.

Security: I found no security issue. The diff adds no new permission, process, storage, HTML, dependency, server, or daemon boundary.

Performance: I found no material performance defect. The common `offsetHeight` measurement prevents alternating box models and repeated state updates.

Architecture: The local branch state is a good direction. The codebase already uses an atom family for per-thread panel state.

The two remaining intents should use the same ownership model.

Evidence: Three GPT-5.6 workers checked security, code quality, and performance. A separate GPT-5.6 reviewer checked their combined findings.

GitHub reports no checks for this SHA. I did not run the fork code because that action would execute untrusted code.

I also did not start a development server or use a browser for the same reason.

The pull request body claims successful tests and packaged browser checks. I could not verify those claims.

Two medium findings remain before merge.

apps/app/src/components/secondary-panel/git-diff/useGitDiffPanel.ts
ratulsarna: Fixed in 48cf661d5. The selected branch is now stored with its environment ID and ignored as soon as the view changes environments. The new production-hook test rerenders from environment A to B and verifies that no B query uses the A branch.

apps/app/src/components/secondary-panel/git-diff/useGitDiffPanel.ts
ratulsarna: Fixed in 48cf661d5. File and commit intents now belong to the pane-local useGitDiffPanel owner and are tagged with the current thread and environment. Tests verify that another owner cannot clear the request, and that an unconsumed file intent cannot revive after navigating away and back.

apps/app/src/components/secondary-panel/git-diff/useGitDiffPanel.test.tsx
ratulsarna: Fixed in 48cf661d5. The copied synchronization effect is gone. The regression now uses useGitDiffPanel, useEnvironmentMergeBase, and useGitDiffPanelState under StrictMode with two independent owners, an unmount/remount cycle, and an in-place environment change. The exact split close/reopen flow is also covered by packaged browser QA.

PR #1263, opened 2026-08-10
Make mobile composer behavior focus-driven

ymichael:
## Summary
- make composer collapse follow settled focus state instead of timeline pointer activity, so scrolling cannot create a focused-but-collapsed prompt box
- collapse correctly when the editor blurs to the page, while preserving focus in composer-owned portaled overlays
- use `interactive-widget=resizes-content` for the standards-based Android keyboard path and limit the Visual Viewport shell workaround to iOS WebKit
- keep the iOS shell aligned to the last reported viewport until keyboard-dismissal geometry arrives, avoiding the composer jumping below the visible viewport on blur

## Testing
- `pnpm exec turbo run test --filter=@bb/app --force` (324 files, 2,447 tests)
- `pnpm exec turbo run lint --filter=@bb/app` (passes with existing warnings)
- Android Chrome mobile smoke test at 390x844: focus expands, timeline touch preserves focus and expansion, blur collapses, and the composer remains within the viewport

### DIFF
--- a/apps/app/src/components/promptbox/FollowUpPromptBox.tsx
+++ b/apps/app/src/components/promptbox/FollowUpPromptBox.tsx
@@ -7,6 +7,7 @@ import {
   useRef,
   useState,
   type ComponentProps,
+  type FocusEvent as ReactFocusEvent,
   type ReactNode,
 } from "react";
 import type {
@@ -122,6 +123,16 @@ const DEFAULT_FOLLOW_UP_COMPOSER_SCOPE = {
   kind: "new-thread",
   projectId: null,
 } as const;
+
+function isKeyboardFocusTarget(target: EventTarget | null): boolean {
+  return (
+    target instanceof HTMLElement &&
+    (target.isContentEditable ||
+      target instanceof HTMLInputElement ||
+      target instanceof HTMLTextAreaElement ||
+      target instanceof HTMLSelectElement)
+  );
+}
 /**
  * Discriminated state for the composer's submit affordances. Replaces the
  * previous canSendFollowUp / canQueueFollowUp / canStopRuntime / onStop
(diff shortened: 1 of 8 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I will check security, code quality, performance, architecture, and the mobile user path.

apps/app/src/components/promptbox/FollowUpPromptBox.tsx
```diff
@@ -460,6 +460,17 @@ function FollowUpPromptBoxWithComposer({
       ) {
         return;
       }
+      // A touch-scroll of the timeline starts with a pointerdown outside the
+      // composer, but mobile browsers can keep the editor and keyboard
+      // focused throughout the gesture. Focus moving outside will emit the
+      // focusin event this listener also handles; pointerdown alone must not
+      // create a focused-but-collapsed composer.
+      if (
+        event.type === "pointerdown" &&
```
SawyerHood: 🚨 `slopcop/review` — This guard can leave the composer expanded after focus leaves.

This handler runs before the browser updates focus. A tap on non-focusable timeline text sees the editor as active and returns here. Chrome then moves focus to `document.body`. No outside `focusin` event occurs, so the composer never collapses.

I reproduced this on the mobile dev page. After a real timeline text click, the active element was `BODY`, but `data-follow-up-composer-expanded` remained set. The new test only calls `outside.focus()`, so it does not cover this path.

Please handle the `focusout` path or defer this check until the focus update completes. Add a regression test for an outside non-focusable target.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** The message box is like a small drawer. This change keeps it open while you move the page.

I found one P2 correctness issue.

A tap on non-focusable timeline text can move focus to the page without a new `focusin` event. The new `pointerdown` guard returns before that focus change. The composer then stays expanded after focus leaves. I left a detailed inline comment with a reproduction.

Security review found no issues. The change does not add a data path, a permission change, or an action trigger.

Performance review found no issues. The change adds one bounded DOM containment check per outside pointer start. It adds no listener, timer, render, or scroll handler.

The architecture scan found no duplicate composer focus implementation. The nearby viewport focus logic has a separate purpose, so a refactor would not help this change.

Validation completed:

- The full `@bb/app` suite passed: 324 files and 2,444 tests.
- The `@bb/app` type check passed.
- The focused prompt-box suite passed all 22 tests.
- The mobile dev page kept the focused composer expanded after a timeline touch start.
- The same page reproduced the missed blur path after a real timeline text click.

Please fix the missed blur path and add a regression test for a non-focusable outside target.

PR #1265, opened 2026-08-10
Freeze agent instructions for thread lifetime

### LINKED ISSUE #1217: Sending a message kills a thread's running background tasks when instructions drifted between turns

SawyerHood:
Fixes #1217.

## Why

bb rebuilt a thread's instructions on every send. A memory write or `AGENTS.md` edit changed the session configuration.

The runtime then sent `thread/resume`. Claude Code replaced the live process and stopped every open background task.

Instructions are thread policy. They should not mutate after the first provider session starts.

## What changed

- Freeze the first resolved instructions through content-addressed snapshots and one thread mapping.
- Reuse that text for every later turn and provider resume.
- Serialize concurrent first resolutions so plugin instruction providers run once.
- Reject unreadable or oversized instruction input without storing a partial value.
- Remove a shared snapshot after its final thread receives a permanent deletion.
- Remove instructions from the live `runTurn` and `steerTurn` contracts.
- Keep the original instructions when execution settings require a session reconfiguration.
- Increase the host daemon protocol version to 91.
- Update the plugin SDK and bb guides to document the fixed instruction lifetime.

## Impact

Memory writes, custom instruction edits, and `AGENTS.md` edits now affect new threads only.

New threads freeze instructions before their first provider session. Pre-migration threads freeze the first runtime configuration resolved after the upgrade.

The complete resolved instruction text has a 256 KiB UTF-8 limit. Shared text uses one database snapshot.

This also keeps the system prompt stable for prompt caching.

## Validation

- Rebased onto `origin/main` after migration 0090; this change now uses migration 0091.
- 373 database tests passed.
- 46 focused server tests passed.
- 36 host daemon contract tests passed.
- 30 focused host daemon dispatch tests passed.
- 26 focused agent runtime tests passed.
- All affected Turbo typecheck tasks passed.
- Generated migrations, templates, and plugin SDK declarations are current.
- GitHub CI passed before the review-fix push and is running again for the rebased head.

The full agent runtime suite passed 849 of 850 tests. The unrelated stderr tail timing test failed twice locally.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull request for security, code quality, architecture, performance, and end-to-end behavior.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull request for security, code quality, architecture, performance, and end-to-end behavior.

apps/server/src/services/threads/thread-runtime-config.ts
```diff
@@ -225,83 +237,85 @@ export async function resolveThreadRuntimeCommandConfig(
...
-      `The following workspace instructions come from ${WORKSPACE_AGENT_INSTRUCTIONS_RELATIVE_PATH}:`,
-      workspaceAgentInstructions,
-    );
+    if (workspaceAgentInstructions) {
+      instructionSections.push(
+        `The following workspace instructions come from ${WORKSPACE_AGENT_INSTRUCTIONS_RELATIVE_PATH}:`,
+        workspaceAgentInstructions,
+      );
+    }
+    instructions = freezeThreadAgentInstructions(deps.db, {
```
SawyerHood: 🚨 `slopcop/review` — [P1] Limit each stored instruction snapshot

This write stores the full instruction text for each thread. The workspace reader permits 25 MiB, and the data-dir reader has no limit.

A repository can grow SQLite and its WAL by about 25 MiB for each started thread. Large synchronous writes also block other database work.

Please apply a UTF-8 byte limit before this write. Reject oversized input or cut complete sections before storage.

apps/server/src/services/threads/thread-runtime-config.ts
```diff
@@ -176,15 +182,21 @@ export async function resolveThreadRuntimeCommandConfig(
   }
 
   const { workspaceProvisionType } = args.environment;
+  const frozenInstructions = getThreadAgentInstructions(
```
SawyerHood, line 242: 🚨 `slopcop/review` — [P1] Serialize the initial instruction resolution

Two first commands can both read no row before the asynchronous work completes. Both calls can then execute `contributeInstructions`.

The database conflict chooses one result only after both callbacks run. This behavior violates the SDK once-only contract and can duplicate callback side effects.

Please use a per-thread asynchronous guard around the initial assembly. Add a concurrent test that checks for one callback call.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: bb now puts the first thread rules in a locked box. Later rule changes do not replace them.

I found two blocking problems.

1. `[P1]` Limit each stored instruction snapshot.

Workspace instructions can reach 25 MiB. Data-dir instructions have no limit.

Each thread copies the full text into SQLite and its WAL. This can fill the disk and block server database work.

Apply a total UTF-8 byte limit before the write. Reject excess input or cut complete sections.

2. `[P1]` Serialize the initial instruction resolution.

Two first resolutions can both call `contributeInstructions` before the database conflict selects one value.

This violates the SDK once-only contract. A stateful plugin can repeat side effects.

Use a per-thread asynchronous guard around the initial assembly. Add a concurrent test that checks for one callback call.

Nonblocking cleanup remains:

- Update stale lifetime text in `docs/configuration.md`, the builtin bb CLI skill, the plugin guide, and internal comments.
- Add `thread_agent_instructions` to the migration replay rewind helpers.
- Update the protocol contract test name and comment from version 90 to version 91.

I found no separate security defect in the intended rule retention across workspace moves.

The database helper does not duplicate an existing persistence path. The per-thread guard is the correct owner for initial assembly.

The security, quality, architecture, and performance reviews ran in parallel. The independent GPT-5.6 gate confirmed both blocking findings.

The fake server-daemon-provider lifecycle passed all six integration tests. A browser cannot inspect this hidden server-to-daemon payload.

Focused tests passed: 370 database, 42 server, 30 daemon, 26 runtime, and 35 contract tests.

GitHub CI is green across the server, packages, integration, app, and package smoke jobs. `git diff --check` also passed.

I posted a comment-only review, as required.

apps/server/src/services/threads/thread-runtime-config.ts
```diff
@@ -225,83 +237,85 @@ export async function resolveThreadRuntimeCommandConfig(
...
-  // built-in update_environment_directory guidance is one of them; plugin
-  // tools are description-only unless they registered a snippet).
-  for (const contribution of dynamicToolContributions) {
-    if (!contribution.instructions) continue;
-    if (contribution.pluginId === null) {
-      instructionSections.push(contribution.instructions);
-    } else {
+  let instructions = frozenInstructions;
+  if (instructions === null) {
+    const dataDirAgentInstructions = readDataDirAgentInstructions(
```
SawyerHood: 🚨 `slopcop/review` — A temporary data-directory read error becomes permanent

`readDataDirAgentInstructions` returns `null` after any read error. This code then stores instructions without the user section.

A temporary permission or I/O error removes user instructions for the thread lifetime. Return an error and freeze only complete input.

packages/templates/src/templates/bb-guide-agent-configuration.md
```diff
@@ -15,14 +15,16 @@ User instructions (<dataDir>/AGENTS.md):
   Add an AGENTS.md file to the bb data dir (usually ~/.bb/AGENTS.md) to give
   every provider-backed thread across all projects default user-level
   instructions. bb reads <dataDir>/AGENTS.md and appends its contents to the
-  thread system prompt for all providers when a provider session starts.
+  thread system prompt for all providers when the first provider session starts.
+  The thread keeps that text. File changes apply only to new threads.
```
SawyerHood: 🚨 `slopcop/review` — Document the upgrade case for existing threads

Migration 0090 creates no snapshot rows for existing threads. Their next turn freezes current files, not the instructions from their first session.

State this exception here. The current text incorrectly says file changes apply only to new threads.

packages/host-daemon-contract/test/contract.test.ts
```diff
@@ -1044,7 +1044,7 @@ describe("host-daemon command schemas", () => {
   // so it silently leaves Plan mode instead of asking, and it would reject the
   // subject if the server sent one back.
   it("uses protocol version 90 for plan-review approvals", () => {
-    expect(HOST_DAEMON_PROTOCOL_VERSION).toBe(90);
+    expect(HOST_DAEMON_PROTOCOL_VERSION).toBe(91);
```
SawyerHood, line 1053: 🚨 `slopcop/review` — Keep the protocol test aligned with version 91

The comment and test name require version 90, but the assertion requires version 91.

Keep the version-90 plan check. Add a separate version-91 check for fixed thread instructions.

apps/server/src/services/threads/thread-runtime-config.ts
SawyerHood: 🚨 `slopcop/review` — Limit and deduplicate instruction snapshots

The workspace reader accepts 25 MiB. The data-directory reader has no size limit. This insert duplicates the full text for each thread.

A shared 1 MiB file across 10,000 threads adds about 10 GiB. Add a byte limit and use content-addressed storage.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: Each thread gets an instruction box. BB locks the box. Later file changes cannot replace what is inside.

I found two medium issues and two low issues.

- Medium: A temporary data-directory read error can permanently freeze an incomplete instruction set.
- Medium: Full snapshots have no instruction-specific size limit. Repeated content can cause large database, WAL, and backup growth.
- Low: The guide omits the behavior for threads that existed before migration 0090.
- Low: The protocol test still describes version 90 while it asserts version 91.

The database lookup uses the primary-key index. The conflict-safe insert handles concurrent first resolutions correctly.

I found no authentication, authorization, cross-project exposure, wire-validation, or query-plan defect. The protocol version correctly increased to 91.

I found no direct duplicate instruction builder. Content-addressed snapshot storage would remove repeated text while preserving one immutable result per thread.

The focused checks passed 165 tests across database, server, runtime, daemon, and contract packages. Six affected package type checks also passed.

The Pi resume case could not run because the local provider returned no models. I found no link between that result and this change.

The real dev app also passed. A Codex thread first returned `FIRST_FROZEN_TOKEN`.

I changed `.bb/AGENTS.md` to `SECOND_CHANGED_TOKEN`. The next turn still returned `FIRST_FROZEN_TOKEN`.

The browser showed the same result in the thread page. A separate restart test also kept one unchanged instruction row.

SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull request for security, code quality, performance, architecture, and end-to-end behavior. I will post the final findings after parallel reviews and practical tests.

packages/db/src/schema.ts
```diff
@@ -564,6 +564,30 @@ export const threads = sqliteTable(
...
+    createdAt: integer("created_at").notNull(),
+  },
+);
+
+export const threadAgentInstructions = sqliteTable(
+  "thread_agent_instructions",
+  {
+    threadId: text("thread_id")
+      .primaryKey()
+      .references(() => threads.id, { onDelete: "cascade" }),
```
SawyerHood, line 583: 🚨 `slopcop/review` — [P2] Clean up snapshots after the last thread link is deleted

This cascade removes the thread-to-snapshot link when `deleteThread()` performs a hard delete, but it never removes the unreferenced snapshot. I reproduced the path: the link count became zero while the snapshot count stayed one.

This retains user, workspace, and plugin instructions after permanent deletion. It can also add 256 KiB for each unique deleted thread. Please remove a snapshot after its last link disappears. Add tests for both unique and shared snapshots.

apps/server/src/services/threads/workspace-agent-instructions.ts
```diff
@@ -92,6 +102,10 @@ export async function readWorkspaceAgentInstructions(
     throw error;
   }
 
+  assertInstructionSize(
```
SawyerHood: 🚨 `slopcop/review` — [P3] Enforce the limit on trimmed text before excess transfer

This check uses the raw file size before decoding and trimming. A whitespace-only file above 256 KiB fails, although the documentation says empty files are absent. The data-directory path checks the trimmed text instead.

The preceding `host.read_file` call can also transfer up to its 25 MiB limit before this rejection. Please use a bounded host read or metadata check. Then validate the trimmed UTF-8 text. Add tests for whitespace-only and padded files.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: Each thread puts its first rules in a locked box. Later rule changes cannot replace them.

I found one required fix and two lower-priority issues.

- Required: permanent thread deletion removes the link but retains the private snapshot. Unique deleted threads can leave 256 KiB each.
- Lower: workspace size checks use untrimmed bytes.
- Lower: a remote host can transfer 25 MiB before the 256 KiB rejection.

Please delete a snapshot after its final link disappears. Add tests for unique and shared snapshots.

The security review found no other authentication, authorization, cross-project, or migration issue.

The architecture scan found no duplicate instruction builder. Content-addressed storage avoids duplicate live snapshots, but it needs reference cleanup.

The browser smoke test created a personal thread and received `ok`. The worker stopped the development server and browser.

Focused tests passed: 34 database, 45 server, 30 daemon, and 26 runtime tests. Seven affected package type checks passed.

GitHub CI passed all jobs. A broader runtime run saw one unrelated process-output timing failure.

The independent GPT-5.6 gate requested the snapshot cleanup. I treated the writer-lock and process-map points as optional.

The side-chat documentation error already exists on `main`. I did not attribute it to this pull request.

I used a comment-only review. I did not approve the pull request or use the request-changes option.

PR #1268, opened 2026-08-10
Freeze thread instructions for the life of a provider session

### LINKED ISSUE #1217: Sending a message kills a thread's running background tasks when instructions drifted between turns

SawyerHood:
Fixes #1217.

## Problem

The server recomposes thread instructions on every turn submission. On the next turn, `reconfigureThreadIfNeeded` in `packages/agent-runtime/src/runtime.ts` compared the incoming instructions string to the session's and issued `thread/resume` on any difference. The claude-code adapter settles every open background task as interrupted on an accepted resume, and the bridge tears down the live CLI session because `baseInstructions` is part of its session-construction config. So any instruction drift — a memory write, an `AGENTS.md` edit, the connect plugin's wall-clock-gated snippet — armed the next user message to kill all running background tasks.

## Fix (first principles)

Instructions are frozen for the life of a provider session. `reconfigureThreadIfNeeded` now compares only execution options; when it does resume (model, permission mode, or another option change), it passes the session's frozen instructions instead of the drifted ones, so the claude-code bridge fast path can keep the live session when nothing it constructs from changed. This also keeps the system prompt stable for prompt caching.

Fresh instructions still apply at every session construction: thread start, cold resume after a daemon restart or environment switch, fork, and the codex account restart (which shuts the process down and resumes). The server still composes instructions per turn because the cold-resume path (`resumeThreadRuntimeIfMissing`) consumes `resumeContext.instructions` from the turn payload. No wire shapes changed, so no `HOST_DAEMON_PROTOCOL_VERSION` bump.

## Contract change

Plugin `contributeInstructions` / `configure` dynamic instructions no longer take effect on the next turn of a live session; they apply when the provider session is next constructed. Updated the `backend-contract.ts` doc (bundled types regenerated) and the `bb-plugin-authoring` skill.

Known caveat: the workflows plugin flips a worker's instructions to a "do not perform more work" notice once the workflow call is terminal (`plugins/workflows/src/service.ts`). That flip no longer reaches a live session; its `bb_workflow_result` tool gating was already session-start-scoped, so enforcement there is unchanged. If mid-session enforcement matters it needs a non-instruction mechanism (follow-up).

## Tests

- New: instruction-only drift produces no `thread/resume` (`runtime.lifecycle.test.ts`).
- Updated: reconfigure-on-options-change tests now assert the resume carries the session's original instructions.
- `turbo run test --filter=@bb/agent-runtime`: 851 passed. Typecheck passes for @bb/agent-runtime, @bb/plugin-sdk, @bb/server.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am now reviewing this pull request for security, code quality, performance, architecture, and end-to-end behavior.

packages/plugin-sdk/bundled-types/bb-plugin-sdk.d.ts
```diff
@@ -12859,7 +12859,11 @@ interface PluginAgents {
      * Contribute a dynamic section appended to thread instructions. The
```
SawyerHood, line 12861: 🚨 `slopcop/review` — Regenerate the plugin declaration template after this bundled type change.

`pnpm exec turbo run typecheck --filter=@bb/templates` fails because `packages/templates/src/generated/plugin-sdk-dts.generated.ts` is stale. Regenerate that file and commit it with this declaration update.

packages/plugin-sdk/src/backend-contract.ts
```diff
@@ -504,7 +504,11 @@ export interface PluginAgents {
    * Contribute a dynamic section appended to thread instructions. The
    * provider runs when a thread's runtime command config is resolved
    * (thread.start / turn.submit); return null to contribute nothing for
-   * that resolution. Must be synchronous and fast — it sits on the
+   * that resolution. A live provider session keeps the instructions it was
+   * constructed with — a changed contribution takes effect when the
```
SawyerHood, line 510: 🚨 `slopcop/review` — Update the related `PluginAgents.configure` instruction contract.

This text now documents session-frozen instructions. However, `configure` still says its dynamic instructions apply to the next turn at line 463. Update that text to use the same session boundary, then regenerate the bundled declarations and template.

packages/agent-runtime/src/runtime.ts
```diff
@@ -810,14 +813,17 @@ function createAgentRuntimeInternal(
     }
 
     const nextOptions = args.options;
-    const nextInstructions = args.instructions ?? currentConfig.instructions;
 
+    // Instructions are frozen for the life of a provider session: drifted
```
SawyerHood, line 817: 🚨 `slopcop/review` — Increase the host daemon protocol version for this instruction-semantics change.

This changes how the daemon interprets `turn.submit` instructions. A version-90 daemon can still connect and use the old instruction-drift behavior. That can still replace a live session and stop background work. Increase `HOST_DAEMON_PROTOCOL_VERSION` to 91, then update its contract test and generated outputs.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: The agent keeps one rule sheet during a session. This protects background work. Three release files still need updates.

I found three blockers.

1. High: Increase the host daemon protocol from 90 to 91.

The daemon now gives a different meaning to turn instructions. An old daemon can still connect and use the old behavior.

2. High: Regenerate the plugin declaration template.

The `@bb/templates` type check fails because `plugin-sdk-dts.generated.ts` is stale.

3. Medium: Correct the `PluginAgents.configure` contract.

It still says that dynamic instructions apply on the next turn. The new behavior applies them at the next session construction.

Security and performance reviews found no other credible issue. The code adds no duplicate instruction loader.

The changed lifecycle file passed all 27 tests. The agent runtime and plugin SDK type checks passed.

The full agent runtime suite passed 850 of 851 tests. One unchanged stderr test failed twice, so I did not link it to this change.

The development app returned status 200 with no browser errors. A real session kept its original instructions after the workspace file changed.

The final GPT-5.6 review gate confirmed these three blockers. I used a comment-only review, with no approval or change request.

PR #1276, opened 2026-08-10
Recover file browser after worktree provisioning

### LINKED ISSUE #1222: All files viewer keeps showing a 409 environment unavailable error
## What happened

When a newly created worktree is opened in the All files viewer, the viewer requests the root directory before the environment is ready. The API returns a 409 environment unavailable response. The viewer shows that error and does not recover, so the file tree never populates.

## What I expected

The viewer should keep showing a loading state or retry the directory request. After the worktree becomes ready, the file tree should populate without requiring a manual reload.

## Steps to reproduce

1. Create or open a new managed worktree.
2. Open the thread while the worktree is still provisioning.
3. Select the All files tab.
4. Wait for provisioning to finish.
5. Observe that the 409 environment unavailable message remains and the file tree is still empty.

## Additional context

The issue is reproducible during initial worktree loading. The dev server and host daemon remain running.

ymichael:
## Summary

- invalidate environment path-list queries when environment workspace state changes
- retry an active failed file-browser request when provisioning reaches ready
- cover the provisioning 409 recovery path and invalidation contract

## Verification

- pnpm exec turbo run test --filter=@bb/app --force -- src/hooks/realtime-cache-effects.test.ts src/hooks/queries/query-helpers.test.ts
- pnpm exec turbo run lint --filter=@bb/app
- git diff --check

Closes #1222

### DIFF
--- a/apps/app/src/hooks/cache-owners/query-cache.ts
+++ b/apps/app/src/hooks/cache-owners/query-cache.ts
@@ -19,6 +19,7 @@ import {
   environmentDiffPatchQueryKeyPrefix,
   environmentFilePreviewQueryKeyPrefix,
   environmentMergeBaseBranchesQueryKeyPrefix,
+  environmentPathsQueryKeyPrefix,
   environmentPullRequestQueryKey,
   environmentQueryKey,
   environmentWorkStatusQueryKey,
@@ -372,6 +373,7 @@ export function getEnvironmentWorkspaceStateInvalidationQueryKeys({
     environmentPullRequestQueryKey(environmentId),
     environmentDiffFilesQueryKeyPrefix(environmentId),
     environmentFilePreviewQueryKeyPrefix(environmentId),
+    environmentPathsQueryKeyPrefix(environmentId),
   ];
 }
 
--- a/apps/app/src/hooks/queries/query-helpers.test.ts
+++ b/apps/app/src/hooks/queries/query-helpers.test.ts
@@ -455,7 +455,7 @@ describe("resolveEnvironmentMergeBaseBranchesPlaceholder", () => {
 });
 
 describe("getEnvironmentWorkspaceStateInvalidationQueryKeys", () => {
-  it("targets workspace-derived status and the observer-backed diff TOC, but never the observer-less patch cache", () => {
+  it("targets observer-backed workspace queries, but never the observer-less patch cache", () => {
     const queryKeys = getEnvironmentWorkspaceStateInvalidationQueryKeys({
       environmentId: "env-1",
     });
@@ -465,6 +465,7 @@ describe("getEnvironmentWorkspaceStateInvalidationQueryKeys", () => {
       ["environmentPullRequest", "env-1"],
       ["environmentDiffFiles", "env-1"],
       ["environmentFilePreview", "env-1"],
+      ["environmentPaths", "env-1"],
     ]);
     // The patch cache is observer-less; invalidation is a no-op for it, so it
     // must be evicted (removeEnvironmentDiffPatchQueries) rather than appearing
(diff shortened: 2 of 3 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I will review security, code quality, performance, architecture, and the main file-browser flow.

I will use three parallel BB review threads. I will post one final review after I combine their results.

apps/app/src/hooks/cache-owners/query-cache.ts
```diff
@@ -372,6 +373,7 @@ export function getEnvironmentWorkspaceStateInvalidationQueryKeys({
     environmentPullRequestQueryKey(environmentId),
     environmentDiffFilesQueryKeyPrefix(environmentId),
     environmentFilePreviewQueryKeyPrefix(environmentId),
+    environmentPathsQueryKeyPrefix(environmentId),
```
SawyerHood, line 376: 🚨 `slopcop/review` — Medium: an active first request can miss this recovery invalidation.

`invalidateQueries` marks an active request stale but does not start a second request when no data exists.

If that request then fails, the query stays in the error state. A direct check made one call and ended with an error.

The new test sends the status event only after the first error. Add a deferred-request test and a trailing refetch.

apps/app/src/hooks/cache-owners/query-cache.ts
```diff
@@ -19,6 +19,7 @@ import {
   environmentDiffPatchQueryKeyPrefix,
   environmentFilePreviewQueryKeyPrefix,
   environmentMergeBaseBranchesQueryKeyPrefix,
+  environmentPathsQueryKeyPrefix,
```
SawyerHood, line 22: 🚨 `slopcop/review` — High: declare this query-key import in the cache-owner registry.

The full `@bb/app` test suite fails because the cache-owner registry does not declare this import.

The test reports `environmentPathsQueryKeyPrefix` as an undeclared import for `query-cache.ts`.

Add this key to the `query-cache.ts` owner list before merge.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** The app now retries an open file search when a new worktree becomes ready.

I found one failing contract and one timing fault that can leave the original defect in place.

**Findings**

1. **High — The cache-owner contract fails.**

   The full `@bb/app` suite has 2,458 passing tests and one failed test.

   The registry does not declare the new `environmentPathsQueryKeyPrefix` import for `query-cache.ts`.

   GitHub's app test job also fails. Add this key to the `query-cache.ts` owner list.

2. **Medium — A ready event can arrive before the first request fails.**

   TanStack marks the active request stale, but it starts no second request when no data exists.

   If the first request then fails, the file search stays failed.

   A direct QueryClient check made one call and ended in the error state.

   Add a trailing refetch and test an event before a deferred first request fails.

3. **Medium — Connection recovery does not include environment paths.**

   The initial-connect and reconnect list contains project paths, but it does not contain environment paths.

   If the client misses the ready event during a socket outage, the active error remains.

   Add an all-environment path prefix to both connection recovery paths.

**Performance**

The shared list also runs for metadata changes and environment actions.

Each active path search causes a recursive workspace scan. The daemon applies the result limit only after this scan.

A targeted path recovery helper would avoid unrelated scans.

**Security**

I found no new material security issue in the changed code.

The query key stays environment-specific. The server also resolves the stored host and workspace for that environment.

**Architecture**

The patch correctly reuses the shared invalidation helper.

However, the live workspace and reconnect lists now differ from it.

A targeted environment-path helper can centralize recovery without broad invalidation.

**Validation**

- I ran the full `@bb/app` test suite. It failed only on the missing cache-owner declaration.

- I ran the `@bb/app` Turbo type check. It passed.

- I opened file search during a 60-second worktree setup and entered `README`.

- The same active search showed `README.md` after provisioning. This browser test passed.

- Three parallel BB threads reviewed security, code quality, architecture, and performance.

I recommend that you fix findings one and two before merge. Please also add the connection recovery coverage.

PR #1277, opened 2026-08-10
Prevent duplicate projects for the same local folder

### LINKED ISSUE #1194: Add Project can create duplicate sidebar projects
## What happened

I ended up with two copies of the same project in the sidebar. Both had the same name and used the same local folder.

At first, I thought this happened because I started two threads within a few seconds of each other. I checked the timestamps, though, and BB created the duplicate project before either thread started. The threads then ended up under different copies of the project.

From what I can tell, this happens when the same local folder is added as a project more than once. I didn't try it again because I don't want more duplicates.

## What I expected

If a folder already belongs to a project, BB should open that project or warn me before it creates another one.

## Environment

- BB desktop 0.36.0
- macOS

ymichael:
## Summary

- return the existing active project when Add Project receives the same normalized host path
- skip redundant host inspection for known projects and serialize the final lookup/create transaction to close concurrent request races
- preserve distinct project identity when the same path string belongs to different machines
- verify that a no-project thread using the folder through the Personal project does not block project creation

## Tests

- pnpm exec turbo run test --filter=@bb/db -- test/data/projects.test.ts
- pnpm exec turbo run test --filter=@bb/server -- test/public/public-projects-local-host.test.ts
- pnpm exec turbo run typecheck --filter=@bb/db

Closes #1194

### DIFF
--- a/apps/server/src/routes/projects.ts
+++ b/apps/server/src/routes/projects.ts
@@ -1,8 +1,9 @@
 import path from "node:path";
 import {
   countProjectSources,
-  createProject,
+  findOrCreateProjectByLocalPathSource,
   getPersonalProject,
+  getPublicProjectByLocalPathSource,
   createProjectSource,
   deleteProjectSource,
   getProjectSourceByHost,
@@ -355,11 +356,22 @@ export function registerProjectRoutes(app: Hono, deps: AppDeps): void {
       requireNonDestroyedHostWithStatus(deps, source.hostId);
       assertUsableHostId(deps, { hostId: source.hostId });
     }
+    const existingProject = getPublicProjectByLocalPathSource(deps.db, source);
+    if (existingProject) {
+      return context.json(
+        buildProjectResponses(deps, existingProject.id)[0],
+        201,
+      );
+    }
     const gitRemoteUrl = await inspectProjectGitRemoteBestEffort(deps, source);
-    const { project } = createProject(deps.db, deps.hub, {
-      name: payload.name,
-      source,
-    });
+    const { project } = findOrCreateProjectByLocalPathSource(
+      deps.db,
+      deps.hub,
+      {
+        name: payload.name,
+        source,
+      },
+    );
     if (gitRemoteUrl !== null) {
       setProjectGitRemoteUrlIfMissing(
         deps.db,
--- a/packages/db/src/data/projects.ts
+++ b/packages/db/src/data/projects.ts
@@ -106,6 +106,77 @@ function getPublicProjectForMutation(
   );
 }
 
+function getPublicProjectWithLocalPathSource(
+  db: DbQueryConnection,
+  source: CreateProjectLocalPathSourceInput,
+) {
+  return (
+    db
+      .select({ project: projects, source: projectSources })
+      .from(projects)
+      .innerJoin(projectSources, eq(projectSources.projectId, projects.id))
+      .where(
+        and(
+          publicProjectFilter(),
+          eq(projectSources.type, source.type),
+          eq(projectSources.hostId, source.hostId),
+          eq(projectSources.path, source.path),
+        ),
+      )
+      .orderBy(asc(projects.sortKey), asc(projects.id))
+      .limit(1)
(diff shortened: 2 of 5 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I started a review under the security, code quality, performance, architecture, and end-to-end guidelines.

apps/server/src/routes/projects.ts
```diff
@@ -355,11 +356,22 @@ export function registerProjectRoutes(app: Hono, deps: AppDeps): void {
       requireNonDestroyedHostWithStatus(deps, source.hostId);
       assertUsableHostId(deps, { hostId: source.hostId });
     }
+    const existingProject = getPublicProjectByLocalPathSource(deps.db, source);
```
SawyerHood, line 359: 🚨 `slopcop/review` — Preserve Git remote repair for an existing project.

An offline first request can create a project with no Git remote. This early return prevents a later online request from inspecting the folder. The project then keeps a missing remote, which disables the normal clone choice. Inspect the folder when the existing project lacks this value.

packages/db/src/data/projects.ts
```diff
@@ -125,45 +196,30 @@ export function createProject(
...
-    return { project: p, source: s };
-  });
+export function findOrCreateProjectByLocalPathSource(
+  db: DbConnection,
+  notifier: DbNotifier,
+  input: CreateProjectInput,
+) {
+  const { created, project, source } = db.transaction(
+    (tx) => {
+      const existing = getPublicProjectWithLocalPathSource(tx, input.source);
```
SawyerHood, line 211: 🚨 `slopcop/review` — Apply the identity rule to all source writes.

Only this helper has the atomic duplicate check. `createProjectSource` and `updateProjectSource` can give two active projects the same `(hostId, path)`. The schema does not prevent that state. Centralize this rule across source creation and updates.

packages/db/src/data/projects.ts
```diff
@@ -106,6 +106,77 @@ function getPublicProjectForMutation(
...
+    db
+      .select({ project: projects, source: projectSources })
+      .from(projects)
+      .innerJoin(projectSources, eq(projectSources.projectId, projects.id))
+      .where(
+        and(
+          publicProjectFilter(),
+          eq(projectSources.type, source.type),
+          eq(projectSources.hostId, source.hostId),
+          eq(projectSources.path, source.path),
```
SawyerHood, line 123: 🚨 `slopcop/review` — Normalize the host path before this comparison.

This exact text check still lets one folder create two active projects. The live route created separate projects for `/tmp/repo` and `/tmp/./repo`. The daemon returns a resolved path, but the server discards it. Compare and store one canonical path.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: One folder should get one project card, like one toy box gets one label. This change handles matching text. Some other forms still make extra cards.

I found three medium issues.

1. The new lookup compares path text. Paths such as `/tmp/repo` and `/tmp/./repo` still create separate projects for one folder.

2. Source creation and source updates bypass the new identity check. They can give two active projects the same host and path.

3. The early return blocks Git remote repair. A project created while offline can keep a missing remote after the host reconnects.

I left an inline comment for each issue.

Performance is acceptable at the current scale. However, the lookup lacks a `(host_id, path)` index and scans all sources for one host.

The duplicate response also returns `201`, although it creates nothing. A clear created-or-existing result would improve the route and CLI contract.

The codebase has an older remote-based duplicate check in onboarding. One server service should own path identity, inspection, lookup, creation, and remote repair.

Validation passed:

- The database test file passed all 11 tests.
- The server route test file passed all four tests.
- The database and server type checks passed.
- `git diff --check` passed.
- A live app browser test returned one ID for sequential and concurrent exact duplicates.
- The live browser test reproduced the path-alias duplicate.
- All GitHub checks passed.

I found no separate authorization or data exposure issue. The path-alias issue can still give two project identities access to one folder.

PR #1278, opened 2026-08-10
Fix terminal stealing new thread focus

### LINKED ISSUE #1160: Focus is stolen by right panel when creating a new session
Repro steps:
1. Existing thread, open right panel with terminal
2. Create a new thread

Expected:
- The new thread input is focused so I can type

Actual:
- The terminal panel remains open and that steals the focus. 
- The terminal panel can stay open, it just shouldn't have focus

ymichael:
## Summary

- distinguish passive terminal and new-tab panel remounts from explicit focus requests
- preserve composer or newer user focus while xterm initializes asynchronously
- retain autofocus for explicit terminal creation, terminal tab activation, and new-tab opens
- consume new-tab autofocus requests so later passive remounts stay inert
- cover both terminal focus race orderings plus passive and explicit new-tab focus behavior

## Testing

- `pnpm exec turbo run test --filter=@bb/app --force -- src/components/secondary-panel/NewTabFileSearch.test.tsx src/components/secondary-panel/NewTabPage.test.tsx src/components/thread/terminal/ThreadTerminalView.test.ts src/components/thread/terminal/ThreadTerminalContent.test.tsx`
- `pnpm exec turbo run typecheck --filter=@bb/app --force`
- `pnpm exec turbo run lint --filter=@bb/app --force` (0 errors; 141 existing warnings)

Fixes #1160

### DIFF
--- a/apps/app/src/components/thread/terminal/ThreadTerminalView.tsx
+++ b/apps/app/src/components/thread/terminal/ThreadTerminalView.tsx
@@ -38,6 +38,28 @@ const TERMINAL_SELECTION_DRAG_DIRECTION_THRESHOLD_PX = 4;
 
 type TerminalFitScheduler = () => void;
 
+interface ShouldFocusTerminalAfterAsyncMountArgs {
+  currentFocusIsAvailable: boolean;
+  hasExplicitFocusRequest: boolean;
+  focusMovedDuringMount: boolean;
+  isPanelOpen: boolean;
+}
+
+export function shouldFocusTerminalAfterAsyncMount({
+  currentFocusIsAvailable,
+  hasExplicitFocusRequest,
+  focusMovedDuringMount,
+  isPanelOpen,
+}: ShouldFocusTerminalAfterAsyncMountArgs): boolean {
+  if (!isPanelOpen) {
+    return false;
+  }
+  if (focusMovedDuringMount && currentFocusIsAvailable) {
+    return false;
+  }
+  return hasExplicitFocusRequest || !currentFocusIsAvailable;
+}
+
 interface WebglRendererAddon extends ITerminalAddon {
   onContextLoss: (listener: () => void) => IDisposable;
 }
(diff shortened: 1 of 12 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I will check security, code quality, performance, architecture, and the user flow.

apps/app/src/components/thread/terminal/ThreadTerminalView.tsx
```diff
@@ -679,8 +713,24 @@ export function ThreadTerminalView({
...
+            currentActiveElement !== null &&
+            currentActiveElement !== document.body &&
+            currentActiveElement.isConnected,
+          hasExplicitFocusRequest: autoFocusRef.current,
+          focusMovedDuringMount: currentActiveElement !== activeElementAtMount,
+          isPanelOpen: isPanelOpenRef.current,
+        })
+      ) {
         terminal.focus();
+        onAutoFocusHandledRef.current?.();
```
SawyerHood, line 733: 🚨 `slopcop/review` — Consume a denied focus request so it cannot affect a later mount.

If focus moves during the asynchronous load, this condition rejects terminal focus. The callback then does not run. Both parents keep the request true. A later terminal mount can reuse that old request and steal focus. Another set to true also cannot start a new request. Consume each request after this decision. Use a request ID and terminal ID to stop an old mount from clearing a new request. Add a lifecycle test for a denied request and a later mount.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

The terminal loads after the page changes. This patch checks whether another input gained focus. It then leaves that input alone.

### Finding

- **P2:** A denied focus request stays active. A later terminal mount can reuse it and steal focus. A later terminal action also cannot start a new request. Consume every request after the focus decision. Use a request ID with the terminal ID. Add a lifecycle test for the rejected request and later mount.

### Other review results

- I found no second security issue.
- I found no performance regression. The new checks use constant time.
- The two parent views duplicate the new request state. A shared hook can own the request ID and consume rules. This refactor is optional.
- The type check passed. The focused tests passed all 16 tests. The lint task passed with zero errors and 141 existing warnings.
- The full app suite passed 2,463 of 2,464 tests. One prompt-box test outside this diff failed.
- The browser smoke test loaded the real app. It did not complete the terminal-to-new-thread focus sequence.
- The final GPT-5.6 review confirmed the P2 finding.

PR #1279, opened 2026-08-10
Publish @get-bb/plugin-sdk to npm

### LINKED ISSUE #1134: @bb/plugin-sdk is documented for external tests but unavailable on npm
## Problem

The Plugin SDK README documents `@bb/plugin-sdk` as the supported package for testing plugins outside the BB monorepo:

```sh
npm install --save-dev @bb/plugin-sdk vitest better-sqlite3 zod
npm install --save-dev react react-dom @testing-library/react jsdom
```

The package on `main` exports executable runtime JavaScript and portable declarations for:

- `@bb/plugin-sdk`
- `@bb/plugin-sdk/app`
- `@bb/plugin-sdk/testing`
- `@bb/plugin-sdk/testing/app`

However, the documented installation currently fails:

```text
$ npm view @bb/plugin-sdk version
npm error code E404
npm error 404 Not Found - GET https://registry.npmjs.org/@bb%2fplugin-sdk
```

The same E404 occurs when trying to install the package in a clean external plugin.

## Context

[PR #716](https://github.com/get-bb/bb/pull/716) merged the external testing distribution, including packed-package tests, portable declarations, and external scaffold fixtures. The current README says installing this package is required when an external plugin imports the testing subpaths.

The repository currently publishes `bb-app`, but I could not find corresponding release automation for `packages/plugin-sdk`.

## Expected behavior

There should be one documented, supported way for an external plugin to install and run the official backend and frontend test harnesses.

If `@bb/plugin-sdk` is intended to be public, publish it and keep its release version aligned with the SDK version embedded in BB.

If the package is intentionally not published independently, update the documentation and scaffolds to point to the actual supported distribution mechanism.

## Acceptance criteria

- The documented install command works in a clean directory outside the BB repository.
- `@bb/plugin-sdk/testing` and `@bb/plugin-sdk/testing/app` load their runtime JavaScript and declarations without resolving unpublished BB workspace packages.
- Release automation verifies the package contents and performs a clean-room install/test before publication.
- The published package version and BB's `engines.bbPluginSdk` guidance remain compatible.
- If public npm publication is not intended, the README clearly documents the supported alternative and does not instruct authors to install a nonexistent package.

## Current code and documentation

- [`packages/plugin-sdk/package.json`](https://github.com/get-bb/bb/blob/main/packages/plugin-sdk/package.json)
- [External plugin tests](https://github.com/get-bb/bb/blob/main/packages/plugin-sdk/README.md#external-plugin-tests)
- [Fidelity boundaries](https://github.com/get-bb/bb/blob/main/packages/plugin-sdk/README.md#fidelity-boundaries)

ymichael:
## Summary

- prepare `@get-bb/plugin-sdk@0.5.0` for public npm publication with public package metadata and synchronized plugin compatibility ranges
- rename the workspace/public package imports to the project-owned `@get-bb` scope while keeping `@bb/plugin-sdk` and `@bb/plugin-sdk/app` as supported host aliases for existing external plugins
- install `cron-parser` as a package dependency and declare Hono as a required peer, so clean npm consumers get both while BB and the harness share one Hono type/runtime
- add a clean npm-consumer smoke test that packs, installs, and exercises both testing runtimes, then make the Trusted Publishing workflow publish that exact validated tarball
- document the independent release process, legacy-import compatibility, and the one-time `@get-bb` npm organization/package bootstrap required before Trusted Publishing can take over

Existing plugins do not need an import-only rewrite: the backend builder/runtime and frontend runtime shim accept both package names. They still need to declare compatibility with SDK `0.5.0` and migrate any genuinely breaking API usage.

The minor bump is intentional: the checked-in public contract has accumulated breaking changes since `0.4.1`, so publishing the current declarations under the old version would make incompatible hosts and plugins report the same SDK version.

## Validation

- `pnpm install --frozen-lockfile`
- `node packages/templates/scripts/generate-templates.mjs --check`
- `node packages/plugin-build/scripts/generate-runtime-export-manifest.mjs --check`
- `pnpm exec turbo run typecheck --force`
  - 58/58 tasks
- affected package test suites
  - 74 plugin SDK tests
  - 23 template/scaffold tests, including a clean external npm install
  - 11 plugin-build tests passed, 1 skipped
  - 383 CLI tests
  - 1,389 server tests
  - 2,458 app tests
- `pnpm exec turbo run smoke:tarball --filter=@get-bb/plugin-sdk --force`
  - clean npm install of the packed tarball
  - SQLite, cron, Hono, React, and jsdom harness coverage
- `npm publish .tmp/plugin-sdk-release/bb-plugin-sdk.tgz --dry-run --access public --tag latest`
  - identifies the artifact as `@get-bb/plugin-sdk@0.5.0`
- `git diff --check`

Fixes #1134


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the Slop Cop. I started a review for security, code quality, performance, architecture, and practical tests.

.github/workflows/publish-plugin-sdk.yml
```diff
@@ -0,0 +1,208 @@
...
+        description: Run validation and npm publish --dry-run only.
+        required: true
+        type: boolean
+        default: true
+
+run-name: Publish @bb/plugin-sdk (${{ inputs.npm_tag }}${{ inputs.dry_run && ', dry run' || '' }})
+
+permissions:
+  contents: read
+  id-token: write
```
SawyerHood, line 30: 🚨 `slopcop/review` — Limit npm publish authority to the publish step.

This job gives each install, build, test, and dry-run step access to GitHub OIDC. A compromised dependency can mint the npm identity.

Use an unprivileged validation job. Give `id-token: write` and the release environment only to a separate non-dry-run publish job.

packages/plugin-sdk/scripts/smoke-tarball.mjs
```diff
@@ -0,0 +1,256 @@
...
+    `${JSON.stringify({ name: "plugin-sdk-smoke", private: true, type: "module" }, null, 2)}\n`,
+  );
+  await runCommand({
+    args: [
+      "install",
+      "--save-dev",
+      "--no-audit",
+      "--no-fund",
+      tarball,
+      "vitest",
```
SawyerHood, line 198: 🚨 `slopcop/review` — Pin the clean-consumer dependency set.

This release test installs current registry versions without a lock file. Some packages also run install scripts.

The same job has npm publish authority. A changed dependency can replace the tarball or publish another package before the intended step.

Use a committed consumer fixture with exact versions and `npm ci`. Run it without publish authority.

.github/workflows/publish-plugin-sdk.yml
```diff
@@ -0,0 +1,208 @@
...
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 24
+          registry-url: https://registry.npmjs.org
+
+      - name: Ensure npm supports trusted publishing
+        run: |
+          npm_version="$(npm --version)"
+          node - "$npm_version" <<'NODE' || npm install --global npm@latest
```
SawyerHood: 🚨 `slopcop/review` — Do not install `npm@latest` in the privileged release job.

This fallback runs new registry code that the commit and lock file do not control. That code receives the job publish authority.

Use a fixed Node release with a supported npm version. Otherwise, install an exact reviewed npm version before the privileged job.

.github/workflows/publish-plugin-sdk.yml
```diff
@@ -0,0 +1,208 @@
...
+
+    steps:
+      - name: Require main branch
+        if: ${{ github.ref_name != 'main' }}
+        run: |
+          echo "::error::Publish @bb/plugin-sdk must run from main, got ${GITHUB_REF_NAME}."
+          exit 1
+
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
SawyerHood, line 51: 🚨 `slopcop/review` — Pin privileged workflow actions to commit SHAs.

The `v4` tag can move. A changed action runs with the npm OIDC permission in this job.

Pin each action to a reviewed full commit SHA. Keep the version name in a comment for update tools.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: This PR puts the BB plugin developer kit on npm and adds a manual release button.

I found two blocking security problems. I also found two related hardening gaps.

## Findings

1. High: The complete job receives `id-token: write`.

Every install, build, test, and dry-run step can request the npm trusted-publisher identity. A compromised dependency can publish before the intended step.

Use an unprivileged validation job. Give OIDC and the `npm-release` environment only to a separate non-dry-run publish job.

2. High: The tarball smoke test installs current registry packages without a lock file.

These packages include install scripts. They run in the same job that has npm publish authority.

Use a committed consumer fixture with exact versions and `npm ci`. Run this fixture without publish authority.

3. Medium: The privileged job uses mutable action tags.

Pin `actions/checkout`, `pnpm/action-setup`, and `actions/setup-node` to reviewed commit SHAs.

4. Medium: The job can install `npm@latest` under publish authority.

Use a fixed Node release with a supported npm version. Otherwise, install one exact reviewed npm version.

## Architecture and performance

The new workflow repeats release setup and validation from `publish-bb-app.yml`. A shared validation script can reduce drift while permission blocks stay explicit.

The two harness dependencies add about 1.5 MB and 597 files to each SDK install. This cost is non-blocking.

The release job also repeats package builds and clean-consumer tests. The standalone smoke path took about 28 seconds with a warm cache.

## Validation

- The focused type checks passed.
- All 74 plugin SDK tests passed.
- All 23 template tests passed.
- The clean-consumer tarball smoke test passed with npm 11.13.0.
- `npm publish --dry-run` passed for the exact tarball.
- The tarball contains 14 files and uses 121,209 bytes.
- `git diff --check` passed.
- No browser test applies because this PR adds no UI route.
- GitHub package CI has one Tasks UI test failure. It appears outside this release path.

.github/workflows/publish-plugin-sdk.yml
ymichael: Thanks. This workflow intentionally follows the existing `publish-bb-app.yml` release model, which also grants OIDC to one install/build/test/publish job. The npm Trusted Publishing guide likewise shows a single job. We are keeping that repository-wide posture here; splitting jobs and handing off artifacts should be evaluated for both publishers in a dedicated hardening change rather than added only to this SDK workflow.

packages/plugin-sdk/scripts/smoke-tarball.mjs
ymichael: The unlocked install is intentional: this smoke test exercises the README commands against the versions a clean consumer resolves today. A committed lockfile would test a different invariant. Given the accepted single-job release posture above, we are keeping this compatibility smoke rather than adding a second fixture.

.github/workflows/publish-plugin-sdk.yml
ymichael: Addressed in 373398b46. Both the SDK and existing bb-app workflows now fail clearly below npm 11.5.1 and never install `npm@latest` in a release job.

.github/workflows/publish-plugin-sdk.yml
ymichael: The repository consistently uses versioned action tags across its workflows, including the existing bb-app publisher. Pinning only this new workflow would create a one-off maintenance policy. We are keeping the current convention; full-SHA pinning can be adopted repository-wide separately.

Issue #1114, opened 2026-08-06
No way to edit a sent message or undo the last turn


iamEvanYT:
## Summary

There's no way to edit a user message after it's been sent, or to undo the last turn (user prompt + assistant reply) and put that prompt back in the composer.

## Environment

- bb 0.35.1

## Problem

In most chat UIs you can tweak the last thing you said — edit it, resend, and the old turn goes away. In bb, once a message is on the timeline it's stuck there.

What I want day-to-day:

1. Realize the last prompt was wrong / incomplete
2. Undo that turn (my message and the reply after it)
3. Get the text back in the composer so I can fix it and send again

Right now the closest option is forking at an earlier `sourceSeqEnd`, which creates a new thread and leaves the old one around. That works as a workaround if you're building a plugin, but it's not an edit/undo affordance in the product — and there's no plugin/SDK API to truncate or rewrite history on the current thread either.

## Expected

Something like:

- Edit on a user message (or an "undo last turn" action) that removes that message and everything after it from the current thread
- Puts the original text back in the composer
- Ideally also exposed to plugins (SDK / composer action surface) so third-party undo/edit tools don't have to fake it with fork + archive

Forking is fine for branching a conversation. It's a rough substitute for "oops, let me rephrase that."

iamhenry:
+1 especially for opencode

ymichael:
thanks for creating this issue! I'm v interested in adding this!

ryanbbrown:
+1 would love for this to be added! for my usage, ideally would work to "rewind" to any past user message rather than only the immediately previous; @ymichael do you think it would be easy to include that?

ymichael:
I added this for codex, claude code and pi (and it should work for any past user message!) - this should enabled for everyone in the next version

Unfortunately for ACP based providers, this isn't supported for now

PR #1280, opened 2026-08-10
Add experimental sent-message editing

ymichael:
## Summary

- add a default-off `editMessages` experiment enforced by both the app and server
- let users edit any eligible completed message without changing history until submission
- reuse the queued-message inline editor styling; opening the editor focuses it without selecting its contents
- support Codex, Claude Code, and Pi through provider-native transcript checkpoints
- expose the atomic edit through the SDK and `bb thread edit-message`
- bump the host daemon protocol

## Behavior

Opening or cancelling the inline editor is client-local and non-destructive. Submitting an edit stages provider history through the turn before the selected message, atomically replaces the selected conversation suffix, keeps workspace changes, and starts the edited turn.

Claude Code uses the latest assistant UUID with `forkSession` and `upToMessageId`. Pi records the session leaf entry and branches through it. Legacy Claude Code or Pi history without a checkpoint rejects safely when a retained turn is required.

## Review hardening

- hash external operation IDs before deriving provider staging identities, preventing Pi path-sanitization collisions
- preserve and validate agent caller identity through the contract, CLI, edit service, persisted replacement turn, and permission policy
- recheck queued messages, pending interactions, and the editable event high-water mark in the atomic rewrite transaction
- keep the timeline edit handler stable so every cached row regains its action after an active turn becomes idle
- let connected realtime own history-rewrite refreshes, retaining mutation invalidation only as the disconnected fallback
- replace the thread-only search-segment index with `(thread_id, source_seq)` and add a generated migration plus exact query-plan coverage

## Validation

- cross-package typechecks across the app, server, runtime, CLI, SDK, contracts, daemon, plugin SDK, templates, database, domain, and thread view
- full affected-package tests: 2,491 app tests, 1,400 server tests, and 857 agent-runtime tests, plus CLI, daemon, contracts, SDK, database, domain, and thread-view suites
- app lint and generated plugin SDK/template checks
- latest review-fix pass: 106 focused app/server/runtime/CLI tests, all 375 database tests, and 227 server-contract/SDK/plugin-SDK/template tests
- all 10 directly affected Turbo typecheck tasks pass; lint reports no errors
- credentialed standalone-dev-server QA passed through the source-built CLI for Codex, Claude Code, and Pi: editing an earlier message replaced the suffix, reran the edited turn, and preserved the replacement context on the next turn
- Pi QA confirmed the rewind session was adopted as the canonical provider session without leaving an operation-suffixed staging file

The Pi smoke pass used `openai-codex/gpt-5.6-luna` because the configured Anthropic route was out of extra usage.

## Simplification pass

- omitting `expectedRequestSequence` on POST `/edit-message` now atomically edits the latest eligible message; the read-only `GET /latest-message-edit` route, `threads.getLatestMessageEdit`, and the CLI's GET-then-POST resolution are removed (the CLI's edit-latest is no longer racy)
- staged rewind forks are keyed by the server-minted per-attempt `leaseId`: each attempt owns exactly one fork, which removes the cross-attempt lease refcounting, join validation, and staging-id hashing, drops `operationId` from both daemon rewind commands, and replaces the runtime's provider allowlist with the `supportsFork` capability
- the edit service parses stored events through `parseStoredEvent` with type narrowing, reuses `hasRootStoredTurnStarted`, builds the rewind-prepare command with a destructure-rest spread, and resolves the editable-turn scan in one query
- the queued and sent-message inline editors share one composer builder and one blocked-reason model; the inline-queued attachment path shares `useDraftAttachmentUploads`; one-caller helpers and redundant prop round-trips are inlined
- render-wiring and duplicate tests were pruned across timeline actions, settings, CLI, and the Pi bridge
- net effect: the PR's hand-written diff (excluding generated Drizzle snapshot/bundled types and the plan doc) is ~5,150 insertions, down from ~5,900 before the pass, with all affected typechecks and full test suites passing (server 1,426, app 2,508, agent-runtime 904, daemon 523, plus CLI/SDK/contracts/templates)

Related to #1114


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

apps/server/src/services/threads/thread-edit-message.ts
```diff
@@ -0,0 +1,542 @@
...
+        }),
+        hostId: readyEnvironment.hostId,
+        timeoutMs: LIVE_DAEMON_COMMAND_TIMEOUT_MS,
+      });
+    stagedProviderThreadId = prepared.providerThreadId;
+  }
+
+  const requestSequence = target.oldMaxSequence + 2;
+  await sendThreadMessage(deps, {
+    beforeAppendInTransaction: ({ tx }) => {
```
SawyerHood: 🚨 `slopcop/review` — Recheck all edit guards after host preparation

The queue and interaction checks occur before the awaited host rewind. Another client can add either state without changing the event high-water mark.

This callback checks only the target events. The edit can commit, and a queued request can later run on the rewritten conversation.

A normal send can also prepare from the old idle thread during this wait. Use one thread lock across edit, send, and queue paths.

Also recheck the queue and pending interactions in this transaction. Add tests that pause the rewind before each concurrent action.

packages/server-contract/src/api/threads.ts
```diff
@@ -272,6 +272,34 @@ export type ContinueAfterProviderRateLimitResponse = z.infer<
   typeof continueAfterProviderRateLimitResponseSchema
 >;
 
+export const editMessageRequestSchema = sendMessageRequestSchema
+  .omit({ mode: true, senderThreadId: true })
+  .extend({
+    operationId: z.string().min(1),
```
SawyerHood, line 278: 🚨 `slopcop/review` — Make the staging identifier collision-safe

The runtime puts this external value into the provider thread ID. Pi replaces each unsafe character with an underscore for its session file.

Thus, `a/b` and `a?b` use the same file. Concurrent edits at different checkpoints can prepare the wrong provider history.

Require a UUID at this boundary, or hash the value for the staging ID. Add a Pi test with two current collisions.

apps/app/src/views/thread-detail/sentMessageEdit.ts
```diff
@@ -0,0 +1,52 @@
+import type { ThreadRuntimeDisplayStatus } from "@bb/domain";
+
+export interface SentMessageEditAvailability {
+  activeBackgroundCommandCount: number;
```
SawyerHood: 🚨 `slopcop/review` — Match the server background-agent guard

The server rejects edits while a background agent runs. This client policy checks workflows and background commands, but not background agents.

The UI can show an edit action that always fails on submit. Add the active background-agent count to this policy and its tests.

packages/db/src/data/events.ts
```diff
@@ -188,6 +195,62 @@ export interface CompletedStoredTurnRow {
...
+  db.delete(promptHistoryEntries)
+    .where(
+      and(
+        eq(promptHistoryEntries.threadId, args.threadId),
+        gte(promptHistoryEntries.requestSequence, args.cutoffSequence),
+        lte(promptHistoryEntries.requestSequence, args.oldMaxSequence),
+      ),
+    )
+    .run();
+  db.delete(threadSearchSegments)
```
SawyerHood, line 226: 🚨 `slopcop/review` — Add an index for the sequence-range delete

This delete filters by `threadId` and a `sourceSeq` range. The schema has only a `threadId` index for this table.

An edit near the end of a long thread can scan every search segment. The scan holds the immediate transaction write lock longer.

Change the index to `(threadId, sourceSeq)` with a generated migration. The new index still supports thread-only lookups.

apps/app/src/views/thread-detail/ThreadDetailPromptArea.tsx
```diff
@@ -1172,6 +1368,60 @@ export function ThreadDetailPromptArea({
...
+          environmentSummary={null}
+          contextWindowUsage={null}
+          execution={compactExecutionConfig}
+          executionReadOnly
+          permission={bottomPermissionConfig}
+          permissionReadOnly
+          typeahead={typeaheadConfig}
+          promptActions={promptActions}
+          zenModeResetKey={`sent-message:${sentMessageEdit.operationId}`}
+          focusEndKey={`${sentMessageEdit.operationId}:${editFocusNonce}`}
```
SawyerHood: 🚨 `slopcop/review` — Select the replacement text when the editor opens

`focusEndKey` puts the caret at the end. It does not select the restored message.

I reproduced this twice in Chromium. The editor had focus, but its selection was empty at the final character.

Add a select-all focus mode for this editor. Add a test that checks the editor selection, not only its text.

apps/app/src/hooks/mutations/thread-runtime-mutations.ts
```diff
@@ -209,6 +215,25 @@ export function useSendThreadMessage() {
...
+  const queryClient = useQueryClient();
+  return useMutation({
+    meta: {
+      errorMessage: "Failed to edit the message.",
+      lifecycleOperation: "edit_message",
+      showErrorToast: false,
+    },
+    mutationFn: ({ id, ...request }: EditMessageMutationRequest) =>
+      sdk.threads.editMessage({ threadId: id, ...request }),
+    onSuccess: (_result, variables) => {
```
SawyerHood, line 228: 🚨 `slopcop/review` — Do not reload history twice

The server sends `history-rewritten` before the HTTP response. Realtime already reloads the timeline, outline, search, histories, and open turn details.

This success handler reloads the same data again for connected clients. Some turn-detail reloads run several database projections.

Use the same realtime connection check as the send mutation. Keep this handler only as the disconnected fallback.

packages/server-contract/src/api/threads.ts
```diff
@@ -272,6 +272,34 @@ export type ContinueAfterProviderRateLimitResponse = z.infer<
   typeof continueAfterProviderRateLimitResponseSchema
 >;
 
+export const editMessageRequestSchema = sendMessageRequestSchema
+  .omit({ mode: true, senderThreadId: true })
```
SawyerHood: 🚨 `slopcop/review` — Preserve the agent caller identity

This schema removes `senderThreadId`, and the new CLI command does not derive the current thread. An agent can target another root thread.

The service then sets `initiator: "user"`. The permission policy returns `ask` instead of the `deny` policy for agent work.

Keep and validate the sender identity. Derive the initiator as the send path does. Test parent, child, self, and unrelated targets.

apps/app/src/views/thread-detail/ThreadDetailView.tsx
```diff
@@ -2593,6 +2803,10 @@ function ThreadDetailViewInternal(props: ThreadDetailViewInternalProps) {
             isThreadTimelinePending,
             timelineError: Boolean(timelineError),
             onForkMessage: isForkAvailable ? handleForkMessage : undefined,
+            onEditMessage: canEditSentMessages
```
SawyerHood, line 2825: 🚨 `slopcop/review` — Restore edit actions on all rows after a turn

This condition removes the handler while a turn runs. After completion, only the new latest row receives the handler.

I reproduced this twice. Each earlier message lost its edit action until a full page reload restored all actions.

Keep a stable handler with a click guard, or make the row cache depend on availability. Add an active-to-idle test with two messages.

packages/agent-runtime/src/runtime.ts
```diff
@@ -1238,6 +1263,165 @@ function createAgentRuntimeInternal(
...
+            const providerThreadId = resolveThreadIdentityResult({
+              result,
+              threadId: stagingThreadId,
+            });
+            if (!providerThreadId) {
+              throw new Error(
+                `${providerId} did not return a provider thread for rewind operation ${operationId}`,
+              );
+            }
+            detachedProviderThreadIds.add(providerThreadId);
```
SawyerHood: 🚨 `slopcop/review` — Release each prepared provider session

Each rewind adds one entry to this set and one entry to `preparedThreadRewinds`. No code removes either entry.

Claude and Pi also start a real staging session. The `finally` block removes only runtime identity state and never stops that session.

Add consume and discard operations. Stop the staging session and remove both entries after success, conflict, error, or a time limit.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: This feature lets a user change an old message. BB removes later chat events and asks the provider to answer again.

I found three defects that must be fixed before merge.

- The edit waits for a host rewind without a shared thread lock. A send, queued message, or interaction can change state during that wait.
- The edit API removes the agent caller identity. An agent can edit another root thread and start the replacement with user permission policy.
- The API accepts any operation ID. Two different IDs can map to the same Pi session file and use the wrong checkpoint.

I also found six other defects.

- Each prepared rewind keeps a provider session and two runtime entries until the daemon restarts.
- The inline editor puts the caret at the end. It does not select the existing message.
- A completed new turn restores the edit action only on the latest message. A page reload restores actions on earlier messages.
- A connected client reloads history after realtime already reloads the same history.
- The search suffix delete has no `(threadId, sourceSeq)` index. It can scan a full long-thread search history under the write lock.
- The client does not check the active background-agent count. It can show an action that the server rejects.

The supported provider list also appears in the client, server, and runtime. A shared provider capability would reduce policy drift.

I tested the feature with the local app and a real Codex thread. A first-message edit worked, and an earlier edit removed the complete later suffix.

The final provider output was `FINAL-REPLY`. The removed secret and second turn were absent from the server timeline.

The five package type checks passed. The 68 focused app, server, and runtime tests passed.

packages/server-contract/src/api/threads.ts
ymichael: Addressed in 3e3c82a70. External operation IDs are SHA-256 hashed before deriving the provider staging thread ID. The runtime test prepares `a/b` and `a?b` concurrently and verifies that both raw staging identities and their Pi-sanitized forms remain distinct.

apps/server/src/services/threads/thread-edit-message.ts
ymichael: Addressed in 3e3c82a70. The commit transaction now rechecks pending interactions and queued messages alongside the existing editable-target and event high-water guard. A regression test injects a pending interaction after rewind preparation and verifies the rewrite is rejected and the staged provider session is discarded. Concurrent sends change the guarded event state and are rejected by the same transaction preflight.

apps/app/src/views/thread-detail/sentMessageEdit.ts
ymichael: Addressed in a5f6f5831. Sent-message edit availability includes the active background-agent count, and the policy test covers that state.

apps/app/src/hooks/mutations/thread-runtime-mutations.ts
ymichael: Addressed in 3e3c82a70. Connected clients now rely on the `history-rewritten` realtime refresh. Mutation success invalidation remains only for disconnected or reconnecting clients, with tests for both paths.

packages/db/src/data/events.ts
ymichael: Addressed in 3e3c82a70. The index is now `(thread_id, source_seq)` with generated migration 0091. A query-plan regression test exercises the exact suffix-delete predicate and verifies SQLite uses the composite index without scanning the table.

apps/app/src/views/thread-detail/ThreadDetailPromptArea.tsx
ymichael: Closing as an intentional product decision. The requested v1 behavior is focus-only when the editor opens; select-all was deliberately backed out after testing, so the current caret behavior is expected.

packages/server-contract/src/api/threads.ts
ymichael: Addressed in 3e3c82a70. `senderThreadId` is retained in the edit contract, derived by the CLI from `BB_THREAD_ID`, validated through the same sender resolver as normal sends, persisted on the replacement request, and used to select the agent permission policy. Self normalizes to user; any other live sender is relation-independent by the existing cross-thread messaging contract. Tests cover agent attribution and permission policy, an invalid sender, and CLI self versus other-thread forwarding.

packages/agent-runtime/src/runtime.ts
ymichael: Addressed in a5f6f5831 and subsequent Pi hardening. Prepared rewinds have explicit consume and discard paths, staging sessions are stopped, identity and prepared-state entries are removed on settlement, failed cleanup is retried, and abandoned preparations expire through the TTL sweep. Tests cover success, conflict/error cleanup, retry, and ambiguous Pi identity cleanup.

apps/app/src/views/thread-detail/ThreadDetailView.tsx
ymichael: Addressed in 3e3c82a70. The timeline receives a stable edit handler that reads current availability and thread state through a ref at click time. The active-to-idle regression test keeps the same cached two-row timeline, verifies both edit actions return, and exercises the earlier message.

PR #1026, opened 2026-08-05
Fix mobile sidebar swipe over timeline prose

ymichael:
## Summary

- allow the compact sidebar swipe gesture to begin over selectable timeline message prose
- cancel a pending sidebar swipe when native text selection appears within that message
- preserve existing exclusions for controls and horizontal scroll regions
- add regression coverage for swipe opening and long-press selection arbitration

## Root cause

Selectable message prose was blanket-marked with `data-no-sidebar-swipe`, so most of the thread timeline was a gesture dead zone. This replaces the blanket opt-out with intent-based selection arbitration.

## Validation

- `pnpm exec turbo run test --filter=@bb/app --force` (312 files, 2,347 tests)
- `pnpm exec turbo run typecheck --filter=@bb/app`
- mobile dev preview verified healthy through bb Connect

PR #1281, opened 2026-08-10
Fix multi-second mobile tap freeze from sidebar swipe layout probe

### LINKED ISSUE #1269: Mobile version hangs for several seconds for each tap
*.getbb.app is close to unusable at the moment for me as each tap takes several seconds to register. Here's a video showing it (each time the circle moves I'm tapping):

https://github.com/user-attachments/assets/e3d1b43b-f2a2-4a8b-a85a-b74a718aa8b3

There is an interesting thing to note if you look at the animations on the thread that are ongoing. They also freeze each time I tap, indicating that something may be blocking the main thread.

SawyerHood:
## Summary

- Defer the sidebar swipe recognizer's horizontal-scroll-region probe off the tap path. It now runs at most once per gesture, at the first `touchmove` that shows horizontal intent — never on `pointerdown`/`touchstart`.
- Keep the probe on the wheel path, but only after the direction test, so vertical trackpad scrolls stop paying it.
- Add regression tests: the tap path performs zero forced-layout reads, and a swipe that starts inside a horizontal scroll region still yields to native scroll.

Fixes #1269

## Root cause

The mobile swipe recognizer listens for `pointerdown` and `touchstart` on `document` (capture). On every tap it called `isInsideHorizontalScrollRegion`, which pairs `getComputedStyle(...).overflowX` with a `scrollWidth`/`clientWidth` read for each ancestor. Each pair forces a synchronous style + layout pass of the whole document, and the handler runs for both events, so one tap paid the flush twice. On a large streamed thread (un-virtualized timeline, styles dirty almost every frame) that flush blocks a mobile main thread for seconds — which also freezes the running animations, exactly as reported.

The path became hot in #1026, which replaced the blanket `data-no-sidebar-swipe` opt-out on message prose with runtime arbitration. After that, taps on prose — most of the mobile screen — fell through to the expensive walk.

## Performance

Measured in Chromium at a 390×844 viewport against a 7.5k-node DOM inside `SidebarInset`, with styles dirtied before each tap (as a streamed thread does), 15 trials per variant on the same DOM:

| Variant | Tap dispatch median | Tap dispatch max |
|---|---|---|
| Before | 75.3 ms | 110.7 ms |
| After | ~0 ms | 0.5 ms |

The pre-fix cost equals one full forced style+layout flush (~75–95 ms on a desktop CPU); phone CPUs are several times slower and real threads have far more nodes, which produces the multi-second freezes in the issue.

## Behavior verification

In a real browser at a compact viewport: a right swipe over message prose still opens the sidebar; a swipe that starts inside a horizontally scrollable region still does not.

## Testing

- `pnpm exec turbo run test --filter=@bb/app` — 326 files, 2,459 tests pass, including two new regression tests in `sidebar.test.tsx`.

### DIFF
--- a/apps/app/src/components/ui/sidebar.test.tsx
+++ b/apps/app/src/components/ui/sidebar.test.tsx
@@ -97,6 +97,41 @@ describe("mobile sidebar text-selection arbitration", () => {
     expect(document.querySelector('[data-sidebar="panel"]')).not.toBeNull();
   });
 
+  it("defers the horizontal-scroll-region probe until horizontal intent", () => {
+    render(
+      <CompactViewportOverrideProvider isCompactViewport>
+        <SidebarProvider>
+          <Sidebar>Sidebar content</Sidebar>
+          <SidebarInset>
+            <div data-testid="scroller" style={{ overflowX: "auto" }}>
+              <div data-sidebar-swipe-selectable>Wide code block</div>
+            </div>
+          </SidebarInset>
+        </SidebarProvider>
+      </CompactViewportOverrideProvider>,
+    );
+    const scroller = screen.getByTestId("scroller");
+    let scrollWidthReads = 0;
+    Object.defineProperty(scroller, "scrollWidth", {
+      get: () => {
+        scrollWidthReads += 1;
+        return 500;
+      },
+    });
+    Object.defineProperty(scroller, "clientWidth", { get: () => 100 });
+    const prose = screen.getByText("Wide code block");
+
+    fireTouch(prose, "touchstart", createTouch(120, 160));
+
+    // The tap path must stay free of forced layout reads (#1269).
+    expect(scrollWidthReads).toBe(0);
+
+    fireTouch(window, "touchmove", createTouch(260, 164));
+
+    expect(scrollWidthReads).toBeGreaterThan(0);
+    expect(document.querySelector('[data-sidebar="panel"]')).toBeNull();
+  });
+
   it("cancels a pending prose swipe when native text selection begins", () => {
     let hasSelection = false;
     let selectionNode: Node | null = null;
--- a/apps/app/src/components/ui/sidebar.test.tsx
+++ b/apps/app/src/components/ui/sidebar.test.tsx
@@ -45,6 +45,53 @@ function fireTouch(
   fireEvent(target, event);
 }
 
+function firePointer(
+  target: Element | Document | Window,
+  type: "pointerdown" | "pointermove",
+  clientX: number,
+  clientY: number,
+) {
+  const event = new Event(type, { bubbles: true, cancelable: true });
+  Object.defineProperties(event, {
+    pointerId: { value: 1 },
+    pointerType: { value: "touch" },
+    isPrimary: { value: true },
+    button: { value: 0 },
+    buttons: { value: 1 },
+    clientX: { value: clientX },
+    clientY: { value: clientY },
+  });
+  fireEvent(target, event);
+}
+
+function renderScrollerSwipeHarness() {
+  render(
+    <CompactViewportOverrideProvider isCompactViewport>
+      <SidebarProvider>
+        <Sidebar>Sidebar content</Sidebar>
+        <SidebarInset>
+          <div data-testid="scroller" style={{ overflowX: "auto" }}>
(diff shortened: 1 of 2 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the Slop Cop. I am reviewing this pull request for security, code quality, performance, architecture, and product behavior.

apps/app/src/components/ui/sidebar.tsx
```diff
@@ -880,6 +886,15 @@ const SidebarInset = React.forwardRef<
           return;
         }
 
+        if (
+          session.startTarget !== null &&
+          isInsideHorizontalScrollRegion(session.startTarget)
```
SawyerHood: 🚨 `slopcop/review` — Cancel the swipe if `startTarget` is no longer connected before this deferred probe.

A live timeline update can replace the target between the gesture start and the first horizontal move. Chromium then returns an empty `overflowX` value for the removed element. The probe fails open, the sidebar opens, and native horizontal scroll stops.

I reproduced this in the mobile app. I removed a CSS-class scroll region after `touchstart`. The probe made zero `scrollWidth` reads, and the sidebar opened on `touchmove`.

Please check `!session.startTarget.isConnected`, clear the session, and return before `preventDefault`. Add touch and pointer tests that replace the scroll region before horizontal intent.

apps/app/src/components/ui/sidebar.test.tsx
```diff
@@ -97,6 +97,41 @@ describe("mobile sidebar text-selection arbitration", () => {
...
+    const prose = screen.getByText("Wide code block");
+
+    fireTouch(prose, "touchstart", createTouch(120, 160));
+
+    // The tap path must stay free of forced layout reads (#1269).
+    expect(scrollWidthReads).toBe(0);
+
+    fireTouch(window, "touchmove", createTouch(260, 164));
+
+    expect(scrollWidthReads).toBeGreaterThan(0);
```
SawyerHood: 🚨 `slopcop/review` — Require exactly one deferred probe.

`toBeGreaterThan(0)` accepts a regression that probes on every move. The intended touch and pointer limit is one probe per gesture. Please use `toBe(1)` and apply the same bound to pointer coverage.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5 summary

The app used to inspect page layout on every mobile tap. This change waits until a person swipes.

The main fix works. However, a live message update can remove the original scroll target during a swipe.

The app can then open the sidebar instead of letting the person scroll code.

### Findings

1. **P2 — Fix the removed-target race before merge.** A live render can disconnect `startTarget` before horizontal intent. Chromium then reports no computed overflow. The guard fails open and captures native horizontal scroll. I reproduced this in the mobile app. Fail closed when the target is not connected, then add touch and pointer regression tests.

2. **P3 — Make the probe-count test exact.** `toBeGreaterThan(0)` allows many layout probes. Use `toBe(1)` to protect the one-probe limit.

### Other review results

The security review found no issue. The architecture scan found no duplicate code that this pull request must refactor.

The wheel path can still probe once per horizontal wheel event. That cost existed before this change. This change removes the cost from vertical wheel events.

A later change can mark horizontal overflow with an observer-backed attribute. The sidebar could then use `closest()` without a layout read.

### Validation

- The focused sidebar tests passed: 5 of 5.
- The `@bb/app` typecheck passed through Turbo.
- A mobile Chromium test used a 390 by 844 viewport.
- Tap starts caused zero `scrollWidth` reads.
- A connected scroll region kept native horizontal scroll.
- Normal prose opened the sidebar.
- A removed scroll region reproduced the race.

I posted this as a comment-only review. I did not approve the pull request or request changes.

apps/app/src/components/ui/sidebar.tsx
SawyerHood: Fixed in a1615d5cf — the deferred probe now cancels the swipe when the start target is no longer connected, with a regression test covering the detach case.

apps/app/src/components/ui/sidebar.test.tsx
SawyerHood: Fixed in a1615d5cf — the touch and pointer tests both assert exactly one probe per gesture (toBe(1)), including a second move event to catch per-move probing.

PR #1290, opened 2026-08-10
Flatten CSS in the Vite dev pipeline

ymichael:
## Summary

- use Vite Lightning CSS transformation in development to match production CSS lowering
- declare Lightning CSS as a direct app development dependency
- flatten Tailwind generated nested conditional declarations for older Safari versions

## Why

Tailwind emits derived theme tokens as nested `@supports` declarations. Production lowers those declarations, but the Vite dev server previously sent them to the browser as nested CSS. Safari versions that support `color-mix()` but mishandle nested conditional declarations then kept fallback ink-colored borders, making the development app look unthemed.

## Verification

- started the standalone Vite dev server and inspected `/src/app.css`; the theme `@supports` rule is top-level

### DIFF
--- a/apps/app/package.json
+++ b/apps/app/package.json
@@ -136,6 +136,7 @@
     "bb-plugin-automations": "workspace:*",
     "eslint": "^9.39.3",
     "eslint-plugin-react-hooks": "^7.0.1",
+    "lightningcss": "^1.32.0",
     "mdast-util-to-hast": "^13.2.1",
     "sharp": "^0.34.5",
     "tailwindcss": "^4.3.0",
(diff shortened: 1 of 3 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** The development server now makes complex CSS rules flat before browsers read them. Older Safari versions can then use theme colors correctly.

I found no correctness, security, quality, or performance problems. I do not have an inline comment.

The development-only option has the correct scope. The dependency and lockfile entries match. A repository search found no duplicate CSS flattening design.

The median CSS transform cost increased by 3.25 ms. This small development cost is reasonable for the compatibility fix.

Typecheck, build, Prettier, the diff check, and all CI checks passed. A live CSS request confirmed the required flat rule structure. Chromium also used the mixed border color.

The full app route check was inconclusive because temporary Vite caches returned 504 responses. This result does not show a regression.

PR #1295, opened 2026-08-10
Serve compressed plugin frontend assets

### LINKED ISSUE #1092: Plugin frontend assets bypass response compression
## Summary

Plugin frontend assets are served without content compression even when the client advertises Brotli and gzip. Core hashed app assets are Brotli-compressed under the same server.

This makes first load of frontend plugins noticeably slow over remote links such as Tailscale.

## Reproduction

On BB 0.35.1:

```bash
curl -sSI -H 'accept-encoding: gzip, br' \
  http://127.0.0.1:38886/assets/<current-core-bundle>.js
# content-encoding: br

curl -sSI -H 'accept-encoding: gzip, br' \
  'http://127.0.0.1:38886/api/v1/plugins/now/assets/app.js?h=<current-hash>'
# no content-encoding
```

Measured examples:

| Bundle | Bytes transferred | gzip size |
| --- | ---: | ---: |
| `now/app.js` | 386,655 | 79,029 |
| `wiki/app.js` | 512,192 | 99,458 |

The plugin response transfers the full raw bundle despite `Accept-Encoding`.

## Expected

Compressible plugin `app.js` and `app.css` responses should use the same negotiated compression behavior as core static assets while preserving their immutable content-hash cache policy.

## Current owner seam

The response is produced by `apps/server/src/routes/plugins.ts` at `GET /plugins/:id/assets/:file` via `context.body(new Uint8Array(bytes), ...)`. Global `compress()` middleware is registered in `apps/server/src/server.ts`, but these responses remain identity-encoded.

Current upstream `main` still uses this response path. This issue does not request plugin-local compression or a proxy workaround; compression should remain owned by BB's asset-serving boundary.

ymichael:
## Summary

- explicitly negotiate Brotli and gzip for plugin app.js and app.css responses
- preserve immutable/no-store caching while adding Vary, Content-Encoding, and exact Content-Length headers
- cache compressed variants in a bounded per-asset LRU and deduplicate concurrent compression work
- share Accept-Encoding quality ranking with the core static asset path
- cover Brotli JavaScript, gzip CSS, rejected encodings, decompressed body integrity, HEAD/GET header parity, and cache lifecycle behavior

## Investigation

The issue reproduction uses curl -I, which sends HEAD. Hono's global compression middleware intentionally skips HEAD, explaining the missing header there. A real GET on current main is gzip-compressed, but unlike core hashed assets it cannot negotiate Brotli.

This change owns compression explicitly at the plugin asset-serving boundary. That closes the real transfer-size gap and also makes HEAD advertise the same representation headers as GET. Plugin app asset paths bypass the outer Hono compression middleware so explicitly rejected encodings remain rejected.

## Validation

- focused plugin/static asset tests: 16 passed
- full server suite: 159 files, 1,393 tests passed
- Prettier and git diff checks passed

Fixes #1092

### DIFF
--- a/apps/server/src/routes/plugins.ts
+++ b/apps/server/src/routes/plugins.ts
@@ -14,6 +14,10 @@ import type {
 } from "../services/plugins/plugin-service.js";
 import type { PluginMentionTrigger } from "../services/plugins/plugin-api.js";
 import { PluginSettingsValidationError } from "../services/plugins/plugin-settings.js";
+import {
+  createAppAssetCompressionCache,
+  type AppAssetCompressionCache,
+} from "../services/plugins/app-asset-compression-cache.js";
 import { rankAcceptedAssetEncodings } from "../asset-content-encoding.js";
 import {
   pluginApplyUpdateRequestSchema,
@@ -34,6 +38,7 @@ type WireAuthProblem = BrowserRequestProblem | { status: 401; error: string };
 const compressBrotli = promisify(brotliCompress);
 const compressGzip = promisify(gzip);
 const MIN_COMPRESSED_APP_ASSET_BYTES = 1_024;
+const MAX_CACHED_APP_ASSETS = 64;
 const APP_ASSET_ENCODINGS = [
   {
     encoding: "br",
@@ -55,12 +60,18 @@ const APP_ASSET_ENCODINGS = [
 async function appAssetResponse(
   context: Context,
   bytes: Buffer,
-  headers: { cacheControl: string; contentType: string },
+  args: {
+    assetKey: string;
+    cache: AppAssetCompressionCache;
+    cacheControl: string;
+    contentHash: string;
+    contentType: string;
+  },
 ): Promise<Response> {
   const responseHeaders: Record<string, string> = {
-    "cache-control": headers.cacheControl,
+    "cache-control": args.cacheControl,
     "content-length": String(bytes.length),
-    "content-type": headers.contentType,
+    "content-type": args.contentType,
   };
   if (bytes.length < MIN_COMPRESSED_APP_ASSET_BYTES) {
     return context.body(new Uint8Array(bytes), 200, responseHeaders);
@@ -75,7 +86,12 @@ async function appAssetResponse(
     return context.body(new Uint8Array(bytes), 200, responseHeaders);
   }
 
-  const compressed = await candidate.compress(bytes);
+  const compressed = await args.cache.getOrCreate({
+    assetKey: args.assetKey,
+    compress: () => candidate.compress(bytes),
+    encoding: candidate.encoding,
+    hash: args.contentHash,
+  });
   responseHeaders["content-encoding"] = candidate.encoding;
   responseHeaders["content-length"] = String(compressed.length);
   return context.body(new Uint8Array(compressed), 200, responseHeaders);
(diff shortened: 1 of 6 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am reviewing this pull request for security, code quality, architecture, performance, and end-to-end behavior.

apps/server/src/routes/plugins.ts
```diff
@@ -28,6 +31,56 @@ export interface PluginRoutesDeps {
...
+    return context.body(new Uint8Array(bytes), 200, responseHeaders);
+  }
+
+  responseHeaders.vary = "Accept-Encoding";
+  const candidate = rankAcceptedAssetEncodings(
+    context.req.header("accept-encoding"),
+    APP_ASSET_ENCODINGS,
+  )[0];
+  if (candidate === undefined) {
+    return context.body(new Uint8Array(bytes), 200, responseHeaders);
```
SawyerHood, line 86: 🚨 `slopcop/review` — Respect rejected encodings

This return sends identity when the helper finds no accepted format. The global `compress()` middleware can then add gzip.

I reproduced this issue at the exact head. A request with `br;q=0, gzip;q=0` still received `Content-Encoding: gzip`.

Please give this route full compression control. You can also make the global middleware use the same parsed result.

Please add a test for `br;q=0, gzip;q=0`.

apps/server/src/routes/plugins.ts
```diff
@@ -28,6 +31,56 @@ export interface PluginRoutesDeps {
...
+  }
+
+  const compressed = await candidate.compress(bytes);
```
SawyerHood: 🚨 `slopcop/review` — Reuse compressed asset variants

Each GET and HEAD request runs Brotli or gzip again. The server does not store compressed output by the immutable asset hash.

My local check used a 1,042,895-byte bundle. Identity took about 3 ms, while Brotli took about 15 ms.

The route is public. App startup also requests several plugin assets in parallel.

Please build sidecars during the plugin build or install. A promise cache per hash, file, and format can also work.

HEAD can use the stored size without compression.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** A plugin is like a small toy added to bb. This change puts its files in smaller boxes for the browser. The boxes move faster. However, the server makes each box again for every request.

I found two issues.

1. **Encoding rejection — medium.** The route selects identity when both formats have `q=0`. The global middleware then adds gzip. This result conflicts with the client request and can stop the asset load.
2. **Repeated compression — medium.** Every GET and HEAD request compresses the full asset again. This public route has no cache or work limit. Parallel startup requests increase CPU and memory use.

The new helper removes duplicate format ranking from the static asset path. However, two components now control the same response. The route and the global middleware can produce different results.

The existing static asset path uses build-time sidecars. The plugin build or install path should use the same design.

I found no new path traversal, symlink, MIME, or cache-poisoning issue.

I ran the server typecheck on the exact head. I also ran all 13 focused plugin asset tests. They passed before my temporary diagnostic assertion.

The diagnostic request used `br;q=0, gzip;q=0`. It failed because the response used gzip.

I started the exact-head development app. A browser loaded 12 plugin assets with Brotli and status 200. The page had no console or page errors.

I left two inline comments with the fixes. I am leaving this review as a comment only.

apps/server/src/routes/plugins.ts
ymichael: Addressed in 7fd116d71. Plugin app.js/app.css paths now bypass the outer Hono compression middleware, so the route has sole ownership of encoding. The regression test sends br;q=0, gzip;q=0 and verifies an identity response with the original body.

apps/server/src/routes/plugins.ts
ymichael: Addressed in 7fd116d71 with a bounded 64-asset LRU keyed by asset, content hash, and encoding. Compression promises enter the cache before zlib starts, so concurrent GET/HEAD requests share the same work; hash changes replace variants, old assets are evicted, and failed compression is retried. Dedicated tests cover all of those behaviors.

PR #1296, opened 2026-08-10
Avoid full keyboard settings row rerenders

ymichael:
## Summary
- derive row-specific keyboard command view models and memoize rows by their visible state
- give each shortcut recorder its own semantic memo boundary and the stable shared dispatcher
- keep one shared mutation owner for draft, optimistic mutation, rollback, and cross-window behavior
- propagate global mutation pending state through a fieldset instead of row props
- add regression coverage for row and recorder isolation, cache/refetch identity churn, conflicts, cross-window updates, pending state, and reset-all rollback

## Render verification
- before: 255 row renders for one record/assign/clear/reset flow (51 rows x 5), with no interaction remounts
- after: 55 row renders (51 initial plus 4 updates to the edited row only)
- an unchanged recorder now also bails out when its row rerenders for a conflict-warning update

## Validation
- pnpm exec turbo run test --filter=@bb/app --force (2,480 tests on the base optimization)
- focused final tests: 19 passed
- pnpm exec turbo run build --filter=@bb/app
- pnpm exec turbo run lint --filter=@bb/app (0 errors; existing warnings only)

### DIFF
--- a/apps/app/src/components/settings/KeyboardSettingsSection.tsx
+++ b/apps/app/src/components/settings/KeyboardSettingsSection.tsx
@@ -619,7 +619,13 @@ export function KeyboardSettingsSection() {
         />
         {/* Native disabled propagation keeps pending state out of every row. */}
         <fieldset
-          className="m-0 min-w-0 space-y-5 border-0 p-0"
+          className={cn(
+            "m-0 min-w-0 space-y-5 border-0 p-0",
+            // Preserve the interaction lock without visually flashing every
+            // otherwise-enabled row while a shortcut mutation is pending.
+            isKeyboardSettingsPending &&
+              "[&:disabled_button:not([disabled])]:opacity-100",
+          )}
           disabled={disabled}
         >
           {visibleGroups.map((group) => (
--- a/apps/app/src/components/settings/KeyboardSettingsSection.tsx
+++ b/apps/app/src/components/settings/KeyboardSettingsSection.tsx
@@ -1,4 +1,12 @@
-import { useMemo, useState, type KeyboardEvent } from "react";
+import {
+  memo,
+  useCallback,
+  useLayoutEffect,
+  useMemo,
+  useRef,
+  useState,
+  type KeyboardEvent,
+} from "react";
 import {
   defaultAppSettings,
   type AppCommandId,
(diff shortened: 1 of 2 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I will review this pull request under the review rule.

I will check security, code quality, performance, architecture, duplication, and the user path. Three parallel reviewers will help with this review.

apps/app/src/components/settings/KeyboardSettingsSection.tsx
```diff
@@ -210,130 +230,215 @@ function KeyboardCommandRow({
...
+      command,
+      conflicts,
+      customized,
+      desktopDefaultShortcut,
+      desktopOnly,
+      shortcut,
+      webDefaultShortcut,
+    } = model;
+    const metadata = getAppCommandMetadata(command);
+    const availableOnClient = activeDefaultShortcut !== null;
```
SawyerHood: 🚨 `slopcop/review` — Preserve commands that have no default shortcut during the required rebase.

Current `main` adds assignable commands with null defaults. It now uses `isAppCommandAvailableForClient` and `resetCommandShortcutOverride` for these commands.

If conflict resolution keeps this check, Rename thread and Archive thread stay disabled. Keep the new availability and reset rules in the row model.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** The keyboard page now changes only the rows that need new data. This saves work when one shortcut changes.

I found one rebase concern. GitHub reports conflicts with current `main` in both changed files.

Current `main` supports assignable commands with no default shortcut. Preserve its availability and reset rules during conflict resolution.

The security review found no issues. The code quality review found no other correctness, duplication, or architecture issues.

The performance review confirmed that row memo rules cover all displayed row data. The shared fieldset keeps pending state outside row properties.

I ran the focused test at the exact head commit. All eight tests passed, and the app typecheck passed.

I also used the dev app to edit a shortcut, search rows, check the Custom badge, and reset the shortcut. Each step passed.

Please rebase, preserve the null-default command behavior, and run the checks again. I used a comment review, as required by this rule.

PR #1300, opened 2026-08-10
Add pnpm seed:perf to build a large realistic perf-test database

SawyerHood:
## What changed

- Add `pnpm seed:perf`, a checked-in command that seeds a dev data dir with a large, realistic BB database: 12 projects, ~1,200 threads, and ~400k event rows in about 12 seconds (~500 MB).
- Event payloads match production shapes: turns, reasoning deltas, command output, file-change diffs, token usage, rate limits, search segments (FTS via triggers), and prompt history.
- A unit test seeds an in-memory database and parses every seeded event with the production `parseStoredThreadEvent` parser, so fixture drift cannot make seeded timelines return 500s.
- Document the workflow in `docs/debugging-and-qa.md`.

## Why

Mobile slowness reports only reproduce at scale. This makes a big, reusable fixture one command away, with deterministic `--seed` runs and append mode for jumbo threads (e.g. `--projects 1 --threads 8 --events 60000`).

## Notes

- Default target is the checkout's dev data dir; the command refuses to touch `~/.bb`.
- Boot dev once before seeding so the fixture adopts the real daemon host id; `--reset` also removes `auth.json` so the daemon re-enrolls under the persisted host id.
- Used to produce the mobile perf findings that will land as issues from this investigation.

### DIFF
--- a/packages/scripts/src/commands/seed-perf-db.ts
+++ b/packages/scripts/src/commands/seed-perf-db.ts
@@ -1,0 +1,224 @@
+import { execFileSync } from "node:child_process";
+import { existsSync, mkdirSync, readFileSync, rmSync } from "node:fs";
+import { homedir } from "node:os";
+import { dirname, join, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+import { createConnection, migrate } from "@bb/db";
+import {
+  resolveCurrentDevInstanceConfig,
+  resolveDataDirDatabasePath,
+  resolveProdDataDir,
+} from "@bb/config/runtime";
+import { seedPerfFixture } from "../lib/seed-perf-fixture.js";
+import { bold, cyan, dim, green, log, endStep } from "../lib/script-helpers.js";
+
+const commandDir = dirname(fileURLToPath(import.meta.url));
+const repoRoot = resolve(commandDir, "..", "..", "..", "..");
+
+const HOST_ID_FILE_NAME = "host-id";
+
+interface SeedCommandArgs {
+  dataDir: string | null;
+  eventCount: number;
+  projectCount: number;
+  randomSeed: number;
+  reset: boolean;
+  threadCount: number;
+}
+
+function renderHelpText(): string {
+  return `
+  ${bold("bb seed-perf-db")}
+
+  Seed a large, realistic BB database for performance testing.
+
+  ${dim("Usage")}
+    pnpm seed:perf [-- options]
+
+  ${dim("Options")}
+    --data-dir <path>  Target data dir (default: this checkout's dev data dir)
+    --projects <n>     Project count (default: 12)
+    --threads <n>      Thread count (default: 1200)
+    --events <n>       Approximate total event rows (default: 400000)
+    --seed <n>         Deterministic random seed (default: 1)
+    --reset            Delete the existing database file before seeding
+
+  ${dim("Notes")}
+    The command refuses to touch the production data dir (~/.bb).
+    Without --reset the fixture is added to the existing database.
+    Start the dev app once before seeding so the fixture attaches to
+    the real local host; otherwise a synthetic offline host is used.
+\n`;
+}
+
+function parsePositiveInteger(name: string, rawValue: string): number {
+  const value = Number(rawValue);
+  if (!Number.isInteger(value) || value <= 0) {
+    throw new Error(`${name} must be a positive integer`);
+  }
+  return value;
+}
+
+function parseArgs(argv: string[]): SeedCommandArgs | null {
+  const args: SeedCommandArgs = {
+    dataDir: null,
+    eventCount: 400_000,
+    projectCount: 12,
+    randomSeed: 1,
+    reset: false,
+    threadCount: 1_200,
+  };
+  for (let index = 0; index < argv.length; index += 1) {
+    const argument = argv[index];
+    const readValue = (): string => {
+      const value = argv[index + 1];
+      if (value === undefined) {
+        throw new Error(`${argument} requires a value`);
+      }
+      index += 1;
+      return value;
+    };
+    switch (argument) {
+      // pnpm forwards the "--" separator to the script; skip it.
+      case "--":
+        break;
+      case "--help":
+      case "-h":
+        return null;
+      case "--data-dir":
+        args.dataDir = resolve(readValue());
+        break;
+      case "--projects":
+        args.projectCount = parsePositiveInteger("--projects", readValue());
+        break;
+      case "--threads":
+        args.threadCount = parsePositiveInteger("--threads", readValue());
+        break;
+      case "--events":
+        args.eventCount = parsePositiveInteger("--events", readValue());
+        break;
+      case "--seed":
+        args.randomSeed = parsePositiveInteger("--seed", readValue());
+        break;
+      case "--reset":
+        args.reset = true;
+        break;
+      default:
+        throw new Error(`Unknown option: ${argument}`);
+    }
+  }
+  return args;
+}
+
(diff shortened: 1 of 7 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the Slop Cop. I am now reviewing this pull request for security, code quality, performance, duplication, architecture, and end-to-end behavior.

packages/scripts/src/commands/seed-perf-db.ts
```diff
@@ -0,0 +1,224 @@
...
+): Promise<void> {
+  const args = parseArgs(argv);
+  if (args === null) {
+    process.stdout.write(renderHelpText());
+    return;
+  }
+
+  const dataDir = resolveTargetDataDir(args);
+  const prodDataDir = resolveProdDataDir({ homeDir: homedir() });
+  if (resolve(dataDir) === resolve(prodDataDir)) {
```
SawyerHood, line 142: 🚨 `slopcop/review` — High: canonicalize the target before the production safety check

`resolve()` compares path text. It does not compare file identities. I confirmed that `~/.BB` and `~/.bb` share one inode on this host. The case alias passes this check. A symlink alias also passes it. With `--reset`, either alias deletes the production database and `auth.json`.

Use `realpathSync.native()` for the existing data directory. Also reject a database symlink that targets the production database. Add tests for case and symlink aliases.

packages/scripts/src/lib/seed-perf-fixture.ts
```diff
@@ -0,0 +1,1084 @@
...
+        { type: "text", text: turnUserText, mentions: [] },
+      ]),
+      createdAt: state.at,
+    });
+
+    emit({ type: "turn/started", turnId, data: { providerThreadId } });
+    emit({
+      type: "turn/input/accepted",
+      turnId,
+      data: { providerThreadId, clientRequestId: `creq_${rng.idSuffix()}` },
```
SawyerHood, line 442: 🚨 `slopcop/review` — High: reuse the request ID for the accepted input

This event creates a new `clientRequestId`. It does not use the related `client/turn/requested.requestId`. I checked one fixture with seven requests and seven accepted inputs. None of the IDs matched.

The timeline uses this ID to connect a user message to its turn. The fixture therefore measures pending, thread-level messages instead of accepted turn messages. Create one request ID for each turn and use it in both events. Add a contract test for this relation.

packages/scripts/src/lib/seed-perf-fixture.ts
```diff
@@ -0,0 +1,1084 @@
...
+  let weightSum = 0;
+  for (let index = 0; index < threadCount; index += 1) {
+    const weight = Math.exp(rng.next() * 4.2);
+    weights.push(weight);
+    weightSum += weight;
+  }
+  return weights.map((weight) =>
+    Math.max(
+      20,
+      Math.min(9_000, Math.round((weight / weightSum) * eventCount)),
```
SawyerHood, line 764: 🚨 `slopcop/review` — Medium: preserve the requested total after the per-thread cap

The 9,000-row cap removes events without moving the excess to another thread. I ran `--projects 1 --threads 8 --events 60000`. The command created 39,772 events. This result is 33.7 percent below the requested total.

Remove the cap, or move the excess to threads that remain below the cap. Add a test with a clear tolerance for the final event count.

packages/scripts/src/commands/seed-perf-db.ts
```diff
@@ -0,0 +1,224 @@
...
+      projectCount: args.projectCount,
+      threadCount: args.threadCount,
+      eventCount: args.eventCount,
+      randomSeed: args.randomSeed,
+      onProgress: (message) => log(dim("○"), dim(message)),
+    });
+    db.$client.pragma("wal_checkpoint(TRUNCATE)");
+    // Real directories keep the host daemon's git probes from failing when a
+    // seeded thread or environment is opened.
+    for (const workspacePath of result.projectWorkspacePaths) {
```
SawyerHood, line 183: 🚨 `slopcop/review` — Medium: create every ready environment path

This list contains only the root project paths. The fixture also records separate managed-worktree paths as ready environments. The default fixture creates many such environments, but the command creates none of their directories.

A host command for one of these threads fails with `path_not_found`. This conflicts with the note that agents can run after the seed. Return and create all ready paths, or attach live threads only to the root environments.

packages/scripts/src/lib/seed-perf-fixture.ts
```diff
@@ -0,0 +1,1084 @@
...
+    ),
+  );
+}
+
+export function seedPerfFixture(
+  db: DbConnection,
+  options: SeedPerfFixtureOptions,
+): SeedPerfFixtureResult {
+  const rng = createRng(options.randomSeed);
```
SawyerHood, line 773: 🚨 `slopcop/review` — Medium: make the documented append mode repeatable

Each call starts the deterministic generator with the same default seed. A second default run creates the same project IDs. The plain inserts then fail with `UNIQUE constraint failed: projects.id`.

I reproduced this failure after one successful no-reset run. Add a unique run namespace, or require a new seed and document that rule. Add a second-run test for the chosen behavior.

packages/scripts/test/seed-perf-fixture.test.ts
```diff
@@ -0,0 +1,115 @@
...
+    expect(Object.fromEntries(parseFailuresByType)).toEqual({});
+
+    // FTS triggers must index the seeded search segments.
+    const ftsRow = db.$client
+      .prepare(
+        `SELECT COUNT(*) AS indexed FROM thread_search_segments_fts
+         WHERE thread_search_segments_fts MATCH 'test'`,
+      )
+      .get();
+    expect(ftsRow).toBeDefined();
```
SawyerHood, line 97: 🚨 `slopcop/review` — Low: assert that the FTS query found rows

A SQLite `COUNT(*)` query always returns one result row. Therefore, `toBeDefined()` also passes when the index count is zero. Assert that `indexed` is greater than zero, or query for one known fixture marker.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

This adds a tool that fills a practice BB with many pretend projects and chats. It is like a large toy box for speed tests. Some safety and data links need repairs before people use it.

### Review

I found two high-severity defects, three medium-severity defects, and one low-severity test gap.

- High: A path alias can bypass the production-data guard. On this host, `~/.BB` and `~/.bb` share one inode. The `--reset` option can then delete the production database and credentials.
- High: Accepted input events use unrelated request IDs. The timeline cannot connect seeded user messages to their turns.
- Medium: The 9,000-event thread cap loses events. A 60,000-event test created only 39,772 rows.
- Medium: Ready managed environments point to directories that the command never creates. Host operations can fail with `path_not_found`.
- Medium: A second default append run fails because it creates the same deterministic project IDs.
- Low: The FTS test checks only that `COUNT(*)` returned a row. It does not check that the count is positive.

I added inline comments with the evidence and a repair for each item.

The default performance path looks reasonable. It created 402,298 events in 15.2 seconds. The database used 496 MiB, and the process used 464 MiB at peak. Inserts use bounded batches and short transactions.

I found no command injection, SQL injection, secret exposure, or duplicate performance seeder. Direct bulk writes fit this tool better than production notifier helpers.

For architecture, reuse `HOST_ID_FILE_NAME` and `HOST_AUTH_FILE_NAME` from `@bb/host-daemon-contract`. A typed request-event helper can also prevent the ID mismatch.

Validation passed for the focused Turbo typecheck and all 86 `@bb/scripts` tests. All pull-request checks pass. The seeded app loaded in a browser, and I opened a seeded thread timeline.

I posted this review as a comment only. I did not approve the pull request or request changes.

Issue #1093, opened 2026-08-06
Make OpenCode custom models and agents clear and selectable


SawyerHood:
## Problem

Users cannot tell how BB maps OpenCode models and agents into its execution controls.

BB already discovers ACP session models and applies a selected model with `session/set_model`. OpenCode custom models should therefore appear under `acp-opencode` when OpenCode advertises them.

However, several gaps make this flow unclear or incomplete:

- The BB guide does not explain how to select an OpenCode custom model.
- The guide does not explain the difference between an OpenCode model and an OpenCode agent.
- BB parses ACP config options, but it only uses `model` and `thought_level`.
- BB ignores the ACP `mode` option, which can represent OpenCode agents.
- The `customModels` fallback only accepts static provider IDs.
- That fallback rejects `acp-opencode` and other dynamic ACP provider IDs.

A user with a custom primary agent, such as an orchestrator, can mistake it for a model. BB gives no useful path for that case.

## Current code

- `packages/agent-runtime/src/acp/bridge/model-catalog.ts` maps ACP model options.
- `packages/agent-runtime/src/acp/bridge/bridge.ts` applies ACP model changes.
- `packages/config/src/bb-app-managed-config.ts` limits `customModels.providerId` to the static provider enum.
- `packages/templates/src/templates/bb-guide-providers.md` documents ACP providers, but not this selection flow.

## Proposed result

BB should give users one clear OpenCode selection flow.

- Show all custom models that OpenCode advertises.
- Let users select a custom OpenCode primary agent when ACP exposes it as a mode.
- Explain that models and agents are different selections.
- Add a supported fallback for a custom model that an ACP agent does not advertise.
- Make the CLI guide show the exact discovery commands.

Example:

```sh
bb provider models acp-opencode --environment "$BB_ENVIRONMENT_ID"
bb thread spawn --provider acp-opencode --model <provider/model>
```

## Acceptance criteria

- A custom OpenCode model appears in the BB model picker.
- The same model appears in `bb provider models acp-opencode`.
- A selected model reaches OpenCode before the first prompt.
- A custom OpenCode primary agent can be selected when ACP advertises it.
- The guide explains models, agents, discovery, and fallback behavior.
- Tests cover custom models, ACP mode selection, and dynamic ACP provider IDs.

## Related issues

Issue #365 covers model-list failure recovery. This issue covers discovery and selection.

Issue #1086 covers Pi extension models. This issue covers OpenCode and generic ACP controls.

[...]

PR #1305, opened 2026-08-11
Accept dynamic ACP provider ids in customModels and document the selection flow

SawyerHood:
## Problem

Part of #1093.

- The `customModels` fallback only accepted the static provider enum (`codex`, `claude-code`, `pi`, `acp-cursor`). An entry for `acp-opencode` or a custom `acp-<slug>` agent failed validation.
- The failure was a throw, not a skip: one invalid entry broke the whole managed-config parse.
- `customModels`, the OpenCode model selection flow, and the model-versus-agent distinction were undocumented.

This PR intentionally does not add ACP `mode` (OpenCode agent) selection.

## Changes

- `customProviderModelSchema.providerId` now accepts the built-in enum or any well-formed `acp-*` provider id.
- `parseBbAppManagedConfig` parses `customModels` entries individually and skips an invalid entry with a warning, matching the `customAcpAgents` boundary behavior.
- No reasoning-ladder change was needed: `getAgentProviderServerCapabilities` already serves the shared ACP ladder for every `acp-*` id. A new test locks this in.
- Docs on all three discoverable surfaces: `bb-guide-providers.md` (with regenerated bundle), a new "Custom Models" section in `docs/configuration.md`, and the `bb-cli` skill.

## Testing

- New config tests: dynamic ids parse, malformed `acp-` ids fail, invalid entries warn and skip.
- New server test: an `acp-opencode` custom model appends with the shared ACP reasoning ladder.
- Verified against a live `opencode acp` (1.17.9) session that OpenCode advertises custom models over ACP, so ACP-advertised models need no `customModels` entry; the fallback covers unadvertised models.
- `turbo typecheck`/`test` green for `@bb/config`, `bb-app`, `@bb/templates`, and the `@bb/server` system suite.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am now reviewing this pull request for security, code quality, architecture, performance, tests, and the selection flow.

docs/configuration.md
```diff
@@ -370,6 +370,43 @@ can write `~/.bb/config.json` can cause bb to run that command as the local user
 when the provider is used. Treat `config.json` write access as the trust
 boundary.
 
+## Custom Models
+
+Register extra picker models by editing top-level `customModels` in
+`~/.bb/config.json`. Use this for a model the provider catalog does not
+advertise, such as a non-public preview id or an ACP agent model that the
+agent does not list. Like `customAcpAgents`, this list has no set/unset CLI
```
SawyerHood: 🚨 `slopcop/review` — This entry becomes selectable, but bb does not apply it for OpenCode.

`selectAcpNativeModel` calls `session/set_model` only when the session advertises the selected ID. An unlisted model therefore keeps the session default. Please apply explicit unlisted IDs, or narrow this claim.

packages/config/src/bb-app-managed-config.ts
```diff
@@ -150,7 +160,7 @@ const bbAppManagedConfigBoundarySchema = z
   .object({
     config: bbAppManagedConfigValuesSchema.optional(),
     customAcpAgents: z.array(z.unknown()).optional(),
-    customModels: z.array(customProviderModelSchema).optional(),
+    customModels: z.array(z.unknown()).optional(),
```
SawyerHood, line 163: 🚨 `slopcop/review` — This boundary change leaves a failing smoke test.

`tests/integration/fake/smoke/custom-models.test.ts` still expects HTTP 422. The new parser returns HTTP 200 and skips the invalid entry. The focused Turbo run failed at line 65. Please update the test and confirm that valid entries survive.

packages/config/src/bb-app-managed-config.ts
```diff
@@ -217,21 +227,46 @@ function parseCustomAcpAgents(
...
+
+  const customModels: CustomProviderModel[] = [];
+  for (const [index, entry] of entries.entries()) {
+    const result = customProviderModelSchema.safeParse(entry);
+    if (!result.success) {
+      options.logger?.warn(
+        { error: result.error.message, index },
+        "Ignoring invalid custom model config entry",
+      );
+      continue;
```
SawyerHood, line 246: 🚨 `slopcop/review` — A later configuration write can delete skipped entries.

`readManagedConfigForWrite` preserves raw invalid `customAcpAgents`, but it does not preserve raw invalid `customModels`. After this skip, `bb-app config set` or `unset` rewrites the file without these entries. Please preserve raw entries during writes, or define and test this deletion contract.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: This change adds new model labels. It is like adding labels to toy boxes. One label shows, but bb still chooses the old toy.

I found three code issues and one merge blocker.

1. **High — bb does not apply an unlisted OpenCode model.**

The model appears in the picker. However, `selectAcpNativeModel` only sends `session/set_model` when the session advertises that model. The session keeps its default model instead.

2. **High — the custom-model integration test fails.**

The new parser skips an invalid entry and returns HTTP 200. The existing smoke test still expects HTTP 422. The focused Turbo run failed at `custom-models.test.ts:65`.

3. **Medium — another configuration write can delete skipped entries.**

`readManagedConfigForWrite` preserves raw invalid `customAcpAgents`. It does not preserve raw invalid `customModels`. A later `config set` or `unset` write removes those entries.

4. **Merge blocker — the pull request conflicts with current `main`.**

GitHub reports `CONFLICTING`. The conflicts affect the provider guide and its generated template. Keep both documentation changes during the resolution.

I found no material security or performance problem. The new work runs only during configuration loads or small model-list operations.

The new ACP provider expression duplicates related identifier policy. This is a low-priority maintenance issue. A shared schema can prevent later differences.

Validation results:

- The configuration tests passed: 18 tests.
- The server selection tests passed: 27 tests.
- The template tests passed: 23 tests.
- The affected Turbo typechecks passed.
- `git diff --check` passed.
- The focused integration test failed: one test passed and one test failed.
- The final GPT-5.6 review gate returned `REQUEST CHANGES`.

This change has no direct browser route. I used the real configuration reload and server integration path instead.

Changes are required before merge. I submitted this as a comment review. I did not use the approval or request-changes option.

docs/configuration.md
SawyerHood: Fixed in e8ea328f68 — and verified live: OpenCode rejects an unlisted id on both session/set_config_option and the session/set_model fallback with -32602, so the failure is a rejected spawn, not a kept default. The docs on all three surfaces now say the provider must still accept the id, and OpenCode models belong in the OpenCode config (bb discovers them automatically).

packages/config/src/bb-app-managed-config.ts
SawyerHood: Fixed in e8ea328f68 — the smoke test now asserts the warn-and-skip contract: reload returns 200, the invalid entry is skipped, and the valid entry survives into the model list.

packages/config/src/bb-app-managed-config.ts
SawyerHood: Fixed in e8ea328f68 — readManagedConfigForWrite now preserves the raw customModels array across writes, mirroring customAcpAgents, with a launcher test covering config set on a file with an invalid entry.

Issue #1302, opened 2026-08-10
sidebar-bootstrap ships 138KB and refetches wholesale on thread status changes


SawyerHood:
Found while profiling the mobile web frontend against a seeded 1,200-thread database (`pnpm seed:perf`, #1300).

## Numbers

- `GET /api/v1/sidebar-bootstrap` returns **138 KB of JSON** (11.7 KB gzip) for 131 unarchived threads: ~1.1 KB per thread across **29 fields**, while the sidebar renders roughly 8 of them (title, status, project, unread, pinned, provider, branch, attention time).
- Sending one message refetched the entire payload; it refires on thread lifecycle transitions. With several concurrently active threads, every turn start/stop reparses and re-renders the whole sidebar tree — this multiplies with the drawer mount cost in #1261.
- Server time is fine (~8 ms warm); the cost is wire size, JSON parse, and the re-render it triggers on mobile CPUs.

## Repro

1. `pnpm seed:perf` on a dev instance.
2. Open the app with the network panel filtered to `sidebar-bootstrap`.
3. Send a message to any thread; watch the full payload refetch.

## Suggested approach

- Trim the wire shape to the fields the sidebar renders.
- On status-change push events, patch the cached thread row instead of refetching the world; if a refetch is needed, scope it to the affected project.
- The same push-driven patching would remove most of the ~10-request fan-out currently observed after a single send.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

amadad:
Supporting scale evidence from a production 0.37.0 install: 1,071 unarchived threads are present (985 idle, 86 error). Origins include 228 automation threads and 75 workflow threads.

Sidebar virtualization in 0.37.0 improves rendering, but it does not remove the payload/refetch invariant described here. At this scale, patching the affected thread row from lifecycle events is materially different from repeatedly transferring and parsing the full bootstrap. I support keeping payload reduction and incremental cache patching as the durable fix.

SawyerHood:
**Reproduction & root-cause report:** https://get-bb.github.io/reports/issues/1302.html

- **Verdict:** REPRODUCED · root-cause confidence: high
- **Root cause (short):** status-changed realtime pushes carry no row data and the app's registry rule for them invalidates the single monolithic sidebarNavigation query (full ThreadListEntry for every visible thread of every project, ~1 KB/row) with an immediate flush, so every turn start/end re-downloads the whole 134 KB bootstrap instead of patching one row.
[...]

PR #1307, opened 2026-08-11
Virtualize the sidebar thread list and keep the mobile drawer mounted

### LINKED ISSUE #1261: Large active thread lists keep all sidebar rows mounted
## Summary

On a long-running self-hosted instance, the web/PWA sidebar keeps every active thread row mounted, including rows far outside the scrollport and when the sidebar is collapsed.

In Chrome I measured:

- 1,213 active thread links
- 16,251 DOM elements
- ~600 MB renderer memory, with a 1.5 GB recorded peak

Collapsing the sidebar changed visibility but did not reduce the mounted tree.

I may simply be keeping more old threads active than the intended workflow. How do established users normally manage older threads? Is regular archiving expected, and is there a recommended way to archive older threads in bulk today beyond per-thread actions or `bb environment archive-threads`?

If large active lists are expected to remain supported, offscreen containment or sidebar virtualization may help. Related: #1159, #1131, #1205.

SawyerHood:
Fixes #1261.

## Problem

Two costs, measured on the seeded 1,200-thread perf database (`pnpm seed:perf`, 174 unarchived threads, 129 visible rows):

1. The sidebar mounted every unarchived thread row, no matter the scrollport. The reporter measured 1,213 links / 16k DOM nodes / ~600 MB renderer memory. Collapsing the sidebar did not unmount or stop painting the tree.
2. The mobile drawer unmounted on close. Every open replayed the full list mount as one synchronous long task (~1.2 s at desktop speed, ~5.3 s at 4x throttle in the issue's measurement).

## Change

- **Window every thread-item list against the sidebar scrollport** (`SidebarWindowedItems`, wired into `SectionThreadTreeItems` and section children — the one choke point all modes render through). Offscreen items render as fixed-height placeholder divs in normal flow, so the CSS sticky-header stack, `space-y` gaps, dnd DOM order, and the scrollbar behave exactly as before. A pre-paint layout pass mounts the initial viewport with no placeholder flash; an IntersectionObserver with a 480 px margin promotes/demotes during scroll inside `startTransition`. Placeholder heights come from measured row heights, with a collapse-aware row-count estimate before first measure. Lists without a usable scrollport (jsdom tests, previews) render fully, and lists under 8 items skip windowing.
- **Keep the item that holds the active thread always mounted**, so the DOM-driven `thread.next`/`thread.previous` navigation keeps its anchor.
- **Replace the vaul mobile drawer with a persistent panel.** Closing translates it off-screen, then delayed `visibility: hidden` plus `inert` take it out of paint, hit-testing, focus order, and the accessibility tree. Reopening replays no mount work and keeps the scroll position (it used to reset to 0). Drag-to-close (distance + fling, mirroring vaul's thresholds), swipe-to-open, Escape, and backdrop dismissal all work; the swipe helpers' DOM contract (`data-sidebar="panel"`, `data-vaul-drawer-direction`, inline-style overrides) is unchanged.
- **Hide the collapsed desktop offcanvas panel** with the same delayed-visibility trick, so a collapsed sidebar stops painting its rows.
- **Memoize the thread-search controller in `AppSidebar`** so drawer toggles no longer defeat `ProjectList`'s memo and re-render every group.

## Measurements

Seeded DB (12 projects, 1,200 threads, 174 unarchived), headless Chromium, 390x844 viewport, long tasks via PerformanceObserver.

| Metric | Before | After |
| --- | --- | --- |
| Drawer open, desktop speed (every open) | 424–528 ms long task | **0 long tasks** |
| Drawer open, 4x CPU throttle (every open) | 1,955–2,304 ms | **134–155 ms** |
| Mounted thread links (drawer open) | 129 | **70** (viewport + 480 px overscan) |
| DOM nodes with drawer closed | 146 (unmounted) | 1,647 (mounted, inert, hidden) |
| Drawer scroll, full sweep at ~1,500 px/s | baseline smooth | 5/305 frames > 50 ms, p95 33 ms |
| Drawer scroll position after close/reopen | reset to 0 | **preserved** (1200 → 1200) |
| Desktop sidebar mounted links / DOM nodes | 129 / 2,223 | **71 / 1,523** |

Mounted rows now scale with the viewport, not the thread count: at the reporter's 1,213-thread scale the mounted-row count stays ~70.

## Manually verified

On the seeded database in headless Chromium (mobile 390x844 and desktop 1400x900):

- Project grouping, pinned group, unread dots, project collapse/expand, sticky headers mid-scroll — all three organization modes (project / manual / machine).
- Keyboard navigation DOM contract: 9 shortcut targets present, `thread.next` adjacency works, and the active thread's row stays mounted when scrolled far away.
- Search mode (25 results, Escape clears then closes) and the row context menu.
- Drag and drop: project-section reorder, and thread drag in manual mode (drag overlay activates and completes on windowed rows).
- Mobile gestures: swipe-to-open, drag-to-close, small drag stays open, vertical scroll inside the drawer does not dismiss, Escape and backdrop close, closed panel does not intercept taps (`inert` + hit-test verified).
- Desktop offcanvas collapse: panel `visibility: hidden` after the transition, instant re-show, rows stay mounted.
- Plugin thread lists (`experimental_threadList`): the host path is untouched; plugin lists render their own scroll-area content as before. Not exercised against a real plugin.

## Notes and follow-ups

- `thread.next`/`thread.previous` walk mounted rows. With windowing, an offscreen unmounted row cannot be a neighbor; the active item stays mounted so adjacency from the active thread still works. In practice navigation cycles through the mounted (near-viewport) rows.
- The closed drawer now keeps ~1,600 DOM nodes mounted by design; they are inert, unpainted, and skip repeat-mount cost. Realtime updates re-render only the ~70 mounted rows.
- The sidebar-bootstrap payload work stays separate (#1302).


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I will use three parallel BB review agents. I will post one final review after I verify their findings.

apps/app/src/components/sidebar/ProjectRow.tsx
```diff
@@ -1682,6 +1705,51 @@ interface SectionThreadTreeItemsProps {
...
+    [collapsedThreadIds, collapsedEnvironmentIds, collapsedSectionKeyList],
+  );
+  const estimateRows = useCallback(
+    (index: number) => {
+      const item = items[index];
+      return item ? countProjectThreadItemRows(item, rowCountContext) : 1;
+    },
+    [items, rowCountContext],
+  );
+  const alwaysMountedKeys = useMemo(() => {
```
SawyerHood, line 1799: 🚨 `slopcop/review` — [P1] Keeping only the active item mounted breaks thread navigation commands.

`AppSidebar` builds `thread.next`, `thread.previous`, and numbered jump targets from mounted anchors. Offscreen rows now have no anchors. With 300 ungrouped threads, I measured 38 targets at the top and 43 at the bottom. The target sets changed by 33 and 38 threads after one scroll.

Please build a complete command sequence from the thread model and navigate by thread ID. Do not derive command order from the virtual DOM.

apps/app/src/components/ui/sidebar.tsx
```diff
@@ -721,6 +709,542 @@ const Sidebar = React.forwardRef<
...
+        if (event.key === "Escape" && !event.defaultPrevented) {
+          onOpenChange(false);
+        }
+      };
+      window.addEventListener("keydown", handleKeyDown);
+
+      // The drawer is modal: take the main content out of focus order and
+      // hit-testing while it is open. Portalled popovers (row context menus)
+      // render under document.body, so they stay interactive.
+      const inset = document.querySelector('[data-sidebar="inset"]');
```
SawyerHood: 🚨 `slopcop/review` — [P1] The replacement mobile modal lets keyboard focus leave the drawer.

This effect makes only `SidebarInset` inert. `SidebarTriggerOverlay` is its focusable sibling. In a 390-pixel browser check, I focused the last drawer control and pressed Tab. Focus moved to the external Toggle sidebar button.

Please keep the Vaul or Radix focus scope with force-mounted content. Another fix must contain focus and isolate all outside controls. Add a keyboard test for Tab and Shift+Tab.

apps/app/src/components/sidebar/ProjectRow.tsx
```diff
@@ -1703,26 +1771,47 @@ function SectionThreadTreeItems({
...
-      renderTopLevelSectionHeaderActions={renderTopLevelSectionHeaderActions}
-      sectionDnd={sectionDnd ?? undefined}
+  const { itemKeys, estimateRows, alwaysMountedKeys } = useWindowedThreadItems({
+    items,
+    collapsedThreadIds,
+    collapsedEnvironmentIds,
+    selectedThreadId,
+  });
+  const rows = (
+    <SidebarWindowedItems
```
SawyerHood, line 1840: 🚨 `slopcop/review` — [P1] Windowing only top-level groups leaves large worktrees fully mounted.

`SidebarWindowedItems` receives top-level `ProjectThreadItem` values. One environment item can contain hundreds of nodes. `EnvironmentThreadGroupRow` still renders them with `nodes.map`. I reproduced this with one worktree and 300 threads. The sidebar mounted all 300 thread rows and created zero window wrappers.

Please flatten the expanded tree into virtual rows, or window each expanded child list. The threshold must use rendered-row count, not top-level group count.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** This change puts sleeping thread rows aside and keeps the phone drawer ready. Three bugs make the list lose threads or controls.

### Findings

1. **P1 — Large worktree groups bypass the window.** One worktree with 300 threads mounted all 300 rows and made zero window wrappers. The code windows top-level groups, while each environment group still renders every node.

2. **P1 — Thread commands use only mounted rows.** `thread.next`, `thread.previous`, and number jumps read DOM anchors. At the list top, 38 targets existed. At the bottom, 43 targets existed. The sets changed by 33 and 38 threads.

3. **P1 — The mobile modal loses focus containment.** The code makes only `SidebarInset` inert. Tab moved from the final drawer control to the external sidebar trigger.

4. **P2 — Old measurements override new row estimates.** A changed collapse state can retain an old placeholder height. This can change scroll space and move the scroll anchor.

5. **P2 — The new TypeScript file contains a literal NUL byte.** Git treats `SidebarWindowedItems.tsx` as binary. GitHub hides its complete source diff. Replace the byte with an escaped delimiter.

### Architecture

Use one flat visible-row model for rendering, drag identities, and command order. The repository already uses `@tanstack/react-virtual`. Use it for range and measurement logic.

Vaul and Radix expose `forceMount`. Keep their modal and focus behavior instead of maintaining a second drawer behavior stack.

The current window also keeps one wrapper and observer target per item. Its height, row, and callback maps retain removed keys. These costs share the same design cause.

### Security and validation

I found no direct authorization, server, dependency, or protocol issue. The NUL byte does reduce review and scanner visibility.

The app type check passed. All 2,486 app tests passed. The current GitHub checks passed. The mobile mount, inert state, backdrop close, and focus state passed product checks.

The test suite does not enter the real virtual path because its scrollport has zero height. Add large flat-list, grouped-list, command-order, modal-focus, and stale-height tests.

The GPT-5.6 review gate found the same five defects. The three P1 defects need fixes before merge. I posted a comment-only review.

apps/app/src/components/sidebar/ProjectRow.tsx
SawyerHood: Fixed in a39683ded. Environment-group node lists and parent-thread child lists now window through the same SidebarWindowedItems, so every sibling list level gates on its own item count. Verified against a 53-thread worktree group: ~40 rows mounted near the viewport, the rest render as placeholders.

apps/app/src/components/sidebar/ProjectRow.tsx
SawyerHood: Fixed in a39683ded. Windowed-out placeholders now carry their threads in visual order (data-sidebar-windowed-nav), and getSidebarThreadNavigationTargets expands them, so thread.next/thread.previous see the complete list from any scroll position (129 targets at top and at bottom on the seeded DB; element-less neighbors navigate by id). I kept the DOM walk rather than a model-driven order because the documented plugin thread-list contract derives shortcut order from the DOM, and plugin sidebars have no host-visible model. The numbered 1-9 jumps stay mounted-rows-only deliberately: their badges render on the rows, so an unmounted row cannot present a shortcut.

apps/app/src/components/ui/sidebar.tsx
SawyerHood: Fixed in a39683ded. While the drawer is open, every sibling of the panel at its container level goes inert (inset, pinned trigger overlay, other chrome), not just the inset; the backdrop stays live for dismissal. Verified in-browser at 390px: Tab from the last drawer control can no longer reach the external toggle. Body-portalled popovers opened from the drawer stay interactive. The unit test asserts the sibling inert set across open/close. On force-mounting vaul instead: vaul 1.1.2 defines no persistent closed-state styling for force-mounted content, and the existing swipe-open code already bypasses vaul's animator with inline transforms, so a persistent panel with one animation path ended up smaller than reconciling vaul's presence animation with a second mounted state.

SawyerHood:
All five review findings are addressed in a39683ded:

1. **P1 worktree groups** — environment-group node lists and parent-thread child lists now window through the same `SidebarWindowedItems`; verified with a 53-thread worktree group (~40 rows mounted, rest placeholders).
2. **P1 thread commands** — placeholders advertise their threads in visual order via `data-sidebar-windowed-nav`; `thread.next`/`thread.previous` now see all 129 threads from any scroll position and navigate element-less neighbors by id. Numbered jumps stay mounted-rows-only because their badges render on the rows.
3. **P1 focus containment** — every sibling of the open panel goes inert, not just the inset; Tab can no longer reach the trigger overlay (browser-verified, unit-tested).
4. **P2 stale heights** — a measured placeholder height now records the row count it was measured under and is discarded when the count changes; windowing maps prune keys that leave the list.
5. **P2 NUL byte** — the key-signature join now uses the escaped backslash-u0000 form; the file diffs as text again.

Perf after the fixes is unchanged: 0 long tasks per drawer open at desktop speed on the seeded database, mounted links ~70. Lint, typecheck, and the full app test suite pass.

SawyerHood:
Additional at-scale validation, matching the reporter's ~1,200-row scenario: I unarchived the full perf fixture (966 visible threads in the sidebar, ~31,000 px scroll height) and added a 30-child parent thread plus a 53-thread worktree group.

- Drawer open at 390x844: **0 long tasks**, 36 mounted thread links, 930 threads represented by placeholders, 1,841 total DOM nodes.
- `thread.next`/`thread.previous` see all 966 threads from any scroll position.
- The 30-child parent renders with its sticky parent row and chevron; children window (31 nested wrappers) and collapse/expand cycles behave.

PR #1310, opened 2026-08-11
fix(codex): render Imagegen outputs inline

oscardobsonbrown:
## Summary

Codex Imagegen outputs now appear as direct inline chat rows before the final assistant response. Each image has a lazy preview, an accessible lightbox, and a thread-scoped content URL; collapsing the work summary no longer hides the output. Codex hook lifecycle events also stop appearing as misleading unhandled rows.

The provider boundary now uses deterministic contracts from a pinned Codex release. It normalises schema defaults and proves compatibility against the official `0.136.0` generated `ThreadItem` union. The server resolves each image from its persisted thread event and owning host, then enforces encoded-size, raster-signature, MIME, cache, and content-sniffing controls before returning bytes.

Session-settled decisions carried from planning: direct timeline output, path-only CLI and SDK parity, persisted event provenance for authorisation, explicit boundary defaults, and version-gated contract generation.

## Validation

- A live Codex thread generated two images. The persisted timeline returned two direct `generated-image` rows at source sequences 82 and 83 before the final assistant response.
- Both thread-scoped content requests returned valid 1254×1254 PNG files with `200`, `Content-Type: image/png`, `Cache-Control: no-store`, `X-Content-Type-Options: nosniff`, and exact content lengths.
- The full affected typecheck matrix passed 13 of 13 Turbo tasks.
- Domain, host-daemon contract, server contract, plugin SDK, thread-view, agent-runtime, CLI, DB, app, and templates suites passed. The app result was 2,449 tests; the generated-image server route and response result was 85 tests.
- Five unrelated server plugin and artefact tests exceeded their fixed five-second timeout during the concurrent full matrix. The generated-image server suites passed separately.
- Contract generation completed twice from Codex `0.147.0-alpha.1.2` with stable output.
- App lint completed with no errors. It reported 146 existing warnings outside the changed component.
- Interactive lightbox and collapse QA could not run because this session had no connected browser instance. Component tests cover loading, error, multiple-image, identity-change, Escape, focus-return, and collapsed-work behaviour.

## Residual risks

- A persisted event authorises a mutable host path. Replacing that file later with another valid raster changes the historical image bytes.
- The route validates encoded bytes and raster signatures, but the browser remains responsible for structural decoding and decoded-dimension limits.

## Post-Deploy Monitoring & Validation

- **Window and owner:** The release owner monitors the first 24 hours after deployment and validates one live Codex Imagegen thread after all server, host-daemon, and app clients are updated.
- **Healthy signals:** Generated images appear before the final response, content requests return `200` with the expected security headers, and no generated-image row becomes `Generated image unavailable`.
- **Failure signals:** Watch server logs and request metrics for increases in `404`, `413`, `415`, or `422` responses on `/threads/:id/generated-images/content`; host-offline errors; `invalid-message` reconnect loops; and unexpected `provider/unhandled` Imagegen events.
- **Rollback:** Revert this PR and deploy the server, host daemon, contracts, and app together. Preserve the protocol gate during rollback so mixed daemon versions cannot reconnect with an incompatible wire contract.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)


SawyerHood:

PR #1311, opened 2026-08-11
Enable plugins when installed

ymichael:
## Summary

- enable every successful plugin install, including reinstalls of disabled path plugins
- add regression coverage proving reinstallation activates the plugin runtime

## Testing

- pnpm exec turbo run test --filter=@bb/server -- test/services/plugins/plugin-service.test.ts

### DIFF
--- a/apps/server/src/services/plugins/plugin-registration.ts
+++ b/apps/server/src/services/plugins/plugin-registration.ts
@@ -236,7 +236,7 @@ export function createPluginRegistration(context: PluginRegistrationContext) {
           activeArtifactId: args.activeArtifactId ?? null,
           rootDir: args.rootDir,
           version: manifest.version,
-          enabled: existing?.enabled ?? true,
+          enabled: true,
         });
         const row = getInstalledPlugin(deps.db, manifest.id);
         if (row) {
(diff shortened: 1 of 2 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** Installing a plugin is like putting a toy on a shelf. This change also turns the toy on again.

I found no blocking or non-blocking issues.

Security: Existing source checks and install confirmation still apply. Managed updates and startup checks still keep a disabled choice.

Code quality: The shared `registerInstalled` function is the correct place for this rule. I found no duplicate code that needs a refactor.

Performance: The change only adds the intended plugin load cost. Existing operation and lifecycle locks prevent duplicate starts.

Validation passed:

- All 18 focused server tests passed.
- The server type check passed.
- A live CLI test installed, disabled, and reinstalled a path plugin.
- The result had `enabled: true` and `status: running`.
- The browser showed the enabled switch for the reinstalled plugin.
- All current GitHub checks passed.

PR #1314, opened 2026-08-11
Ignore Codex raw response completion events

ymichael:
## Summary

- classify Codex `rawResponse/completed` notifications as provider noise instead of rendering an unhandled-event row
- cover both raw visibility classification and adapter suppression using the reported usage payload shape
- bump the host daemon protocol to 97 so older daemons update before forwarding the changed event stream

## Tests

- `pnpm exec turbo run typecheck --filter=@bb/agent-runtime --filter=@bb/host-daemon-contract`
- `pnpm exec turbo run test --filter=@bb/agent-runtime --force -- src/codex/adapter.test.ts src/provider-visibility.test.ts` (165 passed)
- `pnpm exec turbo run test --filter=@bb/host-daemon-contract --force` (49 passed)
- `git diff --check`

## Broader suite note

The full `@bb/agent-runtime` suite passed 875 tests but twice failed the unrelated `runtime.process-lifecycle.test.ts` stderr-tail timing assertion. The changed Codex adapter and provider-visibility suites pass cleanly.

### DIFF
--- a/packages/agent-runtime/src/codex/visibility.ts
+++ b/packages/agent-runtime/src/codex/visibility.ts
@@ -12,7 +12,11 @@ import {
 } from "../shared/provider-visibility-helpers.js";
 import type { ServerNotification } from "./generated/codex-app-server/schema/ServerNotification.js";
 
-type CodexServerNotificationMethod = ServerNotification["method"];
+type CodexServerNotificationMethod =
+  | ServerNotification["method"]
+  // Visibility can lead the independently refreshed generated schema when a
+  // newer installed Codex starts emitting a notification.
+  | "rawResponse/completed";
 
 interface CodexNotificationRawEvent {
   kind: "notification";
@@ -80,6 +84,7 @@ const CODEX_SERVER_NOTIFICATION_METHODS = {
   "model/rerouted": true,
   "process/exited": true,
   "process/outputDelta": true,
+  "rawResponse/completed": true,
   "rawResponseItem/completed": true,
   "remoteControl/status/changed": true,
   "serverRequest/resolved": true,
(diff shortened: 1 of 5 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I will post one final review after all checks finish.

packages/agent-runtime/src/codex/visibility.ts
```diff
@@ -158,6 +163,9 @@ const CODEX_NOTIFICATION_COVERAGE = {
   // there is nothing to surface — left "unknown" intentionally.
   "process/exited": "unknown",
   "process/outputDelta": "unknown",
+  // Internal per-response accounting; thread/tokenUsage/updated carries the
+  // user-facing token state.
+  "rawResponse/completed": "noise",
```
SawyerHood, line 168: 🚨 `slopcop/review` — P3: Stop this unused notification at the source.

BB enables `experimentalRawEvents` because it needs `rawResponseItem/completed` for command output recovery.

However, BB does not use `rawResponse/completed`.

This patch drops the notification only after Codex creates, serializes, sends, and BB parses it.

Codex supports an exact method filter through `initialize.params.capabilities.optOutNotificationMethods`.

Please add `rawResponse/completed` to that list and test the initialize request.

This change keeps required item events and avoids work for each upstream response.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

This change hides an extra machine receipt.

The user still sees the useful token total.

It is like discarding a duplicate shop receipt.

### Findings

- P3: BB can stop `rawResponse/completed` at Codex initialization.
- The current patch drops each event after Codex sends it.
- Codex supports an exact filter through `optOutNotificationMethods`.
- I found no security or correctness issue.
- The protocol bump to 97 is correct because the daemon output changes.

### Architecture

The notification method map duplicates the coverage map.

The code can derive membership from the coverage map and remove one source of stale keys.

This note does not identify a current bug.

### Checks

- Both affected package type checks passed.
- All 165 focused agent-runtime tests passed.
- All 49 host-daemon-contract tests passed.
- GitHub CI passed at the exact head SHA.
- The full runtime suite passed 875 of 876 tests.
- One unrelated process-tail test failed twice outside this pull-request diff.

I did not start a development server because this provider notification has no direct browser action.

The adapter tests isolate the event path more accurately.

This is a comment-only review. I found no blocking issue.

PR #1317, opened 2026-08-11
perf(app): replace react-markdown with Streamdown (PoC)

smsunarto:
Note: this is more of a PoC than anything -- I realize this whole system is pretty integral and wired together to a lot of things, so I can't say with confidence that this PR is mergeable. The goal here is to demonstrate 1) feasability and 2) perf gain is there.

Also, I'll add that it's likely most ideal in the long run to move all markdown rendering to Streamdown to avoid the additional bundle size overhead from 2 markdown renderers.

## Summary

- replace `react-markdown` with Streamdown while preserving BB's custom Markdown components, directives, mentions, raw HTML policy, and local-file routing
- stream assistant Markdown in stable blocks so completed content does not reparse on every token
- keep static rendering for queued messages and non-streaming surfaces without Streamdown's unused block-lexer pass
- restore safe URL protocol filtering and one-line queued-message truncation

## Performance evidence

Measured in the real source-built BB app with React Scan 0.5.7 and React Profiler enabled. Both runs used the same prompt and rendered 25 level-two headings, 25 lists, and 50 paragraphs. The Streamdown response was 8.1% longer, so render work per KiB is the fairest comparison.

All benchmark numbers from MBP M5 Max -- numbers are likely to differ in other devices esp. mobile which should be much worse.

| React Profiler metric | react-markdown | Streamdown | Change |
| --- | ---: | ---: | ---: |
| Render work per KiB | 103.92 ms | 21.84 ms | 79.0% lower |
| Total render work | 1,321.4 ms | 300.1 ms | 77.3% lower |
| Median commit | 9.4 ms | 0.6 ms | 93.6% lower |
| p95 commit | 21.8 ms | 1.8 ms | 91.7% lower |
| Maximum commit | 28.3 ms | 3.1 ms | 89.0% lower |
| Commits above 16.7 ms | 20 | 0 | 20 fewer |

Streamdown produced more commits (393 versus 134) because it updates smaller blocks; the individual commits were much cheaper. React Scan adds overhead, so these are instrumented component timings rather than clean frame-rate measurements. Model and network time are excluded from renderer conclusions.

### Recordings

See below recordings of the re-renders with React Scan enabled to visualize component renders more clearly.

#### Before: react-markdown

https://github.com/user-attachments/assets/b0a282dc-c471-4b84-9bfa-1cbc0df66c67

#### After: Streamdown

https://github.com/user-attachments/assets/53b85e4e-f8a5-4f66-b568-58d09a52c860

## Test plan

- `pnpm exec turbo run test --filter=@bb/app --force` — 331 files and 2,524 tests passed
- `pnpm exec turbo run typecheck --filter=@bb/app --force`
- `pnpm exec turbo run build --filter=@bb/app --force`
- `node apps/app/scripts/check-bundle-budget.mjs` — 431.0 KB Brotli boot payload against a 435.3 KB budget

## Notes

- unsafe Markdown URL schemes remain blocked in both static and streaming modes
- local `file:` links remain allowed only after BB's workspace containment checks
- reference links, footnotes, loose lists, fences, raw HTML blocks, and BB message directives keep full-document semantics


SawyerHood:

PR #1323, opened 2026-08-11
Inject the bb CLI into every automation script run

SawyerHood:
Script automations failed on this host with `bb CLI not found on PATH or in common install locations` — even when the script never called `bb`. A scheduled merge that only used `gh` and `jq` died before its first line.

## Two bugs

**1. The server had no pointer to the bb CLI.** `createServerEnv` never passed `BB_CLI_DIR`, though `createDaemonEnv` always has. bb lives in the daemon bundle directory, which is on no shell PATH, and `resolveBbBinary` only tried bare `bb` plus `/opt/homebrew/bin/bb` and `/usr/local/bin/bb`. On Linux nothing resolved. On macOS it worked only if a stray `/usr/local/bin/bb` happened to exist.

Observed on the live server process — no `BB_CLI`, no `BB_CLI_DIR`, and no bundle directory on `PATH`:

```
BB_APP_VERSION=0.36.0  BB_DATA_DIR=/home/sawyer/.bb  BB_HOST_DAEMON_PORT=38887 ...
```

**2. Failing to resolve was fatal.** `resolveBbBinary` threw, and `executeStoredScript` called it before spawning anything. Injection is a convenience for scripts that call `bb`, not a precondition for running one.

## The fix

- `createServerEnv` passes `BB_CLI_DIR`, matching `createDaemonEnv`.
- `resolveBbBinary` consults `BB_CLI`, then `BB_CLI_DIR`, then `PATH`, then the macOS paths, and returns `null` instead of throwing.
- An unresolved CLI now costs only the injection. The script runs and its output opens with `[bb] warning: ...`, so a script that does call `bb` fails on that line with its own output intact.
- Scripts also receive `BB_CLI`, matching how agent shells are set up.

## Latent PATH bug fixed along the way

When bb resolved off `PATH` as the bare name `bb`, `dirname("bb")` is `"."` — so the old code prepended `.` to `PATH`. The script cwd is the automation scripts directory, so a file named `git` or `node` next to a script would have shadowed the real binary. Only absolute paths are prepended now.

## Testing

- New unit tests for candidate ordering and the PATH construction, including the `"."` case.
- Verified end to end against the real binary: a script now reports `bb on PATH -> /home/sawyer/projects/bb/apps/host-daemon/dist/bb`, `bb --version -> 0.36.0`, `BB_CLI -> ...`.
- Verified degradation with bb unreachable: exit 0, output `[bb] warning: ...` followed by `the real work ran`.
- `turbo run typecheck test` passes for `bb-plugin-automations` (61) and `bb-app` (64).
- One `@bb/server` test fails locally (`internal-skill-trees`, a 0644-vs-0664 file-mode assertion). Confirmed identical on a clean tree — pre-existing and umask-dependent.

No `HOST_DAEMON_PROTOCOL_VERSION` bump: nothing on the server/daemon wire changes.

### DIFF
--- a/packages/bb-app/src/launcher.ts
+++ b/packages/bb-app/src/launcher.ts
@@ -2419,7 +2419,14 @@ function createServerEnv(args: CreateServerEnvArgs): NodeJS.ProcessEnv {
     // The daemon bundle holds the bb CLI. Server-side features that shell out
     // — script automations put it on the script's PATH — otherwise have no way
     // to find it: bb lives in the bundle directory, which is on no shell PATH.
-    // Matches createDaemonEnv, which has always passed this through.
+    // BB_CLI_DIR matches createDaemonEnv, which has always passed it through.
+    //
+    // BB_CLI is set rather than inherited on purpose. Launching bb-app from an
+    // agent shell brings that shell's BB_CLI along, pointing at whichever
+    // install spawned it. That binary can be older than this bundle and still
+    // answer `--version`, so an inherited value would quietly win over the
+    // bundle actually being run.
+    BB_CLI: join(args.context.daemonBundleDir, "bb"),
     BB_CLI_DIR: args.context.daemonBundleDir,
     BB_DATA_DIR: args.context.dataDir,
     BB_HOST_DAEMON_PORT: String(args.context.daemonPort),
--- a/packages/bb-app/src/launcher.ts
+++ b/packages/bb-app/src/launcher.ts
@@ -2416,6 +2416,11 @@ function createServerEnv(args: CreateServerEnvArgs): NodeJS.ProcessEnv {
     ...args.env,
     BB_APP_VERSION: args.context.appVersion,
     [APP_SURFACE_ENV_NAME]: APP_SURFACE_WEB,
+    // The daemon bundle holds the bb CLI. Server-side features that shell out
+    // — script automations put it on the script's PATH — otherwise have no way
+    // to find it: bb lives in the bundle directory, which is on no shell PATH.
+    // Matches createDaemonEnv, which has always passed this through.
+    BB_CLI_DIR: args.context.daemonBundleDir,
     BB_DATA_DIR: args.context.dataDir,
     BB_HOST_DAEMON_PORT: String(args.context.daemonPort),
     BB_SERVER_PORT: String(args.context.serverPort),
--- a/plugins/automations/src/script-runner.ts
+++ b/plugins/automations/src/script-runner.ts
@@ -1,7 +1,8 @@
 import { execFile } from "node:child_process";
-import { dirname, isAbsolute, join } from "node:path";
+import { constants } from "node:fs";
+import { delimiter, dirname, isAbsolute, join } from "node:path";
 import { promisify } from "node:util";
-import { mkdir } from "node:fs/promises";
+import { access, mkdir, stat } from "node:fs/promises";
 import {
   AUTOMATION_SCRIPT_TIMEOUT_MAX_MS,
   type AutomationScriptInterpreter,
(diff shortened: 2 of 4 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I will run the security, quality, and performance reviews in parallel. I will post one final review after I combine and verify all results.

packages/bb-app/src/launcher.ts
```diff
@@ -2416,6 +2416,11 @@ function createServerEnv(args: CreateServerEnvArgs): NodeJS.ProcessEnv {
     ...args.env,
     BB_APP_VERSION: args.context.appVersion,
     [APP_SURFACE_ENV_NAME]: APP_SURFACE_WEB,
+    // The daemon bundle holds the bb CLI. Server-side features that shell out
+    // — script automations put it on the script's PATH — otherwise have no way
+    // to find it: bb lives in the bundle directory, which is on no shell PATH.
+    // Matches createDaemonEnv, which has always passed this through.
+    BB_CLI_DIR: args.context.daemonBundleDir,
```
SawyerHood, line 2430: 🚨 `slopcop/review` — Override an inherited `BB_CLI` for the server.

`createServerEnv` keeps `args.env.BB_CLI` and adds only the current `BB_CLI_DIR`. The script resolver selects `BB_CLI` first.

A server that starts from an agent shell can therefore inject the parent installation CLI. That CLI can be old but still pass `--version`.

Set `BB_CLI` to `join(args.context.daemonBundleDir, "bb")` here. Keep `BB_CLI_DIR` if other server features need it.

plugins/automations/src/script-runner.ts
```diff
@@ -27,16 +31,70 @@ async function commandWorks(command: string, args: string[]): Promise<boolean> {
...
+ */
+export async function resolveBbBinary(
+  env: NodeJS.ProcessEnv = process.env,
+): Promise<string | null> {
   if (resolvedBbPath !== null) return resolvedBbPath;
-  const candidates = ["bb", "/opt/homebrew/bin/bb", "/usr/local/bin/bb"];
-  for (const candidate of candidates) {
+  for (const candidate of bbBinaryCandidates(env)) {
     if (await commandWorks(candidate, ["--version"])) {
       resolvedBbPath = candidate;
```
SawyerHood, line 110: 🚨 `slopcop/review` — Store the absolute executable path, not the candidate text.

When `PATH` supplies the CLI, this line stores `bb`. A relative `BB_CLI_DIR` also produces a relative value.

Later code exports this value as `BB_CLI`, but the contract promises an absolute path. A script can change `PATH` before it calls `"$BB_CLI"`.

The probe and the script can also use different directories. Resolve each successful candidate to a validated absolute path before you store it.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: Each automatic job now gets a `bb` tool. It is like the correct key in every toolbox.

I found two P1 defects:

- The launcher keeps an inherited `BB_CLI`. The resolver can select an old parent CLI instead of the current bundle CLI.
- The PATH fallback stores `bb` instead of an absolute path. This breaks the documented `BB_CLI` contract after a PATH change.

The two P1 defects should have fixes before merge. I left inline comments with concrete fixes.

Security review: I found no new privilege boundary or command injection issue. The script already runs server-side code under the server account.

Performance review: I found no new material regression. Failed CLI resolution still repeats probes because the cache stores only successful results.

Architecture review: The plugin repeats path rules from `runtime-shell-env.ts`. A small shared helper near `@bb/config` could prevent future drift.

Tests: Turbo tests and type checks passed for both packages. All 125 tests passed, and `git diff --check` passed.

End-to-end: I created and ran a real script automation. `BB_CLI` and `command -v bb` returned the same absolute path.

The script also returned bb version `0.36.0`. The Automations plugin settings route loaded without an error in the browser.

I removed the temporary automation and project. They contained no user data. I also stopped the dev server.

I used a comment review only. I did not approve or request changes.

SawyerHood:
Both P1 findings are fixed in 680548c7a.

**Inherited `BB_CLI`** — confirmed real. `createServerEnv` spreads the parent env, and the resolver consults `BB_CLI` first, so a shell-inherited value would outrank the bundle actually running. The server now sets `BB_CLI` from `daemonBundleDir`.

**Bare `bb` stored as the resolved path** — also real, and the fix is broader than the inline suggestion. Rather than resolving after the fact, `bbBinaryCandidates` now expands `PATH` itself and yields only absolute paths. Relative env pointers are skipped instead of being resolved against a process cwd unrelated to either, and empty `PATH` entries are dropped — those mean the cwd, which during a run is the automation scripts directory, so a file named `bb` beside a script could have stood in for the CLI.

Verified with `BB_CLI`/`BB_CLI_DIR` unset and only the bundle dir on `PATH`: resolves to the absolute path, and `$BB_CLI` still points at the right binary after a script rewrites `PATH`.

**On the repeated probes** — this got worse before it got better: expanding `PATH` makes the candidate list long, so a host without bb would pay a process launch plus timeout per entry. Candidates are now stat-ed for an executable file before being spawned. I left the negative result uncached deliberately, so installing bb takes effect without a server restart.

**On the shared helper near `@bb/config`** — agreed in principle, not doing it here. `bb-plugin-automations` depends only on `@bb/plugin-sdk`, so sharing with `runtime-shell-env.ts` means either a new dependency edge or a new shared package. That is worth its own PR rather than riding along on a bug fix.

PR #1338, opened 2026-08-11
Continue archived environments in new threads

ymichael:
## Summary

- add a **Continue in new thread** action to destroyed managed-environment banners
- open the root composer with the original project, **New worktree**, and **Continue from: `<branch>`** selected
- prefill the prompt with a rich `Continue from @thread:<source>` mention so the new agent can inspect the source conversation
- introduce a create-only `{ type: "continue", sourceEnvironmentId }` request across the app, SDK, and `bb thread spawn --continue-from-environment`
- preserve the archived environment's merge base and reuse its local branch name when Git permits
- when the branch is already checked out, create the normal generated branch at the exact same commit; fail explicitly if the archived branch no longer exists

## Lifecycle and compatibility

- keep the archived source environment destroyed and create a fresh managed environment for the new thread
- persist `continuation_branch_name` so provisioning retries cannot lose the archived branch intent before the actual branch result is stored
- add `continueFromBranchName` to managed host provisioning commands and bump `HOST_DAEMON_PROTOCOL_VERSION` from 100 to 101
- retain the ordinary environment provisioning lifecycle, setup script, result settlement, and cleanup paths

## Validation

- affected package typechecks pass
- app lint passes with zero errors
- 84 focused app tests pass
- 377 database tests pass, including migration replay
- 24 host-workspace provisioning tests pass
- 29 focused server tests pass
- 64 CLI tests pass
- 39 SDK tests pass
- 49 host-daemon contract tests pass
- 17 host-daemon environment dispatch tests pass
- `git diff --check` passes


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I started a security, code quality, architecture, performance, and end-to-end review for this pull request.

I will add the complete result after the parallel checks finish.

packages/host-workspace/src/provisioning.ts
```diff
@@ -180,6 +187,71 @@ async function ensureExistingWorkspaceMatches(
...
+        return exactResult;
+      }
+      if (!isBranchAlreadyCheckedOutFailure(exactResult)) {
+        const detail = exactResult.stderr.trim();
+        throw new WorkspaceError(
+          "git_command_failed",
+          `git worktree add failed${detail ? `: ${detail}` : ""}`,
+        );
+      }
+      return runGit(
```
SawyerHood, line 302: 🚨 `slopcop/review` — High: `-B` can move an existing fallback branch and lose its commits.

When the preferred branch is checked out, this code runs `git worktree add -B`.

If the fallback branch exists, Git resets it to the preferred branch commit. This can make unique commits unreachable.

Use `-b`, or verify the existing fallback ref before use. Add a test that proves the ref never moves.

apps/server/src/services/threads/thread-request-eligibility.ts
```diff
@@ -74,6 +92,71 @@ function requireHostEnvironmentId(
...
+  const mergeBaseBranch = resolveEnvironmentMergeBaseBranch(sourceEnvironment);
+  if (mergeBaseBranch === undefined) {
+    throw new ApiError(
+      409,
+      "invalid_request",
+      "Archived environment has no merge base branch to preserve",
+    );
+  }
+  assertUsableHostId(deps, { hostId: sourceEnvironment.hostId });
+  const localSource = getProjectSourceByHost(
```
SawyerHood, line 148: 🚨 `slopcop/review` — Medium: Continuation does not verify the original source repository.

This query loads the current project source for the host. The archived environment does not store its original repository identity.

After a source change, a matching branch can continue the wrong repository. Its setup script can also run.

Store and compare a stable source identity before continuation.

apps/app/src/components/promptbox/banner/ThreadPromptContextBanner.tsx
```diff
@@ -883,11 +890,17 @@ export function ThreadPromptContextBanner({
...
-        }
+        statusLabel={environmentGoneCopy?.label ?? ARCHIVED_THREAD_STATUS_LABEL}
+        showStatusActionWithParent={Boolean(
+          environmentGoneSection?.onContinueInNewThread,
+        )}
         statusAction={
-          archivedSection?.onUnarchive && !environmentGone ? (
+          environmentGoneSection?.status === "destroyed" &&
+          environmentGoneSection.onContinueInNewThread ? (
+            <ContinueInNewThreadAction
```
SawyerHood, line 902: 🚨 `slopcop/review` — Medium: The only continuation action disappears in a narrow prompt shell.

`BannerActionSlot` always has `data-promptbox-hide-tiny`. The 22-rem container rule hides this action.

The archived environment also replaces the composer. The user then has no continuation control.

Keep this action visible, or add the same action to a compact menu.

packages/host-workspace/src/provisioning.ts
```diff
@@ -180,6 +187,71 @@ async function ensureExistingWorkspaceMatches(
   return true;
 }
 
+function isBranchAlreadyCheckedOutFailure(result: GitCommandResult): boolean {
```
SawyerHood: 🚨 `slopcop/review` — Low: This fallback depends on English Git error text.

Git inherits the host locale. A translated checked-out message will not match this expression.

The continuation will fail instead of creating the fallback branch. Inspect worktree state under the lock, or force a stable Git locale.

apps/server/src/services/threads/thread-request-eligibility.ts
```diff
@@ -74,6 +92,71 @@ function requireHostEnvironmentId(
...
+    environment.sourceEnvironmentId,
+  );
+  if (sourceEnvironment.projectId !== projectId) {
+    throw new ApiError(
+      409,
+      "invalid_request",
+      "Environment belongs to a different project",
+    );
+  }
+  if (
```
SawyerHood, line 125: 🚨 `slopcop/review` — Medium: Deleted work can pass this archived-environment check.

Thread deletion also destroys its environment. A caller that retained the environment ID can submit a continuation request after cleanup.

This creates a new thread from work that the user deleted. Require an archive-specific destruction reason or a durable continuation flag.

apps/app/src/components/plugin/new-thread-environment-seed.ts
```diff
@@ -29,6 +29,9 @@ export function newThreadEnvironmentArgsToSeed(
   if (environment.type === "project-default") {
     return null;
   }
+  if (environment.type === "continue") {
```
SawyerHood, line 51: 🚨 `slopcop/review` — Low: This breaks the documented plugin composer round trip.

`defaultEnvironment` promises that an unchanged submitted environment produces an equivalent request, except for its listed limits.

The new public `continue` variant silently selects the normal default instead. Support this seed or document the exception in the public contract.

apps/app/src/views/RootComposeView.tsx
```diff
@@ -1333,10 +1413,39 @@ export function RootComposeView() {
...
     onCreateBranch: handleCreateBranch,
     onCreateBranchFrom: handleCreateBranchFrom,
+    onContinueFrom: handleContinueFrom,
   } = useScopedBranchSelection({
     environmentValue: effectiveEnvironmentValue,
+    initialSelection: continueEnvironmentSeed
+      ? {
+          name: continueEnvironmentSeed.branchName,
+          isNew: false,
+          continueFromEnvironmentId:
```
SawyerHood, line 1372: 🚨 `slopcop/review` — Medium: Cold continuation waits for an unnecessary full branch refresh.

This selection enters worktree mode. The existing branch query then blocks submission while the daemon fetches and scans the repository.

A continuation request only needs `sourceEnvironmentId`. Disable the branch query gate until the user selects an ordinary branch.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

Think of an archived workspace as a toy box that someone put away. This change opens a new box with the saved toys.

It keeps the old box closed.

### Review result

I reviewed exact SHA `b7593e1595`. The pull request now points to a newer SHA, so these comments describe the requested commit.

I found one high-risk issue, four medium-risk issues, and two low-risk issues.

- High: The fallback `git worktree add -B` can reset an existing local branch. This can make its unique commits unreachable.
- Medium: A destroyed environment from a deleted thread can pass the archive continuation check.
- Medium: Continuation uses the current project source. A source change can make it open a different repository with the same branch name.
- Medium: The narrow prompt layout hides the only continuation action.
- Medium: A cold continuation waits for a full branch refresh that the request does not need.
- Low: The plugin composer drops the new `continue` environment, although its public contract promises an equivalent round trip.
- Low: The checked-out branch fallback parses English Git error text and can fail under another host language.

I found no command injection or cross-project access path. The server derives the host, branch, and merge base from stored data.

### Architecture

The public request correctly carries only the source environment ID. The server owns the Git details.

I recommend an explicit checkout union for new work and continued work. This will replace the optional `continueFromBranchName` field.

The environment should also store the cleanup reason and original source identity. These fields will prevent both provenance errors.

### Validation

- `git diff --check` passed.
- The affected Turbo test run completed all six tasks.
- The host contract passed 49 tests. The host workspace passed 197 tests.
- The app passed 2,506 tests in 332 files.
- A later broad server run passed 1,390 tests. Sixteen tests hit npm cache permissions and file-watch limits.
- The browser flow passed on the exact SHA.
- The flow archived a real worktree and created a new thread from it.
- The new environment kept the branch, commit, added file, and merge base.

I left inline comments for each finding. I did not approve the pull request or request changes.

apps/app/src/components/plugin/new-thread-environment-seed.ts
ymichael: Fixed in 3fc366200 by supporting the round trip. The plugin composer resolves the archived environment ID, seeds its host and branch with continuation intent, and submits the original continue request untouched. Its integration test also verifies the normal branch query stays disabled.

apps/app/src/views/RootComposeView.tsx
ymichael: Fixed in 3fc366200. Both root and plugin composers detect the continuation branch seed, disable the ordinary project branch query, and remove the managed-worktree availability gate until the user selects an ordinary branch.

packages/host-workspace/src/provisioning.ts
ymichael: Fixed in 3fc366200. Continuation now checks whether the generated fallback ref already exists. An existing free fallback is reused without moving its ref; an occupied fallback is rejected; only a missing fallback is created with -b. The new host-workspace regression test gives the fallback a unique commit and asserts both HEAD and the ref remain unchanged.

apps/app/src/components/promptbox/banner/ThreadPromptContextBanner.tsx
ymichael: Fixed in 3fc366200. Read-only status actions can opt out of the tiny-shell hiding rule, and the archived continuation CTA now remains visible below 22rem. The banner test asserts the hide-tiny marker is absent.

apps/server/src/services/threads/thread-request-eligibility.ts
ymichael: Fixed in 3fc366200. Continuation now requires the destroyed environment to still have at least one archived, non-deleted thread. A regression test archives and deletes the source thread, then verifies the retained environment ID is rejected with 409.

apps/server/src/services/threads/thread-request-eligibility.ts
ymichael: Fixed in 3fc366200. Managed environments now persist the project source checkout path used at creation. Continuation compares that provenance with the current host source before creating anything and rejects a remap. Migration 0093 backfills existing managed rows, and the server test covers a changed source checkout.

packages/host-workspace/src/provisioning.ts
ymichael: Already fixed by 7052fb6bf and retained here. The implementation no longer parses Git stderr. It inspects git worktree list --porcelain under the provisioning lock and compares exact branch refs, so host locale does not affect fallback behavior.

PR #1340, opened 2026-08-11
Restore provider-specific composer selections

### LINKED ISSUE #1288: Automatically select a model and reasoning level when switching composer providers
## Summary

Switching providers in the composer currently requires a second action to select a model or reasoning level. Selecting a provider tab should immediately switch providers and automatically select that provider's previous model and reasoning level.

## Expected behavior

When a user selects a provider tab:

- Switch to that provider immediately.
- Select its last-used model and reasoning level.
- Close the picker or keep it open with the restored values selected.
- Do not require the user to select a model or reasoning level to complete the provider switch.

If the provider has no previous selection, select its default model and reasoning level.

## Example

1. Codex previously used **5.6 Sol -> High**.
2. Claude Code previously used **Opus 5 -> Max**.
3. Select the Claude Code tab.
4. The composer immediately switches to **Claude Code -> Opus 5 -> Max**.

ymichael:
## Summary

- commit provider changes as soon as a composer provider tab is selected
- persist model and reasoning preferences independently for each provider
- restore prior provider selections, or use the provider catalog defaults when no history exists
- retain equivalent in-memory history for component-local/plugin composers without changing root-composer preferences
- migrate the existing unscoped model/reasoning preferences without leaking them to a newly selected provider

## Testing

- `pnpm exec turbo run test typecheck --filter=@bb/app --force`
- 332 test files passed, 2,505 tests passed

Closes #1288

### DIFF
--- a/apps/app/src/hooks/useThreadCreationOptions.ts
+++ b/apps/app/src/hooks/useThreadCreationOptions.ts
@@ -67,6 +67,11 @@ type ReasoningLevelSelectionSetter = (value: ReasoningLevel) => void;
 type PermissionModeSelectionSetter = (value: PermissionMode) => void;
 type ClearSelectionHandler = () => void;
 
+interface ProviderModelReasoningSelection {
+  model: string;
+  reasoningLevel: ReasoningLevel;
+}
+
 export interface UseThreadCreationOptionsResult<TExecutionInputSources> {
   executionOptionsRouting: SystemProvidersQuery;
   selectedProviderId: string;
@@ -163,12 +168,8 @@ export function useThreadCreationOptions(
   } = options ?? {};
   const { setValue: setStoredProviderId, value: storedProviderId } =
     usePromptBoxProviderPreference();
-  const { setValue: setStoredSelectedModel, value: storedSelectedModel } =
-    usePromptBoxModelPreference();
   const { setValue: setStoredServiceTier, value: storedServiceTier } =
     usePromptBoxServiceTierPreference();
-  const { setValue: setStoredReasoningLevel, value: storedReasoningLevel } =
-    usePromptBoxReasoningLevelPreference();
   const { setValue: setStoredPermissionMode, value: storedPermissionMode } =
     usePromptBoxPermissionModePreference();
   const {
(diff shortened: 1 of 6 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am reviewing this PR now. I will check security, code quality, performance, architecture, and the affected app workflow.

apps/app/src/hooks/useThreadCreationOptions.ts
```diff
@@ -327,6 +327,29 @@ export function useThreadCreationOptions(
     return providers[0]?.id ?? "";
   }, [providers, rawSelectedProviderId]);
 
+  const { setValue: setStoredSelectedModel, value: storedSelectedModel } =
+    usePromptBoxModelPreference(effectiveProviderId);
```
SawyerHood, line 343: 🚨 `slopcop/review` — Keep provider handoffs atomic

These setters bind to `effectiveProviderId` for this render. `RootComposeView` applies a fork seed by calling `setSelectedProviderId`, `setSelectedModel`, and `setReasoningLevel` in one effect. React batches those calls, so the model and reasoning setters still write the source provider atoms. The fork then uses the target provider default or old selection. It also replaces the source provider preferences. Please add an atomic provider, model, and reasoning setter, plus a fork regression test.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: BB now remembers a separate model and reasoning choice for each provider. Changing providers and returning restores the prior choice.

I found one blocking correctness defect. The new model and reasoning setters bind to the provider from the current render. The fork path sets the provider, model, and reasoning in one React batch. The later setters therefore write to the old provider. The fork uses the target provider default or prior choice, and the source preference changes. I left an inline comment with the repair direction and test case.

Security review found no security or privacy issue. Performance review found no material cost. The architecture review found no other material duplication or refactor need.

Validation passed:

- The full app suite passed: 332 files and 2,505 tests.
- The Turbo app type check passed.
- `git diff --check` passed.
- All GitHub checks passed.
- The browser test restored Codex 5.6-Terra with Medium and Claude Sonnet 5 with Medium after provider changes.

apps/app/src/hooks/useThreadCreationOptions.ts
ymichael: Fixed in 4c11a330a. Fork seeding now uses a single setProviderModelReasoning(...) operation that writes model/reasoning to the target provider atoms before switching the active provider, so batched React updates cannot bind those writes to the source provider. The same operation handles component-local selection state atomically. Added a regression test that seeds a different provider/model/reasoning tuple, then switches between providers to verify both the target tuple and the source provider remembered values are preserved.

Validation: full @bb/app test suite passes (332 files / 2,506 tests), typecheck passes, and lint has 0 errors (existing warnings only).

PR #1341, opened 2026-08-11
Add an OG card so shared bb links unfurl with an image

brsbl:
Sharing a getbb.app link produced a blank preview — the site set `og:title` and `og:description` but never an `og:image`, so X and Slack fell back to a placeholder icon.

## The card

New 1200×630 (@2x) image at `apps/web/public/og.png`, built from the landing page's own tokens: white background, `oklch(0.3211 0 0)` ink, Inter, and the bb mark. It carries the hero headline and the closer's "Free, open source, and local-first." The regenerable source is checked in at `assets/og-card.html` — render it with `shot assets/og-card.html og.png --width 1200 --height 630 --scale 2`.

## The tags

Moved into a shared `unfurlMeta()` helper in `landing/site.ts` so the landing and changelog routes stay in sync:

- `og:image` + `og:image:width` / `:height` / `:alt`
- `og:url`, `og:site_name`
- `twitter:card=summary_large_image` — without it X renders a small thumbnail instead of the large card
- `twitter:title` / `:description` / `:image`

Image URLs are absolute (`https://getbb.app/og.png`) because scrapers fetch the tags with no base URL to resolve against.

## Copy

Every string is lifted verbatim from the landing page. The unfurl title is `bb` alone and the description is the hero sub, so no line repeats between the card image and the text a platform prints beside it. The browser tab `<title>` is unchanged apart from matching the hero h1.

## Verification

Built the site and curled `/` and `/changelog` from `vite preview` — all tags render in the SSR HTML that scrapers fetch, and `/og.png` serves 200. `pnpm typecheck` passes. (`src/server/api.test.ts` has 29 pre-existing failures on a clean tree, unrelated to this change.)

After deploy, X and Slack cache old unfurls — revalidate at https://cards-dev.twitter.com/validator or by re-pasting the link.


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

apps/web/src/landing/site.ts
```diff
@@ -36,6 +24,35 @@ export function downloadMacosHref(placement: CtaPlacement): string {
   return `${DOWNLOAD_MACOS_REDIRECT_PATH}?placement=${placement}`;
 }
 
+export const SITE_URL = "https://getbb.app";
```
SawyerHood: 🚨 `slopcop/review` — Staging links always identify themselves as production.

The staging worker sets `APP_URL` to `https://vibecodethis.site`. This constant sets both `og:url` and `og:image` to `https://getbb.app`. A staging share cannot validate its own card, and it identifies itself as production. Please derive the origin from the deployment configuration or request. Then pass that origin into `unfurlMeta`.

assets/og-card.html
```diff
@@ -0,0 +1,85 @@
+<!doctype html>
+<!-- Source for apps/web/public/og.png (the link-unfurl card).
+     Regenerate with a 1200x630 @2x headless-Chrome screenshot of this file,
+     e.g.: shot "file://$PWD/og-card.html" ../apps/web/public/og.png \
```
SawyerHood: 🚨 `slopcop/review` — The card generation command is not reproducible from a clean checkout.

This command requires an undeclared `shot` program and an `assets` working directory. The repository cannot reliably reproduce the committed PNG. Please add a package script with a fixed screenshot tool and stable paths.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** This PR gives shared bb links a picture card. The card works, but staging uses production links.

I reviewed commit `53bcc7bae2`. The pull request moved to `c255b37e39` during the review.

The new head resolves the old merge conflicts. All current GitHub checks pass.

I found two issues:

1. **P2:** Staging returns production values for `og:url`, `og:image`, and `twitter:image`.
2. **P3:** The card generation command depends on an unavailable `shot` program and a specific current directory.

I left inline comments with concrete fixes for both issues.

The security review found no issue. The metadata uses fixed values, and the PNG contains no private metadata.

The architecture review found no duplicate metadata helper. The new helper centralizes the Open Graph and Twitter tags.

The end-to-end test covered `/`, `/changelog`, and `/og.png`. Each route returned the correct metadata and no browser error.

The Turbo build and type check passed. All 46 tests passed at the requested commit.

The current head also passed the Turbo build, type check, 55 tests, and all GitHub checks.

The PNG is 411,633 bytes at 2400×1260. A smaller file is possible, but the review gate did not classify this size as a defect.

apps/web/src/landing/site.ts
brsbl: Fixed in 0c1388e — you're right, and staging couldn't validate its own card.

`CLOUDFLARE_ENV` already selects the wrangler env at build time (deploy-web.yml sets it for the production build), so the origin is now read back out of that env's `APP_URL` via wrangler's own `unstable_readConfig` — which handles the JSONC and the production env's inheritance — and injected as a build-time constant.

Verified by building both ways and grepping the bundle:

- default env → `https://vibecodethis.site`
- `CLOUDFLARE_ENV=production` → `https://getbb.app`

`resolveSiteOrigin` throws on a missing or malformed `APP_URL` rather than defaulting: a wrong origin here is invisible until someone shares a link, so a failed build is the cheaper failure. Covered by `src/server/site-origin.test.ts`.

assets/og-card.html
brsbl: Fixed in 0c1388e. The template moved to `apps/web/scripts/og-card.html` with a generator beside it:

```
pnpm --filter @bb/web og:card
```

- Every path resolves from the script file, so the working directory doesn't matter (verified by running it from `/tmp`).
- The font comes from the `@fontsource-variable/inter` dependency via `require.resolve`, replacing the hand-written `node_modules` path — which was already wrong under pnpm.
- No new dependency: it drives whatever Chrome is installed (`CHROME_PATH` or the usual locations) over DevTools. Not `--headless --screenshot` — recent Chrome hangs on that switch and never writes the file.

One thing worth flagging, since it argues for this being a script rather than a doc comment: the first two runs produced a card with a **broken-image box** where the logo should be. Two causes — a `file://` document may only read its own directory, and the template's own comment names the `__LOGO_URL__` placeholder, so a first-occurrence `.replace()` rewrote the comment and left the real `src` as the literal token. Nothing downstream would have caught either; the tags and the build were green with a visibly broken card.

So the generator now waits for the load event, then fails if the webfont or any image didn't decode. It reproduces the committed PNG byte-for-byte (sha256 `ad7332fb…`, unchanged in this commit).

PR #1349, opened 2026-08-11
Add thread hover cards example plugin

brsbl:
Adds the local thread hover-card plugin as an official reference example.

- ports the server RPCs, content-script UI, styles, and behavior tests
- uses stable section and project IDs without expanding the public plugin API
- keeps active previews bounded and removes the full-sidebar background warmer
- adapts lifecycle cleanup and theme color mixing to current bb conventions
- documents bb-plugins as the canonical implementation and this example as an adapted downstream port
- documents the example in the plugin authoring skill and bb guide plugins

Verification:
- pnpm exec turbo run typecheck test build --filter=bb-plugin-thread-hover-cards --force
- pnpm exec turbo run typecheck --filter=@bb/app --force
- pnpm exec turbo run test --filter=@bb/app --force -- src/components/sidebar/ProjectListSectionHeader.test.tsx src/components/sidebar/ProjectRow.interactions.test.tsx
- pnpm exec turbo run test --filter=@bb/server --force -- test/services/plugins/plugin-authoring-docs.test.ts
- installed and reloaded in the isolated dev app; the hover card rendered from the final bundle with six handler calls and zero errors
- final staged architecture, implementation, and test re-review: clean

> AGENT GENERATED: by GPT-5.6


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am now reviewing this pull request.

I will check security, code quality, performance, duplicate code, architecture, and end-to-end behavior.

examples/plugins/thread-hover-cards/server.ts
```diff
@@ -0,0 +1,1466 @@
...
+    output: threadTimingSchema,
+  },
+  threadPullRequest: {
+    input: z.object({ threadId: z.string().min(1) }).strict(),
+    output: threadPullRequestSchema,
+  },
+  sectionSummary: {
+    input: z
+      .object({
+        name: z.string().min(1).max(256),
```
SawyerHood: 🚨 `slopcop/review` — Use stable IDs for section identity

This RPC uses display names as identity. Project names are not unique. A custom section can also use a built-in name such as `Pinned`.

Duplicate project names make this handler reject a valid row. A custom `Pinned` section makes the built-in group show unrelated counts.

Please add stable section and project IDs to the sidebar row contract. Send those IDs through this RPC.

examples/plugins/thread-hover-cards/server.ts
```diff
@@ -0,0 +1,1466 @@
...
+        const loadOutlineMessage = (
+          stageName: "messageOutline" | "messageOutlineFallback",
+        ): Promise<string | null> =>
+          measureStage(
+            recorder,
+            stageName,
+            () =>
+              within(
+                safely(
+                  bb.sdk.threads.conversationOutline({
```
SawyerHood: 🚨 `slopcop/review` — Avoid a full-thread projection on active hovers

`conversationOutline` reads and projects the complete thread history. Its cache changes after each new event.

The client refreshes active summaries after two seconds. A long active thread can therefore cause repeated full-history work from a pointer hover.

Please use a bounded output query. Add a bounded latest-assistant-preview SDK method if active text requires one.

examples/plugins/thread-hover-cards/server.ts
```diff
@@ -0,0 +1,1466 @@
...
+  // from memory and spends its whole budget on the one query it needs.
+  bb.background.service("section-directory-warm", {
+    async start(signal) {
+      while (!signal.aborted) {
+        const startedAt = monotonicNow();
+        sectionDirectory.delete("sections");
+        sectionDirectory.delete("projects");
+        const [sections, projects] = await Promise.all([
+          sectionDirectory.get("sections", () =>
+            within(safely(bb.sdk.threadSections.list({ signal })), 10_000),
```
SawyerHood: 🚨 `slopcop/review` — Do not warm sections through the full sidebar bootstrap

`threadSections.list()` calls `/sidebar-bootstrap`. That route builds all public projects and their threads.

This service runs at plugin load and every five minutes. It performs that work even when no user opens a card.

Please add a lightweight section-list route. Stable section IDs on sidebar rows could also remove this warmer.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

This change adds small cards beside each thread. The cards work like name tags. They show what an agent did without opening the thread.

The feature works, but two matching names can confuse it. Some hovers also read much more data than they need.

### Findings

I found three important issues.

1. **High — active hovers can project the complete thread history.** `conversationOutline` reads every relevant event. Active writes invalidate its cache. The client refreshes active summaries after two seconds.
2. **Medium — display names act as section and project IDs.** Project names are not unique. Built-in group names can also match custom section names.
3. **Medium — the background warmer reads the complete sidebar bootstrap.** It loads all public projects and threads at startup and every five minutes.

Two additional performance risks need follow-up. A zero hover delay starts server work during brief pointer passes. Section summaries also fetch full thread rows for simple counts.

### Architecture and duplicate code

The client repeats the same RPC request and envelope code four times. A typed helper can own this behavior.

The thread and section controllers repeat card setup, timers, position listeners, and cleanup. A small shared card shell can reduce that code.

The plugin also copies private sidebar rules and selectors. Stable host row IDs and a bounded summary SDK method would remove most fragile code.

### Security

I found no security issue. The code uses text nodes for message content. The RPC schemas bound input, output, messages, and caches.

### Verification

- The type check passed.
- Both test files passed. They contain two tests in total.
- The plugin build passed.
- `git diff --check` passed.
- The live plugin loaded with a compatible bundle and a running service.
- The browser test passed pointer hover, keyboard focus, viewport placement, disable cleanup, and re-enable recovery.

The visual result is clean, and the lifecycle cleanup is strong. The identity and hot-path issues need correction before this example becomes a reference.

examples/plugins/thread-hover-cards/server.ts
brsbl: Fixed in caf82d49c. Sidebar rows now expose stable internal section/project IDs, the content script sends those IDs, and the RPC queries by them. Regression coverage includes duplicate project display names and a custom section named Pinned.

examples/plugins/thread-hover-cards/server.ts
brsbl: Fixed in caf82d49c. Summary previews now use a one-segment, non-nested timeline query for active and idle threads; conversationOutline and its full-history fallback are no longer used. Tests assert the exact bounded arguments and zero outline calls.

examples/plugins/thread-hover-cards/server.ts
brsbl: Fixed in caf82d49c. The section directory cache and five-minute warmer are removed. Section summaries use the stable section ID directly with a targeted threads.list query, so plugin activation no longer calls threadSections.list or sidebar-bootstrap.

brsbl:
Follow-up on the review summary: the three blocking identity/query/warmer findings are fixed in caf82d49c, along with section-anchor replacement cleanup found during the review loop.

I also reviewed the non-blocking observations. The zero hover delay remains intentional source behavior, section counts now come from a targeted section/project thread query, and the duplicated request/controller code plus private selector coupling are broader refactoring notes rather than required changes for this example port. This PR does not expand the public plugin API.

brsbl:
Folded into #1474 — the sidebar hover-target integration and the Thread Hover Cards plugin now ship together in the official-plugins layer; the vendored example-plugin copy was removed earlier in favor of the pinned catalog entry.

PR #1353, opened 2026-08-11
Fix stale terminal tabs that cannot be closed

### LINKED ISSUE #1252: Cannot close already opened tab
Sometimes (cannot property understand when), i cannot close already opened tabs in the right panel.
May be it happens, when there is an error with other tabs.
But it have already happened multiple times in different projects and situations

For example in my case, i tried to open recent documents, there is an error (maybe, because i was switching between projects, doesn't matter in this particular case)
and now i cannot close zsh tab

Image

ymichael:
## Summary

- make daemon terminal closes idempotent by acknowledging terminals already absent from memory
- reconcile daemon-owned terminal rows to exited when close acknowledgement fails or times out
- return the reconciled terminal state to clients so panel tab synchronization removes the stale tab
- bump the host-daemon protocol version and add daemon, server, contract, and app regression coverage

## Why

If the server still recorded a terminal as running after the daemon had forgotten its PTY, the daemon silently ignored terminal.close. The server timed out without changing persisted state, and terminal synchronization kept restoring the tab indefinitely.

## Testing

- 25 host-daemon terminal manager tests
- 35 server public terminal route tests
- 12 app terminal tab synchronization tests
- 35 host-daemon contract tests
- Turbo typechecks for @bb/app, @bb/server, @bb/host-daemon, and @bb/host-daemon-contract

Closes #1252

### DIFF
--- a/apps/server/src/services/terminals/terminal-session-lifecycle.ts
+++ b/apps/server/src/services/terminals/terminal-session-lifecycle.ts
@@ -91,6 +91,7 @@ interface PendingTerminalRpcKey extends PendingRpcKey {
 }
 
 interface PendingTerminalCloseKey extends PendingRpcKey {
+  closeReason: TerminalSessionCloseReason;
   daemonSessionId: string;
   terminalId: string;
 }
@@ -511,7 +512,12 @@ export class TerminalSessionLifecycle {
         "Timed out attaching terminal session",
       ),
     });
-    this.pendingCloses = new PendingRpcRegistry({
+    this.pendingCloses = new PendingRpcRegistry<
+      PendingTerminalCloseKey,
+      TerminalSessionRow
+    >({
+      onFail: (pending, error) =>
+        this.finalizeFailedTerminalClose(pending, error),
       timeoutMs: closeTimeoutMs,
       timeoutError: terminalTimeoutError(
         "terminal_close_timeout",
@@ -952,6 +958,7 @@ export class TerminalSessionLifecycle {
     }
 
     const pending = this.pendingCloses.claim({
+      closeReason: args.payload.reason,
       daemonSessionId: current.daemonSessionId,
       rpcKey: terminalRpcKey(current.daemonSessionId, current.id, current.id),
       terminalId: current.id,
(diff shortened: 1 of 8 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

apps/server/src/services/terminals/terminal-session-lifecycle.ts
```diff
@@ -1735,6 +1757,52 @@ export class TerminalSessionLifecycle {
...
+      scope: {
+        daemonSessionId: pending.daemonSessionId,
+        kind: "daemon",
+        statuses: DAEMON_OWNED_TERMINAL_STATUSES,
+        terminalId: pending.terminalId,
+      },
+      update: {
+        closeReason: pending.closeReason,
+        exitCode: null,
+        kind: "exit",
```
SawyerHood, line 1780: 🚨 `slopcop/review` — A failed close can hide a live terminal process.

The `onFail` hook also runs for `host_disconnected`, not only for the timeout. The `!sent` branch rejects the close before it marks the daemon sessions as disconnected. This update then changes the row to `exited` and clears its daemon session ID. However, the daemon did not receive the close request. The reconnect cleanup selects only `disconnected` rows, so it cannot stop this process. Please restrict this finalization to `terminal_close_timeout`, or keep the row `disconnected` until the reconnect cleanup runs.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** The change removes a terminal tab when the app loses its terminal. One path can remove the tab while the terminal still runs.

I found one important issue in `fbebe88`. The failure hook can mark a terminal as exited when the server never sent the close request. This clears the daemon session ID. The reconnect cleanup only selects disconnected rows. A terminal process can then run without any user control. I left an inline comment with the exact path and a proposed fix.

The security review found no authorization bypass, cross-host access, or information leak. The daemon reply remains limited by both the daemon session ID and terminal ID.

The performance review found no new loop or unbounded work. The normal missing-terminal path now returns immediately. The timeout path adds one database update and limited notifications.

The architecture review traced the pending RPC registry, all close failures, reconnect cleanup, and daemon process control. The code does not need a new shared abstraction. The failure hook must distinguish a timeout from a delivery failure, or preserve the disconnected state.

I ran the four changed test areas through Turbo. All 107 tests passed. I also ran Turbo typechecks for the app, server, host daemon, and host-daemon contract. All typechecks passed. `git diff --check` also passed.

I started the dev app and created a real thread terminal. I opened its tab, closed it in the browser, and confirmed that the terminal list became empty. This test covered the normal close flow. The focused server test covered the missing acknowledgement timeout.

The pull request head moved to `f786f96` after this review started. This review and its marker cover the required `fbebe88` commit only.

PR #1354, opened 2026-08-11
Prioritize exact user-command matches in slash command suggestions

### LINKED ISSUE #1306: Prioritize exact user-command matches in slash command suggestions
## Summary

Slash command suggestions group skills before user commands, even when the typed value exactly matches a user command. This hides the intended command below unrelated skill matches and makes the first selected result misleading.

**Screenshot: First match shown are skills**

Image

**Screenshot: User command shown at the bottom**

Image

## Reproduction

1. Open a prompt composer that has the `/plan` user command.
2. Type `/plan`.
3. Observe that skills with “plan” in their name or description appear first.
4. Scroll to the bottom to find the exact `plan` user command.

## Expected behavior

The exact `plan` user-command match appears first and is selected by default.

An exact user-command name match must take priority over non-exact skill matches. The existing section order can remain for results with the same match quality.

## Actual behavior

The Skills section appears first. The exact `plan` user command appears in the User commands section at the bottom of the results.

This makes Enter or Tab select an unrelated skill unless the user scrolls to the exact command.

## Acceptance criteria

- An exact user-command name match appears before non-exact matches from other sections.
- The exact match is the initial keyboard selection.
- Keyboard navigation order matches the visible result order.
- Other results remain grouped under their existing section labels.
- The behavior is consistent in all prompt composers that use slash command suggestions.

ymichael:
Fixes #1306

## Problem

The slash command typeahead sorted purely by menu section (`agent-command` → `skill` → `project-command` → `user-command`). Typing `/plan` — the full name of a user command — put every skill that merely mentions "plan" above it, so the exact match sat at the bottom of the list and Enter/Tab applied an unrelated skill. Query relevance only broke ties *within* a section, so it could never rescue a command from a section that sorts late.

## Fix

`orderCommandSuggestions` (formerly `orderCommandSuggestionsBySection`) now takes the active query and ranks in two steps:

1. **Match quality outranks section.** One ladder for how directly the query names a command: whole name → name prefix → matched only through description/argument hint. So `/plan` puts the `plan` user command first, and `/pla` does too, ahead of skills that only mention "plan" in prose. Matches of equal quality keep the existing `PROVIDER_COMMAND_SECTIONS` order, and an empty query prefix-matches everything, leaving today's pure section order untouched. Namespaced skills (`ottonomous:review`) count `review` as naming them, matching the existing prefix-ranking rule.
2. **Collapse sections so their rows are contiguous**, ordered by first appearance — each section lands under its own best match. This is what keeps the hoist honest: `MentionMenu` renders one header per section, so a section whose rows were scattered through the flat list would paint in a different order than the composer walks for Arrow/Enter. Hoisting `plan` therefore brings its "User commands" header to the top with the rest of that section behind it, rather than splitting the section into two headers.

`PromptBoxInternal` ranks against the query under the caret and is the single place every composer's command list is ordered before it reaches rendering, keyboard navigation, and apply — so the follow-up composer, new-thread composer, embedded chat, root compose view, and plugin composer all get the same behavior. `useCommandSuggestions` shares the same comparator so its pre-ordered output agrees.

## Acceptance criteria

- ✅ Exact user-command match appears before non-exact matches from other sections
- ✅ Exact match is the initial keyboard selection (selection resets to index 0 on every query change)
- ✅ Keyboard navigation order matches visible order (guaranteed by the contiguous-section collapse)
- ✅ Other results stay under their existing section labels, one header each
- ✅ Consistent across every composer, since all of them render through `PromptBoxInternal`

Prefix ranking across sections goes slightly beyond the issue's wording, which asked only about exact matches — without it `/pla` still reproduces the original complaint.

## Testing

- New `command-suggestion-order.test.ts` covers the exact and prefix hoists, section contiguity, equal-quality section fallback, namespaced skills, casing/whitespace, and the unchanged empty-query order.
- New `PromptBoxInternal` test renders `/plan` with suggestions in server order and asserts the rendered row order, one header per section, that `plan` is the highlighted row, and that Enter applies it. Reverting the three source files to `main` while keeping that test reproduces the bug: rows render `planner, planning-doc, plan, plan-b`.
- `pnpm exec turbo run typecheck --filter=@bb/app` and the full `@bb/app` suite (2530 tests) pass; lint warning count unchanged from baseline.

### DIFF
--- a/apps/app/src/components/promptbox/mentions/types.ts
+++ b/apps/app/src/components/promptbox/mentions/types.ts
@@ -135,63 +135,53 @@ function commandSuggestionSearchNames(
 }
 
 /**
- * Relevance order for a lowercased, trimmed query. Typing a command's whole
- * name is an unambiguous request for that command, so an exact name match
- * outranks every non-exact match no matter which section it lives in — a `/plan`
- * user command must not sit below skills that merely mention "plan". Matches of
- * equal quality keep the `PROVIDER_COMMAND_SECTIONS` order, and within a section
- * a name-prefix match beats a description-only match. An empty query ties every
- * row, leaving pure section order.
+ * How directly the query names a command. Lower wins: the whole name, then a
+ * name prefix, then a row that only matched through its description or argument
+ * hint. An empty query prefix-matches everything, so it ranks every row alike.
+ */
+function commandSuggestionMatchRank(
+  suggestion: ComposerCommandSuggestion,
+  normalizedQuery: string,
+): number {
+  const names = commandSuggestionSearchNames(suggestion);
+  if (names.includes(normalizedQuery)) {
+    return 0;
+  }
+  return names.some((name) => name.startsWith(normalizedQuery)) ? 1 : 2;
+}
+
+/**
+ * Relevance order for a lowercased, trimmed query. How directly the query names
+ * a command outranks which section that command lives in: typing `/plan` in full
+ * is an unambiguous request for the `plan` user command, and even a partial
+ * `/pla` names it more directly than a skill that merely mentions "plan" in its
+ * description. Matches of equal quality keep the `PROVIDER_COMMAND_SECTIONS`
+ * order, so an empty query — which prefix-matches every row — leaves pure
+ * section order.
  */
 export function compareCommandSuggestions(
   left: ComposerCommandSuggestion,
   right: ComposerCommandSuggestion,
   normalizedQuery: string,
 ): number {
-  const leftExact = isExactCommandNameMatch(left, normalizedQuery);
-  const rightExact = isExactCommandNameMatch(right, normalizedQuery);
-  if (leftExact !== rightExact) {
-    return leftExact ? -1 : 1;
-  }
-
-  const bySection = compareCommandSuggestionSections(left, right);
-  if (bySection !== 0) {
-    return bySection;
-  }
-
-  const leftPrefix = hasCommandNamePrefixMatch(left, normalizedQuery);
-  const rightPrefix = hasCommandNamePrefixMatch(right, normalizedQuery);
-  if (leftPrefix !== rightPrefix) {
-    return leftPrefix ? -1 : 1;
-  }
-  return 0;
-}
-
-function isExactCommandNameMatch(
-  suggestion: ComposerCommandSuggestion,
-  normalizedQuery: string,
-): boolean {
-  return commandSuggestionSearchNames(suggestion).includes(normalizedQuery);
-}
-
-function hasCommandNamePrefixMatch(
-  suggestion: ComposerCommandSuggestion,
-  normalizedQuery: string,
-): boolean {
-  return commandSuggestionSearchNames(suggestion).some((name) =>
-    name.startsWith(normalizedQuery),
-  );
+  const byMatch =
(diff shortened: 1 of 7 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the Slop Cop. I am reviewing this PR now.

I will check security, code quality, performance, architecture, and the user workflow.

apps/app/src/components/promptbox/mentions/types.ts
```diff
@@ -109,22 +111,111 @@ export function toProviderCommandSuggestion(
...
+    return leftPrefix ? -1 : 1;
+  }
+  return 0;
+}
+
+function isExactCommandNameMatch(
+  suggestion: ComposerCommandSuggestion,
+  normalizedQuery: string,
+): boolean {
+  return commandSuggestionSearchNames(suggestion).includes(normalizedQuery);
```
SawyerHood: 🚨 `slopcop/review` — A namespaced skill alias still beats the literal user command

For `[skill("plugin:plan"), userCommand("plan")]`, the query `plan` makes both rows exact here. The section rank then puts the skill first. Enter applies the wrong row.

Give a full-name match a better rank than a bare skill alias. Add this case to the tests.

apps/app/src/hooks/useCommandSuggestions.ts
```diff
@@ -216,8 +195,9 @@ export function useCommandSuggestions(
         ),
       trimmedQuery,
     );
-    return orderCommandSuggestionsBySection(
+    return orderCommandSuggestions(
```
SawyerHood, line 198: 🚨 `slopcop/review` — The command list is ordered three times for each query

`filterCommandSuggestions` sorts the list first. This call sorts it again. `PromptBoxInternal` sorts the same result a third time.

The repeated sorts give the same output. They add small CPU and allocation costs to each query. Keep `PromptBoxInternal` as the single order owner.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: Think of a toy box. If you ask for the red car, that car should come first. This change usually does that. One name trick still puts the wrong toy first.

The main path works, but I found one case that still selects the wrong row.

## Findings

1. **Medium — A skill alias can still beat the literal user command.**

   For a `plugin:plan` skill and a `plan` user command, the query `plan` marks both rows as exact. Section rank then puts the skill first. Enter applies the skill.

   Give a full-name match a better rank than a bare skill alias. Add this case to the tests.

2. **Low — The list gets three equivalent sorts for each query.**

   The filter sorts once. The hook sorts again. `PromptBoxInternal` sorts a third time. Benchmarks show a small constant cost on this input path.

   Keep `PromptBoxInternal` as the single order owner. Remove the other two sorts.

3. **Low — The exact row moves its full section.**

   Rows that only match by description can move above name-prefix rows from another section. The exact row remains first, so this effect is visual.

## Other checks

- I found no new security issue. React escapes command text, the fixed section keys prevent prototype attacks, and the change adds no network work.

- The PR removes duplicate prefix logic. A project-wide search found no old function name. The three order passes remain the main refactor target.

- The full app suite passed: 333 files and 2,527 tests. The app type check passed. Lint found no errors.

- I tested the real app in a browser. `/plan` appeared above the Skills section. Enter selected `plan` and inserted its command pill.

PR #1356, opened 2026-08-11
Speed up iOS Safari drawer interactions

SawyerHood:
## Summary

- move mobile sidebar open and close state updates after their compositor transitions
- keep the sidebar trigger active and contain keyboard focus without changing the large background subtree
- use one translate property for Tailwind and drag motion to prevent animation snaps
- window every sidebar list and remove forced layout reads from prompt and timeline hot paths
- defer heavy right-panel content until the bottom-sheet entrance finishes

## Root cause

WebKit recalculated styles for large subtrees when inert, visibility, and drawer state changed inside tap handlers. These recalculations took about 280 ms and blocked the first animation frame. The right-panel sheet also mounted its full content inside the opening tap.

The updated paths start each compositor animation first. React applies the expensive state change after the panel reaches its resting position.

## Measured impact

- sidebar touch taps fell from about 280 ms to 0-2 ms in WebKit
- right-panel open taps fell from 325-370 ms to about 2 ms in WebKit
- Chrome and WebKit both keep the trigger active, close on a second press, and reopen correctly

## Validation

- `pnpm exec turbo run typecheck test build --filter=@bb/app`
- 336 test files passed
- 2,537 tests passed
- production build passed

### DIFF
--- a/apps/app/src/components/secondary-panel/useDrawerPanelRealization.ts
+++ b/apps/app/src/components/secondary-panel/useDrawerPanelRealization.ts
@@ -1,0 +1,56 @@
+import { useCallback, useEffect, useState } from "react";
+
+// vaul keeps a closing drawer mounted for its ~500ms exit transition and
+// tears it down afterwards. The grace must outlast that teardown so the
+// panel never unmounts mid-slide, and so a quick reopen finds the panel
+// still mounted and skips the mount cost entirely.
+const DRAWER_PANEL_TEARDOWN_GRACE_MS = 700;
+// Realization normally happens when the entrance animation ends. The
+// fallback covers environments where that event never fires (reduced
+// motion, jsdom).
+const DRAWER_PANEL_REALIZE_FALLBACK_MS = 700;
+
+/**
+ * Defers the mount of a bottom drawer's heavy panel until the sheet's
+ * entrance animation has finished. vaul mounts drawer children synchronously
+ * inside the opening tap; on an iPhone the secondary panel's mount plus its
+ * style resolution cost ~430 ms of main-thread time, which froze the
+ * slide-in (see the iOS drawer performance recording behind this change).
+ * A light skeleton rides the entrance instead, and the real panel mounts one
+ * settle callback later, while the sheet is at rest.
+ *
+ * The caller invokes `realizePanel` from its drawer settle callback and
+ * swaps `isPanelRealized ? panel : skeleton` in the drawer body.
+ */
+export function useDrawerPanelRealization({
+  isDrawerOpen,
+  rendersAsDrawer,
+}: {
+  isDrawerOpen: boolean;
+  rendersAsDrawer: boolean;
+}): { isPanelRealized: boolean; realizePanel: () => void } {
+  const [isPanelRealized, setIsPanelRealized] = useState(false);
+  const realizePanel = useCallback(() => setIsPanelRealized(true), []);
+
+  useEffect(() => {
+    if (!rendersAsDrawer) {
+      return;
+    }
+    if (isDrawerOpen) {
+      const timeout = window.setTimeout(
+        () => setIsPanelRealized(true),
+        DRAWER_PANEL_REALIZE_FALLBACK_MS,
+      );
+      return () => window.clearTimeout(timeout);
+    }
+    const timeout = window.setTimeout(
+      () => setIsPanelRealized(false),
+      DRAWER_PANEL_TEARDOWN_GRACE_MS,
+    );
+    return () => window.clearTimeout(timeout);
+  }, [isDrawerOpen, rendersAsDrawer]);
+
+  // A wide viewport renders the panel inline; the drawer flag only has
+  // meaning while the drawer branch renders.
+  return { isPanelRealized: rendersAsDrawer && isPanelRealized, realizePanel };
+}
--- a/apps/app/src/components/sidebar/SidebarWindowedItems.tsx
+++ b/apps/app/src/components/sidebar/SidebarWindowedItems.tsx
@@ -15,12 +15,9 @@ import {
   type SidebarWindowedNavigationEntry,
 } from "./sidebarThreadShortcuts";
 
-// Below this count the windowing bookkeeping costs more than the rows it
-// could skip, so short lists render exactly as before.
-const WINDOW_MIN_ITEMS = 8;
 // Items within this distance of the scrollport stay mounted, so scrolling
 // promotes rows before they become visible.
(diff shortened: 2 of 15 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I started the security, code quality, architecture, performance, and browser review.

apps/app/src/components/ui/sidebar.tsx
```diff
@@ -435,6 +441,76 @@ const SidebarProvider = React.forwardRef<
...
+    // flips) then pays its style recalculation — ~280 ms on iOS Safari for a
+    // large sidebar — after the slide instead of blocking its first frame.
+    // The panel stays `inert` and the backdrop passes taps through until the
+    // commit lands one settle window later.
+    const openMobileSidebar = React.useCallback(() => {
+      if (openMobileRef.current || mobileSettleTimeoutRef.current !== null) {
+        return;
+      }
+
+      applySidebarMobileDragStyles({ progress: 1, settling: true });
```
SawyerHood, line 494: 🚨 `slopcop/review` — The drawer passes taps to the page during its open transition

`openMobile` stays false for 220 ms after this call. The panel remains inert, and the closed backdrop ignores pointer input.

Chromium confirmed that an inert overlay passes a tap to the control below it. A rapid second tap can start an unintended action.

Make the backdrop receive pointer input before motion starts. Keep it active through the state commit.

apps/app/src/components/ui/sidebar.tsx
```diff
@@ -831,50 +959,86 @@ const SidebarMobilePanel = React.forwardRef<
         document.activeElement instanceof HTMLElement
           ? document.activeElement
           : null;
-      panelRef.current?.focus({ preventScroll: true });
+      // A touch-opened drawer does not need DOM focus. On iOS, focusing this
+      // newly interactive subtree can synchronously rebuild style and the
+      // accessibility tree for hundreds of milliseconds. Keyboard and
+      // assistive-technology activation leave the trigger focus-visible, so
+      // keep the modal focus move for those paths.
+      const shouldMoveFocus =
```
SawyerHood, line 967: 🚨 `slopcop/review` — The modal can leave focus in a background input

This check depends on `:focus-visible`, which does not cover a swipe or all assistive input.

A swipe can leave a background input focused. Normal keys can then change that input because the trap handles only Tab and Escape.

Move focus to a small guard at open start. Contain later focus changes until the drawer closes.

apps/app/src/components/secondary-panel/useDrawerPanelRealization.ts
```diff
@@ -0,0 +1,56 @@
...
+  rendersAsDrawer: boolean;
+}): { isPanelRealized: boolean; realizePanel: () => void } {
+  const [isPanelRealized, setIsPanelRealized] = useState(false);
+  const realizePanel = useCallback(() => setIsPanelRealized(true), []);
+
+  useEffect(() => {
+    if (!rendersAsDrawer) {
+      return;
+    }
+    if (isDrawerOpen) {
```
SawyerHood, line 39: 🚨 `slopcop/review` — Reduced motion waits for the full fallback

The animation event can be absent when reduced motion is on. This timer then keeps the skeleton for at least 700 ms.

Browser QA saw real content near one second after open. Realize on the next frame for reduced motion.

The drawer shell can also send the settle callback for zero-duration motion.

apps/app/src/components/thread/timeline/useStickyBottomScroll.ts
```diff
@@ -66,6 +62,26 @@ export function useStickyBottomScroll<TElement extends HTMLElement>({
...
+  useEffect(() => {
+    const element = scrollRef.current;
+    if (!element) return;
+    refreshMaxScrollOffset(element);
+    if (typeof ResizeObserver === "undefined") return;
+
+    const observer = new ResizeObserver(() => {
+      refreshMaxScrollOffset(element);
+    });
+    observer.observe(element);
```
SawyerHood, line 82: 🚨 `slopcop/review` — The resize observer misses content-only height changes

This observer watches the fixed scroll port. An image load or disclosure can change `scrollHeight` without changing this box.

The old maximum can restore bottom attachment while the user is above the real bottom.

Observe an inner content wrapper. You can also refresh the cache when user scroll starts.

apps/app/src/components/sidebar/SidebarWindowedItems.tsx
```diff
@@ -83,8 +80,11 @@ export function SidebarWindowedItems({
   renderItem,
 }: SidebarWindowedItemsProps) {
   const scrollElementRef = useSidebarContentElementRef();
+  // A sidebar can contain many short sibling lists. Rendering each short list
+  // in full makes their aggregate offscreen subtree large, so every non-empty
+  // list participates in the shared-scrollport window.
   const windowingEnabled =
```
SawyerHood, line 86: 🚨 `slopcop/review` — Short mobile lists get the wrong placeholder height

Every short list now uses the window path. Each list starts with 30 px, but mobile rows use 36 or 40 px.

Offscreen lists cannot correct this value before display. Their later size change can move the scrollbar and visible rows.

Use a coarse-pointer CSS value or a shared scroll-port measurement.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

The app opens a side box before it does the hard work. It is like opening a toy box before you fill it.

This makes the box move faster. Some taps and scrolls can still go to the wrong place.

### Findings

- High: The open sidebar stays inert for 220 ms. A second tap can reach and start an action behind it.
- High: A swipe can leave focus in a background input. Normal keys can then change hidden content.
- Medium: Reduced motion can show the panel skeleton for about one second.
- Medium: Short mobile lists use a 30 px placeholder for rows that are 36 or 40 px.
- Medium: The sticky-scroll cache can miss content-only height changes. It can then move a user back to the bottom.

I posted one inline comment for each defect.

### Checks

The full app test passed 336 files and 2,537 tests. The app type check and all GitHub checks passed.

Browser checks used Chromium at 390 by 844. The sidebar motion, close, Escape, and both panel paths worked.

This browser cannot run WebKit. The Chromium check does not replace an iOS Safari check.

The browser also found an old Vaul body-style cost. I did not assign that old cost to this pull request.

### Architecture

The root and thread views copy the same panel settle process. A shared hook can keep these paths equal.

Each short list also creates its own observer. A shared sidebar observer can reduce this work.

This review used comment-only actions. It did not approve the pull request or request changes.

PR #1357, opened 2026-08-11
Fix terminal reconnect cleanup and slash command ranking

ymichael:
## Summary

- preserve disconnected terminal rows when a close request cannot reach the daemon, allowing reconnect cleanup to stop the resident PTY
- rank canonical slash-command names ahead of namespaced skill aliases
- keep PromptBoxInternal as the single suggestion-ordering owner and remove two redundant sorting passes

Follow-up to #1353 and #1354.

## Testing

- server terminal route tests: 36 passed
- focused app tests: 99 passed
- Turbo typechecks for @bb/server and @bb/app
- Turbo lint for @bb/server and @bb/app (0 errors; existing warnings remain)

> AGENT GENERATED: by GPT-5

### DIFF
--- a/apps/server/src/services/terminals/terminal-session-lifecycle.ts
+++ b/apps/server/src/services/terminals/terminal-session-lifecycle.ts
@@ -517,7 +517,7 @@ export class TerminalSessionLifecycle {
       TerminalSessionRow
     >({
       onFail: (pending, error) =>
-        this.finalizeFailedTerminalClose(pending, error),
+        this.finalizeTimedOutTerminalClose(pending, error),
       timeoutMs: closeTimeoutMs,
       timeoutError: terminalTimeoutError(
         "terminal_close_timeout",
@@ -973,13 +973,13 @@ export class TerminalSessionLifecycle {
         },
       );
       if (!sent) {
+        this.disconnectDaemonSessionTerminals({
+          daemonSessionId: current.daemonSessionId,
+        });
         this.rejectPendingClose(
           current.id,
           new ApiError(502, "host_disconnected", "Host is not connected"),
         );
-        this.disconnectDaemonSessionTerminals({
-          daemonSessionId: current.daemonSessionId,
-        });
       }
     }
     try {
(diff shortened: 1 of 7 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull request for security, code quality, performance, architecture, and practical end-to-end behavior.

apps/server/src/services/terminals/terminal-session-lifecycle.ts
```diff
@@ -973,13 +973,13 @@ export class TerminalSessionLifecycle {
         },
       );
       if (!sent) {
+        this.disconnectDaemonSessionTerminals({
```
SawyerHood, line 976: 🚨 `slopcop/review` — Medium: the reconnect cleanup record can disappear before reconnect.

This call marks every daemon terminal as `disconnected`. The client then force-closes each unretained disconnected session.

That second request takes `finishTerminalCloseWithoutDaemon` and changes the row to `exited`.

`expireDisconnectedHostTerminals` scans only disconnected rows. Therefore, the new daemon receives no `terminal.close`, and the PTY stays alive.

A probe reproduced this sequence and received zero terminal close messages.

Please preserve cleanup data after the second close, or send a best-effort close through the replacement daemon.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

This PR helps the server remember lost terminals until a new connection can stop them. Exact slash commands also move above skill aliases.

### Review results

**Medium — reconnect cleanup can still miss resident terminals.**

The server marks all terminals from the lost daemon as `disconnected`. The client then force-closes each unretained disconnected terminal.

The second request changes the row to `exited`. Reconnect cleanup scans only disconnected rows, so it sends no `terminal.close` message.

A probe reproduced this sequence. The replacement daemon received zero terminal close messages, and the PTY could stay alive.

This problem already exists on `main`. However, this PR depends on that path and does not fully meet its cleanup goal.

Please preserve cleanup data after the second close. Another option is a best-effort close through the replacement daemon.

**Low — an automatic close can now show an error toast.**

The changed route returns `502 host_disconnected` after a failed send. The automatic close path then shows `Failed to close terminal.`

Please suppress this toast for automatic closes, or return a successful disconnected result.

**Minor cleanup — one rejection call is now redundant.**

`disconnectDaemonSessionTerminals` already rejects the pending close. The following `rejectPendingClose` call cannot find the same entry.

### Other checks

I found no security problem. The command change reduces command shadowing.

I found no performance regression. The command path removes two duplicate sort passes.

`PromptBoxInternal` now owns the command order. I found no other consumer that needs a separate order pass.

### Tests

- The 99 focused app tests passed.
- The 36 server terminal tests passed.
- The Turbo type checks passed for `@bb/app` and `@bb/server`.
- Browser QA confirmed that `/plan` lists `plan` first and Enter selects it.
- GitHub CI is green.

PR #1358, opened 2026-08-11
Link "Submit to gallery" to the plugin intake form

brsbl:
Gives plugin authors a first-class path into BB's marketplace, and gives BB a structured intake instead of ad-hoc requests.

- **Plugin detail page**: a "Submit to marketplace" link on the detail page of locally-developed plugins, pointing at the intake form.
- **CLI intake**: `bb plugin submit` prints the same intake form link, so authors working from the terminal (or agents) discover the path without opening the app.
- **One source of truth**: the form URL lives in `@bb/domain` as `PLUGIN_SUBMISSION_FORM_URL`; both surfaces read it, so it cannot drift.
- The `bb plugin` help/guide/skill surfaces are updated in the same change.

Follow-up commit scopes an old catalog test's "no marketplace" help assertion to command position: the guard is against a `marketplace` command group returning, not against `submit`'s legitimate mention of the marketplace.

BB-Thread-ID: thr_kuixkqt8n2

> AGENT GENERATED: by GPT-5


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am now reviewing this pull request for security, code quality, performance, duplication, architecture, and end-to-end behavior.

apps/app/src/components/tools/PluginDetail.tsx
```diff
@@ -277,6 +280,7 @@ export function PluginDetail({
   const sourceQuery = usePluginSource(plugin?.id ?? "", {
     enabled: plugin !== null && pluginHasUpdateSurfaces(plugin),
   });
+  const openUrl = useOpenUrlByPreference();
```
SawyerHood: 🚨 `slopcop/review` — This hook cannot honor the desktop in-app browser preference on the Tools route.

`ToolsView` has no `UrlOpenRoutingProvider`. The context stays null, so this action always selects the external browser.

This behavior conflicts with the pull request description. Supply an in-app opener for Tools routes.

Add a click test with the preference enabled.

apps/app/src/components/plugin/plugin-submission-form.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * BB's plugin gallery intake form (GitHub URL, name, description, why it's
+ * useful, email). The form is the entire submission UI for now — bb only links
+ * out to it, so there is no schema, route, or field validation on this side to
+ * keep in sync with its questions.
+ *
+ * A module constant rather than a setting: this is BB's own form, not anything
+ * a user configures, so there is no knob to document in the CLI guide.
+ */
+export const PLUGIN_SUBMISSION_FORM_URL =
```
SawyerHood: 🚨 `slopcop/review` — This client-only constant gives agents no submission path.

[The repository rule](https://github.com/get-bb/bb/blob/main/AGENTS.md#cli-guide-and-skill) requires every end-user feature through the SDK and the `bb` CLI.

Add a shared SDK method and a `bb plugin` command for the canonical submission URL. Document and test both surfaces.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: This change adds a door to a form. The door works, but agents have no matching door.

The desktop setting also cannot choose where this door opens.

I found two high-severity issues.

1. `PluginDetail` cannot honor the desktop in-app browser preference from the Tools route. That route has no `UrlOpenRoutingProvider`.

   The context stays null. The hook therefore chooses the external browser for every selection.

   Supply an in-app opener for Tools routes. Add a click test with the preference enabled.

2. The feature has no SDK or `bb` CLI surface. This conflicts with the repository rule for all end-user features.

   Add a shared SDK method and a `bb plugin` command for the canonical submission URL. Document and test both surfaces.

I also found one low-severity gap. The exhaustive control story omits this new action for direct and local plugins.

The architecture scan found no duplicate submission flow. It found one existing duplicate official-provenance check in `PluginDetail`.

The imported `isOfficialProvenance` helper can replace that local check. This existing duplication does not block this pull request.

Security and performance reviews found no issues.

Validation passed:

- The focused app test passed all 31 tests.
- The app typecheck passed.
- The app lint passed with unrelated warnings.
- `git diff --check` passed.
- The form returned HTTP 200.
- The browser test opened the exact form with the title “Submit a bb plugin.”
- The official plugin menu did not show the submission action.

The final GPT-5.6 review gate returned `REQUEST CHANGES`. I posted this as a comment-only review, as required.

apps/app/src/components/tools/PluginDetail.tsx
brsbl: Fixed in 31ec6bf3f. You're right that the provider never mounts on Tools routes — the in-app browser is a thread-panel surface, so the hook was dead indirection that could only ever pick the external browser. Rather than supplying an in-app opener (which would drag the thread-panel browser surface into Tools), the action now calls `openUrlInExternalBrowser` directly — the same behavior as every other Tools-route link (e.g. SkillsBrowse) — and the comment + PR description now say so instead of implying the preference applies. Added the click test with the in-app preference enabled, asserting the external `window.open`.

apps/app/src/components/plugin/plugin-submission-form.ts
brsbl: Addressed in a781953fe, deliberately at half the asked-for scope: the constant moved to `@bb/domain` (the shared package both the app and CLI already depend on), and `bb plugin submit [--json]` prints the canonical URL, documented in the plugins guide chapter and the bb-cli skill in the same change. The existing guide-coverage CLI test now requires the chapter to mention the subcommand.

No SDK method, on purpose: the Google Form is the entire submission UI for this iteration — there is no server route or behavior for an SDK method to expose, so it would return a constant and imply a programmatic submission path that doesn't exist. The SDK half of the parity rule lands with the real in-app form, which is planned as a separate change. The skill entry directs agents to hand the link to the user, since the form asks for details only the author knows (including their email).

brsbl: Fully addressed in 643a0b06a: the public SDK now exposes sdk.plugins.catalog.submission(), with runtime coverage, public-type snapshot coverage, generated bundled declarations, and guide documentation.

PR #1360, opened 2026-08-12
Graduate Extensions from experiments

brsbl:
Graduates Extensions out of the `toolsHub` experiment: every install now gets the Extensions surface, and the experiment-era forks are deleted rather than left to rot.

- Removes the `toolsHub` experiment gate and the gated legacy Settings tabs it kept alive (−3.4k lines, almost all deletions of the pre-Extensions plugin/skill management UI).
- One management surface remains: Extensions. Settings no longer carries a parallel, half-featured copy.
- Bumps the plugin SDK to **0.5.0**, the version boundary plugins target now that the surface is on for everyone.

BB-Thread-ID: thr_kuixkqt8n2

> AGENT GENERATED: by GPT-5


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

apps/app/src/components/settings/PluginSettingsCompatibilityRoute.tsx
```diff
@@ -1,21 +1,34 @@
...
   const location = useLocation();
-  const toolsHubEnabled = useToolsHubExperiment();
-  if (toolsHubEnabled && location.pathname === SETTINGS_PLUGINS_ROUTE_PATH) {
+  const normalizedPathname = location.pathname.replace(/\/+$/u, "");
+  const detailMatch = matchPath(SETTINGS_PLUGIN_ROUTE_PATH, normalizedPathname);
+  const pluginId = detailMatch?.params.pluginId;
+  if (pluginId !== undefined) {
+    return (
+      <Navigate
+        to={`${getPluginDetailRoutePath({ pluginId })}#configuration`}
```
SawyerHood: 🚨 `slopcop/review` — This route breaks the saved Settings destination.

The app saves every `/settings/*` location before this redirect. It therefore saves this old path.

After the user leaves Extensions, the Settings link opens this path and redirects back to Extensions.

I reproduced this behavior in the browser. Please exclude these paths from `useAppSettingsRouteMemory` and add a regression test.

apps/app/src/App.tsx
```diff
@@ -133,6 +132,50 @@ export function LegacyPluginBrowseRedirect() {
...
+    };
+
+    if (scrollToTarget()) return;
+
+    // Lazy routes and plugin slots may mount after the URL changes. Observe the
+    // app until the destination exists instead of dropping the navigation.
+    const observer = new MutationObserver(() => {
+      if (scrollToTarget()) observer.disconnect();
+    });
+    observer.observe(document.body, { childList: true, subtree: true });
```
SawyerHood, line 184: 🚨 `slopcop/review` — This observer can remain active for the full page visit.

An unknown hash never creates a matching element. The observer then checks every document mutation until the next navigation.

A live thread can cause many such checks through timeline updates. Please limit the wait and disconnect the observer after route setup.

apps/app/src/components/tools/PluginCapabilities.tsx
```diff
@@ -277,7 +348,24 @@ function pluginAppSurfaceItems(
 
 export function PluginIncludes({ plugin }: { plugin: PluginListItem }) {
   const slots = usePluginSlots();
-  const appItems = pluginAppSurfaceItems(plugin.id, slots);
+  const hasSkillCapabilities = plugin.capabilities.some(
+    (capability) => capability.kind === "skill",
+  );
+  const skillsQuery = useProjectSkills(PERSONAL_PROJECT_ID, {
```
SawyerHood: 🚨 `slopcop/review` — This optional link starts a full skill search.

The query always runs again when the component mounts. The server then asks each skill provider and the shared skill source.

I confirmed that the Automations detail page sends this request. Please use a small direct lookup or reuse cached data here.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

This change makes Extensions available to all users. It moves plugin controls from Settings into Extensions and keeps old links active.

### Findings

I found three issues.

1. A legacy plugin link can prevent later access to Settings. The app saves the old path, so the Settings link returns to Extensions.
2. An unknown hash leaves a document-wide observer active. Live thread updates can then cause repeated element searches.
3. A plugin with a skill starts a full skill search for one optional link. This search runs again after each component mount.

I added an inline comment for each issue.

### Security and architecture

I found no security issue. The removed experiment controlled the user interface and did not control server access.

The change cleanly removes the duplicate plugin manager from Settings. I found no stale `toolsHub` reference outside historical migration tests.

One small refactor remains. Three changed files build the same plugin configuration URL with `#configuration`. A shared route helper can prevent drift.

### Checks

- The affected Turbo type checks passed for the app, server, domain, and plugin SDK.
- The affected Turbo tests passed. The app completed 332 test files and 2,502 tests.
- `git diff --check` passed.
- The Extensions plugin list and plugin detail pages worked in the browser.
- The old plugin settings route opened and focused the new Configuration section.
- The browser reproduced the saved Settings destination defect.

apps/app/src/components/settings/PluginSettingsCompatibilityRoute.tsx
brsbl: Fixed in 2a29cb83f. The legacy collection and detail routes are now transient to app route memory, so they preserve both the last real Settings destination and Back to app. Added coverage for warm and cold entry paths.

apps/app/src/App.tsx
brsbl: Fixed in 2a29cb83f. Hash target observation now stops on success, after a 2-second deadline, on route change, or on unmount. The regression test proves later mutations do not trigger target lookups after the deadline.

apps/app/src/components/tools/PluginCapabilities.tsx
brsbl: Fixed in 2a29cb83f. Plugin detail now reads the existing project-skills cache without initiating discovery; when uncached, the capability links to the Skills library. Tests cover both cached exact links and the no-request fallback.

PR #1361, opened 2026-08-12
Fix Codex and Claude subagent projection gaps

brsbl:
## Outcome

Fixes two confirmed subagent projection bugs. The first four commits stay as the failing repros; this change adds the production fixes.

| Provider | Failure | Fix |
| --- | --- | --- |
| Codex | A resumed completed child turn lost `parentToolCallId`, so its completion could clear the root loading state. | Re-arm the original delegation association when `subAgentActivity(kind: "interacted")` arrives for a known terminal agent. |
| Claude | An idle root with a still-running nested agent showed no `Background work running` indicator. | Include `thread.activeBackgroundAgentCount > 0` in the timeline background-only indicator. Nested-row filtering stays unchanged. |

## Codex

Terminal child handling calls `clearTrackedSubAgentLinks`. A later `followup_task` arrived as `interacted` and did not restore the parent, so the resumed `turn/started` had no `parentToolCallId`. `RuntimeTurnState` then treated that child as the root turn and went idle when the child finished.

The adapter now:

- restores the agent-thread mapping and FIFO pending link for that same spawn
- keeps the completed delegation row (no second `spawnAgent` row)
- removes a FIFO slot when a later turn matches the explicit agent-thread mapping, so two concurrently resumed agents cannot cross-link
- counts queued `interacted` follow-ups and re-arms after each terminal child completion while any remain

Confirmed trace: `thr_mcapedchhh`.

## Claude

A nested `Agent → Agent → backgroundTask local_agent` can outlive the root turn. The server already counts it in `activeBackgroundAgentCount` (sidebar was correct). `selectActiveBackgroundCommandMessages` excludes that nested row from the root timeline on purpose. The detail view now uses the server count for the background-only indicator.

Confirmed traces: `thr_4bivjj7tqz`, `thr_zm2anhgygp`, `thr_hxxaw2aabn`.

## Protocol

`HOST_DAEMON_PROTOCOL_VERSION` is 111 so enrolled daemons pick up the Codex adapter change. Version 110 on main already covered complete Codex command output.

## Verification

```sh
pnpm exec turbo run test --filter=@bb/agent-runtime --force -- --run src/codex/adapter.test.ts src/claude-code/task-translation.test.ts src/runtime-turn-state.test.ts
pnpm exec turbo run test --filter=@bb/app --force -- --run src/components/thread/timeline/ThreadTimelinePanelContent.test.tsx
pnpm exec turbo run test --filter=@bb/thread-view --force -- --run test/build-thread-timeline.test.ts
pnpm exec turbo run typecheck --filter=@bb/agent-runtime --filter=@bb/app --filter=@bb/thread-view --filter=@bb/host-daemon-contract
```

- Agent runtime adapter and turn-state suites pass after rebase, including queued follow-up resumes.
- App: nested-agent indicator and count-returns-to-zero companion.
- Thread view: `keeps nested Claude agents out of root active-work rows`.

### DIFF
--- a/apps/app/src/components/thread/timeline/ThreadTimelinePanelContent.tsx
+++ b/apps/app/src/components/thread/timeline/ThreadTimelinePanelContent.tsx
@@ -85,7 +85,8 @@ export function ThreadTimelinePanelContent({
     displayStatus === "provisioning" || displayStatus === "starting";
   const hasActiveBackgroundWork =
     resolvedTimeline.activeWorkflows.length > 0 ||
-    resolvedTimeline.activeBackgroundCommands.length > 0;
+    resolvedTimeline.activeBackgroundCommands.length > 0 ||
+    (threadQuery.data?.activeBackgroundAgentCount ?? 0) > 0;
   const backgroundOnlyIndicatorLabel =
     displayStatus === "idle" && hasActiveBackgroundWork
       ? "Background work running"
--- a/packages/agent-runtime/src/codex/adapter.ts
+++ b/packages/agent-runtime/src/codex/adapter.ts
@@ -1058,6 +1058,8 @@ export function createCodexProviderAdapter(
     CodexPendingDelegationTurnLink[]
   >();
   const pendingDelegationCallIds = new Set<string>();
+  const pendingDelegationProviderThreadIdByCallId = new Map<string, string>();
+  const processedSubAgentInteractionIds = new Set<string>();
   const trackedSubAgentsByCallId = new Map<string, CodexTrackedSubAgent>();
   const trackedSubAgentCallIdsByAgentThreadId = new Map<string, string>();
 
@@ -1201,6 +1203,12 @@ export function createCodexProviderAdapter(
         continue;
       }
       clearTrackedSubAgentLinks(tracked);
+      if (
+        trackedSubAgentCallIdsByAgentThreadId.get(tracked.agentThreadId) ===
+        tracked.callId
+      ) {
+        trackedSubAgentCallIdsByAgentThreadId.delete(tracked.agentThreadId);
+      }
       trackedSubAgentsByCallId.delete(callId);
     }
   }
@@ -1353,36 +1361,47 @@ export function createCodexProviderAdapter(
       pendingLinks,
     );
     pendingDelegationCallIds.add(args.callId);
+    pendingDelegationProviderThreadIdByCallId.set(
+      args.callId,
+      args.providerThreadId,
+    );
   }
 
   function removePendingDelegationCall(callId: string): void {
     pendingDelegationCallIds.delete(callId);
-    for (const [
-      providerThreadId,
-      pendingLinks,
-    ] of pendingDelegationTurnLinksByProviderThreadId) {
-      const remainingLinks = pendingLinks.filter(
-        (pendingLink) => pendingLink.callId !== callId,
-      );
-      if (remainingLinks.length === 0) {
-        pendingDelegationTurnLinksByProviderThreadId.delete(providerThreadId);
-      } else if (remainingLinks.length !== pendingLinks.length) {
-        pendingDelegationTurnLinksByProviderThreadId.set(
-          providerThreadId,
-          remainingLinks,
-        );
-      }
+    const providerThreadId =
+      pendingDelegationProviderThreadIdByCallId.get(callId);
+    pendingDelegationProviderThreadIdByCallId.delete(callId);
+    if (!providerThreadId) {
+      return;
     }
+    const pendingLinks =
+      pendingDelegationTurnLinksByProviderThreadId.get(providerThreadId);
+    if (!pendingLinks) {
+      return;
+    }
+    const remainingLinks = pendingLinks.filter(
+      (pendingLink) => pendingLink.callId !== callId,
+    );
+    if (remainingLinks.length === 0) {
+      pendingDelegationTurnLinksByProviderThreadId.delete(providerThreadId);
+    } else if (remainingLinks.length !== pendingLinks.length) {
+      pendingDelegationTurnLinksByProviderThreadId.set(
+        providerThreadId,
+        remainingLinks,
+      );
+    }
+  }
+
+  function hasPendingNativeTurnStart(providerThreadId: string): boolean {
+    return (
+      (nativeTurnStartClientRequestIdsByProviderThreadId.get(providerThreadId)
+        ?.length ?? 0) > 0
+    );
   }
 
   function clearTrackedSubAgentLinks(tracked: CodexTrackedSubAgent): void {
     removePendingDelegationCall(tracked.callId);
-    if (
-      trackedSubAgentCallIdsByAgentThreadId.get(tracked.agentThreadId) ===
-      tracked.callId
-    ) {
-      trackedSubAgentCallIdsByAgentThreadId.delete(tracked.agentThreadId);
-    }
     if (
       delegationParentToolCallIdsByProviderThreadId.get(
         tracked.agentThreadId,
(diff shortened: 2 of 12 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the Slop Cop. I am reviewing this pull request now.

I will check security, code quality, architecture, duplication, performance, tests, and the applicable end-to-end path.

packages/agent-runtime/src/codex/adapter.ts
```diff
@@ -1524,15 +1530,55 @@ export function createCodexProviderAdapter(
...
+      tracked.agentThreadId,
+      tracked.callId,
+    );
+    if (tracked.agentThreadId !== tracked.parentProviderThreadId) {
+      delegationParentToolCallIdsByProviderThreadId.set(
+        tracked.agentThreadId,
+        tracked.callId,
+      );
+    }
+    enqueuePendingDelegationTurnLink({
```
SawyerHood, line 1578: 🚨 `slopcop/review` — The fallback can attach a child link to a user turn

Rearm adds an unclear link to the root provider queue. The next `turn/started` uses it without a client-request check.

I called `prepareTurnStart` after `interacted`. The accepted user turn received `parentToolCallId: "spawn-1"`.

`RuntimeTurnState` then treats the user turn as a child. Please exclude accepted root turns before FIFO assignment and add this test case.

packages/agent-runtime/src/codex/adapter.ts
```diff
@@ -1582,10 +1629,20 @@ export function createCodexProviderAdapter(
...
         // Messaging an existing agent is activity within the original
-        // delegation, not a new timeline row.
+        // delegation, not a new timeline row. A completed agent can receive
+        // followup_task; re-arm the original parent so the next child turn
+        // is not projected as a root turn.
+        const tracked = findTrackedSubAgentByAgentThreadId(
+          activity.item.agentThreadId,
+        );
+        if (tracked?.terminal) {
+          tracked.pendingFollowups += 1;
```
SawyerHood, line 1664: 🚨 `slopcop/review` — A repeated interaction leaves a stale parent link

This counter does not reject a repeated item ID. Two copies leave one pending follow-up after one resumed turn completes.

I reproduced this case. The next accepted user turn received the old `parentToolCallId`.

Please record processed interaction IDs and clear them with the provider thread. Add a test that sends one interaction twice.

packages/agent-runtime/src/codex/adapter.ts
```diff
@@ -1524,15 +1530,55 @@ export function createCodexProviderAdapter(
...
   }
 
+  function findTrackedSubAgentByAgentThreadId(
+    agentThreadId: string,
+  ): CodexTrackedSubAgent | undefined {
+    const callId = trackedSubAgentCallIdsByAgentThreadId.get(agentThreadId);
+    if (callId) {
+      return trackedSubAgentsByCallId.get(callId);
+    }
+    for (const tracked of trackedSubAgentsByCallId.values()) {
```
SawyerHood: 🚨 `slopcop/review` — Historical agent lookup becomes quadratic

Terminal cleanup removes the reverse index. Each later interaction then scans every historical agent record.

N agents with one resume each cause O(N²) event work. The performance review measured 3.2 seconds for 8,000 resumes.

Please keep a direct all-state index by `agentThreadId`. Remove its records when the thread closes.

apps/app/src/components/thread/timeline/ThreadTimelinePanelContent.tsx
```diff
@@ -85,7 +85,8 @@ export function ThreadTimelinePanelContent({
     displayStatus === "provisioning" || displayStatus === "starting";
   const hasActiveBackgroundWork =
     resolvedTimeline.activeWorkflows.length > 0 ||
-    resolvedTimeline.activeBackgroundCommands.length > 0;
+    resolvedTimeline.activeBackgroundCommands.length > 0 ||
+    (threadQuery.data?.activeBackgroundAgentCount ?? 0) > 0;
```
SawyerHood, line 89: 🚨 `slopcop/review` — Background events do not refresh this value

The `events-appended` rule refreshes timelines and thread lists. It does not refresh the thread-detail query used here.

This count can remain zero after a start. It can remain positive after completion.

Please refresh thread detail when `backgroundActivityChanged` is true. You can also use a cache that these events refresh.

Add cache tests for the task start and completion.

packages/agent-runtime/src/codex/adapter.ts
```diff
@@ -1450,14 +1450,20 @@ export function createCodexProviderAdapter(
...
-          : undefined) ??
-        consumePendingDelegationTurnLink({
+      const mappedFromProviderThread = providerThreadId
+        ? delegationParentToolCallIdsByProviderThreadId.get(providerThreadId)
+        : undefined;
+      if (mappedFromProviderThread) {
+        parentToolCallId = mappedFromProviderThread;
+        // A turn that matches an explicit agent-thread mapping must not also
+        // consume a different agent's FIFO slot on the multiplexed root.
+        removePendingDelegationCall(mappedFromProviderThread);
```
SawyerHood, line 1479: 🚨 `slopcop/review` — Pending-link removal becomes quadratic

Each explicit mapping calls `removePendingDelegationCall`. That helper scans every queue and filters every array.

N mapped resume starts cause O(N²) scans and allocations. The performance review measured 2.37 seconds for 8,000 starts.

Please add a direct call-ID index. A queue node or an inactive generation can preserve FIFO order.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5 summary

This PR keeps resumed helper-agent work under its original timeline item. It also shows when a hidden nested helper still works.

### Review result

I found five issues. The first two can break the intended behavior during normal use.

1. **High:** A queued user turn can consume a resumed-child FIFO link.
   I reproduced this with `prepareTurnStart`. The accepted user turn received the child's `parentToolCallId`.

2. **Medium:** Background events do not refresh the thread-detail query used by the new indicator.
   The indicator can remain absent after a start. It can remain visible after completion.

3. **Medium:** A repeated `interacted` item increments `pendingFollowups` twice.
   I reproduced a stale parent link on the next accepted user turn.

4. **Medium:** Historical agent lookup can cause O(N²) event work.
   The review measured 3.2 seconds for 8,000 resumes.

5. **Medium:** Pending-link removal can cause O(N²) scans and allocations.
   The review measured 2.37 seconds for 8,000 mapped starts.

### Security and architecture

I found no exploitable security-boundary fault. The provider adapter keeps this state within one BB thread process.

I found no direct duplicate function that needs immediate extraction. However, several maps and queues represent the same parent relationship.

A small adapter-owned correlation state object can give each identifier one direct index. It can also centralize cleanup and idempotence.

The server and daemon ownership stays correct. Protocol version 111 correctly covers the changed provider output.

### Verification

- All current GitHub checks pass.
- `git diff --check` passes.
- Local focused tests pass: 189 agent-runtime tests, 4 app tests, 70 thread-view tests, and 35 contract tests.
- The development app opened without page errors or console errors.
- The development data had no nested-agent test thread. Therefore, I could not verify the exact live indicator state in the browser.

Please resolve the inline findings before merge.

SawyerHood:
Addressed the SlopCop findings on the current head:

1. Native user turns no longer consume a resumed-child FIFO parent. A later child turn still gets the original parent.
2. `events-appended` now refreshes thread detail when `backgroundActivityChanged` is true.
3. Duplicate `interacted` item ids increment the follow-up count once.
4. The agent-thread index stays after a child settles; thread close is the cleanup.
5. Pending FIFO removal looks up the provider thread by call id.

Added adapter and cache tests for those cases.

PR #1362, opened 2026-08-12
Add plugin nav sidebar accessories

### LINKED ISSUE #1202: Plugin API: allow a trailing accessory in nav panel sidebar rows
Plugin nav panel rows currently support a title and icon, but no content on the right.

It would be useful to allow a small trailing accessory for live values, such as an open task count:

`Tasks                  12`

Something like an optional `sidebarAccessory` component on `navPanel` could work. The host could limit its size and hide it in compact mode. Existing plugins would be unaffected.

This would avoid plugins needing to modify bb's DOM.

ymichael:
## Summary

- add `experimental_sidebarAccessory` to plugin nav panel registrations, the app/testing registration collectors, bundled SDK types, and authoring docs
- mount accessories in an isolated plugin boundary, omit them on compact viewports, and constrain wide-screen content to a clipped single-line 4rem by 1.25rem box with host-owned 12px accessory typography
- share the trailing action column: show the live accessory at rest, then fade it for the host options button on hover, keyboard focus, or while its menu is open without remounting plugin state
- make Tasks the first consumer with a realtime non-terminal task count backed by a targeted SQLite count query and a count-only, burst-coalesced RPC refresh path
- silently contain accessory crashes and remount their boundary on plugin generation changes so a fixed reload gets a fresh attempt
- cover compact non-mounting, overflow bounds, accessory-less compatibility, registration validation, live Tasks refreshes, burst coalescing, crash recovery, and accessory lifecycle through the options swap

## Design

A no-props component keeps the surface small while letting live accessories own their RPC query and realtime subscription. Updates stay within the accessory subtree instead of lifting state through the sidebar. The accessory is presentational and sits outside the navigation button, preserving the stable accessible name. It shares the trailing action column with the host options button through the existing inverse hover-action CSS, so hover/focus does not remount or reset its live state. The host applies its 12px secondary chrome type scale so accessory values remain subordinate to 13px row labels. Plugin-generation keys reset a latched error boundary after reload, while a non-null empty crash fallback keeps failure local and silent. Tasks single-flights its count-only RPC and reduces any invalidation burst during a request to one trailing refresh. The audit checklist records the component-versus-value, portal, budget, compact lifecycle, and accessibility questions to revisit before stabilization.

No host-daemon protocol bump is needed because this changes the frontend plugin contract and plugin-local RPC response only, not the server/daemon wire contract.

## Testing

- `pnpm exec turbo run test --filter=@bb/app --force` (336 files, 2549 tests)
- `pnpm exec turbo run test --filter=bb-plugin-tasks --force` (35 files, 329 tests)
- `pnpm exec turbo run test --filter=@bb/plugin-sdk --force` (9 files, 75 tests)
- `pnpm exec turbo run test --filter=@bb/server --force -- test/services/plugins/plugin-authoring-docs.test.ts`
- `pnpm exec turbo run test --filter=@bb/templates --force` (4 files, 23 tests)
- `pnpm exec turbo run typecheck --filter=@bb/plugin-sdk --filter=@bb/app --filter=bb-plugin-tasks --filter=@bb/server --filter=@bb/templates --force`
- live dev-server QA for idle, hover, menu-open, realtime count updates, compact mode, final 12px accessory scale, and the dedicated count/summary RPC split
- generated SDK/template checks, Prettier, and `git diff --check`

Fixes #1202

> AGENT GENERATED: by GPT-5.6


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am reviewing this pull request for security, code quality, performance, duplication, architecture, and end-to-end behavior.

apps/app/src/components/plugin/PluginNavSidebarItems.tsx
```diff
@@ -447,6 +450,26 @@ function PluginNavSidebarItem({
     label: panel.title,
   });
   const splitIndicator = usePaneContentSplitIndicator(content, splitEnabled);
+  const SidebarAccessory = panel.experimental_sidebarAccessory;
+  const sidebarAccessory =
+    !isCompactViewport && SidebarAccessory !== undefined ? (
+      <PluginSlotMount
```
SawyerHood, line 457: 🚨 `slopcop/review` — Add the plugin generation to this mount key.

A plugin reload clears the global crash set. The mounted boundary still keeps `state.crashed` because this stable row reuses it. Other slot mounts include `generation` in their React keys. Add a key with `panel.generation`. Add a test that crashes generation one and renders generation two.

plugins/tasks/shell/sidebar-accessory.tsx
```diff
@@ -0,0 +1,37 @@
...
+/** Live count of non-terminal tasks for the host sidebar row. */
+export function TasksSidebarAccessory() {
+  const rpc = useTasksRpc();
+  const connectionState = useRealtimeConnectionState();
+  const requestSequence = useRef(0);
+  const [count, setCount] = useState<number | null>(null);
+
+  const refresh = useCallback(() => {
+    const sequence = ++requestSequence.current;
+    void rpc.call("sidebarSummary").then(
```
SawyerHood: 🚨 `slopcop/review` — Limit refresh work during event bursts.

Each change signal starts another `sidebarSummary` request. The sequence check rejects old results, but it does not limit active requests. Each summary request runs the project aggregation and the new full count scan. Add a count-only RPC. Permit one active request and one later refresh. Test several signals while the first request stays open.

apps/app/src/components/plugin/PluginNavSidebarItems.tsx
```diff
@@ -447,6 +450,26 @@ function PluginNavSidebarItem({
...
+        pluginId={panel.pluginId}
+        slotKind="navPanelSidebarAccessory"
+        slotId={panel.id}
+        crashFallback={null}
```
SawyerHood: 🚨 `slopcop/review` — Use a non-null empty crash fallback here.

`PluginSlotBoundary` uses `fallback ?? <CrashedPluginChip>`. Therefore, this null value selects the default crash chip. An accessory crash then shows clipped crash text instead of hiding the accessory. Add a test that verifies the chip does not appear.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: Plugins can now show a small live label beside a sidebar item. Tasks uses this label to show the open-task count.

I found three concrete issues:

- Medium: A crashed accessory stays crashed after a plugin reload. The mount needs a React key that includes `panel.generation`.

- Medium: Each Tasks change signal starts another broad summary request. Add a count-only RPC and coalesce signals during an active request.

- Low: `crashFallback={null}` selects the default crash chip. Use a non-null empty fallback so a crash hides the accessory.

The authoring skill also says a `headerContent` crash hides the accessory. Change that phrase to “header content.”

I found no new data exposure, authorization error, or unsafe navigation path. The repeated host and test-harness validation follows the current SDK contract design.

Validation passed on the exact head SHA. GitHub CI passed all required checks. Local Turbo type checks and package tests passed.

The browser test passed on wide and compact viewports. It also passed route navigation and a live count update from 9 to 10. I stopped the dev app after the test.

Please fix the three inline findings before merge.

apps/app/src/components/plugin/PluginNavSidebarItems.tsx
ymichael: Addressed in 69b3e5a92. The accessory PluginSlotMount key now includes panel.generation, and the host regression test crashes generation one, resets the plugin crash state, registers generation two, and verifies the healthy accessory renders.

apps/app/src/components/plugin/PluginNavSidebarItems.tsx
ymichael: Addressed in 69b3e5a92. The mount now passes a non-null empty fragment as crashFallback. The crash/reload regression also verifies that no plugin crash chip appears.

plugins/tasks/shell/sidebar-accessory.tsx
ymichael: Addressed in 69b3e5a92. Tasks now uses a dedicated sidebarOpenTaskCount RPC. The accessory runs one request at a time and collapses any burst during that request into one trailing refresh; the new deferred-RPC test sends three signals and pins the call count at two.

PR #1365, opened 2026-08-12
Add plugin panel actions to the New thread screen

### LINKED ISSUE #1085: Plugin threadPanelAction registrations missing from root New thread Actions list
## Summary

A plugin registered with `app.slots.threadPanelAction` appears in the right-panel **Actions** list for an existing thread, but is omitted from the same list on the root **New thread** screen.

## Reproduction

This uses only the standard plugin scaffold.

1. Create a plugin:

```sh
bb plugin new panel-action-repro --app
cd bb-plugin-panel-action-repro
```

2. Replace `app.tsx` with:

```tsx
import type { PluginThreadPanelProps } from "@bb/plugin-sdk";
import { definePluginApp } from "@bb/plugin-sdk/app";

function ReproPanel({ threadId }: PluginThreadPanelProps) {
  return <div className="p-4">Opened for thread: {threadId}</div>;
}

export default definePluginApp((app) => {
  app.slots.threadPanelAction({
    id: "repro-action",
    title: "Repro action",
    component: ReproPanel,
  });
});
```

3. Build and install it:

```sh
npm install
bb plugin build .
bb plugin install . --yes
bb plugin reload panel-action-repro
```

4. In BB:

   1. Open any existing thread.
   2. Open the right panel and select **Actions**.
   3. Confirm **Repro action** is present.
   4. Navigate to **New thread**.
   5. Open the right-panel **Actions** list.

## Actual behavior

**Repro action** appears for the existing thread, but is missing from the root **New thread** Actions list, which only shows the built-in **Start terminal** action.

## Expected behavior

A `threadPanelAction` registration should either:

- appear in the root New thread Actions list with a defined thread-less context; or
- be explicitly documented as thread-only, with a separate SDK slot for actions on the root New thread screen.

## Environment

- BB 0.35.1
- macOS
- Standard scaffolded plugin
- No private or third-party plugin required

I verified this live in BB after building, installing, and reloading the scaffolded plugin. The `bb plugin dev` watcher loop hit a local `EMFILE` file-watcher limit, so the behavior check used the built/install/reload path above.

ymichael:
## Summary

- keep `threadPanelAction` thread-only with its required `threadId` contract
- add `app.slots.experimental_newThreadPanelAction` for the root New thread panel, with explicit `projectId: string | null` context
- wire the new slot through registration collection, the slot store, root panel tabs, capability reporting, SDK declarations, scaffolds, and authoring docs
- pin that root actions appear beside Start terminal while existing thread actions remain scoped to threads

## Design

Widening the established thread action contract would make existing plugins receive an absent thread ID at runtime. The in-repo consumers are all thread-owned surfaces, including Side Chat, GitHub PRs, Tasks, Docs, and Workflows. A separate opt-in slot avoids that migration hazard and gives New thread actions a context that matches the surface.

## Testing

- `pnpm exec turbo run typecheck --filter=@bb/plugin-sdk --filter=@bb/app`
- `pnpm exec turbo run typecheck --filter=@bb/server --filter=@bb/templates`
- focused Turbo tests for app slot rendering/collection, the plugin SDK app harness, plugin authoring docs, and generated templates

Fixes #1085

> AGENT GENERATED: by GPT-5


SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

apps/app/src/components/plugin/PluginPanelActions.tsx
```diff
@@ -84,6 +85,45 @@ function runPluginPanelAction({
   }
 }
 
+interface RunPluginNewThreadPanelActionArgs {
+  action: PluginNewThreadPanelActionSlot;
+  openPluginPanel: OpenPluginPanelHandler;
+  projectId: string | null;
+}
+
+function runPluginNewThreadPanelAction({
```
SawyerHood, line 94: 🚨 `slopcop/review` — Use one shared panel-action runner and renderer.

This runner repeats the thread runner's serialization, error, promise, and open behavior. The new tab component also repeats the lookup and layout path.

A later fix can change only one surface. Extract shared helpers with a typed surface context.

apps/app/src/views/RootComposeView.tsx
```diff
@@ -3036,7 +3050,10 @@ export function RootComposeView() {
     ) : activePluginPanelTab ? (
       <PluginPanelTabContent
         tab={activePluginPanelTab}
-        threadId={rootPanelThreadId}
+        context={{
+          kind: "new-thread",
+          projectId: isProjectless ? null : projectId,
```
SawyerHood, line 3055: 🚨 `slopcop/review` — Keep the project context stable for an open plugin tab.

The root panel uses one persistent state key across project changes. This line gives an open tab the current project ID.

The action received the prior project ID when the user opened the tab. A project change can combine old parameters with the new project.

Store the project ID in the tab. You can instead close these plugin tabs when the project changes.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: This pull request lets a plugin add an action beside the New thread screen. The action opens a plugin panel without a thread.

I found one medium issue and one low design issue.

- Medium: An open panel can change from project A to project B. Its original action data can still belong to project A.
- Low: The new action runner and panel renderer repeat the thread action paths. Shared helpers will prevent later behavior differences.

I found no concrete security issue. Plugin identity, JSON data, error containment, and surface separation remain safe.

I found no material performance issue. The root view adds small repeated subscriptions and linear action scans.

The scan cost was 0.47 milliseconds for 1,000 actions and 100 tabs. It was 4.09 milliseconds for 10,000 actions.

Validation passed:

- The three focused app test files passed 83 tests.
- The plugin SDK harness passed 30 tests.
- Turbo type checks passed for the app, server, plugin SDK, and templates.
- GitHub CI reports all required checks as successful.
- Chromium showed the new action, opened its tab, and supplied the correct projectless context.

I recommend a project-context fix before merge. The shared-helper refactor can follow in this pull request or a small follow-up.

PR #1367, opened 2026-08-12
Fix Claude Code update handling

### LINKED ISSUE #1331: Claude Code does not update
Image

Claude Code does not update

ymichael:
## Summary

- respect Claude Code’s configured stable/latest release channel when checking versions
- only offer managed updates for native or matching npm-global installs, while surfacing external installs as manual updates
- verify that a successful update command actually advances the active executable and report actionable failures otherwise
- bump the host-daemon protocol for the changed provider status semantics

## Test plan

- `pnpm exec turbo run test --filter=@bb/host-daemon --filter=@bb/app --filter=@bb/cli --force`
- `pnpm exec turbo run test --filter=@bb/host-daemon-contract --force`
- `pnpm exec turbo run typecheck --filter=@bb/host-daemon --filter=@bb/app --filter=@bb/cli --filter=@bb/host-daemon-contract`

Fixes #1331

> AGENT GENERATED: by GPT-5.6

### DIFF
--- a/apps/host-daemon/src/provider-cli-health.ts
+++ b/apps/host-daemon/src/provider-cli-health.ts
@@ -18,6 +18,7 @@ import type { HostDaemonLogger } from "./logger.js";
 import { ensureNodePtySpawnHelperExecutable } from "./terminals/terminal-manager.js";
 
 const COMMAND_CHECK_TIMEOUT_MS = 5_000;
+const CLAUDE_DOCTOR_TIMEOUT_MS = 10_000;
 const NPM_VIEW_TIMEOUT_MS = 15_000;
 const NPM_INSTALL_STATE_TIMEOUT_MS = 5_000;
 const CLAUDE_CODE_INSTALL_SCRIPT_URL = "https://claude.ai/install.sh";
@@ -43,6 +44,24 @@ const npmGlobalListResponseSchema = z
       .default({}),
   })
   .passthrough();
+
+const npmDistTagsSchema = z
+  .object({
+    latest: z.string().min(1),
+    stable: z.string().min(1).optional(),
+  })
+  .passthrough();
+
+type ClaudeCodeInstallMethod =
+  | "native"
+  | "npm-global"
+  | "package-manager"
+  | "unknown";
+
+interface ClaudeCodeDoctorStatus {
+  installMethod: ClaudeCodeInstallMethod;
+  updateChannel: "latest" | "stable";
+}
 
 export interface ProviderCliDefinition {
   key: ProviderCliKey;
@@ -179,9 +198,11 @@ interface ResolveProviderCliInstallSourceArgs {
 interface BuildInstallActionArgs {
   definition: ProviderCliDefinition;
   installed: boolean;
+  installSource: ProviderCliInstallSource;
   needsUpdate: boolean;
   versionUnsupported: boolean;
   nodePlatform: NodeJS.Platform;
+  claudeCodeDoctorStatus: ClaudeCodeDoctorStatus | null;
 }
 
 interface CreateCommandResultArgs {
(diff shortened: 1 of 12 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the Slop Cop. I am now reviewing this pull request against the stated security, quality, and performance rules.

I will also check the design, duplicate code, tests, and the available user workflow.

apps/host-daemon/src/provider-cli-health.ts
```diff
@@ -534,7 +609,13 @@ function buildInstallAction({
       command: command.displayCommand,
     };
   }
-  if (needsUpdate) {
+  const canRunUpdate =
+    definition.key !== "claudeCode" ||
+    claudeCodeDoctorStatus?.installMethod === "native" ||
```
SawyerHood, line 614: 🚨 `slopcop/review` — Older native Claude Code releases lose the managed update path when `claude doctor` requires a TTY.

I reproduced this with version 2.1.69. The daemon pipe causes an Ink raw-mode error, so the doctor status becomes null.

A native install also appears as `external`. Therefore, it fails this check and loses the prior `claude update` action.

Please keep a safe native fallback or set a supported minimum version. Add a non-TTY doctor-failure test.

apps/host-daemon/src/provider-cli-health.ts
```diff
@@ -717,6 +802,13 @@ export async function inspectProviderCli({
           args: ["list", "-g", npmPackageName, "--depth=0", "--json"],
           timeoutMs: NPM_INSTALL_STATE_TIMEOUT_MS,
         }),
+    definition.key === "claudeCode"
```
SawyerHood, line 805: 🚨 `slopcop/review` — This unconditional doctor probe also runs through the onboarding status poll every 15 seconds.

One current doctor call used 1.83 seconds of real time and 0.78 seconds of CPU time on this host.

An open onboarding page starts four calls each minute. Please cache this result or use an installed-only onboarding probe.

apps/host-daemon/src/provider-cli-health.ts
```diff
@@ -726,10 +818,21 @@ export async function inspectProviderCli({
...
+          `${claudeDoctorResult.stdout}\n${claudeDoctorResult.stderr}`,
+        )
       : null;
+  const latestVersion =
+    latestResult === null || !isSuccessfulCommand(latestResult)
+      ? null
+      : definition.key === "claudeCode"
+        ? parseNpmDistTagVersion(
+            `${latestResult.stdout}\n${latestResult.stderr}`,
+            claudeCodeDoctorStatus?.updateChannel ?? "latest",
```
SawyerHood, line 833: 🚨 `slopcop/review` — A doctor failure changes an unknown release channel to `latest`.

An older npm-global client can still receive the managed action. However, `claude update` can follow its configured stable channel.

The verifier then expects the latest release and reports a false failure. It can offer the same update again.

Please preserve an unknown channel or avoid the managed action. Add a stable-channel test where the doctor command fails.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

BB now checks which Claude Code track and installer a machine uses. It is like choosing the correct tool for one toy.

The main idea is good. However, some older tools cannot answer the new check, so BB can choose the wrong update action.

### Findings

1. **High:** Older native Claude Code releases can lose the managed update action.

   I reproduced the failure with version 2.1.69. Its non-TTY doctor command reports an Ink raw-mode error.

   The daemon then loses the native install method. It changes the prior `claude update` action to a manual-update state.

2. **Medium:** A doctor failure defaults an unknown release channel to `latest`.

   An older stable npm client can run its stable update. The verifier can still require the latest version and report failure.

3. **Medium:** The onboarding page now runs the doctor command every 15 seconds.

   One call used 1.83 seconds of real time and 0.78 seconds of CPU time on this host.

### Other review areas

I found no security problem. I also found no material duplicate implementation in this change.

The daemon correctly owns the update policy. The UI and CLI correctly consume the typed status.

The architecture should cache the doctor result or use a small targeted probe. The code must also keep the unknown channel explicit.

### Verification

- I reviewed all 12 changed files and their related callers at the stated head SHA.
- `git diff --check` passed.
- The four affected Turbo type checks passed.
- The four affected test suites passed all 3,522 tests.
- All GitHub checks passed.
- The real Updates page loaded in the branch dev app.
- The page showed Claude Code 2.1.228 and one machine in sync.
- The refresh control entered its busy state and completed successfully.

I left this as a comment review, as required. I did not approve the pull request or request changes.

PR #1373, opened 2026-08-12
Handle Claude doctor failures safely

ymichael:
## Summary

- Follow up on review findings from #1367 by preserving managed native updates when older `claude doctor` versions fail outside a TTY.
- Keep an unavailable Claude release channel unknown instead of defaulting stable installations to `latest`; only report a definite update when both possible channels are newer.
- Verify unknown-target updates by requiring the installed version to advance, and present the unknown target accurately in the app.
- Bump the host-daemon protocol to 106 so enrolled machines receive the corrected status and action behavior.

## Testing

- `pnpm exec turbo run test --filter=@bb/host-daemon --filter=@bb/app --filter=@bb/host-daemon-contract --force` (3,141 tests)
- `pnpm exec turbo run typecheck --filter=@bb/host-daemon --filter=@bb/app --filter=@bb/host-daemon-contract`
- `git diff --check`

> AGENT GENERATED: by GPT-5

### DIFF
--- a/apps/app/src/components/provider-cli/provider-cli-install.test.tsx
+++ b/apps/app/src/components/provider-cli/provider-cli-install.test.tsx
@@ -161,6 +161,22 @@ describe("buildProviderCliIssue", () => {
       title: "Claude Code update available",
     });
   });
+
+  it("describes an update without inventing a target for an unknown channel", () => {
+    const actionable = issueForProvider("claudeCode");
+    const issue = buildProviderCliIssue({
+      provider: "claudeCode",
+      status: {
+        ...actionable.status,
+        latestVersion: null,
+      },
+    });
+
+    expect(issue).toMatchObject({
+      description: "1.0.0; newer release available",
+      title: "Claude Code update available",
+    });
+  });
 });
 
 describe("useProviderCliInstallRunner", () => {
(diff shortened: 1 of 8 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the Slop Cop. I am reviewing this pull request for security, code quality, performance, architecture, and end-to-end behavior.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** Claude sometimes cannot tell BB which update choice it uses. This is like a sign with no arrow. The change keeps the choice unknown and offers an update only when other facts make it safe.

I found no blocking issue.

- Security: I found no command injection, access increase, secret exposure, or unsafe boundary change.
- Quality: The unknown channel stays explicit through the contract, daemon, app, and update check. The protocol version increases to 106.
- Performance: The change adds only bounded text parsing, path checks, and version comparisons. It adds no process or network call.
- Architecture: I found no duplicate parser. A future helper could centralize Claude installer policy. One repeated update-action branch can also become one condition.
- Test coverage: A Windows test for the `.local/bin/claude.exe` fallback would add useful coverage. I do not consider this a blocker.

Local validation passed. All 3,141 affected tests passed, and all three affected type checks passed. `git diff --check` also passed.

I did not run a browser test. A browser cannot create the failed `claude doctor` state without a fake host CLI. The command-runner tests exercise that boundary, and the app test checks the visible text.

PR #1370, opened 2026-08-12
Fix new-thread panel tab creation

ymichael:
## Summary

- restore the new-tab button in the new-thread right panel
- enforce the control for both desktop and compact panel layouts
- add regression coverage for the root compose panel contract

## Verification

- `pnpm exec turbo run test --filter=@bb/app --force -- src/views/RootComposeSecondaryContent.test.tsx`
- `pnpm exec turbo run typecheck --filter=@bb/app`
- `pnpm exec turbo run lint --filter=@bb/app`

> AGENT GENERATED: by GPT-5

PR #1375, opened 2026-08-12
Extract shared secondary panel host from the two biggest views

ymichael:
## Summary

Extract the right-hand secondary-panel host that `RootComposeView.tsx` and `ThreadDetailView.tsx` each hand-rolled into one shared implementation:

- `apps/app/src/components/secondary-panel/SecondaryPanelHost.tsx` — panel tab state, terminal create/close/sync, browser deck, file-preview tabs, `panel.*`/`terminal.open`/`file.quickOpen`/`workspace.openPreferred`/`diff.toggle` command handlers, and the panel content switch, all behind an explicit `SecondaryPanelHostCapabilities` object that names every real difference between the two views (`hideNewTab`, `autoOpenNewTabWhenEmpty`, `preserveWorkspaceTabsAcrossContexts`, `closeLoneNewTabByHidingPanel`, `routeBrowserPopupsByPreference`, `registerLegacyOpenNewTab`, `toggleOpensNewTab`).
- `apps/app/src/components/secondary-panel/useSecondaryPanelFileTabs.tsx` — the tab-strip builder that was 115/131 lines identical between the two views (including a byte-identical `RightPanelFileTabIcon`).

## Motivation

These are the #1 and #2 churn files in the repo (181 and 99 changes in the last 3 months). The panel host was written twice, and drift between the copies is a live bug class: three of the last ten commits on `main` (#1370, #1353, #1359) were panel-tab fixes. Every such fix currently has to land twice or silently drifts. After this change it lands once.

## Honest accounting

Repo-wide this is only about **−210 lines net** (1,974 insertions, 2,183 deletions). The two views shed 1,643 lines (`RootComposeView` 3,564 → 2,635; `ThreadDetailView` 2,905 → 2,191), and 1,436 of those come back as the single-copy host + tabs hook (plus a 172-line test). The value is single-copy semantics in the hottest files, not deletion.

## Relationship to recent main PRs

The branch was authored on top of `5a17540` (#1370), so #1370, #1365, #1359, and #1353 are ancestors and the rebase onto current `main` was conflict-free (#1362 and #1373 touch disjoint files). Each was verified through the shared host rather than assumed:

- **#1370 (new-thread panel tab creation)**: the fix lives in `RootComposeSecondaryContent.tsx` (untouched here), which forces `showNewTabButton` on and omits it from the host-provided props. Neither view nor the shared host reintroduces `showNewTabButton: false`; `ThreadSecondaryPanel` defaults it to `true`. The regression test `RootComposeSecondaryContent.test.tsx` still passes.
- **#1365 (plugin panel actions on the New thread screen)**: root feeds the host's `openPluginPanel` into `usePluginNewThreadPanelActions` and passes `newThreadPanelActions` as the host's `pluginPanelActions`; thread detail keeps `usePluginPanelActions`/`threadPanelActions`. Each view keeps its own action registry; the shared tabs hook resolves tab icons from whichever list its view provides.
- **#1362 (plugin nav sidebar accessories)**: entirely in `PluginNavSidebarItems`/plugin-sdk/tasks plugin — disjoint from every file this branch touches, so both contexts get accessories from the shared sidebar chrome unchanged.
- **#1353 (stale terminal tabs)**: the fix's client half lives in the already-shared `terminalPanelTabs.ts` helpers, which the host imports; its tests are unchanged and passing. The host's unified `useCreateTerminal`/`useCloseTerminal` path is behaviorally identical to root's old split mutations (`useCreateEnvironmentTerminal`/`useCloseEnvironmentTerminal` are thin wrappers that build the same request).
- **#1359 (drawer tap-through / sticky-scroll cache)**: timeline + sidebar files, disjoint from this branch.

## Review guide (commit by commit)

1. `f011a8609 refactor(app): share secondary panel file tabs` — extracts the tab-strip builder. The only real differences between the two copies were the `isHidden` flag on root's placeholder tab (now `hideNewTab`) and the actions list each view passes; a normalized diff of the old thread-detail builder against the hook shows only mechanical transforms.
2. `18ffd2633 refactor(app): extract shared secondary panel host` — the big one. Moves panel state, terminal plumbing, browser deck, file openers, command handlers, and panel content rendering into `useSecondaryPanelHost`/`SecondaryPanelCommandHandlers`/`SecondaryPanelHostContent`. Review the two views' `SecondaryPanelHostCapabilities` objects against what each view did before; every flag encodes a pre-existing behavioral difference. Also extends `useComposerTypeahead` so root can use it (`hostId` pass-through, optional `currentThreadId`), replacing root's hand-rolled `usePromptMentions` + `useCommandSuggestions` wiring.
3. `db33b8942 test(app): cover shared panel command routing` — moves/extends `RootComposePanelCommandHandlers.test.tsx` into `SecondaryPanelHost.test.tsx` and deletes the now-unused root-only command bridge.
4. `3d0ea6e84 refactor(app): preserve plugin panel tab icons` — restores the plugin-tab icon lookup dropped in an earlier pass (via a capability flag).
5. `fd32e6be7 refactor(app): drop always-on plugin tab icon capability` *(new in this PR)* — commit 4's flag ended up `true` in both views, i.e. a knob with one position; this removes it and makes the icon lookup unconditional, matching both views on `main`.
6. `f7691b282 fix(app): keep projectless root mentions unscoped` *(new in this PR)* — verification found one real semantic drift in the extraction: `mentionsProjectId: isProjectless ? undefined : projectId` was silently defeated by the hook's `?? projectId` fallback, scoping projectless mention ranking and plugin mention search to `proj_personal` where `main` left them unscoped. `mentionsProjectId` is now a required, explicit argument.

## Remaining duplication

The audit that motivated this branch measured 333 matching lines (runs ≥5) between the two views; after the extraction, 30 remain across 4 runs: a connected-hosts `Set` memo, a working-tree `openWorkspaceFile` callback, a file-search selection handler, and the shared-host mount JSX itself. All are thin per-view glue; deliberately left in place to keep this PR's scope to the panel host.

## Provenance

This branch originated in archived thread `thr_germ7hsseh` (branch `bb/de-slop-1-shared-secondary-panel-host-thr_germ7hsseh`), rebased onto current `main` with original authorship preserved; commits 5–6 were added during rebase verification.

## Verification

- `pnpm exec turbo run typecheck` — 58/58 tasks pass
- `pnpm exec turbo run test --filter=@bb/app --force` — 336 files / 2,551 tests pass, including the new `SecondaryPanelHost.test.tsx`
- `git diff --check` clean


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am now reviewing this pull request for security, code quality, performance, architecture, duplication, and end-to-end behavior.

apps/app/src/components/secondary-panel/SecondaryPanelHost.tsx
```diff
@@ -0,0 +1,1259 @@
...
+    return {
+      kind: "host_path" as const,
+      hostId: terminalTarget.hostId,
+      ...(terminalTarget.cwd === null ? {} : { cwd: terminalTarget.cwd }),
+    };
+  }, [terminalTarget]);
+  const terminalsListQuery = useTerminals(terminalQueryScope, {
+    enabled: isSecondaryPanelOpen && terminalTarget !== null,
+  });
+  const loadedTerminalSessions = resolveSecondaryPanelTerminalSessions({
```
SawyerHood: 🚨 `slopcop/review` — Restore the terminal value memoization.

The shared host now derives filtered sessions on every render. It also derives synchronized tabs again at line 775. These new arrays invalidate the terminal map, tab models, and synchronization effect during unrelated prompt or timeline updates. Both former callers memoized this work.

Please memoize the filtered sessions and synchronized tabs with their data inputs. A rerender test can verify stable references and no repeated state synchronization.

apps/app/src/components/secondary-panel/SecondaryPanelHost.tsx
```diff
@@ -0,0 +1,1259 @@
...
+    [closeTerminal, removeFixedTerminalTab, terminalTarget],
+  );
+  const handleActivateFileTab = useCallback(
+    (tabId: string) => {
+      activateTab(tabId);
+      openCompactDrawer();
+    },
+    [activateTab, openCompactDrawer],
+  );
+  const handleCloseWindowRequest = () => {
```
SawyerHood: 🚨 `slopcop/review` — Keep the desktop close callback stable.

This function changes on each host render. The command handler effect depends on it, so prompt edits and live thread updates replace the desktop close listener. Both former implementations used `useCallback`.

Please restore `useCallback` with complete dependencies. A rerender test can verify that one desktop listener stays registered.

apps/app/src/components/secondary-panel/SecondaryPanelHost.test.tsx
```diff
@@ -75,7 +75,32 @@ afterEach(() => {
...
+        isFocused,
+        registerLegacyOpenNewTab: false,
+      }}
+      onOpenPreferred={() => false}
+    />
+  );
+}
+
+describe("SecondaryPanelCommandHandlers", () => {
   it("routes panel shortcuts and desktop close requests only to the focused New Thread pane", async () => {
```
SawyerHood, line 127: 🚨 `slopcop/review` — Add coverage for the other shared command routes.

This test covers toggle, close, and desktop close routing. The extraction also centralizes new-tab, quick-open, terminal, preferred-open, diff, and legacy desktop new-tab routing. Focus or capability regressions in those paths can pass this test.

Please add table-driven focused and unfocused cases. Add a legacy desktop case where `onAppCommand` is absent.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: Two rooms had their own tool boxes. This change gives them one shared tool box. It works. Some tools move too often. Some buttons need more tests.

I found no blocking issue.

I left three low-severity inline findings:

- Memoize the filtered terminal sessions and synchronized tabs.
- Keep the desktop close callback stable.
- Test all shared command routes and the legacy desktop route.

I found no security issue.

The codebase scan found no second panel state host. The remaining secondary components own layout behavior.

The architecture can improve later. The shared host still imports two thread-detail helpers. Seven booleans can also form invalid capability combinations. The returned controller contract is large.

I completed these checks:

- The app type check passed.
- The focused unit test passed.
- The full app suite passed with 2,551 tests.
- The browser test passed both large views.
- The test covered panel toggle, new tab, file search, and file preview.
- All GitHub CI checks passed.
- The GPT-5.6 review gate returned `APPROVE`.

ymichael:
Addressed the three SlopCop inline findings:

**1. `SecondaryPanelHost.tsx:744` — unmemoized terminal derivations (confirmed, fixed in ecf8deea0)**
Verified against `main`: RootComposeView memoized `loadedTerminalSessions` with `useMemo` (ThreadDetailView used the raw stable query ref), and both views memoized `syncedOrderedSecondaryFileTabs`. The extracted host ran `resolveSecondaryPanelTerminalSessions` (a fresh `.filter` array for host-path targets) and `buildTerminalSyncedSecondaryFileTabs` (always a fresh array when sessions are loaded) on every render, invalidating the `terminalsById` map, the `useSecondaryPanelFileTabs` tab models, and re-running the terminal sync effect on unrelated rerenders. Both are now wrapped in `useMemo` keyed on their data inputs (`terminalsListQuery.data?.sessions` + `terminalTarget`, and `loadedTerminalSessions` + `orderedSecondaryFileTabs` + `retainedTerminalId`), matching the pre-extraction deps. `terminalTarget` is memoized by both callers, so it is a sound dep.

**2. `SecondaryPanelHost.tsx:1121` — unstable desktop close callback (confirmed, fixed in ecf8deea0)**
Both pre-extraction views wrapped `handleCloseWindowRequest` in `useCallback`; the extracted host made it a plain function, and the `onCloseWindowRequest` effect in `SecondaryPanelCommandHandlers` depends on it, so the desktop close listener was torn down and re-registered on every host render. Restored `useCallback` with the complete dep set (`isOpen`, `activeFixedSecondaryTab`, `capabilities.closeLoneNewTabByHidingPanel`, `fixedPanelTabsState.secondary.tabs.length`, `closePanel`, `closeTab`, `handleCloseTerminalTab`). Note the `useAppCommandHandler` registrations were already churn-proof (they hold the handler in a ref); the desktop listener effect was the real exposure.

**3. `SecondaryPanelHost.test.tsx:104` — missing route coverage (confirmed, added in 79fc85def)**
Added table-driven cases for `panel.newTab`, `file.quickOpen`, `terminal.open`, `workspace.openPreferred`, and `diff.toggle`, each dispatched against a focused pane plus a fully-capable unfocused pane (so any focus-gating bug trips the unfocused pane's zero-call assertions), asserting handler calls and handled/unhandled propagation via `defaultPrevented`. Covers the `canStartTerminal` capability gate, the `onOpenPreferred` result passthrough, and the missing-`onToggleDiff` fall-through. Also added the legacy desktop cases: with `onAppCommand` absent, only the focused pane with `registerLegacyOpenNewTab` registers the `onOpenNewTab` listener; with `onAppCommand` present, no legacy listener is registered. Every route in the shared handler carries real gating logic, so none were skipped as pure passthrough. Mutation-checked: removing the `terminal.open` focus gate fails both terminal cases.

Verification: `turbo run typecheck --filter=@bb/app` clean; `turbo run test --filter=@bb/app` all 336 files / 2561 tests pass (11 in this suite).

ymichael:
**Follow-up: CI lint failure and correction of the earlier memoization fix**

CI caught three `react-hooks/preserve-manual-memoization` errors in `SecondaryPanelHost.tsx`, all introduced by ecf8dee ("Restore memoization lost in the secondary panel host extraction"). The manual `useMemo`/`useCallback` wrappers depend on values the React Compiler considers mutable (`orderedSecondaryFileTabs`, `activeFixedSecondaryTab`), so the compiler could not preserve them and **skipped compiling `useSecondaryPanelHost` entirely** — losing auto-memoization for the whole ~1000-line hook.

Correcting my earlier fix comment: the review's perf findings were already handled by the React Compiler. `apps/app` builds with `reactCompilerPreset()` (`vite.config.ts`), and at f7691b2 — before the manual memoization — this exact lint passes with zero errors, meaning the hook compiled cleanly and the flagged derivations (`loadedTerminalSessions`, `syncedOrderedSecondaryFileTabs`, `handleCloseWindowRequest`) were all auto-memoized in the shipped build. The manual memoization was at best redundant, and as written it made things strictly worse by defeating the compiler.

9eb3419 reverts ecf8dee and instead locks in the behavior with a test as the durable artifact: the desktop close listener registers once across rerenders with stable inputs (the host prop object is deliberately rebuilt every render, so keying the effect on the whole object fails the test — mutation-verified), and re-registers exactly when the close-callback identity changes, without leaking the old listener or invoking a stale closure.

Gates on the new head: lint 0 errors (148 pre-existing warnings, unchanged), typecheck clean, 2562 tests pass.

PR #1379, opened 2026-08-12
Speed up the mobile right panel

SawyerHood:
## Summary

- replace the mobile right panel modal path with a persistent portal shell
- avoid a full application-tree style recalculation during panel open and close
- defer the first heavy panel mount by two frames and retain it for later opens
- preserve the backdrop, Escape key, keyboard cleanup, focus, nested panel, and handle drag behavior

## Verification

- [x] 30 focused right-panel tests
- [x] @bb/app and @bb/shared-ui type checks
- [x] @bb/app production build
- [x] iOS 18.4 Simulator first-open and reopen traces show zero potential WebKit hangs

> AGENT GENERATED: by GPT-5.6


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am reviewing this pull request for security, code quality, architecture, and performance.

I will also test the mobile right-panel path in the real application when the local setup permits it.

apps/app/src/views/RootComposeSecondaryContent.tsx
```diff
@@ -393,28 +393,25 @@ export function RootComposeSecondaryContent({
...
-          handleOnly
-          repositionInputs={false}
         >
           <div className="flex min-h-0 flex-1 flex-col overflow-hidden">
-            {/* The panel mounts after the sheet settles: mounting it inside
-                the opening tap costs hundreds of milliseconds of style
-                resolution on iOS and freezes the slide-in. */}
+            {/* Paint the light shell first. The real panel mounts two frames
+                later, while the compositor moves the drawer. */}
             {isPanelRealized ? (
```
SawyerHood, line 408: 🚨 `slopcop/review` — The persistent shell does not retain the active file-tab body.

`ThreadSecondaryPanel` still uses `isOpen || !renderAsDrawer` for its file content. Close removes that subtree, while `isPanelRealized` stays true.

I reproduced this in the mobile dev app. The `Start terminal` node disappeared on close and returned as a new node on reopen. No skeleton covered that mount.

Later opens can still pay the heavy mount cost during the open action. They can also lose local tab state.

Please let the persistent drawer retain this body, or pass an explicit retention mode into `ThreadSecondaryPanel`. This issue also affects the thread-detail caller.

packages/shared-ui/src/components/ui/responsive-overlay.tsx
```diff
@@ -331,3 +334,312 @@ export function ResponsiveDrawerShell({
...
+    return () => window.clearTimeout(timeout);
+  }, [motionDurationMs, open, reportSettled]);
+
+  React.useEffect(() => {
+    if (!open) {
+      return;
+    }
+
+    const handleKeyDown = (event: KeyboardEvent) => {
+      if (event.key === "Escape") {
```
SawyerHood, line 436: 🚨 `slopcop/review` — This capture listener closes the outer drawer before a nested overlay can handle Escape.

The right panel contains portaled menus and pickers. Radix registers its document listener after this shell.

This handler prevents the event first. The nested surface then skips its own close, while the outer right panel closes.

Please use the bubble phase and return when `event.defaultPrevented` is true. The mobile sidebar already uses that pattern.

packages/shared-ui/src/components/ui/responsive-overlay.tsx
```diff
@@ -331,3 +334,312 @@ export function ResponsiveDrawerShell({
...
+      const activeElement = panel.ownerDocument.activeElement;
+      if (
+        event.shiftKey &&
+        (activeElement === first || activeElement === panel)
+      ) {
+        event.preventDefault();
+        last?.focus({ preventScroll: true });
+      } else if (
+        !event.shiftKey &&
+        (activeElement === last || !panel.contains(activeElement))
```
SawyerHood: 🚨 `slopcop/review` — The Tab trap takes focus from portaled child surfaces.

A Radix menu or dialog renders under `document.body`, outside `panelRef`. This condition then moves focus to the first drawer control.

Shift+Tab has the opposite fault. Focus outside the panel can move into the background app.

Please ignore focus outside the panel unless the active element is `document.body`. The sidebar focus trap has this guard.

packages/shared-ui/src/components/ui/responsive-overlay.tsx
```diff
@@ -331,3 +334,312 @@ export function ResponsiveDrawerShell({
...
+        panelRef.current?.focus({ preventScroll: true });
+      });
+    }
+    return () => {
+      document.removeEventListener("keydown", handleKeyDown, true);
+      if (focusFrame !== null) {
+        window.cancelAnimationFrame(focusFrame);
+      }
+    };
+  }, [isPointerCoarse, open, requestClose]);
```
SawyerHood, line 499: 🚨 `slopcop/review` — Parent renders can restart this effect and move focus again.

Both callers pass a new `onOpenChange` function on each render. That changes `requestClose`, which restarts this effect.

On a narrow fine-pointer window, each restart schedules `panelRef.current.focus()`. Stream updates can therefore take focus from an input inside the panel.

Please run initial focus once per open. Keep the latest close callback in a ref, or stabilize both caller callbacks.

packages/shared-ui/src/components/ui/responsive-overlay.tsx
```diff
@@ -331,3 +334,312 @@ export function ResponsiveDrawerShell({
...
+  }, [open]);
+
+  const setDragPosition = React.useCallback(
+    (offsetY: number, animate: boolean) => {
+      const panel = panelRef.current;
+      const backdrop = backdropRef.current;
+      if (panel === null || backdrop === null) {
+        return;
+      }
+      const height = Math.max(panel.clientHeight, 1);
```
SawyerHood: 🚨 `slopcop/review` — This layout read runs for every drag move.

`handleDragMove` calls this function at pointer rate. A dirty timeline can make `clientHeight` force document layout before each transform write.

The panel height stays constant during one handle drag. Please read it once at pointer down and store it in `dragRef`.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** The change starts a light mobile drawer first. It adds the large panel after the drawer starts to move.

It also avoids Radix and Vaul changes that caused WebKit delays.

**Review result for `ad83748`:** I found five problems.

- The drawer still removes the active file-tab body when it closes. A reopen mounts a new node without the skeleton.
- The global Escape handler can close the outer panel before a nested menu closes.
- The Tab trap takes focus from a portaled menu. Shift+Tab can also reach the background.
- Unstable close callbacks restart the focus effect on each parent render in a narrow, fine-pointer window.
- `clientHeight` runs for every drag move and can force layout.

**Security:** I found no security issue or trust-boundary change.

**Architecture:** The shell repeats focus, keyboard, drag, and settle logic from `SidebarMobilePanel`.

Share the common helpers, but keep the vertical and horizontal shells separate.

**Validation:** All 30 changed tests passed. Turbo type checks passed for `@bb/app` and `@bb/shared-ui`.

Chromium mobile QA passed the first open, close, backdrop, and reopen paths at 390 by 844 pixels.

The QA also proved that the `Start terminal` node unmounts on close and mounts as a new node on reopen.

I did not repeat the iOS simulator trace.

The PR advanced to `6bf9fb1` during this review. This review covers the required SHA, `ad83748`, only.

PR #1380, opened 2026-08-12
Fix duplicate mobile sidebar close animation

ymichael:
## Summary

- keep the current mobile sidebar variant mounted through its deferred close transition
- swap app, Settings, and Extensions sidebars only after the drawer is offscreen
- add regression coverage for navigating to Settings and back to the app

## Testing

- pnpm exec turbo run test --filter=@bb/app --force -- src/components/ui/sidebar.test.tsx src/components/layout/AppLayoutSidebar.test.tsx src/components/layout/AppLayout.root-compose-project.test.tsx src/components/layout/AppLayout.plugin-panel-header.test.tsx
- pnpm exec turbo run lint --filter=@bb/app

> AGENT GENERATED: by Codex GPT-5

### DIFF
--- a/apps/app/src/components/layout/AppLayoutSidebar.tsx
+++ b/apps/app/src/components/layout/AppLayoutSidebar.tsx
@@ -1,0 +1,83 @@
+import {
+  useEffect,
+  useState,
+  type MouseEvent as ReactMouseEvent,
+} from "react";
+import { AppSidebar } from "@/components/sidebar/AppSidebar";
+import { SettingsSidebar } from "@/components/settings/SettingsSidebar";
+import { ToolsSidebar } from "@/components/tools/ToolsSidebar";
+import { useSidebar } from "@/components/ui/sidebar.js";
+
+export type AppLayoutSidebarMode = "app" | "settings" | "tools";
+
+interface AppLayoutSidebarProps {
+  mode: AppLayoutSidebarMode;
+  onResizeMouseDown: (event: ReactMouseEvent<HTMLDivElement>) => void;
+  isResizing: boolean;
+  appRoutePath: string;
+  settingsRoutePath: string;
+  toolsBackRoutePath: string;
+  toolsRoutePath?: string;
+}
+
+/**
+ * Keeps the current mobile drawer mounted until its close animation finishes.
+ *
+ * Routes such as Settings replace the entire sidebar. Mobile closes are
+ * intentionally deferred so the compositor can finish the slide before the
+ * expensive React state commit. Swapping sidebar modes during that window
+ * would mount a fresh, still-open panel and replay the close animation.
+ */
+export function AppLayoutSidebar({
+  mode,
+  onResizeMouseDown,
+  isResizing,
+  appRoutePath,
+  settingsRoutePath,
+  toolsBackRoutePath,
+  toolsRoutePath,
+}: AppLayoutSidebarProps) {
+  const { isCompactViewport, openMobile } = useSidebar();
+  const holdCurrentMode = isCompactViewport && openMobile;
+  const [settledMode, setSettledMode] = useState(mode);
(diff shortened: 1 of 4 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop, and I started the review. I will check security, code quality, performance, architecture, and the mobile behavior.

apps/app/src/components/layout/AppLayoutSidebar.tsx
```diff
@@ -0,0 +1,83 @@
...
+  mode,
+  onResizeMouseDown,
+  isResizing,
+  appRoutePath,
+  settingsRoutePath,
+  toolsBackRoutePath,
+  toolsRoutePath,
+}: AppLayoutSidebarProps) {
+  const { isCompactViewport, openMobile } = useSidebar();
+  const holdCurrentMode = isCompactViewport && openMobile;
```
SawyerHood: 🚨 `slopcop/review` — The hold can keep the wrong sidebar after browser navigation.

`openMobile` means that the drawer is open. It does not mean that a close operation is active.

I reproduced this fault at 390 × 844 pixels:

1. Open the app drawer.
2. Select Settings and wait for the drawer to close.
3. Open the Settings drawer.
4. Use browser Back.

The URL becomes `/`. The panel stays open and still shows the Settings navigation. It stays wrong until the user closes it.

Please hold the mode only while a close operation is active. `SidebarProvider` can expose that state from its settle operation.

Add a test that changes `mode` while the drawer stays open. The current test always calls `closeMobileSidebar()` before each mode change.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** This PR keeps the old menu visible while the drawer slides away. It prevents two close slides. One route can now keep the wrong menu open.

I found one required fix.

### Required fix

`AppLayoutSidebar` holds the old mode whenever the mobile drawer is open. It should hold the mode only during an active close.

I reproduced the fault in the local app at 390 × 844 pixels. I opened Settings, reopened its drawer, and used browser Back. The URL changed to `/`, but the open drawer still showed the Settings navigation. The correct app sidebar appeared only after I closed the drawer.

The inline comment has the full reproduction. Please expose a provider-owned close state and use it instead of `openMobile`. Add a test that changes `mode` without a close call.

### Other review results

- **Security:** I found no security issue. The change adds no data boundary, permission change, injection point, or protocol change.
- **Performance:** The close path produced one `translate` transition in the browser. However, the effect at `AppLayoutSidebar.tsx:44` causes a second render after each mode swap. The lint check also reports this new effect. A ref or one provider-owned state machine can avoid that render.
- **Code quality:** The new component removes the large sidebar branch from `AppLayout`. The class assertions in the new test are static and do not prove that a second transition did not occur.
- **Architecture:** The codebase has separate settle paths for navigation, inset swipe, and panel drag. A provider-owned close phase would give this component one correct signal.

### Verification

- All 337 app test files passed. All 2,552 tests passed.
- The app type check passed.
- The app lint check finished with zero errors and 149 warnings. One warning points to the new effect.
- The normal app-to-Settings and Settings-to-app paths each used one close transition.
- The browser Back test reproduced the stale open sidebar.

I did not find another required change.

apps/app/src/components/layout/AppLayoutSidebar.tsx
ymichael: Fixed in 79c6af6a5. `SidebarProvider` now exposes the actual deferred-close phase, and `AppLayoutSidebar` holds the previous mode only during that phase. A route change while the drawer remains open now swaps sidebar content immediately. I also replaced the synchronization effect and added the requested no-close mode-change regression test. The 18 related tests, app typecheck, and app lint pass.

PR #1385, opened 2026-08-12
Harden provider lifecycle and provisioning recovery

ymichael:
## Summary

- harden provider process teardown and replacement so final output drains before cleanup completes
- preserve Codex command output across event reordering and retry transient rename failures during rollout initialization
- make provisioning recovery reject stale events and retain managed branch identity before live RPC dispatch
- make standalone recovery QA deterministic and reusable, and align archive and permission runbooks with current behavior

## Review follow-ups

- close inherited provider pipes at the exit grace deadline and ignore output from superseded processes
- persist the active provisioning attempt before daemon dispatch and reject progress from older attempts on the same environment
- replace the native filesystem timing dependency in the Docs watcher test with an injected, deterministic watcher callback
- stabilize the shared server database-maintenance test clock uncovered while validating the review fixes

## Validation

- `pnpm exec turbo run lint --force` — passed with 0 errors (148 pre-existing warnings)
- `pnpm exec turbo run typecheck --force` — 58/58 tasks passed
- `pnpm exec turbo run build --force` — 14/14 tasks passed
- `pnpm exec turbo run test --force` — 55/55 tasks passed
- `env -u OPENAI_API_KEY pnpm exec turbo run test:integration --force` — 57 runtime tests and 11 real-provider tests passed
- repeated the full Docs package test five additional uncached times after removing its watcher-readiness race
- completed the full `qa/manual-runbook.md` with Codex, Claude Code, and Pi, including provisioning, server/daemon recovery, and pending-interaction scenarios
- verified standalone teardown and cleanup left no processes or temporary roots behind

## Protocol

- bumps `HOST_DAEMON_PROTOCOL_VERSION` from 105 to 108 for daemon/server event contract changes and to resolve an independent version collision with the latest `main`

> AGENT GENERATED: by GPT-5


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the Slop Cop. I am reviewing this pull request for security, code quality, performance, architecture, duplication, and end-to-end behavior.

packages/agent-runtime/src/runtime-provider-process.ts
```diff
@@ -403,13 +436,49 @@ export class RuntimeProviderProcessManager {
...
-        providerProcess,
         signal: signal ?? null,
-      });
+      };
+      exitStatus = status;
+      // `exit` can precede the final stdout/stderr data events. Prefer
+      // `close`, which fires after stdio closes, so final provider output is
+      // consumed before pending requests and diagnostics are settled. Bound
+      // the wait because a descendant can inherit and hold a pipe open.
+      closeGraceTimer = setTimeout(() => {
```
SawyerHood, line 480: 🚨 `slopcop/review` — Close the old provider streams when the grace period ends.

`handleExit` removes the process from the manager. However, the readline handler and both pipe streams remain open. A descendant can keep a pipe open past one second. It can then send stale JSON-RPC after a replacement starts. The open pipe can also keep resources for an unlimited time.

Please close the readline interface and destroy both pipes before finalization. Also ignore callbacks from a process that is no longer current. Add a test with descendant output after replacement.

apps/server/src/internal/events.ts
```diff
@@ -820,30 +822,41 @@ function resolvePostableEventBatchEntries(
...
       rejectedEvents.push({
         eventIndex,
         reason: "thread_not_owned_by_host",
         threadId: entry.threadId,
       });
       continue;
     }
-    const canonicalEnvironmentId =
-      canonicalEnvironmentIdByThreadId.get(entry.threadId) ?? null;
+    if (
```
SawyerHood, line 845: 🚨 `slopcop/review` — Compare the provisioning attempt ID.

This guard checks only the environment ID and status. Attempt B can start on the same environment before the daemon flushes queued progress from attempt A. The server then accepts the old event. A stale cancellation can become the newest row and affect attempt B.

Please persist the active `provisioningId` before dispatch. Require each progress event to match it. Add a test with two attempts on one environment.

plugins/docs/server.test.ts
```diff
@@ -1061,13 +1061,17 @@ describe("Docs vault operations", () => {
...
+        () =>
+          harness.realtimeSignals.some(
+            (signal) =>
+              signal.channel === "vault-changed" &&
+              isRecord(signal.payload) &&
+              signal.payload.vaultId === "personal",
+          ),
+        // Stay below the 10-second polling interval while allowing native
+        // FSEvents and the 250ms debounce to run under full-suite contention.
+        5_000,
```
SawyerHood: 🚨 `slopcop/review` — Make this watcher test deterministic.

This change only increases the timeout. One worker saw this test fail in a broad run and an isolated run. A later isolated run passed. The mixed result shows that the test still depends on native watcher readiness.

A longer timeout cannot recover a missed filesystem event. Please add a watcher-ready barrier or inject the watcher callback in this test.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** This PR makes provider exits and workspace recovery safer. It also preserves Codex command output when events arrive out of order.

I found three issues:

1. **High — old provider pipes remain active.** The one-second deadline finalizes the exit but leaves the old streams and handlers open. A descendant can send stale JSON-RPC after a replacement starts.
2. **High — the stale-event guard ignores the attempt ID.** Old progress can enter a newer attempt on the same provisioning environment.
3. **Medium — the Docs watcher test remains unreliable.** A longer timeout does not fix a filesystem event that occurs before watcher readiness.

I found no additional security, performance, architecture, duplication, or refactor issue that needs a comment.

I checked the exact requested SHA. The PR head later merged `main`, but that merge did not change these lines.

Seven Turbo test tasks passed. Six Turbo typechecks passed. The current GitHub CI passed. A real Codex thread reached idle and returned `ok` in the browser.

The Docs test passed once in my isolated run. Another worker saw it fail in both a broad run and an isolated run.

I submitted this as a comment-only review. I did not approve the pull request or request changes through GitHub.

packages/agent-runtime/src/runtime-provider-process.ts
ymichael: Fixed in 920944e20. When the one-second grace deadline expires, BB now closes the readline interface and destroys both old provider pipes before exit finalization. Stdout and stderr callbacks also verify that their provider process is still current. The regression test starts a descendant that writes after replacement and proves the old streams are destroyed and its stale line is ignored. The agent-runtime typecheck and all 914 package tests pass, as do the final 57 runtime and 11 real-provider integration tests.

apps/server/src/internal/events.ts
ymichael: Fixed in d42746c3a. A generated migration adds active_provisioning_id to environments. The server records the attempt before daemon dispatch, clears it on lifecycle settlement, and accepts streamed provisioning progress only when environment ID, status, and attempt ID all match. The regression test claims attempts A then B on one environment and proves a late cancellation from A is rejected while progress from B is stored. DB/server typechecks, 379 DB tests, and 1,438 server tests pass.

plugins/docs/server.test.ts
ymichael: Fixed in 248c802bf. The Docs service now has a small watcher injection seam with the native watcher as its production default. The test registers a deterministic callback, verifies the watched vault root, triggers the callback only after registration, checks the debounced realtime signal, and verifies shutdown closes the watcher. It no longer depends on an FSEvents readiness race or a longer timeout. The package typecheck/test pass, followed by five additional uncached package-test passes.

ymichael:
Closing this oversized validation branch in favor of focused replacement PRs.

- Codex runtime correctness: #1400
- Provider shutdown races: replacement PR in progress
- Flaky-test and QA/runbook maintenance: replacement PR in progress

The provisioning restart-survival/schema work is intentionally abandoned. The corrected runbook will validate only completed provisioning failure followed by a fresh retry on the next message.

PR #1395, opened 2026-08-12
Show prompt-cache status beside context-window usage

AndrewSB:
Adds a prompt-cache indicator beside the context-window meter, using the cached/input token telemetry providers already report rather than inferring cache state.

Prompt cache indicator states

Left to right: a positive cache hit, a reported miss, and a provider that does not report cached tokens.

## Approach

Cache state is read from `last.cachedInputTokens` / `last.inputTokens` on `thread/tokenUsage/updated`, scoped to the **latest root turn**. Two consequences worth noting:

- A nested subagent turn never contributes its numbers to the thread-level indicator — `listLatestTokenUsageRows` filters nested turn usage at the query level.
- Numbers from an earlier turn are never carried forward. If the newest turn reports no cache telemetry, the indicator degrades to unknown rather than showing stale values.

The contract models the no-telemetry case as an explicit `{ status: "unknown" }` variant rather than an optional field. This is the load-bearing design decision: with an optional `cachedInputTokens`, a provider that never reports cache data is indistinguishable from one reporting a genuine 0-token miss, and both would render as "0% hit" — a claim the data does not support. The discriminated union makes "we don't know" unrepresentable as "we know it was zero", and the UI renders it as "Unavailable".

Rendering lives in the shared `ThreadContextWindowIndicator`, so every view using that chrome picks it up with no per-view wiring.

## Testing

`typecheck` and `test` pass across the five affected packages:

| Suite | Result |
|---|---|
| `typecheck` (app, db, server, server-contract, thread-view) | exit 0 |
| `@bb/thread-view` | 21 files, 352 tests passed |
| `@bb/db` | 28 files, 360 tests passed |
| `@bb/app` | 322 files, 2417 tests passed |

Coverage: hit, reported miss, unknown, and live update on a new usage event (`build-thread-timeline.test.ts`); the three rendered states including aria-label text (`ThreadContextWindowIndicator.test.tsx`); latest-root-turn selection skipping nested subagent turns (`events.test.ts`). The stale-turn and nested-turn cases are the two most likely to regress silently, since both would produce plausible-looking numbers rather than an obvious failure.

## Notes for review

**No `HOST_DAEMON_PROTOCOL_VERSION` bump.** The change is confined to the server→client contract (`threadContextWindowUsageSchema` in the timeline response) and touches no session payload, WebSocket message, host RPC command, or RPC result. The daemon wire format is unchanged, so an older enrolled daemon will not hit an `invalid-message` reconnect loop. Flagging explicitly since the repo rule is strict and this does add a required contract field.

**Generated-types churn.** `packages/plugin-sdk/bundled-types/bb-plugin-sdk.d.ts` carries ~50 lines unrelated to this feature — zod enum key reordering (`idle`/`starting`, `local`/`missing`, `utf8`/`base64`, `append`/`replace`) from a regenerator run. The only substantive change there is the `promptCacheUsage` union. Left as generator output rather than hand-edited; happy to land a clean regeneration if preferred.

**Rename.** The timeline profile stages `context-window-*` became `usage-metadata-*`, since that query now loads both context-window and token-usage rows. Grepped project-wide including docs; no stale references.


AndrewSB:
Closing this.

The indicator reports the **last completed turn's** cached-token split — historical, not predictive. The actionable question before you send a message is "will this request hit cache," and that can't be answered from this telemetry: it depends on time-since-last-turn against a TTL the provider never reports (5 min default, 1 hour opt-in), and on the prompt prefix being byte-identical, which nothing here can detect. A per-turn retrospective percentage doesn't inform any decision the user is about to make.

Separately, the math is wrong for Claude Code threads, which is worth recording before this closes. In `packages/agent-runtime/src/claude-code/sdk-extraction.ts:348`, `cachedInputTokens = cache_read_input_tokens + cache_creation_input_tokens` — cache *creation* is a miss paid at a 1.25x premium, not a hit. And Anthropic's `input_tokens` is the uncached remainder, so `cached / inputTokens` isn't a fraction of the prompt; on a warm turn it exceeds 1.0 and the clamp pins it to "100% hit." Codex is unaffected — its `cachedInputTokens` is a subset of `inputTokens`, which is why the Codex-shaped tests passed.

If prompt-cache status is revisited, two things carry forward: the providers need different arithmetic (the contract should carry cache read and cache write separately rather than one pre-summed `cachedInputTokens`), and a liveness indicator needs an assumed TTL plus a prefix-invalidation signal that does not exist in the current event stream.

PR #1405, opened 2026-08-12
Reduce oversized file read log noise

ymichael:
## Summary

- classify file_too_large as an expected online host RPC failure
- preserve the compact debug lifecycle event and unchanged RPC error response while avoiding repetitive warning stack traces
- add regression coverage for the expected-failure classification

## Testing

- pnpm exec turbo run typecheck --filter=@bb/host-daemon
- pnpm exec turbo run test --filter=@bb/host-daemon --force (543 tests)

> AGENT GENERATED: by GPT-5

### DIFF
--- a/apps/host-daemon/src/command-dispatch-support.test.ts
+++ b/apps/host-daemon/src/command-dispatch-support.test.ts
@@ -16,8 +16,10 @@ vi.mock("@bb/agent-runtime", async (importOriginal) => {
 });
 
 import {
+  CommandDispatchError,
   defaultListModels,
   getErrorCode,
+  isExpectedOnlineRpcFailureError,
   shutdownDefaultListModelsRuntimes,
 } from "./command-dispatch-support.js";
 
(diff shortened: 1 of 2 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am reviewing this pull request for security, code quality, performance, tests, and code reuse.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** BB still rejects a file that is too large. It now avoids a large warning log for this normal error.

I found no actionable issue in this pull request.

- **Security:** The size limits, path controls, RPC error code, and RPC error message remain unchanged.
- **Code quality:** The new code extends the existing expected-failure list. A project-wide search found no duplicate that should replace it.
- **Architecture:** The server has a separate expected-error list for settled commands. The daemon list owns this online RPC log policy, so the separation is correct.
- **Performance:** The change removes warning stack serialization and warning log writes for expected failures. The small debug lifecycle record remains.
- **Protocol:** The wire response does not change. This pull request does not need a daemon protocol version increase.

The host-daemon type check passed. All 543 host-daemon tests passed. The focused five-test suite passed. `git diff --check` passed. All GitHub checks passed.

A browser test does not apply here. The change affects only an internal daemon log branch, and the browser-visible response remains the same.

PR #1407, opened 2026-08-12
Add a global All tasks board

AndrewSB:
## Summary

"All tasks" can now be viewed as a board spanning every project, not just as a list.

- The route grammar carries a view mode for the all-tasks route (`all?view=board`), so the choice is linkable and survives navigation.
- The topbar list/board switch now renders on the all-tasks route (still hidden below `BOARD_MIN_WIDTH`, where the shell falls back to the list).
- `BoardView` accepts a `null` projectId to load tasks across all projects, reusing the list filter bar for status/priority/label filtering. Cards show project identity.
- An unfiltered global board omits canceled work: it sends an explicit `statuses` list to `listTasks`. Selecting Canceled in the shared Status filter is still an explicit inspection path.

## Cross-project drag correctness

Board positions are stored per project. A drag in a global column can land between cards from other projects, so `projectDropNeighbors` maps the visual insertion index to same-project neighbors before computing the `beforeTaskId`/`afterTaskId` sent to the server — cards from other projects influence the visual slot but are never sent as reorder neighbors.

## Testing

Three unit tests cover the new behavior. All CI checks pass; locally the two touched test files pass (38 tests) along with `tsc --noEmit` for the plugin.

What the tests actually verify:

- **`routes.ts` round-trip** (`shell.test.tsx`) — `parseTasksRoute(tasksRouteToSubPath(route))` is identity for both all-tasks view modes, and an empty subPath still parses to the list view. This is the guard against the view mode getting dropped or defaulting to board.
- **Global board rendering** (`shell.test.tsx`) — mounting the panel at `all?view=board` with tasks seeded across two projects renders cards from both, shows each project's name on the cards, exposes the List/Board switch, and renders no canceled card. The fake rpc honors the `statuses` argument, so the canceled assertion is a check on the request the board makes, not on the fake.
- **`projectDropNeighbors`** (`drop-position.test.ts`) — one case where a card is dragged in a column interleaved with another project's cards, asserting the neighbors sent to the server are same-project only.

Limits worth naming: the drag coverage is a single unit test of the neighbor calculation, not a rendered drag interaction, and there is no test for the filter bar wiring (priority/label filters) on the global board. I did not run the plugin's `cli/` suite locally — those e2e tests were timing out in this sandbox — but they are untouched by this change and pass in CI.

The first CI run failed on the canceled-card assertion: status filtering is server-side, and the test's fake rpc originally ignored `statuses`, returning the canceled task regardless of what the board requested. Making the fake filter by `statuses` fixed it.

### DIFF
--- a/plugins/tasks/shell/app-shell.tsx
+++ b/plugins/tasks/shell/app-shell.tsx
@@ -162,7 +162,11 @@ function RouteOutlet({
 }) {
   switch (route.kind) {
     case "all":
-      return <ListView projectId={null} />;
+      return route.view === "board" && boardUsable ? (
+        <BoardView projectId={null} />
+      ) : (
+        <ListView projectId={null} />
+      );
     case "active":
       return <ListView projectId={null} activeOnly />;
     case "manage":
@@ -244,7 +248,7 @@ function TasksAppShellContent({ subPath }: PluginNavPanelProps) {
     // Routes are plain data; keying on subPath tracks every route change.
   }, [subPath]);
   const backFromTask = () =>
-    navigation.go(lastBrowseRouteRef.current ?? { kind: "all" });
+    navigation.go(lastBrowseRouteRef.current ?? { kind: "all", view: "list" });
   const onTaskRoute = route.kind === "task";
   const backRef = useRef(backFromTask);
   backRef.current = backFromTask;
--- a/plugins/tasks/shell/routes.ts
+++ b/plugins/tasks/shell/routes.ts
@@ -7,7 +7,7 @@ export const PANEL_PATH = "tasks";
 export type TaskViewMode = "list" | "board";
 
 export type TasksRoute =
-  | { kind: "all" }
+  | { kind: "all"; view: TaskViewMode }
   | { kind: "active" }
   | { kind: "manage" }
   | { kind: "project"; projectId: string; view: TaskViewMode }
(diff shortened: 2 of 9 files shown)


GusevV1987:
We built the same feature independently (private fork of the tasks plugin) before finding this PR, and while comparing implementations we hit one edge case worth reporting here.

**`projectDropNeighbors` computes wrong neighbors for same-column downward reorders.** `dropIndexForPointer` returns an insertion slot over the column *with the dragged card excluded* (per `dropNeighborsForIndex`'s doc comment), but `column.slice(0, visualDropIndex)` counts over the column that still contains the dragged card. Failing case: column `[a(P1), b(P1), x(P2)]`, drag `a` to the slot below `b` (visualDropIndex 1): returns `{before: null, after: "b"}` — a no-op — instead of `{before: "b", after: null}`. The existing unit test doesn't catch it because the card at the slice boundary (`d`) belongs to another project; with a same-project card there, the count comes up one short. Note this also affects ordinary per-project boards, not just the global one: `commitDrop` now routes every board through `projectDropNeighbors`, and with all cards sharing one projectId the same off-by-one makes a downward same-column reorder a no-op (e.g. `[a, b, c]`, drag `a` below `b`: `{before: null, after: "b"}` instead of `{before: "b", after: "c"}`). Filtering the dragged card out of `column` before slicing fixes it; a regression test with a same-project boundary card pins it. Happy to share our version of the helper and its boundary test if useful.

Also, in case it helps the rebase: since this PR was opened, #1704 added the per-project List/Board persistence store (a reserved non-ULID scope key could persist the All-tasks choice the same way) and #1788 moved the Tasks navigation into host-owned fixed tabs — both touch files in this diff.

SawyerHood:

PR #1408, opened 2026-08-12
Fix provider retry for Claude rate limits

ymichael:
## Summary

- canonicalize hard Claude Code rate-limit events into one accepted failed turn with one terminal provider error
- allow a later 429-only failure to inherit the active blocked subscription window, so a premature manual retry does not destroy recovery eligibility
- distinguish manual retries from automatic post-reset continuations and enforce automatic eligibility again at the server boundary
- keep provider-retry timers process-local, serialize all per-thread scheduling/cancellation operations, and attempt a reset window at most once per plugin process
- bump the host daemon protocol so enrolled daemons receive the adapter behavior change

## Root cause

Claude Code reports a hard subscription limit before its synthetic assistant/result sequence. The adapter previously persisted the rate-limit state and an error before `turn/started`, then persisted a second error when the result arrived. Recovery only inspected the failed turn range, so it missed the blocked state. An immediate manual continuation could then fail with a plain 429 and no new rate-limit state, which removed the plugin timer.

The retry plugin also treated reconciliation, cancellation, supersession, deletion, and timer release as separate asynchronous operations. A reconciliation already in flight could therefore recreate an entry after cancellation. They now use one per-thread operation queue, while pending timers remain intentionally in memory and disappear on server/plugin restart.

## Review follow-up

- cancellation cannot be undone by an in-flight reconciliation
- automatic release rechecks both the failed request and `candidate.automatic`; the server independently rejects automatic continuation of a manual-only candidate
- a later Claude `allowed` update clears a pending hard rejection
- durable waiting/attempt keys and restore-time scanning were removed rather than made authoritative
- deleted threads clear their in-process attempt guard
- guide text and generated templates describe the process-local behavior

## Validation

- 10 affected package typechecks passed through Turbo
- agent-runtime suite: 46 files / 917 tests passed
- focused Claude adapter suite: 149 tests passed
- provider recovery server suite: 11 tests passed
- provider-retry scheduler suite: 14 tests passed
- affected CLI, SDK, contract, template, and plugin package suites passed (9 Turbo tasks)
- all 24 plugin-fixture cases that failed only when the entire server suite ran concurrently passed when their four files were isolated
- standalone server + daemon from this worktree with the real Claude Code provider:
  - provider emitted a current `allowed` rate-limit state
  - a real Claude turn completed successfully
  - no false provider error row or retry entry was created
  - plugin reload retained no timer, as intended
  - no server or daemon warnings/errors were produced

The Claude account's external limit had reset before final standalone QA, so the exact hard-limit path is covered deterministically by the real-message adapter and end-to-end server/plugin tests rather than claimed as a live rate-limited run.

> AGENT GENERATED: by GPT-5


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

plugins/provider-retry/src/service.ts
```diff
@@ -196,6 +316,7 @@ export class ProviderRetryService {
       retryAtMs = this.retryAt(candidate.resetsAtMs);
     }
 
+    await this.rememberWaitingThread(threadId);
```
SawyerHood: 🚨 `slopcop/review` — The reconcile operation can restore a cancelled retry.

`reconcileDirect()` reads state, then waits for storage before `upsert()`. A concurrent cancel, supersede, or delete can finish during either wait. The old operation then writes the durable marker and timer again. Please serialize all state changes for a thread, or reject stale writes with a generation value. Add a race test that cancels during reconciliation.

plugins/provider-retry/src/service.ts
```diff
@@ -358,13 +485,26 @@ export class ProviderRetryService {
...
+      const resetsAtMs = entry.candidate.resetsAtMs;
+      if (resetsAtMs === null) {
+        this.remove(threadId);
+        await this.forgetWaitingThread(threadId);
         return false;
       }
       await this.bb.sdk.threads.continueAfterRateLimit({
         threadId,
         failedRequestId,
+        mode: "automatic",
```
SawyerHood, line 406: 🚨 `slopcop/review` — Automatic release does not verify current automatic eligibility.

`release()` only compares the failed request ID. A candidate can become manual-only while the timer waits. The server also ignores `mode` during both eligibility checks. Require `candidate.automatic === true` here and in both server checks.

packages/agent-runtime/src/claude-code/translate-message.ts
```diff
@@ -984,29 +1009,23 @@ export function translateClaudeSdkMessage(
...
+      const rateLimitsEvent: ThreadEvent = {
         type: "provider/rateLimits/updated",
         threadId,
         providerThreadId: "",
         scope: threadScope(),
         rateLimits: normalizeClaudeRateLimits(message),
-      });
+      };
       if (!isHardClaudeRateLimitRejection(message)) {
+        events.push(rateLimitsEvent);
```
SawyerHood, line 1020: 🚨 `slopcop/review` — A later allowed limit still leaves the turn failed.

This branch keeps `pendingHardRateLimitRejection`. A later successful result then becomes a failed turn. Clear the flag when a later allowed state supersedes the rejection. Add a reject, allowed, and success sequence test.

plugins/provider-retry/src/service.ts
```diff
@@ -132,10 +239,11 @@ export class ProviderRetryService {
     return entry === undefined ? null : toView(entry);
   }
 
-  cancel(threadId: string): boolean {
+  async cancel(threadId: string): Promise<boolean> {
     const entry = this.entries.get(threadId);
     if (entry === undefined || entry.releasing) return false;
     this.remove(threadId);
+    await this.forgetWaitingThread(threadId);
```
SawyerHood: 🚨 `slopcop/review` — Cancellation reports success after storage cleanup fails.

`forgetWaitingThread()` catches the delete error, so `cancel()` still returns `true`. A reload can then restore the cancelled retry. Propagate the failure or write a durable cancellation record. Add a storage-failure test.

plugins/provider-retry/src/service.ts
```diff
@@ -58,11 +70,105 @@ function toView(entry: WaitingEntry): ProviderRetryView | null {
...
+      this.bb.log.warn(
+        `Could not restore pending provider retries: ${errorMessage(error)}`,
+      );
+      return;
+    }
+    for (const key of keys) {
+      const threadId = key.slice(WAITING_THREAD_KEY_PREFIX.length);
+      if (threadId.length === 0) continue;
+      try {
+        await this.reconcile(threadId);
```
SawyerHood: 🚨 `slopcop/review` — Restore can block plugin activation.

The plugin loads every stored key and awaits each server request in sequence. A large store can exceed the plugin activation limit. Move restore work to a bounded background process. Add a batch limit and an abort check.

plugins/provider-retry/src/service.ts
```diff
@@ -58,11 +70,105 @@ function toView(entry: WaitingEntry): ProviderRetryView | null {
...
+      } catch (error) {
+        this.bb.log.warn(
+          `Could not restore provider retry for ${threadId}: ${errorMessage(error)}`,
```
SawyerHood: 🚨 `slopcop/review` — Deleted threads can leave permanent restore keys.

A deleted thread returns an error here, but this catch keeps its waiting marker. Each activation then repeats the request and warning. Remove the marker after a confirmed missing-thread response. Also clean old attempted-window keys.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: Claude tells bb when a usage limit blocks work. This change remembers the wait and restarts work after the reset.

I found these issues:

1. **High:** A reconciliation can recreate a retry after a user cancels it.
2. **High:** Automatic release does not confirm that the current candidate still permits automatic recovery.
3. **High:** A later allowed limit does not clear the pending hard rejection.
4. **High:** Cancellation reports success when durable storage cleanup fails.
5. **Medium:** Deleted threads can leave permanent restore keys.
6. **Medium:** Restore reads every key in sequence during plugin activation.
7. **Low:** The built-in guide templates still describe memory-only timers and lost state after reload.

I left inline comments for each code issue. Please also update both guide templates and regenerate `templates.generated.ts`.

I found no duplicate retry implementation. The main architecture risk comes from state changes across plugin waits and server policy checks.

The focused exact-head suite passed 228 tests. Five affected package typechecks passed. All current GitHub checks passed.

Browser E2E does not fit this change because it adds no browser route. The review exercised the nearest CLI and server API paths.

plugins/provider-retry/src/service.ts
ymichael: Fixed in 007c8fd1e. Reconciliation, cancellation, supersession, deletion, and timer release now share one per-thread operation queue. The new race test pauses reconciliation, queues cancellation, resumes reconciliation, and verifies that no entry or timer survives.

plugins/provider-retry/src/service.ts
ymichael: Fixed in 007c8fd1e. Release now requires the fresh candidate to match the failed request, remain automatic, and retain a reset time. Both server-side eligibility checks also reject automatic mode for a manual-only candidate. Tests cover the candidate changing to manual-only before release and the server returning 409.

packages/agent-runtime/src/claude-code/translate-message.ts
ymichael: Fixed in 007c8fd1e. A later allowed update for the active turn clears the pending hard rejection. The added rejected → allowed → successful-result test verifies a completed turn with no provider error.

plugins/provider-retry/src/service.ts
ymichael: Resolved by simplifying the design in 007c8fd1e: pending retries and attempted windows are no longer written to KV. Cancellation removes the in-process entry under the same per-thread operation queue, so there is no durable marker that can resurrect it after reload.

plugins/provider-retry/src/service.ts
ymichael: Resolved by simplifying the design in 007c8fd1e: there are no durable waiting or attempted-window keys anymore. Thread deletion removes the in-process timer and attempted-window guard under the per-thread queue, leaving nothing to scan on a later activation.

plugins/provider-retry/src/service.ts
ymichael: Resolved by simplifying the design in 007c8fd1e: restore and the KV key scan were removed. Timers are intentionally process-local, so plugin activation performs no retry restoration work.

PR #1410, opened 2026-08-12
Polish active child thread banner

ymichael:
## Summary

- align the active child-thread banner with background activity cards by showing the primary child title and a `+N more` count
- use the thread mention `UserRound` icon with the shared active icon shimmer while keeping the banner text static
- retain the expandable child list and update story/test coverage for single and multiple running children

## Validation

- `pnpm exec turbo run test --filter=@bb/app -- ThreadPromptContextBanner`
- visual QA in the Context banner Ladle story at desktop and compact widths

> AGENT GENERATED: by GPT-5

### DIFF
--- a/apps/app/src/components/promptbox/banner/ThreadPromptContextBanner.tsx
+++ b/apps/app/src/components/promptbox/banner/ThreadPromptContextBanner.tsx
@@ -20,7 +20,6 @@ import {
 import {
   activityIconClass,
   activityRowClass,
-  activityTextClass,
 } from "@/components/ui/activity-row-styles";
 import { WorkspaceChangesList } from "@/components/thread/WorkspaceChangesList";
 import {
@@ -250,7 +249,6 @@ function ChildThreadIcon({ className }: { className?: string }) {
 }
 
 interface SectionToggleButtonProps {
-  active?: boolean;
   id: string;
   controlsId: string;
   ariaLabel?: string;
@@ -263,7 +261,6 @@ interface SectionToggleButtonProps {
 }
 
 function SectionToggleButton({
-  active = false,
   id,
   controlsId,
   ariaLabel,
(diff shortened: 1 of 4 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull request for security, code quality, performance, architecture, and the user flow.

apps/app/src/components/promptbox/banner/ThreadPromptContextBanner.tsx
```diff
@@ -721,28 +707,32 @@ function ActiveChildThreadsCard({
...
-          <span
-            className={activityTextClass(
-              "active",
-              "min-w-0 flex-1 truncate text-left",
-            )}
-          >
-            {label}
+          <span className="min-w-0 flex-1 truncate text-left">
+            <span className="text-muted-foreground">
+              Running child thread:{" "}
```
SawyerHood: 🚨 `slopcop/review` — This copy calls each active child “running,” but the active set includes `waiting-for-host`.

`isRunningThreadRuntimeDisplayStatus()` returns false for that state. Use neutral active copy, or pass the status and show exact copy. Add a test for `waiting-for-host`.

apps/app/src/components/promptbox/banner/ThreadPromptContextBanner.tsx
```diff
@@ -721,28 +707,32 @@ function ActiveChildThreadsCard({
...
           aria-expanded={isExpanded}
           aria-controls={SECTION_IDS.childThreads.body}
-          aria-label={label}
+          aria-label={`${groupLabel}: ${primary.title}`}
           onClick={onToggle}
           className={CHILD_THREADS_HEADER_BUTTON_CLASS}
         >
           <Icon
-            name="CircleDashed"
+            name="UserRound"
```
SawyerHood, line 715: 🚨 `slopcop/review` — The icon catalog still describes `UserRound` as a parent-thread icon.

Update the `icon.stories.tsx` usage entry because this line now uses the icon for a child-thread banner.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: This change makes the child-work banner quieter and shows the first child title. The list still opens to show all children.

I found two issues:

- P2: The banner says every active child is running. The active set includes `waiting-for-host`, which the runtime model says is not running. Use neutral active copy or status-specific copy. Add a test for this state.
- P3: The icon catalog still describes `UserRound` as a parent-thread icon. Update that description for its new child-banner use.

Security review found no issue. Performance review found no regression and confirmed less animation work.

The new header resembles `ThreadBackgroundCommandsCard`. Existing shared activity styles cover the common behavior, so this change does not require a new abstraction.

Browser QA covered the desktop and mobile story fixtures. The toggle updated `aria-expanded`, opened the region, and showed all child links. Only the small status icon used the shine animation.

The focused 15 tests passed. The app typecheck and `git diff --check` also passed. I posted this review as a comment only.

apps/app/src/components/promptbox/banner/ThreadPromptContextBanner.tsx
ymichael: Fixed in 1a907a20e. The banner now uses neutral “Active child thread” copy, and the regression test explicitly covers `waiting-for-host` as banner-active without calling it running.

apps/app/src/components/promptbox/banner/ThreadPromptContextBanner.tsx
ymichael: Fixed in 1a907a20e. The `UserRound` catalog entry now includes thread mentions, parent-thread indicators, and the active child-thread banner.

PR #1416, opened 2026-08-12
Fix lowercase Pi tool call rendering

ymichael:
## Summary

- Recognize Pi lowercase `read`, `grep`, and `glob` calls in structured timeline projection.
- Preserve the title-case aliases emitted by Claude.
- Add end-to-end timeline coverage for semantic activity intents and humanized labels.

## Verification

- `pnpm exec turbo run test --filter=@bb/thread-view --force` — 358 tests passed.
- `pnpm exec turbo run typecheck --filter=@bb/thread-view` — passed.
- Live Pi smoke thread emitted lowercase `read` and rendered as `Read package.json` in the dev UI.

> AGENT GENERATED: by GPT-5

### DIFF
--- a/packages/thread-view/src/tool-call-parsing.ts
+++ b/packages/thread-view/src/tool-call-parsing.ts
@@ -7,9 +7,13 @@ const DELEGATION_TOOL_NAMES = new Set([
   "spawnAgent",
   "resumeAgent",
 ]);
-const STRUCTURED_READ_TOOL_NAMES = new Set(["Read"]);
-const STRUCTURED_SEARCH_TOOL_NAMES = new Set(["Grep"]);
-const STRUCTURED_LIST_TOOL_NAMES = new Set(["Glob"]);
+// Claude names these built-ins with title case while Pi uses lowercase names.
+// Keep this as an explicit alias list instead of normalizing every tool name:
+// plugin-contributed tool names are case-sensitive and may not share the
+// built-ins' semantics.
+const STRUCTURED_READ_TOOL_NAMES = new Set(["Read", "read"]);
+const STRUCTURED_SEARCH_TOOL_NAMES = new Set(["Grep", "grep"]);
+const STRUCTURED_LIST_TOOL_NAMES = new Set(["Glob", "glob"]);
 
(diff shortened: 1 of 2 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am reviewing this pull request against the security, quality, performance, architecture, and test rules.

packages/thread-view/src/tool-call-parsing.ts
```diff
@@ -7,9 +7,13 @@ const DELEGATION_TOOL_NAMES = new Set([
...
   "resumeAgent",
 ]);
-const STRUCTURED_READ_TOOL_NAMES = new Set(["Read"]);
-const STRUCTURED_SEARCH_TOOL_NAMES = new Set(["Grep"]);
-const STRUCTURED_LIST_TOOL_NAMES = new Set(["Glob"]);
+// Claude names these built-ins with title case while Pi uses lowercase names.
+// Keep this as an explicit alias list instead of normalizing every tool name:
+// plugin-contributed tool names are case-sensitive and may not share the
+// built-ins' semantics.
+const STRUCTURED_READ_TOOL_NAMES = new Set(["Read", "read"]);
```
SawyerHood, line 14: 🚨 `slopcop/review` — Lowercase aliases can mislabel plugin and MCP tools.

The parser treats every matching base name as a native tool. Plugins may register these names. The parser also removes server prefixes before this test. A plugin `read` call can therefore appear as a file read. Scope these aliases to native provider tools, or reserve the names. Add negative tests for plugin and MCP tools.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** Pi uses small letters for tool names. This change makes three Pi file tools show clear timeline labels.

I found one P2 correctness issue. The new aliases also match plugin and MCP tools with the same base names. Those tools can then receive false file-read, search, or list labels. The inline comment gives the required fix and test cases.

Security checks found no code execution, access control, injection, or data access issue. Performance checks found no material cost from the three constant-size sets. The code search found no duplicate classifier or useful refactor.

The `@bb/thread-view` type check passed. All 358 package tests passed. The pull request app loaded in a browser without console errors. A Pi feature test could not start because the test app had no Pi model catalog.

I am leaving a comment-only review, as required.

PR #1418, opened 2026-08-12
Keep a foreign DOM mutation from blanking the app

SawyerHood:
A user reported a white screen. Their console held repeated `NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.`

## Root cause

The stack decodes exactly against a local production build — the columns match, so this is not a guess:

| Frame | Decodes to |
|---|---|
| `dist-<hash>.js` | the react-dom chunk |
| col `98077` | `hostParent.removeChild(node)` in `commitDeletionEffectsOnFiber`, the **plain-parent** branch (not the portal-container one) |
| col `97489` | the recursive call in `recursivelyTraverseDeletionEffects` |
| col `98395` | the branch for function, forwardRef, memo, and lazy components |

React deleted a subtree and the parent element no longer held the node.

Our own code never moves React-owned nodes. A scan across every package and plugin finds DOM mutation only on `document.head`, `document.body`, or elements the code created itself. cmdk (used by the Tasks plugin) reorders within one parent, and `removeChild` throws only on a parent mismatch, so it cannot produce this either. React 19 does not produce it from duplicate keys — it leaks a node instead. That leaves an agent outside React, and the report's console shows an extension content script live in the page (`ObjectMultiplex`, Node `EventEmitter` warnings).

The blank window is ours, though: bb had no error boundary above the router, so React had nowhere to send the error and unmounted the whole root.

## What changed

- **`lib/foreign-dom-mutation-guard.ts`** wraps `Node.prototype.removeChild` and `insertBefore`. Each wrapper changes behavior only in the case that would otherwise throw, so every well-formed call reaches the native method unchanged. It counts suppressed calls and logs the first three, so a burst with no extension in play points at a bug of ours. This is the workaround React's maintainers publish for translated pages ([facebook/react#11538](https://github.com/facebook/react/issues/11538)).
- **`components/AppErrorBoundary.tsx`** is the last boundary above the router. It shows a message, the stack behind a details toggle, and a reload button. It pulls in no shared-ui component, no query client, and no router hook, so whatever crashed cannot take the recovery screen down too. Colors come from `--canvas`/`--ink`-derived tokens, so custom palettes re-tint it.
- **`main.tsx`** installs both, raises `Error.stackTraceLimit` to 50, and logs the component stack from `onUncaughtError`. V8's default of 10 frames was entirely reconciler internals, which is why the original report named no bb component.

## Validation

- New tests reproduce the exact failure: mount a node, reparent it the way an extension does, then unmount. The guard test asserts React escalates nothing. **The boundary test runs the same reproduction with the guard switched off** and asserts the recovery screen renders — so the white screen is closed even if my reading of the cause is wrong.
- Rendered the boundary in a real browser, light and dark, against the built CSS.
- `@bb/app`: 338 test files, 2552 tests pass. Typecheck passes. Lint reports 0 errors. Production build passes.

## Notes and risks

- Patching `Node.prototype` is global. Both wrappers are no-ops for well-formed calls, and `removeChild`/`insertBefore` are the only two patched.
- Not included, as a product call for a reviewer: `translate="no"` / `<meta name="google" content="notranslate">`. The browser's page translator is the other common cause of this error, and the opt-out is standard for developer tools, but it removes UI translation for non-English readers. The guard stops the crash either way.

### DIFF
--- a/apps/app/src/lib/foreign-dom-mutation-guard.ts
+++ b/apps/app/src/lib/foreign-dom-mutation-guard.ts
@@ -1,0 +1,134 @@
+/**
+ * Keeps foreign DOM mutation from killing the app.
+ *
+ * React owns every node it renders and remembers which parent each one belongs
+ * to. On unmount it calls `parent.removeChild(node)` directly. When another
+ * agent in the page — a browser extension's content script, the browser's own
+ * page translator, a bookmarklet — moves or removes one of those nodes first,
+ * React's call throws:
+ *
+ *   NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be
+ *   removed is not a child of this node.
+ *
+ * React catches that throw in the commit phase and escalates it to the nearest
+ * error boundary. bb has no boundary above the router, so React tears the whole
+ * root down and the user gets a blank page — a total loss for a mutation that
+ * only ever affected one subtree.
+ *
+ * This guard makes both calls non-fatal. `removeChild` returns the node instead
+ * of throwing when it already has a different parent (the node is gone from
+ * where React expected it, which is exactly the outcome React wanted).
+ * `insertBefore` appends instead of throwing when the reference node has moved
+ * away, so the content still reaches the DOM rather than vanishing.
+ *
+ * This is the workaround React's own maintainers publish for translated pages
+ * (facebook/react#11538). It is deliberately narrow: both wrappers only change
+ * behavior in the case that would otherwise throw, so every well-formed call
+ * reaches the native method unchanged.
+ *
+ * Suppressed calls are counted and the first few are logged, because a burst of
+ * them from a build where no extension is involved would point at a real bug in
+ * our own rendering instead.
+ */
+
+/** Log at most this many suppressions, then only keep counting. */
+const MAX_LOGGED_SUPPRESSIONS = 3;
+
+type RemoveChild = <T extends Node>(child: T) => T;
+type InsertBefore = <T extends Node>(node: T, child: Node | null) => T;
+
+interface InstalledGuard {
+  removeChild: RemoveChild;
+  insertBefore: InsertBefore;
+}
+
+let installed: InstalledGuard | null = null;
+let suppressedCount = 0;
+let loggedCount = 0;
+
+/** How many throwing DOM calls the guard has absorbed this session. */
+export function foreignDomMutationCount(): number {
+  return suppressedCount;
+}
+
+function describe(node: Node): string {
+  if (node instanceof Element) {
+    const id = node.id ? `#${node.id}` : "";
+    const testId = node.getAttribute("data-testid");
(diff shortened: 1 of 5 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull request for security, code quality, architecture, performance, and real product behavior.

apps/app/src/lib/foreign-dom-mutation-guard.ts
```diff
@@ -0,0 +1,134 @@
...
+ * to. On unmount it calls `parent.removeChild(node)` directly. When another
+ * agent in the page — a browser extension's content script, the browser's own
+ * page translator, a bookmarklet — moves or removes one of those nodes first,
+ * React's call throws:
+ *
+ *   NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be
+ *   removed is not a child of this node.
+ *
+ * React catches that throw in the commit phase and escalates it to the nearest
+ * error boundary. bb has no boundary above the router, so React tears the whole
```
SawyerHood: 🚨 `slopcop/review` — The comment describes the old tree in the present tense.

This pull request adds `AppErrorBoundary` above the router. Please use past tense here, so the text matches the final tree.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** A browser tool can move a page part. This change catches the problem, like a safety net catches a toy.

I found no security, correctness, architecture, or performance defects. I found no blocker.

**Review**

- Security: The guard adds no new rights. React escapes the stack text on the recovery screen.
- Quality: The guard handles the known DOM fault. The root boundary handles other render and commit faults.
- Architecture: No other `Node.prototype` guard exists. The plugin error boundaries have smaller scopes and cannot replace this boundary.
- Performance: A live test measured about 0.22 microseconds of extra cost per remove-and-insert pair. This cost is acceptable.
- E2E: The app booted and stayed visible. Manual guard checks passed for foreign and valid DOM operations.

The live React fault injection did not complete because its test node did not render. The focused unit tests reproduce that exact commit fault.

**Validation**

- Six focused tests passed.
- The app type check passed.
- The app production build passed.
- Lint finished with zero errors. Existing files produced 148 warnings.
- All GitHub checks passed.

I left one non-blocking inline note. A module comment still describes the old tree in the present tense.

The pull request text also overstates custom palette behavior. The code uses existing theme tokens and adds no color literal.

apps/app/src/lib/foreign-dom-mutation-guard.ts
SawyerHood: Fixed in f1aca7f76. The paragraph now reads in past tense, states that `AppErrorBoundary` catches it, and says why the guard still earns its place above the boundary: a recovery screen still costs the user their place.

PR #1420, opened 2026-08-12
fix(pi): allow disabling thinking for models that support off

fgrehm:
Pi models that advertise a thinking-off level (for example, models with `off` in their `thinkingLevelMap`, plus non-reasoning models) can now be set to "Reasoning: None" in the picker. Three gaps previously dropped the level:

- `getPiReasoningEfforts` never emitted `none`, only `low` through `max`.
- The Pi bridge schema excluded `off`, with no `none` → `off` translation.
- Server-side plan validation used a coarse Pi provider ladder that lacked `none`.

The picker now exposes `none` when Pi reports `off`, the adapter maps `none` to `off` on every session launch path (start, resume, and fork), and the server accepts the selection. Models without `off`, such as `minimax-m2.7`, remain unchanged. Non-reasoning models now surface a truthful single "Reasoning: None" entry instead of a misleading `low`.

Sparse reasoning ladders preserve their previous default: `medium` is preferred, then the first non-`none` level. For example, a model reporting `[off, high]` still defaults to `high` while offering `none` as an explicit choice.

Server-defined custom Pi models inherit the coarse ladder, including `none`, and continue to default to `medium`.

## Protocol compatibility

This changes the `supportedReasoningEfforts` reported by the host daemon, so `HOST_DAEMON_PROTOCOL_VERSION` is bumped from 113 to 114.

- New server + old daemon: benign feature omission; the old daemon does not report `none`, so the picker does not offer it.
- New daemon + old server: the old server can parse `none` in the model list, but rejects a turn using it against its old Pi provider ladder.

The version bump prevents the broken second pairing and triggers the normal enrolled-daemon update path.

## Persisted preferences

No persisted-preference migration is needed. Models that genuinely support `low` retain it. Non-reasoning models previously advertised `low`, but Pi already clamped that selection to `off`; reconciliation now displays `none` without changing execution behavior. A provider-wide `low` → `none` migration would incorrectly alter preferences for other Pi models.

## Validation

Post-rebase validation on `origin/main`:

- Typechecks: `@bb/agent-providers`, `@bb/agent-runtime`, `@bb/domain`, `@bb/server`, `@bb/host-daemon-contract`, and `@bb/host-daemon`.
- Focused regression tests:
  - Pi runtime and bridge: 101
  - server execution options and runtime validation: 58
  - provider catalog: 6
  - persisted reasoning reconciliation: 11
  - host-daemon contract: 35
- `git diff --check`

Original implementation by @fgrehm; compatibility hardening and regression coverage added during maintainer review.

> AGENT GENERATED: by glm-5.2 (ollama-cloud)

### DIFF
--- a/packages/agent-providers/src/catalog.ts
+++ b/packages/agent-providers/src/catalog.ts
@@ -182,7 +182,12 @@ const PI_SERVER_CAPABILITIES: ProviderServerCapabilities = {
   supportsWorkflows: false,
   supportsExecutionOverride: false,
   backsHostDaemonAiServices: false,
-  reasoningLevels: ["low", "medium", "high", "xhigh", "max"],
+  // "none" is the thinking-off level: some Pi models (e.g. Ollama Cloud models
+  // whose `thinkingLevelMap` advertises `off`, and non-reasoning models) expose
+  // it per-model via `supportedReasoningEfforts`. The picker only offers it for
+  // those models, but this coarse ladder must include it so the plan-path
+  // validation accepts a `none` selection the picker legitimately offered.
+  reasoningLevels: ["none", "low", "medium", "high", "xhigh", "max"],
 };
 
 // ACP agents manage reasoning effort internally; "medium" is the single
--- a/packages/agent-providers/src/catalog.ts
+++ b/packages/agent-providers/src/catalog.ts
@@ -184,9 +184,11 @@ const PI_SERVER_CAPABILITIES: ProviderServerCapabilities = {
   backsHostDaemonAiServices: false,
   // "none" is the thinking-off level: some Pi models (e.g. Ollama Cloud models
   // whose `thinkingLevelMap` advertises `off`, and non-reasoning models) expose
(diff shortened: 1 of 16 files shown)


packages/agent-providers/src/catalog.ts
```diff
@@ -182,7 +182,12 @@ const PI_SERVER_CAPABILITIES: ProviderServerCapabilities = {
   supportsWorkflows: false,
   supportsExecutionOverride: false,
   backsHostDaemonAiServices: false,
-  reasoningLevels: ["low", "medium", "high", "xhigh", "max"],
+  // "none" is the thinking-off level: some Pi models (e.g. Ollama Cloud models
+  // whose `thinkingLevelMap` advertises `off`, and non-reasoning models) expose
+  // it per-model via `supportedReasoningEfforts`. The picker only offers it for
+  // those models, but this coarse ladder must include it so the plan-path
+  // validation accepts a `none` selection the picker legitimately offered.
```
fgrehm: Sorry but looks like we got some comment noise here, happy to trim these down before you folks merge. just LMK!

PR #1421, opened 2026-08-12
Prevent bleed-through in sticky open-in-split thread rows

ymichael:
## Summary

- resolve the unfocused open-in-split thread tint against the sidebar so it is opaque
- preserve the stronger hover and menu-open treatment
- add regression coverage for sticky rendering of that related thread-row state

## Scope clarification

This is a follow-up issue found while auditing the reported no-split bleed-through; it is not the root cause of that incident. The selected-parent path shown in the original report was fixed by #1183. Git ancestry confirms that fix is included in desktop 0.37.0 but not 0.36.0, so a user who has not updated can still reproduce the old bug without using splits.

The current no-split paths were rechecked with a seeded two-level hierarchy: selected parents, ordinary parents, pinned trees, project grouping, manual grouping, and machine grouping all compute to opacity 1 with an opaque backing surface while descendant rows overlap them. Project, section, and worktree headers also use opaque fills; leaf thread rows do not become sticky.

The audit did find one adjacent gap: an unfocused thread open in a split uses a separate 50% accent tint. If that thread is also a sticky parent, descendants can show through it. This PR closes that gap.

## Validation

- pnpm exec turbo run test --filter=@bb/app -- src/components/sidebar/ThreadRow.test.tsx src/components/sidebar/sidebarRowClasses.test.ts src/components/ui/theme.test.ts (94 tests)
- pnpm exec turbo run lint --filter=@bb/app (0 errors; existing warnings only)
- seeded an isolated dev database with 80 threads and a two-level parent/child hierarchy
- verified no-split sticky overlap in project, manual, and machine organization modes, including selected, ordinary, and pinned parent rows
- verified the related sticky open-in-split tint stays opaque and the stronger hover tint still wins

> AGENT GENERATED: by GPT-5

### DIFF
--- a/apps/app/src/components/sidebar/ThreadRow.test.tsx
+++ b/apps/app/src/components/sidebar/ThreadRow.test.tsx
@@ -10,6 +10,7 @@ import { afterEach, describe, expect, it, vi } from "vitest";
 import { ThreadRow, type ThreadRowOptions } from "./ThreadRow";
 import { SidebarThreadTitleMentionResourcesProvider } from "./SidebarThreadTitleMentions";
 import {
+  SIDEBAR_ROW_OPEN_IN_SPLIT_STATE_CLASS,
   SIDEBAR_SUCCESS_STATUS_COLOR_CLASS,
   SIDEBAR_WORKING_STATUS_COLOR_CLASS,
 } from "./sidebarRowClasses";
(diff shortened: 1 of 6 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the Slop Cop. I am reviewing this pull request now.

I will check security, code quality, performance, architecture, and end-to-end behavior. Three BB reviewers will inspect the main areas in parallel.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: This change puts a solid cover on a sticky thread row. It works like a card that hides paper beneath it.

I found no issue that needs a change.

Security: The change uses fixed CSS and theme values. It adds no input, network, data, or access-control path.

Code quality: The state class reaches the sticky row directly. The tests cover the component class and the CSS paint rules.

Architecture: A project-wide search found no duplicate production implementation. The selected-row state already uses the same solid background-image method.

Performance: The change adds static color work and scoped selectors. It adds no React work, layout work, filter, animation, or broad selector.

Validation:

- All 94 focused tests passed.
- The `@bb/app` type check passed.
- The pull request diff format check passed.
- The browser test used two split panes and a sticky parent with scrolling children.
- A child moved behind the parent row during the test.
- The parent kept opacity `1` and a solid `linear-gradient` background.
- Hit testing selected the parent above the overlapping child.

PR #1427, opened 2026-08-12
Fix false active ACP turns and rejected steers

SawyerHood:
## Summary

- prevent late ACP updates from opening synthetic turns
- map the ACP no-active-turn response to stale state and preserve the user send as a new turn
- store rejected turn requests and show failed steers as terminal timeline rows
- close legacy failed steers through their stored command failure event

## Tests

- `pnpm exec turbo run typecheck --filter=@bb/agent-runtime --filter=@bb/thread-view --filter=@bb/server --filter=@bb/app`
- `pnpm exec turbo run test --filter=@bb/agent-runtime --force`
- `pnpm exec turbo run test --filter=@bb/thread-view --force`
- `pnpm exec turbo run test --filter=@bb/domain --force`
- `pnpm exec turbo run test --filter=@bb/server --force -- test/threads/thread-send-dispatch.test.ts`

## Notes

This change does not alter the server and host daemon wire contract.

> AGENT GENERATED: by GPT-5


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the Slop Cop. I am now reviewing this pull request for security, quality, performance, architecture, and end-to-end behavior.

packages/thread-view/src/build-thread-timeline.ts
```diff
@@ -828,14 +833,78 @@ function buildPendingSteerRowsFromEvents(
...
+    if (
+      event.type === "client/turn/requested" &&
+      (event.target.kind === "auto" || event.target.kind === "steer") &&
+      event.target.expectedTurnId !== null
+    ) {
+      unresolvedSteerRequestIds.push(event.requestId);
+      continue;
+    }
+    if (event.type === "turn/input/accepted") {
+      const index = unresolvedSteerRequestIds.indexOf(event.clientRequestId);
```
SawyerHood: 🚨 `slopcop/review` — Replace the quadratic request lookup.

Each accepted or rejected request uses `indexOf` and `splice` on the unresolved array. The work grows with the square of the request count.

A 40,000-request check took 1.54 seconds. Use a Set with an ordered queue or stack, and add a scale test.

packages/domain/src/thread-events.ts
```diff
@@ -16,6 +16,7 @@ import { clientTurnRequestIdSchema } from "./protocol-ids.js";
 export const systemEventTypeValues = [
   "client/thread/start",
   "client/turn/requested",
+  "client/turn/rejected",
```
SawyerHood, line 19: 🚨 `slopcop/review` — Regenerate the shipped SDK declarations.

This event changes the realtime event type. The plugin SDK typecheck fails because its committed declaration is stale.

Regenerate `packages/plugin-sdk/bundled-types` and the template copy. Run both generator checks.

apps/server/src/services/threads/thread-lifecycle.ts
```diff
@@ -747,6 +747,19 @@ function settleThreadCommandFailure(
   if (hasExpectedTurnCompletedEvent(args.deps, args.command)) {
```
SawyerHood: 🚨 `slopcop/review` — Record the failed request before this early return.

When the target turn has completed, this return skips the new rejection event. The failed message then has no accepted or rejected terminal event.

Move request settlement before this turn-lifecycle guard. Add a test where the target turn completes before a failed `turn.submit` report.

packages/thread-view/src/build-thread-timeline.ts
```diff
@@ -828,14 +833,78 @@ function buildPendingSteerRowsFromEvents(
     context: acceptedClientRequestContext,
     events: orderedEvents,
   });
+  const rejectedClientRequestIds = new Set(
```
SawyerHood: 🚨 `slopcop/review` — Carry rejection state across timeline page limits.

This set sees only the selected event page. The server fetches extra request context for accepted steers, but not for rejected steers.

A page can therefore hide a rejected steer or show it as pending. Extend the request context with rejection state and add a page-boundary test.

packages/agent-runtime/src/runtime.ts
```diff
@@ -1941,16 +1943,27 @@ function createAgentRuntimeInternal(
...
+                providerThreadId: adapterCommand.providerThreadId,
+                threadId,
+              },
+            });
+          } catch (error) {
+            if (
+              error instanceof JsonRpcResponseError &&
+              error.code === ACP_BRIDGE_NO_ACTIVE_TURN_ERROR_CODE
+            ) {
+              turnState.clearThread(threadId);
```
SawyerHood, line 1964: 🚨 `slopcop/review` — Clear the ACP adapter turn state before the fallback starts.

This branch clears only the runtime turn tracker. The ACP adapter still holds the old turn identifier.

The fallback can then assign new input to the old turn without a second `turn/started` event. Add an adapter reset hook and a fallback-turn test.

apps/server/src/services/threads/thread-lifecycle.ts
```diff
@@ -747,6 +747,19 @@ function settleThreadCommandFailure(
...
     return emptyCommandResultSideEffects();
   }
+  if (args.command.type === "turn.submit") {
+    appendThreadEventInTransaction(args.deps.db, {
+      threadId: thread.id,
+      environmentId: thread.environmentId,
+      type: "client/turn/rejected",
```
SawyerHood, line 787: 🚨 `slopcop/review` — Do not reject an accepted request.

A provider can emit `turn/started` before its command response. That event can accept the request before a later response timeout reaches this branch.

Check the request identifier for an accepted input before appending the rejection. Test a started turn whose command response never arrives.

packages/agent-runtime/src/runtime.ts
```diff
@@ -1941,16 +1943,27 @@ function createAgentRuntimeInternal(
...
+              recovery: {
+                providerId: pid,
+                providerThreadId: adapterCommand.providerThreadId,
+                threadId,
+              },
+            });
+          } catch (error) {
+            if (
+              error instanceof JsonRpcResponseError &&
+              error.code === ACP_BRIDGE_NO_ACTIVE_TURN_ERROR_CODE
```
SawyerHood, line 1962: 🚨 `slopcop/review` — Limit this error code to ACP.

JSON-RPC code `-32001` belongs to the server-specific range. Another bridge can use it for an unrelated failure.

The generic runtime can then clear valid turn state and resend input. Add an adapter-specific classifier and a non-ACP test.

packages/thread-view/src/build-thread-timeline.ts
```diff
@@ -828,14 +833,78 @@ function buildPendingSteerRowsFromEvents(
...
+    if (event.type === "client/turn/rejected") {
+      const index = unresolvedSteerRequestIds.indexOf(event.requestId);
+      if (index >= 0) unresolvedSteerRequestIds.splice(index, 1);
+      continue;
+    }
+    if (
+      event.type === "system/error" &&
+      event.code === "thread_command_failed"
+    ) {
+      const requestId = unresolvedSteerRequestIds.pop();
```
SawyerHood: 🚨 `slopcop/review` — Do not consume the next pending steer.

The settlement appends a rejection and then a `thread_command_failed` error. This scan removes the rejected request, then pops a different pending request.

I reproduced this with two steer requests. One failed request produced two rejected rows.

Skip the companion error after an explicit rejection. Add a two-request regression test.

packages/agent-runtime/src/acp/bridge-protocol.ts
```diff
@@ -213,6 +213,9 @@ export const ACP_UPDATE_METHOD = "acp/update";
 export const ACP_FS_WRITE_METHOD = "acp/fs/write";
 export const ACP_WARNING_METHOD = "acp/warning";
 
+/** The bridge has a session, but its agent prompt has already ended. */
+export const ACP_BRIDGE_NO_ACTIVE_TURN_ERROR_CODE = -32001;
```
SawyerHood, line 217: 🚨 `slopcop/review` — Increment the host daemon protocol version.

This change alters a `turn.submit` result across the server and daemon boundary. A version-110 daemon can keep the old false-active behavior.

Set `HOST_DAEMON_PROTOCOL_VERSION` to 111. Update the contract test.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

BB can think a helper still works after it stops. This change adds a stop sign and protects the next note.

However, one broken note can still mark another note as broken. It is like putting the wrong name on two toy boxes.

### Review

I found nine issues. I recommend fixes before merge.

High findings:

- The ACP fallback clears the runtime turn tracker, but it leaves the adapter tracker on the old turn.
- A completed target turn can stop the server before it records a failed request.
- One explicit rejection and its error can also reject the next pending steer.
- The host behavior changes without a host daemon protocol version increase.
- The committed plugin SDK declaration and its template copy remain stale.

Medium findings:

- The server can reject a request that already has an acceptance event.
- Rejection state does not cross timeline page limits.
- The generic runtime treats JSON-RPC code `-32001` as an ACP result for every adapter.
- Repeated array searches make unresolved-request work grow with the square of the request count.

The request-state code now exists in two projection paths. A shared request-state index can remove this duplication and address the page and performance issues.

I found no direct data-access or permission issue. The generic error handling can still clear valid state and resend input.

### Checks

- I inspected all 25 changed files and the related runtime, server, timeline, contract, and SDK code.
- `git diff --check` passed.
- The selected agent-runtime, thread-view, and server tests passed.
- The plugin SDK declaration check failed because the committed output is stale.
- GitHub CI is green.
- A local dev server passed browser checks for the home page, new-thread composer, and provider picker.
- The browser showed no console error.
- Direct reproductions confirmed the ACP state fault and both timeline faults.
- A full local server run had unrelated timeouts during heavy parallel work.

SawyerHood:
I addressed the SlopCop findings in 345ca0360.

- ACP stale recovery now clears both runtime and adapter turn state.
- The ACP error check now applies only to ACP providers.
- Failed requests now settle once, even after turn completion.
- Accepted requests cannot receive a later rejection.
- Legacy rejection matching is linear and does not consume the next steer.
- Rejection context now crosses timeline page limits.
- The host daemon protocol version is now 111.
- The generated SDK and template types were updated in 6c15a875d.

The full type check and all related regression suites pass.

PR #1432, opened 2026-08-12
Settle provider prompts that complete without starting a turn

### LINKED ISSUE #1431: claude-code: a `/clear` turn never settles — the thread stays `active` until some later turn completes
**bb 0.37.0** · provider `claude-code` · Claude Code CLI 2.1.226 · macOS 26.5.2.
Also seen on 0.36.0 and 0.35.1 / CLI 2.1.225. Source references are against
[`b5d89f3`](https://github.com/get-bb/bb/commit/b5d89f34141fc22c2dc3afaf4a003202b354531d).
The CLI probe below was run on 2.1.229.

## What I was trying to do

Send `/clear` to a thread and have the thread go back to idle afterwards, so a
script could clear a worker thread and carry on.

## What happens

The clear itself succeeds, but the thread never leaves `active`. It sits there
indefinitely, showing as working when nothing is running. The turn produces no
`turn/started` and no `turn/completed`:

```
11  turn/completed         <- the previous, ordinary turn
12  client/turn/requested  {"source":"tell","input":[{"type":"text","text":"/clear"}]}
13  provider/unhandled     {"rawType":"sdk/conversation_reset", …}
14  thread/identity        {"providerThreadId":"3886608c-…"}
                           <- nothing further; status stays "active"
```

It settles only when something *else* closes the turn. An ordinary later turn
does it. So does any unrelated interrupt, which is the part that makes this hard
to see: whether one arrives is chance.

## What I expected

`status` back to `idle` once the CLI reports the run finished — and it does
report it. Driving the CLI directly shows the whole sequence:

```sh
echo '{"type":"user","message":{"role":"user","content":"/clear"}}' \
  | claude --print --input-format stream-json --output-format stream-json --verbose
```

```
conversation_reset  {}
system              {"subtype":"init"}
result              {"subtype":"success","num_turns":0,"is_error":false,"result":""}
```

It is a zero-work run — the command resolves before any model call — but the
provider still closes it.

## Impact

`bb thread wait --status idle` never returns, so any script that clears a thread
and waits will hang. It exits 2 with `Error: Timed out waiting for thread <id> to
reach status idle.` if you bound it, and blocks for the default 1200s if you do
not. Anything keyed on the idle transition does not fire for these turns either.

It also costs an interactive user the whole thread, under a message that names
the wrong cause. One of my threads held a turn open for **14h 26m** after a
`/clear`, and the next event was not a settle:

```
2206  Aug 11 17:35:43  client/turn/requested       {"input":[{"text":"/clear"}]}
2207  Aug 11 17:35:44  provider/unhandled          {"rawType":"sdk/conversation_reset", …}
2208  Aug 11 17:35:44  thread/identity             {"providerThreadId":"c7ccfbd2-…"}
                                                   <- nothing for 14h 26m
2209  Aug 12 08:02:10  system/error                {"code":"thread_command_failed",
                                                    "message":"Thread interrupted because
                                                     the host daemon disconnected",
                                                    "detail":"Please retry the thread to continue."}
2210  Aug 12 08:02:10  system/thread/interrupted   {"reason":"host-daemon-restarted"}
```

Three things that adds:

- **The error names the wrong cause.** The daemon restart did not break the
  thread. It is what finally *closed* the turn the `/clear` left open 14 hours
  earlier, and it reported itself as the cause. A user reading "the host daemon
  disconnected" goes looking at their network or their machine sleeping. I only
  found the real cause by reading the event log.
- **A queued message never drains.** I typed `hi` five seconds after the clear.
  bb queued it correctly, behind a turn it believed was running. Queued messages
  drain on turn *completion*, and this turn ended as an interrupt, so the queue
  held. `bb thread queue delete` was the way out — findable only once you know
  the queue is what holds it.
- **It survived an upgrade.** That clear was on 0.36.0. bb replaced itself with
  0.37.0 six hours before the daemon restart and the open turn came through
  untouched.

The same thread settled an identical `/clear` four days earlier, but only by
luck: a trailing interrupt from a stop I had already requested happened to land
one second later. That thread has now taken three clears, across 0.35.1, 0.36.0
and 0.37.0, and settled none of them itself.

## Why it happens

That `result` reaches bb and is dropped.
`packages/agent-runtime/src/claude-code/translate-message.ts` gates the whole
`result` handler on there already being a turn to complete:

```ts
    case "result": {
      // …
      if (state.currentTurnId) {
        // …
        events.push({ type: "turn/completed", /* … */ });
        args.turnState.finishTurn({ state, threadId: stateKey });
      }
      break;
    }
```

`currentTurnId` is set by `ensureTurnStarted`, which the adapter calls when it
translates an assistant message, a stream event, a tool call, or a compaction
status. A `/clear` produces none of those, so the one message that would have
closed the turn takes the empty branch.

From there the domain behaves exactly as written. In
`packages/domain/src/thread-lifecycle.ts`, `active` has three exits:

```ts
  active: {
    "run.succeeded": "idle",
    "run.failed": "error",
    "stop.requested": "stopping",
  },
```

`turn/completed` is what becomes `run.succeeded`
(`apps/server/src/internal/turn-completed-events.ts`). None was emitted, so
there is no transition to take, and the table's own note applies: *"Absent cell
= the event is a no-op in that status."* Nothing reconciles it later either —
`system/provider-turn-watchdog` is marked *"Legacy persisted watchdog
diagnostic; retained for read/decode/render only, with no current producer"*
(`packages/domain/src/thread-events.ts`).

The same missing `onTurnStart` explains the queue symptom, since accepted input
is drained there. The thread above carries four `turn/input/accepted` events and
not one for either of its clears.

## Reproduction

```sh
bb thread spawn --project <id> --model claude-haiku-4-5-20251001 \
  --prompt "Reply with only READY"
bb thread wait <thread-id> --status idle

bb thread tell <thread-id> "/clear"
bb thread wait <thread-id> --status idle --timeout 60
# -> times out; `bb thread show <thread-id>` reports "active" indefinitely

bb thread tell <thread-id> "hello"       # any ordinary turn settles it
bb thread stop <thread-id>               # or this: same effect, no model call
```

I also have a scripted case that asserts the absence of `turn/started` and
`turn/completed` rather than just the status, and bounds its wait so it fails
fast against a bb that has fixed this. It lives in a plugin of mine that is not
public yet; happy to share it if that is useful.

## A fix

A PR follows this issue. It settles the turn the way the synthetic no-response
message already does one case earlier in the same file, gated on
`pendingAcceptedUserMessages` so that a result whose turn has already closed
still emits nothing. `@bb/agent-runtime` is green at 918 tests on `b5d89f3`, and
the new test fails without the change.

jerrison:
Fixes #1431.

The original diagnosis and narrow Claude Code fix are Jerrison's work. This
branch preserves that commit and credits Jerrison on the generalized change.

## Problem

Claude Code handles `/clear` locally and returns a successful zero-work result:
there is no assistant message, stream event, or tool call to start a bb turn.
The result handler previously required `state.currentTurnId`, so it emitted no
`turn/completed`. The thread remained `active`, `bb thread wait --status idle`
hung, and accepted input queued behind the abandoned turn could not drain.

The underlying lifecycle problem is broader than Claude's `/clear`: a provider
can accept a prompt and then report a terminal signal before emitting any of the
ordinary activity that starts a bb turn.

## Contract

Provider terminal translators now resolve the turn through one shared rule:

1. Complete the currently open turn when one exists.
2. Otherwise, start and complete a synthetic turn only while accepted input is
   still pending.
3. Emit nothing when neither condition holds.

The pending accepted-input queue is the ownership proof that keeps this narrow.
`onTurnStart` drains it, so a late terminal signal cannot open a second empty
turn after the real turn has already closed.

## Provider coverage

- **Claude Code:** a terminal `result` can settle accepted input even when no
  earlier SDK message started the turn. This covers the `/clear` regression.
- **ACP:** `turn/completed` follows the same terminal-turn rule.
- **Pi:** `agent_end` follows the same rule, and the bridge reports when
  `session.prompt()` settles without producing an SDK event.

Codex is unchanged because its native lifecycle already settles this shape.

The synthetic lifecycle also drives the existing server completion path, so a
queued message is sent after the zero-work turn completes instead of remaining
stuck.

## Deliberate non-goals

- Process-detach recovery remains separate and is covered by #1234.
- This does not add a general stuck-turn watchdog or reconciler.
- This does not introduce a provider-independent `/clear` command or a new
  first-class clear timeline event; it fixes settlement when a provider handles
  a prompt without ordinary turn activity.

## Wire compatibility

`HOST_DAEMON_PROTOCOL_VERSION` is bumped to **115** because Pi now sends a
prompt-settled bridge event that can change host-daemon event output.

## Verification

- `@bb/agent-runtime`: 941 tests passed across 46 files.
- `@bb/server`: 1,470 tests passed across 162 files.
- `@bb/host-daemon-contract`: 49 tests passed across 3 files.
- Typechecks passed for `@bb/agent-runtime`, `@bb/server`, `@bb/host-daemon`,
  and `@bb/host-daemon-contract` through Turbo.
- Live reproduction against the source-built app:

  ```sh
  bb thread tell <id> "/clear"
  bb thread wait <id> --status idle --timeout 60
  ```

  The thread reached `idle` in 1.6 seconds, and its accepted-message queue was
  empty afterward.

### DIFF
--- a/apps/server/test/internal/internal-event-append-ownership.test.ts
+++ b/apps/server/test/internal/internal-event-append-ownership.test.ts
@@ -1,5 +1,5 @@
 import { eq } from "drizzle-orm";
-import { events, getThread } from "@bb/db";
+import { events, getThread, listQueuedThreadMessages } from "@bb/db";
 import { threadScope, turnScope } from "@bb/domain";
 import {
   groupHostDaemonEvents,
@@ -18,6 +18,7 @@ import {
   seedEvent,
   seedHostSession,
   seedProjectWithSource,
+  seedQueuedMessage,
   seedThread,
   seedThreadRuntimeState,
 } from "../helpers/seed.js";
@@ -400,6 +401,75 @@ describe("internal event append ownership", () => {
           .map((row) => row.sequence)
           .sort((left, right) => left - right),
       ).toEqual([1, 2]);
+    } finally {
+      await harness.cleanup();
+    }
+  });
+
+  it("auto-sends a queued message after a zero-work turn completes", async () => {
+    const { environment, harness, session, thread } = await setupEventRoute();
+    try {
+      seedThreadRuntimeState(harness.deps, {
+        environmentId: environment.id,
+        inputText: "/clear",
+        providerThreadId: "provider-zero-work",
+        threadId: thread.id,
+      });
+      seedQueuedMessage(harness.deps, {
+        content: [
+          {
+            type: "text",
+            text: "queued behind zero-work turn",
+            mentions: [],
+          },
+        ],
+        threadId: thread.id,
+      });
+
+      const response = await postEventBatch({
+        harness,
+        sessionId: session.id,
+        events: [
+          {
+            threadId: thread.id,
+            event: {
+              type: "turn/started",
+              threadId: thread.id,
+              providerThreadId: "provider-zero-work",
+              scope: turnScope("turn-zero-work"),
+            },
+          },
+          {
+            threadId: thread.id,
+            event: {
+              type: "turn/completed",
+              threadId: thread.id,
+              providerThreadId: "provider-zero-work",
+              scope: turnScope("turn-zero-work"),
+              status: "completed",
+            },
+          },
+        ],
+      });
+
+      expect(response.status).toBe(200);
+      const queuedTurn = await waitForQueuedCommand(
+        harness,
+        ({ command }) =>
+          command.type === "turn.submit" && command.threadId === thread.id,
+      );
+      expect(queuedTurn.command).toMatchObject({
+        type: "turn.submit",
+        input: [
+          {
+            type: "text",
+            text: "queued behind zero-work turn",
+            mentions: [],
+          },
+        ],
+      });
+      expect(listQueuedThreadMessages(harness.db, thread.id)).toEqual([]);
+      expect(getThread(harness.db, thread.id)?.status).toBe("active");
     } finally {
       await harness.cleanup();
     }
--- a/packages/agent-runtime/src/acp/adapter.test.ts
+++ b/packages/agent-runtime/src/acp/adapter.test.ts
@@ -774,6 +774,40 @@ describe("acp adapter event translation", () => {
     ]);
   });
 
+  it("settles accepted input when completion arrives before an update", () => {
+    const adapter = createAdapter();
+    adapter.translateAcceptedCommand({
+      command: {
+        type: "turn/start",
+        clientRequestId: "creq_222222228e",
+        input: [promptTextInput({ text: "/agent-local-command" })],
+        options: fullProviderExecutionContext,
+        providerThreadId: "sess-1",
+        threadId: "thread-1",
+      },
+    });
+
+    const events = adapter.translateEvent(
+      {
+        jsonrpc: "2.0",
+        method: "acp/turn/completed",
+        params: { threadId: "thread-1", stopReason: "end_turn" },
+      },
+      THREAD_CONTEXT,
+    );
+
+    expect(events.map((event) => event.type)).toEqual([
+      "turn/started",
+      "turn/input/accepted",
+      "turn/completed",
+    ]);
+    expect(events.at(-1)).toMatchObject({
+      type: "turn/completed",
+      scope: turnScope("turn-1"),
+      status: "completed",
+    });
+  });
+
   it("translates ACP usage updates into exact context-window usage", () => {
     const adapter = createAdapter();
     startTurn(adapter);
--- a/packages/agent-runtime/src/acp/adapter.ts
+++ b/packages/agent-runtime/src/acp/adapter.ts
@@ -63,6 +63,7 @@ import { createStandardAdapterMembers } from "../shared/standard-adapter-members
 import { completeStartedToolItem } from "../shared/tool-item-translation.js";
 import { buildUnhandledProviderEvents } from "../shared/provider-unhandled-event.js";
 import { createScopedItemIdFactory } from "../shared/scoped-item-ids.js";
+import { resolveProviderTerminalTurn } from "../shared/provider-terminal-turn.js";
 import {
   createProviderTurnStateRegistry,
   finishOpenProviderTurn,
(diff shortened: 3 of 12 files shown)


jerrison:
Thanks @ymichael for the hardening commit — I have left it untouched.

Two updates: this is verified end-to-end now, and the branch is unconflicted again.

## Verified against a running app

The caveat in the description no longer holds. I still have not built bb from source, so instead I applied the equivalent edit to the shipped daemon bundle of **bb 0.37.0** on macOS, immediately before the minified gate `if(i.currentTurnId){` in `host-daemon/dist/daemon-bundle.mjs`:

```js
i.currentTurnId||i.pendingAcceptedUserMessages.length>0&&e.ensureTurnStarted({events:n,state:i,threadId:r});
```

I have an automated case in my own repo — a harness that drives the `bb` CLI against a real thread — which *asserts the defect*: send `/clear`, then expect the turn not to settle within 45s and the thread to still be `active`. On 2026-08-12 it flipped, on the same machine and bb version, with the patch as the only change:

| | Before | After |
|---|---|---|
| `bb thread wait --status idle --timeout 45` | timed out | reached `idle` |
| thread status | `active` | `idle` |
| wall clock for the case | 54.2s | 5.5s |

The 5.5s is the part I would look at: it is no longer waiting out the bound at all. `sdk/conversation_reset` still arrives first and the context is genuinely empty afterwards, so the Clear itself is unchanged — the turn now just ends with it. No second, empty turn opened on any run.

**What this does not cover.** It validates the logic, not your compiled artifact: my edit is the minified equivalent of the diff, applied by script and checked with `node --check`, not a source build. One machine, one provider (`claude-code`), model `claude-haiku-4-5`. Please read it as a strong signal rather than a substitute for CI.

## Merged main, and took protocol version 113

#1427 landed with `HOST_DAEMON_PROTOCOL_VERSION = 112` while this branch was also sitting on 112, so both sides claimed the same number and the same block of `contract.test.ts`. Resolved in 2fe2b9542 by taking **113** and keeping #1427's 112 line in the version history, so the two read in landing order.

Worth flagging for anything else racing a protocol bump: `commands.ts` merged *clean* and was still wrong. Both sides had made the byte-identical 111 → 112 edit, so git saw no conflict and only the contract test caught the collision.

I merged rather than rebased so as not to rewrite your commit. Happy to squash or rebase if you would rather the branch were linear.

## One question, and one ask

**Question.** #1427 adds `clearActiveTurnState?()` for the same *class* of bug — a provider reporting no active turn — but keys off `isAcpProviderId` and an ACP bridge error code, so it does not reach the claude-code `result` path. I have kept this on `ensureTurnStarted`. If you would rather it route through the new hook, say so and I will rewrite it.

**Ask.** The workflow run on 2fe2b9542 is sitting at `action_required` — fork PRs here need "Approve and run". CI was green on 1eaac7c63, and the only changes since are the merge and the 112 → 113 bump.

ymichael:
Sorry my agent was too eager and pushed before I told it to - working on some changes here so we get this fix in for all providers - ty for your patience

PR #1434, opened 2026-08-12
Keep plugin sidebar overflow menus visible on mobile

SawyerHood:
## Summary

- keep the `…` menu visible on the plugin nav rows at the top of the sidebar on a phone
- reuse the existing `data-sidebar-hover-actions-mobile="always"` flag that project and section rows already use
- leave thread-row menus hover-only; a long press would conflict with drag-to-reorder on these chrome rows

A phone has no hover, so hide/show for Extensions, Docs, Tasks, and the other plugin rows was unreachable.

## Validation

- `pnpm exec turbo run test --filter=@bb/app -- --run src/components/plugin/PluginNavSidebarItems.test.tsx`
- local dev app started for phone QA

## Risks

Small. The `…` glyph now stays on those few chrome rows on coarse pointers, matching project and section chrome.

### DIFF
--- a/apps/app/src/components/plugin/PluginNavSidebarItems.test.tsx
+++ b/apps/app/src/components/plugin/PluginNavSidebarItems.test.tsx
@@ -143,6 +143,19 @@ describe("PluginNavSidebarItems", () => {
     ).toBeNull();
   });
 
+  it("keeps the panel options trigger visible on mobile", () => {
+    registerPanel("docs", "Docs");
+
+    renderSidebarItems();
+
+    expect(
+      screen
+        .getByRole("button", { name: "Docs panel options" })
+        .closest("[data-sidebar-hover-actions-mobile]")
+        ?.getAttribute("data-sidebar-hover-actions-mobile"),
+    ).toBe("always");
+  });
+
   it("bounds and truncates a long sidebar accessory", () => {
     registerPanel("tasks", "Tasks", () => (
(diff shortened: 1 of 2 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop. I am reviewing this pull request for security, code quality, performance, architecture, and mobile behavior.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** A phone cannot hover over a button. This change keeps the plugin menu button visible, so a person can tap it.

I found no blocking issue.

The security review found no new data path, action, request, or trust boundary. The new attribute uses a fixed internal value.

The code quality review found that the change uses the shared mobile action contract. The test matches the existing contract test style.

The performance review found no new state, handler, observer, layout work, or DOM node. The existing CSS changes only opacity and pointer input.

The architecture scan found the same shared contract in section, project, thread, and toggle rows. This change reuses that contract and adds no duplicate system.

I also checked the exact head SHA with these steps:

- The focused test passed all 12 cases.
- The `@bb/app` Turbo type check passed.
- A 390 × 844 touch browser showed full opacity and active pointer input.
- A tap opened the mobile action sheet with “Hide from sidebar.”

GitHub reported the pull request as mergeable. Most checks passed, while two checks still ran during this review.

I did not add inline comments because I found no line-specific issue.

PR #1435, opened 2026-08-12
Keep tab overflow caret beside fixed controls

brsbl:
## Summary

- Keep the tab overflow caret in a fixed slot immediately beside Diff, outside the horizontal tab viewport.
- Reuse one persistent directional button while tabs overflow and collapse its slot completely when they fit.
- Preserve smooth step scrolling, wheel and trackpad behavior, active-tab visibility, dragging, selection, closing, and keyboard focus.

## Validation

- `pnpm exec turbo run test --filter=@bb/app -- --run src/components/secondary-panel/SecondaryPanelTabStrip.test.ts src/components/secondary-panel/ThreadSecondaryPanel.collapseControl.test.tsx src/components/secondary-panel/ThreadSecondaryPanel.test.ts`
- `pnpm exec turbo run lint --filter=@bb/app` (0 errors; existing warning baseline only)
- Live Ladle verification at representative overflowing and fitting widths, including directional caret scrolling

> AGENT GENERATED: by GPT-5.6

### DIFF
--- a/apps/app/src/components/secondary-panel/SecondaryPanelTabStrip.tsx
+++ b/apps/app/src/components/secondary-panel/SecondaryPanelTabStrip.tsx
@@ -92,8 +92,8 @@ interface SortableFileTabProps {
  * The middle, horizontally-scrolling region of the secondary panel tab strip.
  *
  * Only the file tabs scroll; the leading Info/Diff controls and trailing
- * new-tab/panel controls stay anchored outside this component. Edge
- * fades and scroll buttons appear only on a side that has more tabs, and the
+ * new-tab/panel controls stay anchored outside this component. A single fixed
+ * overflow button sits beside Diff and points toward the remaining tabs. The
  * active tab is auto-scrolled into view on mount and whenever it changes
  * (covering pointer, keyboard, and programmatic selection).
  */
@@ -107,8 +107,7 @@ export function SecondaryPanelTabStrip({
   const viewportRef = useRef<HTMLDivElement>(null);
   const contentRef = useRef<HTMLDivElement>(null);
   const activeTabRef = useRef<HTMLDivElement>(null);
-  const leftScrollButtonRef = useRef<HTMLButtonElement>(null);
-  const rightScrollButtonRef = useRef<HTMLButtonElement>(null);
+  const scrollButtonRef = useRef<HTMLButtonElement>(null);
   const [overflow, setOverflow] = useState<TabStripOverflowState>(
     INITIAL_OVERFLOW_STATE,
   );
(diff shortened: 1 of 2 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the Slop Cop. I am reviewing this change for security, code quality, performance, architecture, and end-to-end behavior.

apps/app/src/components/secondary-panel/SecondaryPanelTabStrip.tsx
```diff
@@ -375,6 +358,7 @@ export function SecondaryPanelTabStrip({
   const chevronNoDragClass = usesDesktopChrome
     ? MACOS_APP_REGION_NO_DRAG_CLASS
     : null;
+  const scrollDirection = overflow.canScrollRight ? "right" : "left";
```
SawyerHood: 🚨 `slopcop/review` — Keep the direction stable until the control reaches the opposite edge.

At the right edge, the first left click moves the viewport from 322px to 182px. This line then selects right because right travel remains. A second click returns the viewport to 322px. Mouse and keyboard users cannot make a second left step through a long tab list.

Keep the selected direction until the opposite edge. Add a test with more than two scroll steps. Verify that each intermediate click continues left.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: This change replaces two side buttons with one fixed caret. The caret points toward hidden tabs.

I found one high-severity interaction defect. At the right edge, one left click moves the tab row only one step. The caret then changes to right before it reaches the left edge. A second click reverses the first click. Users cannot use repeated clicks to reach older tabs in a long list.

The browser check reproduced this result. The viewport moved from 322px to 182px, then returned to 322px after the second click. The current test cannot detect this fault because its full range is 120px, while one click requests 140px.

Keep the selected direction until the control reaches the opposite edge. Add a test that needs more than two steps. Verify that each intermediate click continues in the selected direction.

I found no security issue, performance regression, or duplicate architecture. The focused test and the app type check passed. The full app suite also passed with 2,572 tests.

I left one inline comment on the direction calculation. I used a comment review, as required.

apps/app/src/components/secondary-panel/SecondaryPanelTabStrip.tsx
brsbl: Fixed in 494fb9f33. The caret now keeps its selected direction at intermediate offsets and flips only at the opposite hard edge. The regression range is 360px, so it requires multiple 140px steps; the real browser pass moved 288 → 148 → 8 → 0 while staying left, then flipped right and moved 140 → 280.

SawyerHood:
@brsbl idk if you saw but michael just fixed this

brsbl:
oooh i missed that, ok closing!

PR #1437, opened 2026-08-13
Attribute event-loop stalls to the work that blocked them

SawyerHood:
## Why

The packaged server logs at `info`. Slow-query instrumentation used `debug`, so 82,590 production log lines covering 2,609 event-loop stall windows had **zero** slow-query lines. The stall report also named no unit of work, so a stall could not be attributed to a route, sweep, or plugin handler.

This is layer 1 of the event-loop stall stack. Later layers (pragmas, timeline, outline cache) need these lines so their effect can be measured rather than argued.

## What

- Log `Slow DB query` at `info` (threshold stays 100 ms).
- Track a labeled work stack on the server event loop.
- Include `currentWork`, `lastWork`, and `lastWorkMs` on each `Event loop stalled` line.
- Label HTTP `/api/v1` and `/internal` requests, periodic sweeps, daemon websocket messages, timeline/outline builds, and plugin handler invocations.

`currentWork` is sampled after the stall: the interval cannot run while the loop is blocked, so the in-flight label is the unit still on the stack. `lastWork` is the unit that just finished.

## Test plan

- [x] `@bb/db` typecheck + tests
- [x] `@bb/server` stall-monitor, periodic-sweep, timeline, and daemon-message tests

Prerequisite: none (targets `main`).

### DIFF
--- a/apps/server/src/services/system/event-loop-stall-monitor.ts
+++ b/apps/server/src/services/system/event-loop-stall-monitor.ts
@@ -1,7 +1,7 @@
 import { monitorEventLoopDelay } from "node:perf_hooks";
 import { roundDurationMs } from "../lib/duration.js";
 import type { ServerLogger } from "../../types.js";
-import { getEventLoopWorkSnapshot } from "./event-loop-work.js";
+import { takeEventLoopWorkWindowSnapshot } from "./event-loop-work.js";
 
 export interface EventLoopStallMonitorOptions {
   logger: Pick<ServerLogger, "info">;
@@ -30,6 +30,7 @@ export function startEventLoopStallMonitor(
 
   const interval = setInterval(() => {
     const maxDelayMs = nanosecondsToMilliseconds(histogram.max);
+    const work = takeEventLoopWorkWindowSnapshot();
     if (maxDelayMs >= DEFAULT_EVENT_LOOP_STALL_LOG_THRESHOLD_MS) {
       // `info`, not `debug`: the packaged app runs at `info`, so a `debug` line
       // here is unreachable in production — which is exactly where a stalled
@@ -38,9 +39,9 @@ export function startEventLoopStallMonitor(
       // dynamic tool call and interactive request, so it delays real agent
       // work, not just UI refreshes. Threshold-gated, so a healthy server
       // stays silent.
-      // Sampled after the stall: the interval cannot run while the loop is
-      // blocked, so currentWork is the unit still on the stack and lastWork
-      // is the unit that just finished. Both are labels, not CPU samples.
+      // currentWork is still in flight. lastWork is the latest finish.
+      // slowestWork is the longest unit in this histogram window, so a later
+      // heartbeat cannot hide the block that produced histogram.max.
       options.logger.info(
         {
           intervalMs: DEFAULT_EVENT_LOOP_STALL_MONITOR_INTERVAL_MS,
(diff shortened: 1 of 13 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am SlopCop, and I am reviewing this pull request now. I will check security, code quality, performance, architecture, and practical behavior.

apps/server/src/services/system/event-loop-stall-monitor.ts
```diff
@@ -49,6 +53,7 @@ export function startEventLoopStallMonitor(
           ),
           resolutionMs: DEFAULT_EVENT_LOOP_STALL_MONITOR_RESOLUTION_MS,
           thresholdMs: DEFAULT_EVENT_LOOP_STALL_LOG_THRESHOLD_MS,
+          ...getEventLoopWorkSnapshot(),
```
SawyerHood: 🚨 `slopcop/review` — High: keep the blocking work linked to the maximum delay

`histogram.max` covers the full five-second window, but this snapshot reads only the later work state. A worker reproduced a 650 ms block followed by a short heartbeat. The report named the heartbeat and showed `lastWorkMs: 0`. Record time-stamped candidates for the sample window, or log slow synchronous work directly. Please test a real block with later work before the report.

apps/server/src/services/system/event-loop-work.ts
```diff
@@ -0,0 +1,70 @@
...
+function formatWorkStack(frames: readonly EventLoopWorkFrame[]): string {
+  return frames.map((frame) => frame.label).join(" > ");
+}
+
+function enterEventLoopWork(label: string): void {
+  workStack.push({ label, startedAt: performance.now() });
+}
+
+function leaveEventLoopWork(): void {
+  const frame = workStack.pop();
```
SawyerHood: 🚨 `slopcop/review` — High: remove the correct frame when concurrent work ends

`runEventLoopWork()` remains active across `await`, but this pop always removes the newest global frame. Request A can finish before request B and remove B's frame. A direct probe then reported B as complete and A as active. Track each frame by identity, and store parent links for each asynchronous task. Please add a test where sibling tasks finish out of order.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** This change gives each server job a name tag. When the server freezes, the log should show the responsible job. Two timing errors can show the wrong job instead.

I found two high-confidence correctness problems. I added inline comments for both.

- The global stack breaks during normal request overlap. One request can remove another request's frame and reverse their labels.
- The five-second maximum delay does not remain linked to its cause. Later quick work can replace the recorded label.

I found no medium-or-higher security issue. SQL logs omit bound values, limit SQL text, and use structured fields.

The architecture scan found related gaps. Manual labels omit plugin background services and startup recovery. The host daemon has a similar monitor, but server and daemon policy should stay separate. An identity-based frame registry and time-stamped samples would reduce incorrect labels and manual gaps. Some test variables still use old debug-log names, but that issue has no runtime effect.

Validation passed:

- All `@bb/server` and `@bb/db` tests passed. The server suite ran 1,461 tests.
- Both package type checks passed.
- The app loaded in a browser without page errors. The projects API returned HTTP 200.
- A direct concurrency probe reproduced reversed labels.
- A real-stall probe reproduced the unrelated heartbeat label.
- All required GitHub checks passed.

The checks do not cover the two core timing failures. I recommend fixing both inline findings before the team relies on these logs.

apps/server/src/services/system/event-loop-stall-monitor.ts
SawyerHood: Fixed on 3bf1a86. The stall tick now snapshots `slowestWork` from every unit in the histogram window before it resets that window. A later heartbeat still shows as `lastWork`; the long block remains `slowestWork`. Added a test that records a 650 ms sweep then a 1 ms heartbeat.

apps/server/src/services/system/event-loop-work.ts
SawyerHood: Fixed on 3bf1a86. Frames are now keyed by id and stored in AsyncLocalStorage, so one request cannot pop another request's frame. Added a test where sibling requests finish out of order.

PR #1438, opened 2026-08-13
Set SQLite cache, mmap, and WAL synchronous pragmas

SawyerHood:
## Why

`better-sqlite3` is synchronous and every query runs on the server event loop. The live database is ~1.9 GB with a 2 MiB page cache, no mmap, and `synchronous = FULL` (fsync on every commit). Cold page reads and per-commit fsync are the main source of sub-second stalls, not a missing index.

This is layer 2 of the event-loop stall stack. It depends on layer 1 (#1437) so the new pragmas can be confirmed from stall and slow-query logs.

## What

`createConnection` now sets:

| pragma | value | reason |
| --- | --- | --- |
| `cache_size` | `-262144` (256 MiB) | replace the 2 MiB default in front of a multi-GB file |
| `synchronous` | `NORMAL` | WAL-standard; no fsync per commit. Power loss can drop the last transactions; it cannot corrupt the file |
| `mmap_size` | `1073741824` (1 GiB) | avoid a `pread` + copy on every page |
| `busy_timeout` | `5000` | wait for a lock instead of failing immediately |
| `temp_store` | `MEMORY` | keep temp sorts off disk |

`wal_autocheckpoint` is left unchanged. A 38 MB WAL is a reader-starvation problem, not an autocheckpoint setting problem.

## Test plan

- [x] File-backed pragma test asserts the live values
- [x] `@bb/db` typecheck + full test suite
- [x] Server database-maintenance sweep tests (busy_timeout save/restore)

Prerequisite: #1437.

### DIFF
--- a/packages/db/src/connection.ts
+++ b/packages/db/src/connection.ts
@@ -41,6 +41,11 @@ interface TimedStatementOperationArgs<TValue> {
 }
 
 const DEFAULT_SLOW_DB_QUERY_LOG_THRESHOLD_MS = 100;
+/** 256 MiB page cache. Negative cache_size is kibibytes. */
+export const SQLITE_CACHE_SIZE_KIB = 262_144;
+/** Memory-map the first 1 GiB of the database file. */
+export const SQLITE_MMAP_SIZE_BYTES = 1_073_741_824;
+export const SQLITE_BUSY_TIMEOUT_MS = 5_000;
 const MAX_LOGGED_SQL_LENGTH = 1_000;
 const SQL_TRUNCATION_SUFFIX = "...";
 // Keep ORM-generated quoted identifiers intact. SQLite accepts double-quoted
@@ -171,6 +176,14 @@ export function createConnection(
   sqlite.pragma("journal_mode = WAL");
   // Enable foreign keys
   sqlite.pragma("foreign_keys = ON");
+  // WAL + NORMAL: no fsync per commit. Power loss can drop the last
+  // transactions; it cannot corrupt the file. cache_size/mmap_size replace
+  // the 2 MiB default cache so a multi-GB database is not re-read per query.
+  sqlite.pragma("synchronous = NORMAL");
+  sqlite.pragma(`cache_size = -${SQLITE_CACHE_SIZE_KIB}`);
+  sqlite.pragma(`mmap_size = ${SQLITE_MMAP_SIZE_BYTES}`);
+  sqlite.pragma(`busy_timeout = ${SQLITE_BUSY_TIMEOUT_MS}`);
+  sqlite.pragma("temp_store = MEMORY");
   instrumentSqliteClient(sqlite, options);
 
   const db = drizzle({ client: sqlite, schema });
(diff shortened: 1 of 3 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** This pull request gives SQLite more memory and reduces slow disk work.

I am the SlopCop. I am reviewing this pull request for security, code quality, architecture, performance, and product behavior.

packages/db/src/connection.ts
```diff
@@ -171,6 +176,14 @@ export function createConnection(
...
   // Enable foreign keys
   sqlite.pragma("foreign_keys = ON");
+  // WAL + NORMAL: no fsync per commit. Power loss can drop the last
+  // transactions; it cannot corrupt the file. cache_size/mmap_size replace
+  // the 2 MiB default cache so a multi-GB database is not re-read per query.
+  sqlite.pragma("synchronous = NORMAL");
+  sqlite.pragma(`cache_size = -${SQLITE_CACHE_SIZE_KIB}`);
+  sqlite.pragma(`mmap_size = ${SQLITE_MMAP_SIZE_BYTES}`);
+  sqlite.pragma(`busy_timeout = ${SQLITE_BUSY_TIMEOUT_MS}`);
+  sqlite.pragma("temp_store = MEMORY");
```
SawyerHood: 🚨 `slopcop/review` — `temp_store = MEMORY` removes the normal disk spill path.

The main connection serves broad thread searches. That query groups and sorts matches before its final limit. Large matches can keep temporary indexes in process memory. The new page cache can also use 256 MiB. Please keep the default file store, or add a measured memory limit.

packages/db/src/index.ts
```diff
@@ -1,4 +1,9 @@
-export { createConnection } from "./connection.js";
+export {
+  createConnection,
+  SQLITE_BUSY_TIMEOUT_MS,
```
SawyerHood: 🚨 `slopcop/review` — These exports add internal tuning policy to the package API.

No production code imports these constants. The test already imports them from `connection.ts`. Please remove the three root exports. This change will keep the public contract small.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

**ELI5:** This change gives SQLite more memory and reduces disk waits. It can run faster, but memory-only scratch work can grow too much.

I found two items:

- **Medium:** `temp_store = MEMORY` removes the disk spill path for temporary indexes. Broad thread searches can create data-dependent temporary work. Keep the default file store, or add a measured memory limit.
- **Low:** The root package exports three tuning constants that no production code uses. Keep this policy private to the connection module.

The security review found no injection issue. All new pragma values use fixed constants. The `NORMAL` durability loss is explicit and matches SQLite behavior. A system failure can remove recent commits, but an application failure will not.

The architecture scan found one main database connection path. Plugin databases use a separate policy, so a shared setup helper would mix different contracts. The 5-second timeout already matches the `better-sqlite3` default. Thus, this line does not add a new delay.

Validation passed:

- The `@bb/db` type check passed.
- All 380 `@bb/db` tests passed.
- All current GitHub checks passed.
- The dev server started from this commit.
- A browser opened the app and showed the empty thread view.

I posted two inline comments. I used a comment review only.

packages/db/src/connection.ts
SawyerHood: Removed `temp_store = MEMORY` on 64d5e58. The connection keeps the default file store so large thread-search sorts can spill to disk.

packages/db/src/index.ts
SawyerHood: Removed the three root exports on 64d5e58. The test still imports the constants from `connection.ts`.

PR #1454, opened 2026-08-13
Rework Extensions into a sidebar-navigated browse surface

brsbl:
Moves plugin configuration into Settings, where configuration lives — and is the top layer of the native GitHub stack for the Extensions rework (#1358 → #1360 → #1471 → #1472 → #1473 → #1474 → here).

- **/settings/plugins/:id** hosts each plugin's settings form, under a Plugins sidebar group where each row uses the plugin's own icon.
- **Mirrored anatomies**: the plugin settings page mirrors the Extensions detail page's header (icon + title + one-line description) and section stack; its "Plugin details" section is one sentence linking across to the plugin page, and the detail page's "Configuration" section is one sentence linking here — each page owns its half and points at the other.
- The per-plugin settings route participates in Settings route memory like any other section; only the bare /settings/plugins list remains a legacy redirect to Extensions.

Bundle budget: this integration layer raises the compressed boot allowance from 435.3 KB to 438.0 KB to account for the persistent plugin rows added to Settings. The final measured boot payload is 436.2 KB Brotli; the raw boot payload remains within its existing limit.

BB-Thread-ID: thr_kuixkqt8n2

> AGENT GENERATED: by GPT-5


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the Slop Cop. I started this review for security, code quality, architecture, performance, and the main product route.

apps/app/src/components/tools/SkillsBrowse.tsx
```diff
@@ -121,113 +129,103 @@ function RegistrySkillSourceItemSkeleton({ skillName }: { skillName: string }) {
...
-                skill={skill}
-                onFork={onFork}
-                onSelect={onSelect}
-              />
-            ),
-          )}
-        </ResourceBrowseGrid>
-      )}
+      <div className={cn("space-y-4", TOOLS_PAGE_BAND_CLASSES)}>
+        {hasError ? (
```
SawyerHood: 🚨 `slopcop/review` — [P2] Keep prior cards after a later page error

A later request can fail after page one remains in local state. This branch then replaces all prior cards with the error state. Keep the cards and show a retry row after them. Add a test with a successful first page and a failed second page.

apps/app/src/components/tools/SkillsLibrary.tsx
```diff
@@ -164,9 +159,36 @@ export function SkillsLibrary() {
     refetchOnWindowFocus: false,
     staleTime: REGISTRY_LIST_STALE_TIME_MS,
   });
+  // Loaded pages accumulate for infinite scroll; a new search starts a fresh
+  // accumulation. Dedupe by id so a refetched page cannot double its rows.
+  const trimmedRegistrySearch = registrySearch.trim();
+  const [loadedRegistry, setLoadedRegistry] = useState<{
```
SawyerHood, line 174: 🚨 `slopcop/review` — [P2] Keep pages with different rank meanings separate

The accumulation key omits the rank mode. The server can change from `trending` to `all-time` after a fallback. The list can then combine counts with different meanings. Include the rank mode in this identity and reset the accumulated pages when it changes.

apps/server/src/services/plugins/builtin-registry.ts
```diff
@@ -144,6 +144,79 @@ export const BUNDLED_PLUGINS: readonly BundledPluginDefinition[] = [
...
+    pluginId: "thread-hover-cards",
+    category: "Interface",
+    gitSource:
+      "git:https://github.com/brsbl/bb-plugins.git@plugin/thread-hover-cards",
+    displayName: "Thread Hover Cards",
+    description:
+      "Preview thread status, the latest agent update, and repository or PR context from the sidebar.",
+    icon: "ZoomIn",
+    bbEngineRange: ">=0.0.34",
+    bbPluginSdkRange: "^0.5.0",
```
SawyerHood, line 191: 🚨 `slopcop/review` — [P1] Fix the copied SDK ranges before these entries ship

Each catalog entry declares SDK `^0.5.0`, but each current branch manifest declares `^0.4.1`. Runtime uses SDK `0.5.0`, so all three installs fail compatibility validation. The catalog test reads only these copied values. Validate each actual source manifest in CI.

examples/plugins/thread-hover-cards/README.md
```diff
@@ -0,0 +1,35 @@
...
+pnpm exec turbo run build --filter=bb-app
+pnpm exec turbo run build --filter=bb-plugin-thread-hover-cards
+bb plugin install ./examples/plugins/thread-hover-cards
+```
+
+After editing the plugin, run `bb plugin reload thread-hover-cards` or use `bb plugin dev` from this directory.
+
+## Source of truth
+
+The canonical implementation lives at `bb-plugins/plugins/thread-hover-cards`. Make behavior changes there first, then port them into this example.
```
SawyerHood: 🚨 `slopcop/review` — [P2] Replace the manual two-repository copy process

This PR adds about 8,450 copied lines from an external source. This instruction requires manual ports, and the copies already have different SDK data. Generate the example from one source, or add an automatic synchronization check that fails on drift.

apps/app/src/components/tools/SkillsCollection.tsx
```diff
@@ -204,39 +205,68 @@ function skillMutationDisabledReason(skill: SkillSummary): string {
...
-          />
-        ) : undefined
-      }
-      description={description}
-      onOpen={onSelect}
-      trailingVisual={<ResourceRowDetailChevron />}
-    />
+    // ResourceRow keeps its narrow prop surface; the intent listeners live on
+    // a wrapper (React delegates both events, so this covers the whole row).
+    <div onPointerEnter={prefetch} onFocus={prefetch}>
```
SawyerHood: 🚨 `slopcop/review` — [P2] Add hover intent before the detail prefetch

Pointer entry immediately requests the file list and full `SKILL.md`. One browser sweep across seven rows made 14 requests and read 168 KB. The same sweep repeated after six seconds. Add a delay with cancel support, or prefetch only the file list.

apps/server/src/services/plugins/builtin-registry.ts
```diff
@@ -144,6 +144,79 @@ export const BUNDLED_PLUGINS: readonly BundledPluginDefinition[] = [
...
+ * here — the store renders these static fields, not the remote manifest.
+ * Install refs are the repo's published per-plugin branches.
+ */
+export const GIT_OFFICIAL_PLUGINS: readonly GitOfficialPluginDefinition[] = [
+  {
+    name: "thread-hover-cards",
+    pluginId: "thread-hover-cards",
+    category: "Interface",
+    gitSource:
+      "git:https://github.com/brsbl/bb-plugins.git@plugin/thread-hover-cards",
```
SawyerHood: 🚨 `slopcop/review` — [P1] Pin official plugin sources and enforce catalog identity

These entries use mutable branches but receive official catalog provenance. A later branch change can run full-trust code in the server process. `installGitCatalogPlugin` also does not pass the expected `pluginId` to validation. Pin each source to a reviewed commit. Reject an ID mismatch before the build or load step.

packages/shared-ui/src/components/ui/resource-pagination.tsx
```diff
@@ -190,6 +192,126 @@ export function useResourcePagination<Item>(
...
+  const resetKey = options.resetKey ?? "";
+  // Render-phase reset, mirroring useResourcePagination's anchor: a new
+  // projection (search, filter, sort) starts back at one page's worth.
+  const [anchor, setAnchor] = useState({ resetKey, pages: 1 });
+  if (anchor.resetKey !== resetKey) {
+    setAnchor({ resetKey, pages: 1 });
+  }
+  const pages = anchor.resetKey === resetKey ? anchor.pages : 1;
+  const visibleItems = useMemo(
+    () => items.slice(0, pages * pageSize),
```
SawyerHood: 🚨 `slopcop/review` — [P2] Do not remove loaded rows after a viewport resize

The state stores a page count, but this slice uses the current page size. If the size falls from ten to five, two loaded pages fall from twenty rows to ten. Preserve the maximum loaded item count across page-size changes. Add a resize test.

apps/app/src/components/tools/SkillsLibrary.tsx
```diff
@@ -212,12 +234,19 @@ export function SkillsLibrary() {
...
+   * reason, but on the trending ranking the entry is also the only source of a
+   * lifetime install count — so ask for every card there rather than letting
+   * install correctness depend on whether the list happened to carry a summary.
+   */
   const registryDescriptionSkills = useMemo(
     () =>
-      (registryQuery.data?.skills ?? []).filter(
-        (skill) => skill.summary === null,
+      loadedRegistrySkills.filter(
+        (skill) => skill.summary === null || registryRanking === "trending",
```
SawyerHood: 🚨 `slopcop/review` — [P2] Remove the request for each trending card

This condition selects every card when the rank is `trending`. `useQueries` then starts one entry request for each card. The browser test made 24 requests for 24 cards. It made about 130 requests after 127 cards loaded. Include the needed count in the list response, or add one batch request.

apps/cli/src/commands/skill.ts
```diff
@@ -297,13 +324,30 @@ export function registerSkillCommands(
...
             registry,
             await enrichRegistryStars(registry, result.skills),
-          ),
+            result.ranking,
+          );
+        // Carried into the payload, not just the table: on the trending
+        // ranking `installs` is a lifetime total for every row EXCEPT these,
+        // which still hold the 24h figure. Without the list, an agent reading
+        // `ranking: "trending"` would misread the rows this change fixed.
+        const enrichedResult = {
```
SawyerHood: 🚨 `slopcop/review` — [P2] Keep the JSON count contract unambiguous

The shared contract says `installs` follows `ranking`. This JSON keeps `ranking: trending` while successful values become lifetime totals. Unresolved values remain 24-hour totals in the same field. Preserve one meaning, or add a separate `lifetimeInstalls` field.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

## ELI5

This PR gives Extensions a clearer sidebar and browse page. The new routes and main user flows mostly worked in browser tests.

However, the three new official plugins cannot install. Their mutable sources also create a full-trust supply-chain risk.

The Skills browse page sends too many requests. Several paging contracts can also lose or mix data. CI currently fails.

## Findings

### P1

- Pin each external official plugin to a reviewed commit. Also verify the fetched manifest ID against the catalog ID before build or load.
- Fix or remove all three external catalog entries. The catalog declares SDK `^0.5.0`, but every current source manifest declares `^0.4.1`.
- Make the bundle-budget check pass. A clean build creates a 345-byte boot chunk without Brotli output, so the required CI check fails.

### P2

- Remove the entry request for each trending skill card. The browser made 24 requests for 24 cards and about 130 after 127 cards.
- Keep prior cards visible when a later page request fails. Show an inline retry control after the existing cards.
- Preserve the loaded row count when the measured page size changes. A smaller page size can remove rows after a viewport resize.
- Reset accumulated skill pages when the rank mode changes. Otherwise, the list can combine `trending` and `all-time` count meanings.
- Keep the CLI JSON count contract clear. The current result mixes lifetime and 24-hour values under `ranking: trending`.
- Treat `/settings/plugins/:id` as a real Settings route. The route-memory hook still excludes it as a compatibility route.
- Use source-specific install text. The dialog says remote Git plugins come bundled with BB.
- Add hover intent before skill detail prefetch. One sweep across seven rows made 14 requests and read 168 KB.

## Architecture and duplication

The shared showcase components are a good boundary. I found no second active showcase engine.

The thread-hover example adds about 8,450 copied lines. Its README requires manual changes in two repositories, and the copies already differ.

Use one generated source or add a synchronization check. The server and daemon boundary remains correct, so this PR needs no protocol version change.

## Verification

- App typecheck passed.
- The app build passed.
- All 2,630 app tests passed across 336 files.
- `git diff --check` passed.
- Desktop browser tests used 1440×900. Compact tests used 390×844.
- The tests covered Extensions routes, legacy redirects, sidebar state, Back, the hero composer, keyboard tabs, search, and reduced motion.
- No tested route produced a console error. The test server and browser daemon are stopped.
- Prettier reports a format issue in `AppLayout.tsx`.
- GitHub CI passes all other completed jobs, but the required bundle-budget job fails.

The GPT-5.6 review gate returned `REQUEST CHANGES`. I posted this review with the comment option only.

apps/app/src/components/tools/SkillsBrowse.tsx
brsbl: Addressed in the rebased stack at 796265dab: later-page failures now preserve already loaded results, with regression coverage.

apps/app/src/components/tools/SkillsLibrary.tsx
brsbl: Addressed in the rebased stack at 796265dab: rank-mode pagination accumulates and deduplicates results across pages.

apps/server/src/services/plugins/builtin-registry.ts
brsbl: Addressed in 0a2991155: catalog metadata requires SDK 0.5-compatible immutable refs. This PR remains unmerged while the plugin source stack publishes its final refreshed refs.

examples/plugins/thread-hover-cards/README.md
brsbl: Addressed in the final stack: the copied SDK example was removed.

apps/app/src/components/tools/SkillsCollection.tsx
brsbl: Addressed in 796265dab: hover intent is delayed by 150ms, preventing accidental card transitions.

apps/server/src/services/plugins/builtin-registry.ts
brsbl: Addressed in 0a2991155: installs use immutable commit sources and validate the expected plugin identity.

packages/shared-ui/src/components/ui/resource-pagination.tsx
brsbl: Addressed in 796265dab: loadedCount preserves fetched rows when the viewport shrinks.

apps/app/src/components/tools/SkillsLibrary.tsx
brsbl: Addressed in 796265dab: list metadata is fetched in a batch instead of issuing one request per plugin.

apps/cli/src/commands/skill.ts
brsbl: Addressed in 796265dab: the CLI distinguishes lifetime installs from the current installed count.

PR #1456, opened 2026-08-13
Stop plugin content scripts from crashing the app on remove

SawyerHood:
## Why

A user plugin (File Reveal) wraps host links and buttons in a new span. React still treats the old parent as the owner. The next remove or reorder of that same node throws `NotFoundError` during commit and can blank the window.

PR #1418 already made `removeChild` and `insertBefore` non-fatal. That was not enough. The plugin still stole React-owned nodes, leftover wrappers stayed in the tree, and `replaceChild` could still throw.

## What

- Refuse a content-script move of a React-owned node to a new parent. Mount, dispose, and observers created during mount all run under that rule.
- The plugin can still insert its own sibling control. The host node stays in React's tree.
- Extend the foreign DOM guard to `replaceChild` and swallow `NotFoundError` on the existing paths.

## Test plan

- [x] `pnpm exec turbo run typecheck --filter=@bb/app`
- [x] `pnpm exec vitest run --config vitest.config.ts src/lib/foreign-dom-mutation-guard.test.tsx src/lib/plugin-frontend-reload.test.ts` (30 passed)

## Risks

The expensive fiber check runs only while a plugin content script (or its observer) is moving a node. React commits hit one integer check and then the native method. Same cost class as the #1418 guard.

### DIFF
--- a/apps/app/src/lib/foreign-dom-mutation-guard.ts
+++ b/apps/app/src/lib/foreign-dom-mutation-guard.ts
@@ -3,57 +3,79 @@
  *
  * React owns every node it renders and remembers which parent each one belongs
  * to. On unmount it calls `parent.removeChild(node)` directly. When another
- * agent in the page — a browser extension's content script, the browser's own
- * page translator, a bookmarklet — moves or removes one of those nodes first,
- * React's call throws:
+ * agent in the page — a browser extension's content script, a plugin content
+ * script, the browser's own page translator, a bookmarklet — moves or removes
+ * one of those nodes first, React's call throws:
  *
  *   NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be
  *   removed is not a child of this node.
  *
  * React catches that throw in the commit phase and escalates it to the nearest
- * error boundary. bb had none above the router, so React tore the whole root
- * down and the user got a blank page. `AppErrorBoundary` now catches it and
- * shows a recovery screen, but a recovery screen still costs the user their
- * place for a mutation that only ever affected one subtree. This guard keeps
- * the fault from reaching that boundary at all.
+ * error boundary. List updates are worse: `insertBefore` during placement is
+ * not wrapped, so a stolen reference node blanks the window even when
+ * `AppErrorBoundary` would have caught a deletion. This guard keeps both
+ * faults from reaching that boundary.
  *
- * This guard makes both calls non-fatal. `removeChild` returns the node instead
- * of throwing when it already has a different parent (the node is gone from
- * where React expected it, which is exactly the outcome React wanted).
- * `insertBefore` appends instead of throwing when the reference node has moved
- * away, so the content still reaches the DOM rather than vanishing.
+ * `removeChild` returns the node instead of throwing when it already has a
+ * different parent. `insertBefore` appends instead of throwing when the
+ * reference node has moved away. `replaceChild` does the same when the node
+ * it would replace is gone. Wrappers only change the case that would throw.
+ *
+ * Plugin content scripts are a second, tighter layer. While one of those
+ * scripts (or a MutationObserver it created) runs, the guard also refuses to
+ * move a React-owned node to a new parent. That stops the wrap-then-remove
+ * crash at the source: the plugin can still insert its own sibling controls,
+ * but it cannot steal a host button or link out of React's tree.
  *
  * This is the workaround React's own maintainers publish for translated pages
- * (facebook/react#11538). It is deliberately narrow: both wrappers only change
- * behavior in the case that would otherwise throw, so every well-formed call
- * reaches the native method unchanged.
+ * (facebook/react#11538), plus a host-side fence for trusted plugin scripts
+ * that still share the document.
  *
- * Suppressed calls are counted and the first few are logged, because a burst of
- * them from a build where no extension is involved would point at a real bug in
- * our own rendering instead.
+ * Suppressed calls are counted and the first few are logged, because a burst
+ * of them from a build where no extension or plugin is involved would point
+ * at a real bug in our own rendering instead.
  */
 
 /** Log at most this many suppressions, then only keep counting. */
 const MAX_LOGGED_SUPPRESSIONS = 3;
 
+const REACT_FIBER_PREFIXES = [
+  "__reactFiber$",
+  "__reactInternalInstance$",
+] as const;
+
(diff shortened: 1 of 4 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the Slop Cop. I am reviewing this pull request now. I will check security, code quality, performance, architecture, and product behavior.

apps/app/src/lib/foreign-dom-mutation-guard.ts
```diff
@@ -102,35 +218,180 @@ export function installForeignDomMutationGuard(): void {
...
+    resolveParent: (self: Element) => Node | null,
+  ): AppendLike =>
+    function adjacent(this: Element, ...nodes: Array<Node | string>): void {
+      const parent = resolveParent(this) ?? this;
+      const kept = filterAppendNodes(parent, nodes);
+      if (kept.length === 0) return;
+      native.apply(this, kept);
+    };
+
   Node.prototype.removeChild = guardedRemoveChild;
```
SawyerHood, line 395: 🚨 `slopcop/review` — Common DOM move operations bypass this wrapper list.

`replaceChildren`, `insertAdjacentElement`, and `Range.insertNode` remain native. A plugin can also call `host.replaceWith(group)` and then append the detached host node.

These paths move a React-owned node while isolation is active. Please cover each production move path with Chromium tests. Consider an architecture that does not depend on a partial method list.

apps/app/src/lib/foreign-dom-mutation-guard.ts
```diff
@@ -102,35 +218,180 @@ export function installForeignDomMutationGuard(): void {
...
+    (self) => self.parentNode,
+  );
+  Document.prototype.append = guardedParentAppend(nativeDocumentAppend);
+  Document.prototype.prepend = guardedParentAppend(nativeDocumentPrepend);
+  DocumentFragment.prototype.append = guardedParentAppend(nativeFragmentAppend);
+  DocumentFragment.prototype.prepend = guardedParentAppend(
+    nativeFragmentPrepend,
+  );
+
+  if (NativeMutationObserver !== null) {
```
SawyerHood: 🚨 `slopcop/review` — The MutationObserver replacement breaks native constructor behavior.

This function always returns a native observer and ignores `new.target`. Therefore, an observer subclass does not create an instance of that subclass.

The replacement also works without `new`. Please preserve both native constructor rules and subclass identity. Add a real browser test for both cases.

apps/app/src/lib/plugin-frontend-reload.test.ts
```diff
@@ -688,6 +695,58 @@ describe("reconcilePluginFrontends", () => {
...
+      }),
+    );
+
+    await reconcilePluginFrontends(state, deps);
+    expect(link!.parentNode).toBe(reactParent);
+    expect(container.querySelector("button")).not.toBeNull();
+
+    act(() => root.unmount());
+    container.remove();
+    uninstallForeignDomMutationGuardForTest();
```
SawyerHood: 🚨 `slopcop/review` — A failed assertion can leave global DOM prototypes changed.

This cleanup runs only after all earlier assertions pass. A failure will affect later tests in the same file.

Please move the uninstall call to `afterEach` or use a `finally` block.

apps/app/src/lib/plugin-frontend.ts
```diff
@@ -506,38 +507,45 @@ async function mountWithTimeout(
   let timeoutId: ReturnType<typeof setTimeout> | undefined;
   let timedOut = false;
   const mountPromise = Promise.resolve().then(() =>
```
SawyerHood, line 509: 🚨 `slopcop/review` — Supported asynchronous mounts escape the DOM isolation.

The SDK permits `mount` to return a Promise. This wrapper clears its state before the first awaited continuation runs.

I reproduced this path in Chromium. A mount used `await Promise.resolve()` and then moved the React-owned Settings link into a new span. The guard logged no refusal.

Please apply the same policy to supported asynchronous work and later callbacks. Add production-path tests for an async mount, an async disposer, and an event callback.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: React owns pieces on a board. This patch stops one plugin move, but other supported moves still get through.

I found four reliability defects that should block the merge. I left a comment-only review, as required.

1. Async plugin work escapes the guard. The SDK permits async mounts and cleanup functions.

   A Chromium test confirmed the gap. The synchronous plugin move failed, and the Settings link stayed with React.

   The same plugin used one `await`. It then moved the Settings link into its wrapper, and the guard logged no refusal.

2. Common DOM move APIs bypass the wrapper list. These APIs include `replaceChildren`, `insertAdjacentElement`, and `Range.insertNode`.

   A plugin can also detach a React node first. The guard then permits the later append because the node has no parent.

3. The MutationObserver replacement breaks native constructor rules. An observer subclass no longer creates an instance of that subclass.

4. The SDK test host and plugin documentation do not match the new production restriction.

The new integration test also needs reliable cleanup. An early assertion failure currently leaves global DOM prototypes changed.

The narrow `replaceChild` fallback looks useful. The broader isolation layer needs a clear contract and one shared conformance test set.

I found no separate security defect. Content scripts already run as trusted same-origin code, so this guard is a reliability control.

I found no duplicate DOM guard. The architectural risk comes from a partial global method list and different production and test-host behavior.

The performance worker measured a 14 percent median increase in a raw append-and-remove benchmark. This result does not prove user-visible delay.

Validation passed for the focused 30 tests and the `@bb/app` Turbo typecheck. All current GitHub checks also pass.

The final GPT-5.6 review gate returned `REQUEST CHANGES`. I did not use the GitHub request-changes option.

SawyerHood:
Addressed the SlopCop review on this head:

- Async mount/dispose now stay isolated until the promise settles. A timed-out mount drops the fence so a stuck plugin cannot freeze later React inserts.
- Covered `replaceChildren`, `insertAdjacentElement`, and `Range.insertNode`. A detached React node can no longer attach to a foreign parent.
- MutationObserver is a real subclass again, so `new` and `instanceof` match native behavior.
- Plugin-frontend tests now uninstall the guard in `afterEach`.

I am not copying this fence into `@bb/plugin-sdk/testing`. That harness is a unit helper. Isolation is host policy in the app, not an SDK contract.

PR #1457, opened 2026-08-13
Fix side chat activation latency

### LINKED ISSUE #1835: Side chat activation projects 20 timeline segments to read one message
## Summary

Opening a reply side chat on a long thread synchronously projects the default 20 timeline segments before the fork and panel can open, even though side chat reads only the newest conversation message. The unnecessary work adds user-visible activation latency on an interactive path.

Implementation: [PR #1457](https://github.com/get-bb/bb/pull/1457)

## Versions and environment

- Present on `main` at `fae5560b38f6f90255cecb06922b87ac65d210a5`
- Surface: built-in side-chat plugin
- Measured with in-memory SQLite using 40 turns × 30 items

## Steps to reproduce

1. Create or open a thread with at least 20 populated timeline segments.
2. Reply to a message in a side chat.
3. Profile the synchronous work before the side-chat fork is persisted and its panel opens.
4. Observe that `createSideChat` requests the timeline without a `segmentLimit`, so the route projects its default 20 segments.

## Expected vs actual

Expected: activation reads only enough history to determine the newest conversation message.

Actual: the measured fixture projects 620 rows / 455,036 bytes in 8.6 ms on average. Requesting one segment projects 31 rows / 22,785 bytes in 1.2 ms and returns the same seed text.

## Evidence

- The default timeline limit is [20 segments](https://github.com/get-bb/bb/blob/fae5560b38f6f90255cecb06922b87ac65d210a5/apps/server/src/services/threads/timeline.ts#L159).
- Pagination takes the [newest segments](https://github.com/get-bb/bb/blob/fae5560b38f6f90255cecb06922b87ac65d210a5/apps/server/src/services/threads/timeline-pagination.ts#L193).
- The implementation bounds the lookup to [one segment](https://github.com/get-bb/bb/blob/11ef8cff865e3b021aecf0cb025a71953c30cc74/plugins/side-chat/server.ts#L180-L190).

## What you ruled out

- Not a reply-seed semantic change: the newest segment contains the newest conversation row.
- Not a wire or daemon protocol change.
- No exact open or closed issue matched the symbol, PR URL, originating thread, title, or performance mechanism.

Medium priority, low effort: the slowdown scales with thread size; the implementation is a bounded request-shape change with focused regression coverage.

Originating implementation thread: [thr_6zqwdhihs2](https://brsbl.getbb.app/threads/thr_6zqwdhihs2)

brsbl:
## What was wrong

`createSideChat` reads exactly one thing from the source thread's timeline: its last conversation message, which the reply-seed rule compares against the anchor (`resolveReplySeedText` → `lastConversationMessageText`). It requested that timeline with no `segmentLimit`, so it got the route default of [`THREAD_TIMELINE_DEFAULT_SEGMENT_LIMIT = 20`](https://github.com/get-bb/bb/blob/fae5560b3/apps/server/src/services/threads/timeline.ts#L159) segments of fully nested rows. On a long thread that is a large synchronous projection sitting directly on the side-chat activation path, before the fork can be persisted and the panel can open.

No issue was filed for this; it was found while profiling side-chat activation.

## What changed

`plugins/side-chat/server.ts` — the reply-seed timeline lookup now passes `segmentLimit: "1"`.

That is sound because of an invariant on the server side: [`isTimelineSegmentAnchorRow`](https://github.com/get-bb/bb/blob/fae5560b3/apps/server/src/services/threads/timeline-pagination.ts#L101) anchors every logical segment at a user message, and [`paginateTimelineRows`](https://github.com/get-bb/bb/blob/fae5560b3/apps/server/src/services/threads/timeline-pagination.ts#L193) takes `segments.slice(-segmentLimit)` — the *newest* segments. So the newest segment always contains the thread's newest conversation row, user or assistant, and the seed answer is unchanged. `segmentLimit` also bounds the underlying event read via `resolveTimelineWindowBounds` rather than only slicing the projected result, so the work is genuinely avoided and not merely discarded.

No wire changes, no CLI or settings surface, no `HOST_DAEMON_PROTOCOL_VERSION` bump needed.

## How you verified

`plugins/side-chat/server.test.ts` gains a test pinning the request shape — it fails if the `segmentLimit` argument is dropped or changed.

To confirm the projection cost was real rather than assumed, I measured `buildThreadTimeline` directly against an in-memory SQLite thread of 40 turns × 30 items:

| `segmentLimit` | rows | bytes | avg ms |
| --- | ---: | ---: | ---: |
| 20 (route default) | 620 | 455,036 | 8.6 |
| 1 | 31 | 22,785 | 1.2 |

`lastConversationMessageText` returned the identical string in both cases, at that size and at 25 turns × 12 items. That benchmark was scaffolding for this investigation and is not part of the diff.

- `pnpm exec turbo run test typecheck lint --filter=bb-plugin-side-chat` — 29 tests passed, typecheck and lint clean
- `pnpm exec turbo run test --filter=@bb/server` — 1725/1726 passed. The single failure, `plugin-service.test.ts > reports one anonymous plugin_installed event per user install`, is a 5s-timeout flake under full-suite load: it passes in isolation and touches nothing in this change.

Fixes #1835

BB-Thread-ID: thr_6zqwdhihs2


ymichael:
Commandeered this to reproduce the problem and verify the fix. The one-line `segmentLimit: "1"` change is correct and I've kept it; I rewrote the PR description around it and trimmed the rest. Summary of what I changed and why, so nothing is silently dropped:

**Reproduced the cost.** Measured `buildThreadTimeline` directly on an in-memory thread (40 turns × 30 items): 620 rows / 455 KB / 8.6 ms at the default 20-segment limit vs 31 rows / 22.8 KB / 1.2 ms at `segmentLimit: 1`, with an identical `lastConversationMessageText` result. Same shape as the 863-event / 677 KB sample in the original description.

**Confirmed why it's safe.** Every logical segment is anchored at a user message (`isTimelineSegmentAnchorRow`) and `paginateTimelineRows` slices the *newest* segments, so the newest segment always holds the newest conversation row — the only thing the seed rule reads. `segmentLimit` also bounds the event read in `resolveTimelineWindowBounds`, so this avoids the work rather than discarding it after the fact.

**Removed the measurement scaffolding** — the three debug timing logs, the `try/finally` restructure of the fork fallback, and the `panelHydrationStarts` map plus its `useEffect` in `app.tsx`. It was ~55 lines of permanent instrumentation supporting a one-line fix, and the end-to-end timings it produced were dominated by machine contention rather than by this code path, as the original description noted.

One of those lines was also a small bug worth recording: `panelHydrationStarts` was only cleaned up when `openPanel` returned `false`, but the panel-launcher path calls `PluginThreadPanelActionContext.openPanel`, which returns `void` ([app-contract.ts#L353](https://github.com/get-bb/bb/blob/fae5560b3/packages/plugin-sdk/src/app-contract.ts#L353)). Any panel that never mounted would leak its map entry for the life of the session.

**Also merged current `main`** — the branch was well behind. Pushed as a fast-forward, so the original commits are intact.

Net: `+25 −0`, down from `+80 −24`.

PR #1459, opened 2026-08-13
Let UI themes declare matching Pierre code themes

### LINKED ISSUE #1239: Let a UI theme declare its own code theme
## Summary

A custom UI theme can restyle every surface except the code and diff viewers, which stay on the library's default `pierre-dark`. I'd like a UI theme to be able to declare its own code theme.

## Proposed API

One optional field on the appearance schema (`packages/domain/src/app-theme.ts`, today `{ themeId, customCss, faviconColor }`), where each side is **a bundled Shiki theme name or a path to a theme file**:

```ts
codeTheme: z
  .object({
    dark: z.string().optional(),
    light: z.string().optional(),
  })
  .optional(),
```

Both sides are optional and resolved independently: an omitted `dark` falls back to `pierre-dark`, an omitted `light` to `pierre-light`. A dark-only theme declares `dark` alone and light mode is left untouched.

### Built-in code theme

```jsonc
// plugin manifest — bb.themes entry
{
  "id": "my-theme",
  "name": "My Theme",
  "css": "themes/my-theme.css",
  "codeTheme": { "dark": "github-dark", "light": "github-light" }
}
```

```sh
bb theme code-theme github-dark github-light
```

Resolves through Shiki's bundle. Roughly 60 of those themes already ship as lazy chunks in the app (`github-dark`, `nord`, `monokai`, `vitesse-dark`, …) and only the two `pierre-*` are ever requested today, so this case adds no payload.

### Custom code theme shipped with the UI theme

A bundled name gets close, but a UI theme ported from an editor theme wants *its own* colors — so the same field should also take a path, resolved the way `css` already is, relative to the plugin directory:

```jsonc
// plugin manifest — bb.themes entry
{
  "id": "my-theme",
  "name": "My Theme",
  "css": "themes/my-theme.css",
  "codeTheme": { "dark": "themes/code-theme.json" }
}
```

The file is a standard VS Code / TextMate theme JSON — `{ name, type, colors, tokenColors }` — passed to `loadTheme()` and referenced by its registered `name`. Anyone porting an editor theme already has this file, so shipping it alongside the CSS is a copy step rather than new work, and it makes a UI theme self-contained: chrome and code colors travel together and stay in sync.

### Settings: a second, separate selector

Code theme should be its own control, not folded into the UI theme picker. They are independent choices — someone may want one UI theme with a different code theme, and someone running no custom UI theme at all may still want their editor's syntax colors in diffs.

```
UI theme     [ My Theme                  ▾ ]
Code theme   [ Use UI theme's default     ▾ ]
                ├─ Use UI theme's default
                ├─ ── Built-in ──
                ├─ GitHub Dark
                ├─ Monokai
                ├─ Nord
                ├─ Vitesse Dark
                └─ Custom theme file…        (stretch goal)
```

**"Use UI theme's default" is the default value**, and it means: whatever the active UI theme declares in its `codeTheme` manifest field, falling back to the library default if it declares none. So a UI theme that ships code colors gets them automatically on selection, and selecting one that ships none changes nothing — no migration, no surprise for existing users.

### Resolution and failure

Explicit user selection → active UI theme's `codeTheme` → library default. The user's choice wins because it is the more specific act; leaving the selector on "Use UI theme's default" is what hands control back to the UI theme.

Switching UI themes should not silently discard an explicit code-theme choice — keep it set, since it was chosen against the UI theme, not as part of it.

An unknown bundled name, an unreadable path, or a JSON that fails theme validation should log and fall back to the library default rather than failing the render.

## Stretch goal: a code theme without authoring a UI theme

Everything above ties a code theme to a UI theme: the UI theme declares it — bundled name or shipped file — and the user selects the UI theme. That covers theme authors, but not the more common case — someone who likes bb's stock look and just wants their own editor's syntax colors in diffs. Authoring and installing a plugin to carry one JSON file is a lot of ceremony for that.

So: let a code theme be set on its own, pointing at a file the user already has.

```sh
bb theme code-theme ~/Downloads/my-editor-theme.json
```

In Settings, the **"Custom theme file…"** entry opens a native file picker. On selection the file is copied into `<data-dir>/code-theme/<name>.json` — copied rather than referenced, so the theme survives the source file moving or being deleted — and the entry then reads back as that theme's `name`, listed under a "Custom" group alongside the built-ins:

```
Code theme   [ My Editor Theme            ▾ ]
                ├─ Use UI theme's default
                ├─ ── Built-in ──
                ├─ GitHub Dark
                ├─ …
                ├─ ── Custom ──
                ├─ My Editor Theme
                └─ Custom theme file…
```

Anything already in `<data-dir>/code-theme/` is listed too, mirroring how custom UI themes are discovered from `<data-dir>/theme/<name>/theme.css` — so dropping a file in that directory and picking it from the CLI are the same operation as using the picker.

Reading and validating a theme file is already needed for the case above, so what is left here is the file picker, the copy, and discovery — which is why it can land separately, and why it needs no schema change when it does.

SawyerHood:
## Summary

- let a custom or plugin UI theme ship Pierre / VS Code JSON so diffs and file preview match the palette
- make the code theme follow the active appearance instead of adding a second picker
- map each built-in palette to a matching Shiki pair, with first-party light files for Nord and Dracula
- register those themes with Pierre and push them into the worker pool, which ignores per-component `theme` options
- stamp registered theme names so Pierre does not fall back to `pierre-light`

This is a narrower product choice than the original issue: there is no independent user code-theme selector. Switching the UI theme is what selects the code colors.

## Validation

- `pnpm exec turbo run test --filter=@bb/domain --filter=@bb/server --filter=@bb/cli --filter=@bb/sdk`
- `pnpm exec turbo run typecheck --filter=@bb/domain --filter=@bb/app`
- browser QA on the worktree app: Ocean custom theme, Nord light, and Dracula light file preview / diffs

## Risks

Pierre custom themes must use a `name` that matches the registered id. The worker pool is a singleton, so a theme switch has to call `setRenderOptions` or File/FileDiff keep the previous pair.

Closes #1239


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

apps/app/src/hooks/useAppTheme.ts
```diff
@@ -16,7 +17,14 @@ import {
  */
 export function useAppTheme(): void {
   const { data } = useSystemConfig();
-  const css = data?.appearance ? resolveAppThemeCss(data.appearance) : null;
+  const appearance = data?.appearance;
+  const css = appearance ? resolveAppThemeCss(appearance) : null;
+
+  // Register Pierre files during render so child File/FileDiff mounts and the
+  // worker-pool sync effect see the resolved names on the same pass.
+  if (appearance?.resolvedCodeTheme !== undefined) {
```
SawyerHood: 🚨 `slopcop/review` — This call changes shared state and the DOM during React render.

`applyResolvedCodeTheme` writes document attributes and synchronously notifies mounted external-store users. A live config update can therefore update children while the parent render is active. Please move this work to a layout effect or use a React-owned provider with a safe snapshot.

apps/app/src/lib/pierre-worker-pool-theme.ts
```diff
@@ -0,0 +1,44 @@
...
+  useResolvedCodeTheme,
+  useResolvedCodeThemePair,
+} from "@/lib/code-theme";
+
+const registeredFileNames = new Set<string>();
+
+function registerResolvedCodeThemeFiles(): void {
+  const resolved = getResolvedCodeTheme();
+  for (const [name, theme] of Object.entries(resolved.files)) {
+    if (registeredFileNames.has(name)) continue;
```
SawyerHood, line 16: 🚨 `slopcop/review` — Custom code-theme edits keep the first registered file until the page restarts.

The server reuses a stable name after a user edits the JSON and runs `bb theme set` again. This permanent name set skips the new content. Pierre also rejects duplicate registrations. Please use a content-versioned registration name or another replacement mechanism, and add a two-update test.

packages/domain/src/code-theme.ts
```diff
@@ -0,0 +1,164 @@
...
+
+export const codeThemePairSchema = z
+  .object({
+    dark: codeThemeNameSchema,
+    light: codeThemeNameSchema,
+  })
+  .strict();
+export type CodeThemePair = z.infer<typeof codeThemePairSchema>;
+
+export const vscodeThemeJsonSchema = jsonObjectSchema.refine(
```
SawyerHood, line 27: 🚨 `slopcop/review` — A deeply nested theme file can make the system configuration route return 500.

`jsonObjectSchema.safeParse` can throw a `RangeError` during recursive validation. The server calls this parser outside its catch. I reproduced the failure with a 2,000-level object below the 256 KB limit. Please contain parser exceptions, limit object depth, and validate the actual VS Code theme fields before activation.

plugins/github/app.tsx
```diff
@@ -1589,6 +1589,27 @@ function ChecksSection({ checks }: { checks: PullCheck[] }) {
...
 
+function readHostCodeTheme(): { dark: string; light: string } {
+  const root = document.documentElement.dataset;
+  return {
+    dark: root.bbCodeThemeDark ?? "pierre-dark",
+    light: root.bbCodeThemeLight ?? "pierre-light",
+  };
+}
+
+function useHostCodeTheme(): { dark: string; light: string } {
```
SawyerHood: 🚨 `slopcop/review` — Each open patch creates its own document observer.

The host writes both theme attributes before its equality check. Even same-value writes create mutation records. Every observer then stores a new object, so an unrelated config refresh can rerender every visible patch. Please subscribe once at the panel or SDK boundary and share one stable theme snapshot.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: This change lets each app palette choose matching code colors. Built-in themes work, but custom themes can fail or stay stale.

I found four main defects:

1. P1: A deeply nested JSON file can throw a `RangeError` and make `/system/config` return 500.
2. P2: A custom theme edit keeps the first registered JSON until the page restarts.
3. P2: Theme application changes the DOM and an external store during React render.
4. P2: Each open GitHub patch creates an observer, so same-value writes can rerender every visible patch.

The server also accepts malformed files and unknown bundled names. Pierre then rejects them and keeps the prior code theme active.

The configuration path reads theme files synchronously and sends them with every response. Nord added about 26.7 KB in my local check. Two custom files can add about 512,000 characters.

The change duplicates the worker-pool theme provider in two views. The GitHub plugin also duplicates the host theme reader and observer. A shared provider and an experimental SDK hook would reduce this duplication.

QA results:

- The browser changed Default, Nord, and Dracula without a reload. The host published the correct dark and light names.
- The affected Turbo type checks passed.
- The domain and SDK suites passed.
- The focused server and CLI tests passed.
- The desktop smoke test passed alone after a timeout during the parallel suite.
- `git diff --check` passed.

I would not merge this revision until the P1 and P2 defects are fixed. I posted a comment-only review as required.

PR #1470, opened 2026-08-13
Reduce the thread-open request waterfall

### LINKED ISSUE #1303: Opening a thread fires ~19 API requests, including per-open environment git probes
Found while profiling the mobile web frontend against a seeded database (`pnpm seed:perf`, #1300).

## Numbers

Opening one thread at mobile width issued ~19 requests (StrictMode duplicates excluded): thread, timeline, tabs, interactions, queued-messages, prompt-history, default-execution-options, thread-storage file listing, two child-thread list queries (fork + parent), environment status, and environment pull-request.

- Each response is small and fast locally (mostly <60 ms), but on mobile RTTs of 100-300 ms the waterfall serializes into seconds before the view settles.
- `environments/:id/status` and `environments/:id/pull-request` each took ~200 ms even on a warm local daemon — they run git (and PR lookup) work on **every thread open**, including idle threads whose worktree hasn't changed.

## Repro

1. `pnpm seed:perf` on a dev instance.
2. Open any seeded thread with the network panel open; count requests until settle.
3. Repeat with DevTools network throttling set to "Slow 4G" to feel the waterfall.

## Suggested approach

- Bundle the thread-open critical path (thread, first timeline page, tabs, interactions, queued messages, execution options) into one bootstrap response, mirroring `sidebar-bootstrap`.
- Defer environment status, pull-request, and storage listing until the surfaces that show them are opened.
- Cache environment status on the daemon with a short TTL and invalidate from file-watcher events instead of probing per open.

ymichael:
## Summary

- keep thread-open data in its existing granular query/cache owners; this revision does **not** add an aggregate bootstrap endpoint or a shared readiness gate
- defer the thread-storage directory listing until the secondary panel is visible
- cache successful `workspace.status` results in the host daemon for 5 seconds and coalesce concurrent reads
- invalidate that cache synchronously after daemon-owned workspace mutations, from real watcher/runtime events, and by affected path; invalidation also prevents an in-flight stale result from repopulating the cache
- stop content-only worktree events from triggering pull-request lookups; PR state still refreshes when git refs move

## Why there is no new bootstrap

The history has two different "bootstrap" paths:

1. The current thread-detail bootstrap ([`d136d0a60`](https://github.com/get-bb/bb/commit/d136d0a60f660361d5ccbec157292e4a711ba3eb)) is still present. It is the narrow `GET thread?include=environment,host` read, and #909/#915 already made it start in parallel with the timeline.
2. The old `composer-bootstrap` aggregated default execution options, queued messages, prompt history, and pending interactions. [#335](https://github.com/get-bb/bb/pull/335) deliberately removed the app's dependency on it because the aggregate had become a readiness gate for resources with independent cache ownership and invalidation rules.

The first version of this PR recreated the second pattern and hydrated the granular caches from one response. Even though its payload was small, it could overwrite a newer realtime-owned value when the aggregate response arrived late. It also made the patch much larger through a new route, public contracts, SDK declarations, and a readiness context.

This revision removes all of that. It avoids both #1302 failure modes by having no new aggregate payload and no aggregate invalidation/refetch lifecycle at all.

## Measurements

Measured against the existing `pnpm seed:perf` database: 1,200 threads, 402,298 events, 22,939 search segments, 10,344 prompt-history rows, and a 500 MB `bb.db`. The A/B used the same database and same thread pairs, fresh 390x844 Chromium profiles, and a daemon restart on current `main` and this commit. Counts include `/api/v1/` resource timings; "completed" excludes canceled entries (`transferSize === 0`). Settle runs from the first request start to the last completed response.

| Profile | `main` | This PR | Delta |
| --- | --- | --- | --- |
| warm local SPA switch | 20 raw / 18 completed / 13 distinct, 1,903.3 ms settle | 17 raw / 15 completed / 12 distinct, 864.8 ms settle | 17% fewer completed requests; 8% fewer distinct |
| switch with 529 ms network latency | 18 raw / 14 completed / 13 distinct, 1,735.8 ms settle | 16 raw / 12 completed / 12 distinct, 1,617.6 ms settle | 14% fewer completed requests; 6.8% shorter settle |

The warm settle improvement is not all attributable to this patch: provider execution-option discovery varied from 1,824.9 ms on `main` to 782.4 ms here. That endpoint is still the tail, which confirms the issue's git probes are important but not always dominant. The request-count comparison is the reviewable result: the storage request is always absent until its panel opens, while watcher timing determines whether the status/PR repeat is also present. In the warm A/B, `main` issued status twice (107.9/55.8 ms) and PR twice (128.1/87.5 ms); this PR issued each once.

For the daemon cache itself, a status read after TTL expiry took 70.7 ms and the immediate repeated read took 2.6 ms (96% faster). Only `available` results are cached.

The responsiveness sampler uses lateness of a recurring 250 ms timer, not the Long Tasks API that WebKit does not expose. Across 20 samples in the 529 ms profile, maximum observed lateness was 1.8 ms on `main` and 57.2 ms on this run; neither result is reported as a `longtask`-derived zero.

## First-paint decisions

- Environment status and pull-request state stay eager because the composer context banner renders checkout, changed-file, and PR state on open.
- Thread tabs stay eager because persisted tabs can determine the visible secondary surface.
- Thread storage listing is deferred because no first-paint surface consumes its file list while the secondary panel is closed.

## Cache correctness and protocol compatibility

Watcher invalidation alone is eventual. The previous head's red integration job exposed that race: `workspace.commit` could return, then an immediate `workspace.status` could read the pre-commit cache before the watcher fired. This revision invalidates synchronously before successful mutation commands return, while retaining watcher invalidation for provider/external edits. `fake/smoke/environment-actions.test.ts`, which failed on the previous head, now passes all 6 tests.

No server/host-daemon command, result, session, or WebSocket payload changed. This is daemon-internal memoization of the existing result, so `HOST_DAEMON_PROTOCOL_VERSION` remains unchanged.

## Patch size

The original PR was 38 files and +1,523/-129, including 577 generated lines. The reworked PR is 14 files and +606/-38: +278/-37 production and +328/-1 regression tests. There is no server route, public contract, SDK, template, or generated declaration churn.

## Validation

All slow commands ran through Turbo with output captured to files before inspection.

- `@bb/app`: 341 files, 2,644 tests passed
- `@bb/host-daemon`: 47 files, 551 tests passed
- targeted integration regression: 6/6 passed
- Turbo typechecks passed for `@bb/app` and `@bb/host-daemon`
- `@bb/app` lint passed with 0 errors (149 existing warnings)
- focused cache/storage/realtime suites: 90/90 passed

Tests pin lazy storage loading, PR invalidation only for ref changes, cache TTL/coalescing, environment/path/watcher invalidation, stale in-flight invalidation, and immediate post-commit freshness.

Fixes #1303

### DIFF
--- a/apps/app/src/views/thread-detail/ThreadDetailView.tsx
+++ b/apps/app/src/views/thread-detail/ThreadDetailView.tsx
@@ -627,7 +627,10 @@ function ThreadDetailViewInternal(props: ThreadDetailViewInternalProps) {
   );
   const [browserAddressFocusRequest, setBrowserAddressFocusRequest] =
     useState<BrowserAddressFocusRequest | null>(null);
-  const shouldLoadThreadStorageFiles = thread !== undefined;
+  // Thread storage backs only the secondary file panel. Keep its directory
+  // walk off the thread-open path until that panel is actually visible.
+  const shouldLoadThreadStorageFiles =
+    thread !== undefined && isSecondaryPanelOpen;
   const {
     isThreadStorageFilesLoading,
     refetchThreadStorageFiles,
--- a/apps/host-daemon/src/workspace-status-cache.ts
+++ b/apps/host-daemon/src/workspace-status-cache.ts
@@ -1,0 +1,163 @@
+import path from "node:path";
+import type {
+  HostDaemonOnlineRpcResult,
+  WorkspaceContext,
+} from "@bb/host-daemon-contract";
+
+const DEFAULT_WORKSPACE_STATUS_CACHE_TTL_MS = 5_000;
+const DEFAULT_WORKSPACE_STATUS_CACHE_MAX_ENTRIES = 128;
+
+type WorkspaceStatusResult = HostDaemonOnlineRpcResult<"workspace.status">;
+
+interface WorkspaceStatusCacheEntry {
+  environmentId: string;
+  expiresAtMs: number | null;
+  promise: Promise<WorkspaceStatusResult>;
+  workspacePath: string;
+}
+
+export interface WorkspaceStatusCacheLoadArgs {
+  environmentId: string;
+  load: () => Promise<WorkspaceStatusResult>;
+  mergeBaseBranch?: string;
+  workspaceContext: WorkspaceContext;
+}
+
+export interface WorkspaceStatusCacheAccess {
+  getOrLoad(args: WorkspaceStatusCacheLoadArgs): Promise<WorkspaceStatusResult>;
+  invalidateEnvironment(environmentId: string): void;
+  invalidatePath(changedPath: string): void;
+}
+
+interface WorkspaceStatusCacheOptions {
+  maxEntries?: number;
+  nowMs?: () => number;
+  ttlMs?: number;
+}
+
+function buildWorkspaceStatusCacheKey({
+  environmentId,
+  mergeBaseBranch,
+  workspaceContext,
+}: Omit<WorkspaceStatusCacheLoadArgs, "load">): string {
+  return JSON.stringify([
+    environmentId,
+    workspaceContext.workspacePath,
+    workspaceContext.workspaceProvisionType,
+    mergeBaseBranch ?? null,
+  ]);
+}
+
+function pathsOverlap(firstPath: string, secondPath: string): boolean {
+  const firstRelativeToSecond = path.relative(secondPath, firstPath);
+  const secondRelativeToFirst = path.relative(firstPath, secondPath);
+  return (
+    firstRelativeToSecond === "" ||
+    (!firstRelativeToSecond.startsWith("..") &&
+      !path.isAbsolute(firstRelativeToSecond)) ||
+    (!secondRelativeToFirst.startsWith("..") &&
+      !path.isAbsolute(secondRelativeToFirst))
+  );
+}
+
+export class WorkspaceStatusCache implements WorkspaceStatusCacheAccess {
+  private readonly entries = new Map<string, WorkspaceStatusCacheEntry>();
+  private readonly maxEntries: number;
+  private readonly nowMs: () => number;
+  private readonly ttlMs: number;
+
+  constructor(options: WorkspaceStatusCacheOptions = {}) {
+    this.maxEntries =
+      options.maxEntries ?? DEFAULT_WORKSPACE_STATUS_CACHE_MAX_ENTRIES;
+    this.nowMs = options.nowMs ?? Date.now;
+    this.ttlMs = options.ttlMs ?? DEFAULT_WORKSPACE_STATUS_CACHE_TTL_MS;
+  }
+
+  getOrLoad({
+    environmentId,
+    load,
+    mergeBaseBranch,
+    workspaceContext,
+  }: WorkspaceStatusCacheLoadArgs): Promise<WorkspaceStatusResult> {
+    const key = buildWorkspaceStatusCacheKey({
+      environmentId,
+      ...(mergeBaseBranch !== undefined ? { mergeBaseBranch } : {}),
+      workspaceContext,
+    });
+    const now = this.nowMs();
+    const existing = this.entries.get(key);
+    if (
+      existing &&
+      (existing.expiresAtMs === null || existing.expiresAtMs > now)
+    ) {
+      return existing.promise;
+    }
+    if (existing) {
+      this.entries.delete(key);
+    }
+
+    this.pruneExpiredEntries(now);
+    const loadPromise = load();
+    const promise = loadPromise.then(
+      (result) => {
+        // Identity-check the entry so invalidation during an in-flight git
+        // probe cannot let that stale result repopulate the cache.
+        if (this.entries.get(key) === entry && result.outcome === "available") {
+          entry.expiresAtMs = this.nowMs() + this.ttlMs;
+        } else if (this.entries.get(key) === entry) {
+          this.entries.delete(key);
+        }
+        return result;
+      },
+      (error: unknown) => {
+        if (this.entries.get(key) === entry) {
+          this.entries.delete(key);
+        }
+        throw error;
+      },
+    );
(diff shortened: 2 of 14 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I started the review under the `review` rule.

I will check security, code quality, performance, architecture, and the main product route.

apps/host-daemon/src/workspace-status-cache.ts
```diff
@@ -0,0 +1,136 @@
...
+    if (existing) {
+      this.entries.delete(key);
+    }
+
+    this.pruneExpiredEntries(now);
+    const loadPromise = load();
+    const promise = loadPromise.then(
+      (result) => {
+        if (this.entries.get(key) === entry && result.outcome === "available") {
+          entry.expiresAtMs = this.nowMs() + this.ttlMs;
```
SawyerHood, line 106: 🚨 `slopcop/review` — Invalidate the workspace status cache after this commit

This command changes the workspace before it returns. The cache waits for the file watcher, so an immediate status request can receive the old result.

GitHub CI and a local focused test both returned `untracked` after the commit, instead of `clean`.

Please invalidate this environment after the commit completes. Add a test for the command and immediate status sequence.

apps/app/src/hooks/cache-owners/active-thread-lifecycle-cache-owner.ts
```diff
@@ -21,7 +19,6 @@ import {
 
 const ACTIVE_THREAD_QUERY_ROOTS = new Set<unknown>([
```
SawyerHood, line 22: 🚨 `slopcop/review` — Keep the bootstrap in the browser resume set

`pagehide` cancels all active query fetches. This set now omits the open-bootstrap query.

A canceled cold bootstrap stays unsettled. The layout then keeps its dependent queries disabled, because the bootstrap also disables focus and reconnect fetches.

Please restore this query key here. Add a `pagehide` and `pageshow` test for a cold thread open.

apps/app/src/hooks/cache-owners/thread-detail-cache-owner.ts
```diff
@@ -71,4 +78,22 @@ export function ingestThreadDetailBootstrap({
       upsertHostList({ host, hosts }),
     );
   }
+
+  queryClient.setQueryData(threadTabsQueryKey(thread.id), bootstrap.tabs);
```
SawyerHood: 🚨 `slopcop/review` — Do not replace newer real-time state with an older bootstrap snapshot

A real-time event can invalidate one of these keys while the bootstrap request runs. These unconditional writes then clear that invalidation.

The fresh-bootstrap policy also stops the immediate dedicated read. The UI can therefore show an old queue, interaction, tab, or history value.

Please preserve a cache change that occurs after the bootstrap request starts. Add a test for this event order.

apps/app/src/views/thread-detail/ThreadDetailView.tsx
```diff
@@ -560,7 +560,9 @@ function ThreadDetailViewInternal(props: ThreadDetailViewInternalProps) {
     useToggleThreadSecondaryPanelSelection(threadId, threadId);
   const threadDetailBootstrapQuery = useThreadDetailBootstrap(threadId ?? "");
```
SawyerHood, line 561: 🚨 `slopcop/review` — Use one owner for the open-bootstrap query

`AppLayout` already owns this query. This second owner also uses `refetchOnMount: "always"` through the shared hook.

The browser test observed two successful open-bootstrap requests on a cold direct load. A normal in-app open made one request.

Please keep one owner, or change the shared policy so the second observer cannot start another fetch.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

### ELI5

This change puts several small thread-open requests into one box. It opens threads faster, but four timing problems can show old data or stop the view.

### Findings

1. **P1 — A commit can return stale workspace status.** The new daemon cache does not clear after `workspace.commit`. GitHub CI and a local focused run both returned `untracked` instead of `clean`.
2. **P1 — Browser suspension can stop a cold thread open.** `pagehide` cancels the bootstrap request. The resume set omits that query, while dependent queries wait for it to settle.
3. **P2 — An older bootstrap can replace newer real-time state.** The response writes all granular caches without a freshness check. This write can clear an invalidation that occurred during the request.
4. **P2 — A cold direct load can send two bootstrap requests.** `AppLayout` and `ThreadDetailView` both own the query. Both use an always-refetch mount policy.

### Architecture

The server extraction of `readThreadTabs` is a useful shared service. The frontend should also use one owner for the open-bootstrap transition.

The bootstrap loads queued messages twice. `listThreadPromptHistory` loads them again after `buildThreadOpenBootstrap` already loaded them. This is a small refactor opportunity.

### Security

I found no new authorization, path, or data exposure problem. The new route keeps the existing public-thread check.

The status cache keeps at most 128 entries. I did not treat its lazy expiry as a security finding without measured memory evidence.

### Product and performance test

The browser test confirmed the main request reduction. A normal in-app open sent one `open-bootstrap` request and one timeline request.

It sent no dedicated thread, tabs, interaction, queue, prompt-history, default-option, or storage request. A cold direct load sent two successful bootstrap requests and one development-only canceled request.

The test used the documented development server and a real browser. The server stopped after the test.

### Validation

- Five affected package type checks passed.
- All 68 focused app tests passed.
- All 20 focused daemon tests passed.
- The focused server route test passed.
- All 134 SDK and server-contract tests passed.
- The focused integration test failed at the stale post-commit status check.
- GitHub app, server, package, and smoke checks passed. The same integration check failed on GitHub.

I am leaving a comment-only review, as required by this SlopCop run. I did not approve the pull request or request changes through GitHub.

ymichael:
Closing this iteration while we step back and inventory the thread-open request graph before choosing a replacement design. The branch is retained for reference.

PR #1477, opened 2026-08-13
Cancel the live ACP prompt when a turn is steered

SawyerHood:
## Summary

- ACP `turn/steer` now sends `session/cancel` while a prompt is open
- After the cancelled prompt returns, the bridge continues the same BB turn with the steer text
- `thread/stop` still clears the queue and ends the turn as cancelled
- Partial agent output from the cancelled prompt stays in the timeline

ACP v1 has no mid-loop inject. This is cancel-then-prompt, which is the only interrupt the protocol allows.

## Tests

- `pnpm exec turbo run test --filter=@bb/agent-runtime -- --project @bb/agent-runtime:isolated src/acp/bridge/bridge.test.ts`

## Notes

This change does not alter the server and host daemon wire contract.

### DIFF
--- a/packages/agent-runtime/src/acp/bridge/bridge.ts
+++ b/packages/agent-runtime/src/acp/bridge/bridge.ts
@@ -130,6 +130,10 @@ interface AcpThreadSession {
   pendingInstructions: string | undefined;
   activePromptKind: "turn" | "compaction" | null;
   queuedInputs: PromptInput[][];
+  /** True while a session/prompt request is outstanding. */
+  promptRequestPending: boolean;
+  /** True after a steer sent session/cancel for the current prompt. */
+  cancelRequested: boolean;
   loading: boolean;
   loadingSessionId: string | undefined;
   pendingLoadUsageUpdate: AcpUsageUpdate | undefined;
@@ -1226,7 +1230,11 @@ function handlePermissionRequest(
     return;
   }
 
-  if (session.stopping || session.activePromptKind !== "turn") {
+  if (
+    session.stopping ||
+    session.cancelRequested ||
+    session.activePromptKind !== "turn"
+  ) {
     responder.result({ outcome: { outcome: "cancelled" } });
     return;
   }
@@ -1476,6 +1484,8 @@ async function startAgentSession(
     pendingInstructions: params.instructions,
     activePromptKind: null,
     queuedInputs: [],
+    promptRequestPending: false,
+    cancelRequested: false,
     loading: false,
     loadingSessionId: undefined,
     pendingLoadUsageUpdate: undefined,
@@ -1627,6 +1637,36 @@ async function stopSession(session: AcpThreadSession): Promise<void> {
 // Turn loop
 // ---------------------------------------------------------------------------
 
+function requestSteerCancel(session: AcpThreadSession): void {
+  if (
+    session.stopping ||
+    session.cancelRequested ||
+    !session.promptRequestPending ||
+    session.connection.exited
+  ) {
+    return;
+  }
+  session.cancelRequested = true;
+  cancelPendingPermissions(session);
+  session.connection.notify("session/cancel", {
+    sessionId: session.providerThreadId,
+  });
+}
+
+function finishTurn(
+  session: AcpThreadSession,
+  stopReason: z.infer<typeof acpStopReasonSchema>,
+): void {
+  session.activePromptKind = null;
+  session.queuedInputs = [];
+  session.promptRequestPending = false;
+  session.cancelRequested = false;
+  sendNotification(ACP_TURN_COMPLETED_METHOD, {
+    threadId: session.bbThreadId,
+    stopReason,
+  });
+}
+
 function runTurn(session: AcpThreadSession, firstInput: PromptInput[]): void {
   session.activePromptKind = "turn";
   sendNotification(ACP_TURN_STARTED_METHOD, { threadId: session.bbThreadId });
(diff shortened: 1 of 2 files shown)


SawyerHood:
🚨 **SLOP COP** 🚨 · `review`

I am the SlopCop. I am reviewing this change for security, code quality, performance, architecture, and test coverage. I will also test the user flow when the local environment permits it.

packages/agent-runtime/src/acp/bridge/bridge.ts
```diff
@@ -1627,15 +1637,52 @@ async function stopSession(session: AcpThreadSession): Promise<void> {
...
   session.turnSettled = (async () => {
     let input = firstInput;
     for (;;) {
+      if (session.stopping) {
+        finishTurn(session, "cancelled");
+        return;
+      }
+
       let stopReason: z.infer<typeof acpStopReasonSchema>;
+      session.cancelRequested = false;
```
SawyerHood, line 1683: 🚨 `slopcop/review` — A later stacked steer can wait forever.

A second steer can arrive while the first cancellation remains active. The bridge queues it but sends no second cancel. This reset starts the first queued prompt without applying the remaining steer. If that prompt hangs, the later steer cannot run. Please recheck the queue after each prompt starts, or coalesce pending steers. Add a test with a hanging first steer prompt.

packages/agent-runtime/src/acp/bridge/bridge.ts
```diff
@@ -1627,15 +1637,52 @@ async function stopSession(session: AcpThreadSession): Promise<void> {
...
+  if (
+    session.stopping ||
+    session.cancelRequested ||
+    !session.promptRequestPending ||
+    session.connection.exited
+  ) {
+    return;
+  }
+  session.cancelRequested = true;
+  cancelPendingPermissions(session);
```
SawyerHood, line 1650: 🚨 `slopcop/review` — A steer can leave a stale permission interaction.

This call settles only the ACP responder. The related runtime request remains pending. A late server registration can then block later prompts. Please propagate an explicit interaction cancel through the runtime and daemon. Add a race test that steers during permission registration.

packages/agent-runtime/src/acp/bridge/bridge.ts
```diff
@@ -1658,22 +1707,18 @@ function runTurn(session: AcpThreadSession, firstInput: PromptInput[]): void {
         }
         return;
       }
+      session.promptRequestPending = false;
 
-      if (stopReason !== "cancelled" && !session.stopping) {
-        // Steer inputs queued during the prompt continue the same bb turn.
+      // Hard steer cancels the current prompt, then continues this bb turn.
+      if (!session.stopping) {
         const next = session.queuedInputs.shift();
```
SawyerHood, line 1720: 🚨 `slopcop/review` — The next prompt has no adapter boundary.

This loop starts the steered prompt without closing items from the canceled prompt. The final `end_turn` can mark an old open tool as completed. Assistant and thought items can also cross the user steer. Please emit an internal prompt boundary. It must interrupt old tools and close old text items without closing the BB turn. Add adapter assertions for item status and item identity.

SawyerHood (commented):
🚨 **SLOP COP** 🚨 · `review`

ELI5: BB now stops an ACP answer when you steer it. BB then sends your new message inside the same turn.

I found four state-boundary defects:

- High: A canceled permission can leave a server interaction open. That interaction can block later prompts.
- High: A canceled tool can later appear completed after the steered prompt ends.
- Medium: Assistant and thought items can cross the user steer. The timeline can join output from two prompts.
- Medium: A later stacked steer can wait forever behind a hanging first steer prompt.

The main fix needs an internal prompt boundary. This boundary must interrupt old tools and close old text items.

The permission path also needs an end-to-end cancel signal. A shared cancel helper can reduce duplicate stop and steer logic.

I found no new direct security defect. A started side effect can still finish because ACP cancellation is cooperative.

I found no busy loop or new synchronous hot path. Stale requests and stacked prompts can add unlimited delay.

Validation passed:

- The forced ACP bridge suite passed all 59 tests.
- The full agent-runtime suite passed all 932 tests.
- The agent-runtime type check passed.
- The diff check passed.
- A real OpenCode ACP run accepted a steer after 22 seconds.
- The source app and browser showed the final `STEERED` response.

I recommend fixes before merge. I submitted this as a comment-only review.

packages/agent-runtime/src/acp/bridge/bridge.ts
SawyerHood: Fixed in 62aec49. The turn loop now re-checks `queuedInputs` right after it sends the next `session/prompt`, and sends a second `session/cancel` when a steer is still queued. A new test stacks two steers where the first steered prompt also hangs; it times out without the fix.

packages/agent-runtime/src/acp/bridge/bridge.ts
SawyerHood: Not in this change. `cancelPendingPermissions` already had this shape on the stop path (`stopSession`), so the steer path does not add a new class of bug. An explicit interaction cancel must cross the runtime and host-daemon boundary and needs a `HOST_DAEMON_PROTOCOL_VERSION` bump, which is a separate change. The late registration is also settled: the `sendRuntimeRequest` continuation checks `pendingPermissions.delete(pending)` and drops the result, and `handlePermissionRequest` answers `cancelled` while `cancelRequested` is true, so a new request cannot attach to the dead prompt.

packages/agent-runtime/src/acp/bridge/bridge.ts
SawyerHood: Not in this change. ACP puts this duty on the agent: after `session/cancel` the agent must mark its pending tool calls cancelled and close its items before it returns `stopReason: "cancelled"`. The bridge waits for that response before it starts the steered prompt, so a compliant agent closes its own items. A bridge-side synthetic boundary would duplicate that work and could conflict with the agent's own updates. If a real agent leaves items open, that deserves its own change with the offending agent named.

PR #1468, opened 2026-08-13
Add company social proof to landing page

brsbl:
## Summary

- add a full-width, low-contrast “Used by builders at” marquee after the product demo and before the feature narrative
- preserve the existing hero, “Works with” compatibility row, and surrounding landing-page content
- include a keyboard-accessible pause/resume control and a static, single-list reduced-motion fallback
- normalize authentic company assets through grayscale and opacity while preserving each mark's proportions

The wording describes individual use and deliberately does not claim that any listed company uses, endorses, or is a customer of BB.

## Affiliation evidence

Publicly documented BB use and current-affiliation chains:

- **Meta:** [BB issue #499](https://github.com/get-bb/bb/issues/499); [GitHub profile](https://github.com/krishnakumar-kapil) → [personal site](https://krishnakumar-kapil.github.io/) → [LinkedIn](https://www.linkedin.com/in/krishnakumarkapil)
- **Render:** [BB issue #1041](https://github.com/get-bb/bb/issues/1041); [GitHub profile](https://github.com/Talor-A) → [personal site](https://taloranderson.com/) → [LinkedIn](https://www.linkedin.com/in/taloranderson)
- **Shortcut:** [BB PR #1171](https://github.com/get-bb/bb/pull/1171); [GitHub profile](https://github.com/charpeni) → [personal site](https://charpeni.com/) → [LinkedIn](https://www.linkedin.com/in/nicolas-charpentier-8a2b8a104)
- **Blackstone:** [BB issue comment](https://github.com/get-bb/bb/issues/1114#issuecomment-5222238371); [GitHub profile](https://github.com/ryanbbrown) → [personal site](https://ryanbbrown.com/) → [current-status page](https://ryanbbrown.com/now/)
- **Moody's:** [BB issue #1431](https://github.com/get-bb/bb/issues/1431); [GitHub profile](https://github.com/jerrison) → [LinkedIn](https://www.linkedin.com/in/jerrison)
- **Pendo:** [BB issue #1191](https://github.com/get-bb/bb/issues/1191); [GitHub profile](https://github.com/davekilleen) → [GitHub-owned Dex repository](https://github.com/davekilleen/Dex) → [LinkedIn](https://www.linkedin.com/in/davekilleen)

Explicitly owner-confirmed additions:

[...]

PR #1485, opened 2026-08-13
Restore the landing-page company marquee with a wide-screen fix

SawyerHood:
## Summary

- Reapply the company social-proof marquee from #1468, which #1482 reverted.
- Fix the wide-screen bug that caused the revert: the marquee went blank near the end of each loop.

## Root cause

The track held exactly two logo copies and scrolled by one copy width per cycle. One copy is about 1307px wide. When the viewport is wider than one copy, the track runs out of content near the end of each cycle, and the band goes blank.

## Fix

- Measure the marquee and one logo list with a `ResizeObserver`. Render `ceil(viewport / copyWidth) + 1` copies, with a minimum of 2. Set the count as `--company-proof-copies` on the track.
- The keyframe now translates by `calc(-100% / var(--company-proof-copies, 2))`, so one cycle always travels exactly one copy and the loop stays seamless at every width.
- The animation still moves only `transform`, and the existing `IntersectionObserver` still pauses it off screen. Resize re-renders occur only when the copy count crosses a threshold.

## Verification

- `pnpm exec turbo run typecheck --filter=@bb/web`
- Headless Chromium checks at 390 / 1440 / 1920 / 3440px: 2 / 3 / 3 / 4 copies; the track covers both viewport edges at 21 sample points across 4 animation loops at each width.
- Screenshot at 1920px at 97% of a cycle, the point where the old code went blank, shows a full band.


SawyerHood:
Fixed in 32607402d: the page now server-renders five logo copies, enough to cover a 5K display before hydration or without JavaScript. The `ResizeObserver` measurement trims the count after hydration (verified: 3 copies at 1920px, full coverage across the animation cycle).

PR #1502, opened 2026-08-13
feat(browser): add find in page

smsunarto:
## Summary

Add Find in Page to embedded desktop browser tabs.

- Open with `⌘F`.
- Search as the user types.
- Show the current and total match count.
- Navigate with buttons, Enter, or Shift+Enter.
- Close with Escape or the close button.

## Design rationale

- Use Chromium's native Find in Page engine through Electron's `webContents.findInPage`.
- Do not reimplement text matching, highlighting, or match navigation in BB.
- Render only the Find bar in BB because `WebContentsView` does not include browser chrome.
- Use BB design-system tokens and existing browser controls for the interface.
- Route `⌘F` through BB's command system so it works from BB chrome and the embedded page.
- Scope requests and results by browser tab ID.
- Keep the new desktop APIs optional so older desktop shells remain compatible.
- Validate Find requests and results at the desktop IPC boundary.
- Clear native highlights when the bar closes or the page changes.

## Testing

- 61 focused app, desktop, desktop-contract, and server tests pass.
- Turbo typecheck passes: 4 tasks.
- Turbo build passes: 8 tasks.
- Verified real-time search, match counts, match navigation, and closing in the live Electron app.
- Verified two-tab isolation in both directions with deterministic pages.

> AGENT GENERATED: by Claude Mythos 5


ymichael:
Superseded by #1938

PR #1504, opened 2026-08-13
[issue sweep] Reject home-relative markdown links

### LINKED ISSUE #1182: Chat links starting with ~/ resolve against the project root
## Summary

A chat message that contains a path like `~/.config/example.md` renders as a clickable link. The click opens a file tab in the secondary panel. From that tab, **Open in editor** fails, because the path bb resolved is `<project root>/~/.config/example.md`.

## Steps to reproduce

1. Open a thread whose environment has a workspace, for example `/Users/me/bb`.
2. Post a message that contains `~/.config/example.md`.
3. Click the link. A file tab opens.
4. Click **Open in editor**.

**Expected:** the link resolves to the home directory, or bb leaves it as plain text.

**Actual:** bb joins the literal `~` segment to the workspace root. The editor reports that it cannot open `/Users/me/bb/~/.config/example.md`.

## Cause

`resolveRelativeLocalFileHref` in [`apps/app/src/components/ui/markdown-local-file-link.ts`](https://github.com/get-bb/bb/blob/main/apps/app/src/components/ui/markdown-local-file-link.ts) treats *every* non-absolute href as workspace-relative. It rejects absolute paths, fragments, queries, and URI schemes, but it does **not** handle a leading `~/`.

The join at line 320 produces the literal path above. That path passes `isAbsoluteFilePathWithinRoot`, so `resolveThreadLocalFileLink` classifies the link as a workspace file and opens a workspace tab.

No code expands `~` in this path. Tilde expansion exists in `apps/cli`, `apps/desktop/src/main.ts`, `apps/host-daemon`, and `packages/config`, but *not* in the renderer link code. The renderer has no home directory value at all.

## Possible fixes

1. **Reject a leading `~/`** in `resolveRelativeLocalFileHref`. The link stays plain text. Small and safe.
2. **Expand `~/` to the host home path.** This needs the home path in the renderer, and it routes the result as a host file link. Larger, and it needs a host-daemon field.

I plan to send a pull request for option 1, unless you prefer option 2.

SawyerHood:
## Summary

- Reject Markdown links whose decoded path starts with `~/`.
- Keep these links as plain text.
- Do not expand home paths.

## Testing

- `pnpm exec turbo run test --filter=@bb/app --force`

Fixes #1182

> AGENT GENERATED: by GPT-5

### DIFF
--- a/apps/app/src/components/ui/markdown-local-file-link.test.ts
+++ b/apps/app/src/components/ui/markdown-local-file-link.test.ts
@@ -288,6 +288,16 @@ describe("resolveRelativeLocalFileHref", () => {
     ).toBeNull();
   });
 
+  it("rejects home-relative paths", () => {
+    expect(
+      resolveRelativeLocalFileHref({
+        baseDir: "/workspace",
+        href: "~/.config/example.md",
+        rootPath: "/workspace",
+      }),
+    ).toBeNull();
+  });
+
   it("rejects relative paths that escape the containing root", () => {
     expect(
       resolveRelativeLocalFileHref({
--- a/apps/app/src/components/ui/markdown-local-file-link.ts
+++ b/apps/app/src/components/ui/markdown-local-file-link.ts
@@ -297,6 +297,7 @@ export function resolveRelativeLocalFileHref({
     parsedHref === null ||
     parsedHref.path.length === 0 ||
     parsedHref.path.startsWith("/") ||
+    parsedHref.path.startsWith("~/") ||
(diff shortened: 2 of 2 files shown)


SawyerHood:
Closing this. @arunsathiya was assigned to #1182 and had said they would send a pull request for option 1. An overnight issue sweep picked the issue up without checking the assignee, which was my mistake. The floor is theirs.

Leaving the branch in place in case it is useful as a reference, but this pull request should not be merged.

PR #1508, opened 2026-08-13
[issue sweep] Add a preference for assistant message visibility

### LINKED ISSUE #1355: With a stop hook, most of the Assistant text is hidden inside "Worked for [time]"
When assistant turn contains multiple text blocks, e.g. using a stop hook, `bb` treats **only the last one** as the message, and collapses everything before it into a "Worked for..." section. 

So the visible message is missing a lot, and is often just a reply to the hook failure.

The result is that the answer looks missing. I had to notice the little chevron and expand the 'worked for' section to read the rest.

## Repro

1. Configure a `Stop` hook that blocks — exit `2` with text on stderr, or print `{"decision":"block","reason":"…"}`
2. Ensure its blocking condition is persistently true, so it fires on every turn
3. Ask any question

Claude answers (text block 1) → the Stop hook vetoes the stop and its reason is injected as a user message → Claude emits a short acknowledgement (text block 2) → bb shows only block 2.

Stop hooks are the easiest trigger, but nothing here is hook-specific: any turn ending in `text → tool_use → text` has the same shape.

### Collapsed
What I actually see. The visible text is the model answering the hook, not the user:

Image

### Expanded
The looooong  answer to my questions, which was hidden under "Worked for 45s":

Image

### Another example, collapsed

Image

### Expanded 
Including explicit questions the user was being asked to decide on :(

Image

## Transcript evidence

From `~/.claude/projects/<project>/<session>.jsonl`, one turn:

| row | type | content |
|-----|------|---------|
| 692 | assistant | text — *"Fair pushback. Taking these in order…"* ← the answer |
| 693 | user (`isMeta: true`) | `Stop hook feedback: … no fresh fast-loop verdict for HEAD (5bec6068…)` |
| 695 | assistant | text — *"The verify gate is still open. Nothing changed this turn…"* ← the only part rendered |

Rows 702/709 are the second screenshot, identical shape.

## Expected

Every assistant text block in a turn should be visible at rest. Tool calls and their results are reasonable things to collapse; prose addressed to the user is not.

Possible suggestions to solve it:

1. Render all text blocks in sequence; collapse only tool activity between them
2. Keep the collapsible but leave text blocks outside it, in order
3. If a heuristic must pick one block, pick the longest rather than the last — a 40-word acknowledgement after a 600-word answer is not the message

## Notes

- Claude Code's own REPL prints every text block in order, so the same session reads correctly there and wrong in bb. Same data, different grouping.
- The failure is silent and worse the more the user relies on hooks. A user who doesn't know to expand the "Worked for" section just sees an assistant that stopped answering questions.
- Labelling matters too: "Worked for 45s" reads as a duration stat, not as a container holding the reply. Even with the grouping unchanged, a label that hints at hidden prose would reduce the damage.

**Environment:** macOS 15 (Darwin 25.5.0), bb desktop, Claude Code via the Agent SDK.

SawyerHood:
## Summary

A finished turn kept only its **last** assistant message visible and collapsed every earlier one into the "Worked for …" summary. A turn shaped `text -> tool_use -> text` therefore hid the long answer and showed only the short reply after the tool call. A blocking Stop hook makes that shape common in Claude Code, but the cause is not hook-specific.

This pull request does two things.

1. Fixes the hidden answer: a finished turn now renders every assistant message in order and collapses only the tool activity.
2. Puts that behavior behind one app-wide General preference, `showAllAssistantMessages`, because the collapsing works well for Codex today and the change would otherwise make every Codex turn noisier.

The preference is global, not per provider. It defaults to **true**, which keeps the fix on for everyone. Set it to false to get today's behavior back: the turn then shows only its last assistant message and keeps the earlier ones inside the collapsed work summary.

### The decision most worth challenging

The default. Hiding an answer is a worse failure than extra verbosity, and a preference that defaults to the hiding behavior leaves the #1355 bug on for everyone who never finds the setting. A reviewer who disagrees can flip one value in `packages/domain/src/app-settings.ts` and one default in `packages/db/src/schema.ts`.

### Surfaces

Following the `showSidebarThreadNumbers` pattern from #1174:

- `packages/domain/src/app-settings.ts`: schema field and default.
- `packages/db`: `app_settings.show_all_assistant_messages`, migration `0094_absent_scrambler` generated with Drizzle, and the rewind-column helper in `migrate.test.ts`.
- Server: `PUT /api/v1/settings/general` and `GET /api/v1/system/config` carry it through the existing app-settings contract.
- App: Settings → General → "Show all assistant messages".
- CLI: `bb settings general showAllAssistantMessages <true|false>`.
- Docs: `docs/configuration.md`, the `bb-cli` SKILL, `references/app-settings.md`, and the `bb-guide-customization` template.

### Reaching every consumer

The grouping happens in `@bb/thread-view`, which the server drives through `buildThreadTimelineFromEvents`. The preference is a required field on every server entry point, so the compiler rejects a consumer that omits it:

- the timeline route, including its cache key and its delta params key,
- the work summary details route,
- the conversation outline, including its cache key,
- `bb thread show`, which renders rows from the timeline route.

A consumer that read a different value would describe a different turn. A summary row's source-sequence range changes with the preference, so the details route would fail to match the range the client asked about. `timeline-assistant-message-visibility.test.ts` pins that: details resolve under the matching value and throw under the other one.

### Toggling while a long thread is open

Segment cursors anchor on user-message rows, which both modes produce identically, so an open older-page cursor stays valid and no page has to be re-anchored.

Writing the preference invalidates the timeline, the work summary details, and the conversation outline, so an open thread re-renders in the new shape without a reload. The window that made the change invalidates locally; other windows do the same through the realtime `config-changed` broadcast. The outline was missing from the local path, which would have left the minimap disagreeing with the timeline; this pull request adds it and covers it with a test.

That refetch returns the new shape whole: the preference is part of the cache key and of the delta params key, so the delta path cannot match a snapshot taken under the other value and the server sends the full window instead of a diff. Older pages already loaded in the client are re-requested by the same invalidation, and segment cursors anchor on user-message rows, which both modes produce identically, so no page has to be re-anchored. No row set is ever a mix of the two modes, and no message is lost in either.

### Protocol version

Nothing here crosses the server and host daemon channel: the preference is read on the server and only changes rows sent to clients over the web API. `HOST_DAEMON_PROTOCOL_VERSION` is unchanged.

## Testing

- `packages/thread-view`: the `text -> tool_use -> text` turn, two adjacent text blocks, an assistant message between two commands, and the collapsed counterpart of each. Legacy bb system prose is covered in the collapsed mode, where it must stay visible.
- `apps/server`: new `timeline-assistant-message-visibility.test.ts` covers both modes across the timeline, the work summary details, and the conversation outline, plus the cross-value mismatch. `general-settings.test.ts` round-trips the new field. `timeline-cache.test.ts` carries it in the key.
- `apps/cli`: `bb settings general showAllAssistantMessages false` sends the full settings contract.
- `apps/app`: `system-cache-effects.test.ts` asserts a General settings write invalidates the timeline, the outline, and the work summary details together.
- `packages/db`: migration replay and rewind scenarios pass with the new column.
- `pnpm exec turbo run typecheck` passes for `@bb/thread-view`, `@bb/server`, `@bb/app`, `@bb/cli`, `@bb/db`, `@bb/domain`, `@bb/plugin-sdk`, and `@bb/templates`.
- `pnpm exec turbo run test --force` passes for `@bb/thread-view` (21 files), `@bb/app` (342), `@bb/cli` (47), `@bb/db` (29), `@bb/domain` (24), `@bb/integration-tests` (25), `@bb/templates` (4), and `@bb/plugin-sdk` (9).
- `@bb/server`: 162 of 163 files pass. `internal-skill-trees` compares a file mode (420 vs 436); it also fails on `main` in this worktree and passes in CI, so it is a local umask difference.
- `pnpm exec turbo run lint` passes for the touched packages.
- The regenerated `bb-plugin-sdk.d.ts` also reorders some unrelated enum members. That is the generator's own output for the new schema key; `build-bundled-dts.mjs --check` only passes with it.

Fixes #1355


SawyerHood:
Thank you for the review. Here is how I answered each note.

**Inline: the outline can grow with every assistant block.** I added `lists every assistant block of a long text and tool alternation` to `apps/server/test/services/threads/conversation-outline-performance.test.ts`. It builds a turn with 40 text and command pairs and asserts one outline item per block.

I did not add a limit, pagination, or virtualization. The outline is a mirror of the visible conversation rows, so a cap would hide the same prose this pull request makes visible. A route that reads every event without pagination is a pre-existing property of `buildThreadConversationOutline`; the growth factor here is the number of text blocks in a turn, not a new unbounded scan. A limit is a product decision about the minimap, so it belongs in its own issue.

**`isLegacyUserMessage` has no runtime reader.** Correct, and my change caused it. I removed the field from `EventProjectionAssistantTextMessage` and from `parseLegacyUserMessage`. A legacy manager message is assistant text, so it now follows the same visibility rule. The event-level tests for legacy messages stay in place.

**A direct test for two assistant blocks without a tool.** Added: `keeps both text blocks visible when a turn has no tool activity` in `packages/thread-view/test/completed-turn-summary-rendering.test.ts`.

Verification after these changes: `@bb/thread-view` (21 files), `@bb/app` (339 files), `@bb/cli` (47 files), and `@bb/integration-tests` (25 files) pass. `@bb/server` passes except the pre-existing `internal-skill-trees` file-mode failure, which also fails on `main` in this worktree and passes in CI. Typecheck passes for all four packages.

SawyerHood:
Thank you for the review, and for the dev-app product test. Two blockers, two notes.

## Correctness: an open work summary across a preference change — fixed

You are right, and the failure is reachable. A General settings write invalidates the timeline query and the details query together, and their refetch order is not guaranteed. A details refetch that lands first carries the old row range into the new policy, and the request failed.

I fixed it in `98c8e209e`, but not by adding the policy to the request. A client can hold a stale range from more than one source — a second window, a background refetch, an SDK caller that kept a row — so I made the lookup tolerate any of them: `buildTimelineTurnSummaryDetails` now matches the requested range under the other preference before it fails. The open row keeps working until the client renders the new rows, and no wire field can be sent out of step with the row it describes.

`timeline-assistant-message-visibility.test.ts` covers both directions: a row built under either preference resolves through a details request that reads the other one. That test previously asserted the throw; it now asserts the recovery.

## Performance: the default — kept, with numbers

I reproduced your measurement. One turn of N assistant/tool pairs on one page, both modes:

| pairs | show-all rows | show-all KB | show-all ms | compact rows | compact KB | compact ms |
|---|---|---|---|---|---|---|
| 10 | 20 | 6 | 11 | 3 | 1 | 2 |
| 50 | 100 | 29 | 4 | 3 | 1 | 4 |
| 200 | 400 | 116 | 8 | 3 | 1 | 11 |
| 750 | 1500 | 435 | 23 | 3 | 1 | 18 |

Your 1,500 rows and ~461 KB are right. Two things the numbers also show: the projection itself costs about the same in both modes (23 ms against 18 ms at 750 pairs), and the cost you name is response size plus the loss of the 200-row cache entry.

I am keeping the true default anyway, and this is a maintainer decision rather than my own. The pull request body states the reasoning and flags it as the decision most worth challenging: hiding an answer is a worse failure than extra verbosity, and a preference that defaults to the hiding behavior leaves #1355 on for everyone who never finds the setting. The behavior you measured is now reachable in one click for any user who wants it back.

I did not add a bounded assistant-message page. A bound re-hides prose silently at an arbitrary cut-off, which is the defect this pull request exists to remove. A reviewer who wants the other trade-off has two one-line levers: the default in `packages/domain/src/app-settings.ts` with the column default in `packages/db/src/schema.ts`, or `DEFAULT_MAX_CACHEABLE_ROWS` in `timeline-cache.ts`, which is a cache heuristic rather than a correctness limit. I left the cap alone because raising it trades bounded memory across 128 entries, and that is a separate decision from this issue.

## Notes — both fixed

- The cache key test now covers `showAllAssistantMessages: false`.
- The agent reference said `bb thread show`. `bb thread log` is the command that renders the timeline; corrected.

Verification after these changes: `@bb/server` passes 162 of 163 files, the exception being the pre-existing `internal-skill-trees` file-mode check that also fails on `main` in this worktree and passes in CI. Typecheck and Prettier are clean.

SawyerHood:
Thank you. All three findings were real, and your synthesis named the cause better than my patch did: the row shape did not travel with the rows it describes. `c29837744` makes it travel.

**The shape is now part of the contract.** `ThreadTimelineResponse` reports `showAllAssistantMessages`, and a client asking for a delta sends `afterShowAllAssistantMessages` — the shape its held rows were built under.

1. **Delta across a shape change.** The server computes a delta only when the client names the current shape. A client that names the other one, or names none, gets the full window. Your scenario now ends with window A replacing its compact rows instead of merging an empty patch. `public-thread-timeline-delta.test.ts` covers all three cases.

2. **Loaded older pages.** `useThreadTimelineController` keeps the shape in `LoadedTimelineState`. When the latest response reports a different one it drops the loaded rows and rebuilds from that response, which is the existing reset branch. An older page that arrives under a different shape is discarded rather than prepended, so two shapes never render together. Covered by a merge test with one loaded older page.

3. **The stale-summary fallback.** You are right that `ungrouped` was accepted too early. A matched range under the previous shape now beats an ungrouped window under the current one, and the ungrouped fallback runs only when neither shape matches. I added the two-message, no-tool regression test you asked for; it fails on the previous commit and passes now.

I did not add a separate policy identifier beyond the boolean itself. The value is the identity here, it is already in the cache key and the params key, and one more indirection would not have made these three boundaries any safer.

Note for anyone rebasing on this: `afterSequence` without `afterShowAllAssistantMessages` now yields a full window rather than a delta. Omission has a real meaning — the caller cannot name the shape it holds — and a full window is the only safe answer.

The earlier default-on concern stands where I left it: the measurements are in [my previous reply](https://github.com/get-bb/bb/pull/1508#issuecomment-5282591250), and the default is the maintainer's decision recorded in the body.

Verification: `@bb/app` 342 files, `@bb/cli` 47, `@bb/thread-view` 21, `@bb/integration-tests` 25, `@bb/templates` 4 all pass. `@bb/server` passes 162 of 163, the exception being the pre-existing `internal-skill-trees` file-mode check that also fails on `main` in this worktree and passes in CI. Typecheck passes across the affected packages, and the regenerated plugin SDK types and templates are fresh.

SawyerHood:
Closing without merging. This came out of the 13 Aug 2026 overnight issue sweep. Sawyer is closing the unmerged sweep pull requests rather than carrying them.

The branch is left in place, so this can be reopened or cherry-picked. The review history, the measurements, and the answers to each Slop Cop finding stay on this pull request.

PR #1406, opened 2026-08-12
Add a keyboard shortcut for archiving the focused thread

AndrewSB:
Superseded by #1516 — same branch, reopened there after the PR head failed to track a new push. Closed.

PR #1516, opened 2026-08-13
Add a keyboard shortcut for archiving the focused thread

AndrewSB:
## What

Adds a `thread.archive` app command bound to **Alt+A** by default. It archives the thread you are currently viewing (and its children) via the existing `useThreadActions().archiveThreadAndChildren`, so it reuses the pane-close, URL reconciliation, and undo-toast behavior of the menu item.

- `packages/domain` — new `thread.archive` command id.
- `apps/server/.../app-keybindings.ts` — default binding, scoped to `mainSurface` and suppressed while a modal, terminal, or embedded browser has focus. It intentionally still fires while the composer is focused, matching the existing Alt chords.
- `apps/app/.../ThreadDetailView.tsx` — handler gated on `isFocused`, so in a split it archives the focused pane's thread. Returns `false` for an already-archived thread (archive-only; it will not silently un-archive) and when unfocused, letting the key event fall through.
- `apps/app/src/lib/app-command-metadata.ts` — "Archive thread" entry so it is listed and rebindable in Settings → Keyboard.
- `packages/plugin-sdk/bundled-types/bb-plugin-sdk.d.ts` — regenerated, since it embeds the `AppCommandId` union. The diff also contains incidental zod enum key reordering produced by the current generator.

## Why Alt+A

Every `Mod+Shift+<letter>` chord that reads as "archive" is either already claimed by bb or reserved by the browser — Chrome binds `Mod+Shift+A` to tab search. Alt is the family this codebase already uses for its own actions (`modelPicker.cycleModel`/`cycleReasoning`), it is unclaimed by the browser and both desktop menus, and it needs no separate web/desktop alias. macOS composes Option+A into "å", which `normalizeAppShortcutInputKey` already handles by falling back to the physical key.

## Tests

- `apps/server/test/system/app-keybindings.test.ts` asserts the default chord and its scope. The existing "no other binding uses Alt" assertion now enumerates each Alt binding with its key, which is the invariant that actually matters: distinct keys mean no Alt command can be shadowed by an earlier binding for the same chord.
- `SplitThreadArea.stories.tsx` renders the real `ThreadDetailView`, which now reads the thread actions, so the story is wrapped in `ThreadActionsProvider`.

Typecheck passes for `@bb/domain`, `@bb/plugin-sdk` (including `build-bundled-dts.mjs --check`), `@bb/server`, and `@bb/app`. `@bb/app` tests: 321 files / 2414 tests pass. `@bb/server` tests pass except three plugin install/update files that fail in this sandbox on `npm error code EALLOWSCRIPTS`, unrelated to these changes.

### DIFF
--- a/apps/app/src/lib/app-command-metadata.ts
+++ b/apps/app/src/lib/app-command-metadata.ts
@@ -48,6 +48,11 @@ export const APP_COMMAND_GROUPS: readonly AppCommandGroup[] = [
         "Next thread",
         "Open the next visible sidebar thread.",
       ),
+      command(
+        "thread.archive",
+        "Archive thread",
+        "Archive the thread you are viewing and its child threads.",
+      ),
       ...THREAD_JUMP_APP_COMMAND_IDS.map((id, index) =>
         command(
           id,
--- a/apps/app/src/views/thread-detail/SplitThreadArea.stories.tsx
+++ b/apps/app/src/views/thread-detail/SplitThreadArea.stories.tsx
@@ -16,6 +16,7 @@ import {
 import { maximizedPaneIdAtom, splitLayoutAtom } from "@/lib/split-layout/atoms";
 import type { SplitLayout } from "@/lib/split-layout";
 import { SidebarProvider } from "@/components/ui/sidebar";
+import { ThreadActionsProvider } from "@/components/thread/ThreadActionsProvider";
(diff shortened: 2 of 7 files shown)


AndrewSB:
#1406 has a reason to close this. it's already been done upstream

PR #1507, opened 2026-08-13
[issue sweep] Show active workspace in composer footer

SawyerHood:
## Summary

- Show the active workspace directory beside the project name when the environment path differs from the host-specific project root.
- Use the full active environment path in the shared project label tooltip.
- Keep the label current through the existing live environment detail query.

## What it looks like

These images come from the real dev app and a real Codex thread. The thread started in `project-a` and changed its environment to `project-b`. I did not mock UI data.

| Before the switch | After the switch |
| --- | --- |
|  |  |

The tooltip shows the full active environment path.

The label truncates a long directory name without moving the other footer controls.

The screenshots live on the throwaway branch `screenshots/issue-1237`. They do not add binary files to this pull request.

## Testing

- `pnpm exec turbo run typecheck --filter=@bb/app`
- `pnpm exec turbo run test --filter=@bb/app --force`
- `pnpm exec turbo run lint --filter=@bb/app`
- Used a real Codex thread in `scripts/bb-dev-app current` to change the environment directory without a page reload.

Fixes #1237

> AGENT GENERATED: by GPT-5

PR #1527, opened 2026-08-13
[issue sweep] Load the diff renderer and KaTeX on demand

### LINKED ISSUE #1072: Thread route still statically imports the 2.15 MB shared chunk (+ katex): FilePreview → @pierre/diffs, markdown-preview → rehype-katex, PromptBox → tiptap should be lazy

SawyerHood:
## Summary

The thread page downloaded the diff renderer, Shiki and KaTeX before it could show anything. `@pierre/diffs` reached the thread route through four static imports and `rehype-katex` reached it through `markdown-preview`. None of that code is on the boot path, so the existing boot budget could not see it.

### The issue's diagnosis is wrong, and #1072 should be corrected

**`@/lib/workspace-checkout-display` never imported the editor.** The whole module is 85 lines: one `import type { GitCheckoutRef } from "@bb/domain"`, a `WorkspaceCheckoutDisplay` interface, and a `switch` that returns label strings. It has no value imports at all, so it cannot pull anything into a bundle.

**The 1.7 MB was a chunk name, not an import edge.** Rolldown names a shared chunk after one of the modules inside it. `workspace-checkout-display.ts` is shared by five lazy entries, so it landed in the big shared chunk and the chunk took its name. The issue read that name as a cause. The chunk's actual contents on the base commit were `@pierre/diffs` (645 KB), the promptbox composer's TipTap and ProseMirror (~450 KB), Shiki (~240 KB), and the markdown pipeline.

**So the four listed importers are innocent.** `ThreadDetailView.tsx`, `ThreadMetadataContent.tsx`, `PluginThreadChat.tsx` and `ThreadEnvironmentSummary.tsx` import a formatter and a type. This change does not touch any of them — which also keeps it clear of #1507, which edits `ThreadEnvironmentSummary.tsx`.

**Splitting that module would have moved zero bytes.** The issue's step 1, "split into a light type-only module and a heavy implementation module", has no heavy implementation to separate. Doing it would have renamed the chunk and changed nothing else.

**Cytoscape was already lazy** before this change, reachable only through mermaid's dynamic import. Verified on the base build; not touched.

**The real cause was five static imports**, none of them named in the issue: `FilePreview.tsx`, `DiffFilesPanel.tsx` (via `DiffFileCard`) and `TimelineFileDiffBlock.tsx` reaching `@pierre/diffs`; `ThreadDetailWorkerPoolProvider.tsx` and `PluginPanelView.tsx` reaching it through the worker-pool provider; and `markdown-preview.tsx` reaching `rehype-katex`. Plus `ThreadSecondaryPanel.tsx` importing `GitDiffCard` for one constant.

The issue's target — "`workspace-checkout-display` under 500 KB" — is therefore not a meaningful goal, because that chunk's name does not describe its contents. The goal this PR meets instead is the one the issue's own headline asks for: no editor, Cytoscape or KaTeX on the thread page's load path.

What this change does:

- `FilePreviewLazy`, `DiffFilesPanelLazy` and `TimelineFileDiffBlockLazy` load the diff renderer when a user opens a file or a diff.
- `markdown-katex` loads KaTeX and its stylesheet only for a body that can contain math. `remark-math` runs with `singleDollarTextMath` off, so that means a body holding `$$` — almost no message.
- `git-diff-view-options` holds the shared option object, so the secondary panel no longer imports `GitDiffCard` for one constant.
- The worker pool needed more care. Its provider wrapped the whole thread workspace, so it could not simply become lazy: a wrapper mounted later remounts the workspace and loses composer state. Each diff surface now supplies the pool to its own subtree through `DiffWorkerPoolProvider`, which also carries the code-theme sync that #1459 added. `ThreadDetailWorkerPoolKeepAlive` renders alongside the workspace to hold the reference count above zero between surfaces. A thread whose user never opens a diff now spawns no diff workers at all.
- `vite-bundle-stats` and `check-bundle-budget` gained a route-level package rule. Nothing else can catch a barrel import putting this code back on the thread route.

### Numbers

Rebased onto `596aab429`. Both sides measured from rolldown's own chunk metadata on a production build of that base and of this branch.

Thread route static closure — the entry chunk, the route chunk, and everything they reach without crossing a dynamic import:

| | chunks | raw bytes |
| --- | --- | --- |
| before | 22 | 5,051,210 |
| after | 40 | 4,090,840 |
| change | | **−960,370 (−19.0%)** |

The chunk the issue names:

| | raw bytes |
| --- | --- |
| `workspace-checkout-display-*.js` before | 2,167,788 |
| `workspace-checkout-display-*.js` after | 1,048,153 |
| change | **−1,119,635 (−51.6%)** |

What is left in that chunk is the thread timeline and the promptbox composer. Both render on arrival, so both belong there. The issue's 500 KB target assumed the editor was the bulk of it.

Modules of each package still statically reachable from the thread route, counted on the real module graph:

| package | before | after |
| --- | --- | --- |
| `@pierre/diffs` | 152 | 0 |
| `shiki` | 3 | 0 |
| `katex` | 2 | 0 |
| `rehype-katex` | 1 | 0 |
| `cytoscape` | 0 | 0 |

Boot payload, the cost of this change:

| | raw | brotli | chunks |
| --- | --- | --- | --- |
| before | 1,695,458 | 442,880 | 12 |
| after | 1,700,344 | 452,505 | 22 |

The raw growth is the new boundary modules, about 4 KB. The brotli growth is larger because the new dynamic entries make rolldown split shared boot code into 22 chunks instead of 12, and more small chunks compress worse. `maxBootBrotliBytes` therefore rises from 445,742 to 460,800, with the reason recorded in `bundle-budget.json`. `maxBootBytes` is unchanged and still has 7 KB of headroom.

**This is the one part of the change that makes something worse, and it is worth watching.** The boot payload now arrives as 22 requests instead of 12 for 9.6 KB more compressed bytes. On HTTP/2 to the local bb server that should be noise next to 938 KB removed from the thread route, but I measured bytes, not wall clock, so I cannot claim it. What a reader should check before trusting the trade:

- Whether 22 boot chunks cost more in request and per-response overhead than the 19% route saving returns — especially on a cold cache over a slow link, and for the desktop app where boot is served from disk and the route saving matters less.
- Whether the boot chunk count keeps climbing. Each future lazy boundary re-partitions the shared graph, and this PR shows the effect is not free. If it reaches the low 30s, rolldown's `advancedChunks` grouping is the lever; I tried its global `minSize` here and it did not merge boot chunks, so it would need explicit groups.
- Whether `maxBootBrotliBytes` should come back down once the count settles. It is a ratchet, and 460,800 currently leaves about 8 KB of slack.

## Testing

- `pnpm exec turbo run typecheck --filter=@bb/app` — passes.
- `pnpm exec turbo run test --filter=@bb/app --force` — 2,672 tests pass across 343 files.
- `pnpm exec turbo run lint --filter=@bb/app` — 0 errors.
- `node apps/app/scripts/check-bundle-budget.mjs` — passes, including the new thread-route rule. The rule is not inert: it failed on `@pierre/trees` and `@xterm/xterm` when I first listed them, which is how I learned the file tree and the terminal are still on the thread route. Moving those off is separate work, so they are deliberately absent from the list and noted there.

Existing tests caught a real bug during the work. `useState(loadedKatexRehypePlugin)` treated the unified plugin — a function — as a lazy initializer and stored its return value. Both the initializer and the setter now use the callback form.

Four tests became async because their surfaces now cross a boundary. The negative math assertions stay synchronous on purpose: a body with no `$$` must never load KaTeX at all.

### In the running app

Driven through the dev app on a real thread whose agent edited two files, checking each on-demand surface actually renders. Re-run after the rebase, since `596aab429`'s ancestry changed how the pool receives its theme:

- Secondary-panel diff tab: expanded `README.md`, `diff-worker-pool.ts` and `git-diff-view-options.ts`. All render with line numbers, `+`/`−` tints and Shiki highlighting. Highlighting proves the worker pool reaches the renderer through the new per-surface provider, and that the code-theme sync still applies.
- File viewer: opened `diff-worker-pool.ts` in a preview tab, full syntax highlighting.
- Timeline inline diff: expanded the agent's `Created SMOKE.md` row, diff renders with its added lines.
- KaTeX: opened a markdown file holding `$$E = mc^2$$` and a display block. Both render as KaTeX, and the KaTeX stylesheet loads with the chunk.
- No page errors on any of these, and the composer survives, which is what the keep-alive design protects.

Fixes #1072

---

## Slop Cop review

All five findings fixed in `3c2d521`.

**P1 — Thread plugin panels lose the diff worker context.** Correct, and the sharpest finding: a plugin rendering `FileDiff` inside a thread would have silently dropped to unhighlighted output.

I first tried a lazy provider inside `PluginSlotMount`. That broke 40 tests, because a Suspense boundary around every composer banner, message directive and header action changes plugin mount timing — too blunt. The shipped fix has no boundary at all: `plugin-frontend`, the chunk every plugin component comes from, already reaches `@pierre/diffs` statically, so it publishes the provider to a light registry when it evaluates. A slot can only render after that module has produced a registration, so `PluginSlotMount` reads it synchronously.

It applies to the slot kinds that previously inherited a pool — the thread workspace slots plus `navPanel` — and not to homepage, settings, sidebar or thread-list slots. That boundary is deliberate: the provider spawns workers eagerly, and those slots never had a pool, so widening it would start workers on a page showing nothing but a plugin chip. `PluginPanelView` now uses the same registry instead of its own wrapper, so there is one mechanism rather than two. `PluginSlotMount.workerPool.test.tsx` covers both halves.

**P2 — The first math frame creates invalid HTML.** Correct, and I had missed it. `remark-math` now joins the pipeline only once KaTeX is loaded, so the TeX stays literal text for that frame instead of becoming `<pre><code>` inside a paragraph. Regression test added: it asserts no `p div` / `p pre` and an empty `console.error` before and after the chunk resolves.

**P2 — A failed KaTeX import cannot recover.** Correct. The repo already had `createRetryingModuleLoader` for this exact problem, so I reused it rather than writing a second cache. It moved from `plugin-frontend-lazy.ts` to `retrying-module-loader.ts` with its three existing tests, since two callers now share it. The rejection path also logs instead of leaving an unhandled rejection.

**P2 — Later thread routes can start unused workers.** Correct. The permanent flag is gone. `DiffWorkerPoolProvider` publishes a demand count while a diff surface is mounted, and the thread keep-alive latches on that count per mount. A thread area engages only after a diff appears inside it, so opening a diff in one thread no longer makes the next thread spawn workers.

**P2 — The route budget fails open.** Correct. The check now iterates `forbiddenRoutePackages` and fails when a guarded route is missing from the stats, naming `TRACKED_ROUTES` in the message. Verified by adding a bogus route name: the check fails with that message, and passes again once removed.

I did not take the optional suggestion to prefetch the renderer when the diff tab activates. It is a real improvement, but it is a new behaviour rather than a fix to this change, and it belongs in its own PR.

Re-verified after the fixes: 2,672 tests pass, typecheck and lint clean, bundle budget passes, and the dev app still renders diffs with highlighting and math with KaTeX — with no console errors, including no DOM-nesting errors on the first math frame.

### Slop Cop re-review

The first pass only saw the pre-rebase SHA, so a second pass ran against `3c2d521`. It raised one P2, fixed in `8d0f4aa`.

**P2 — The worker latch does not reset after workspace navigation.** Correct, and it is the residual of my own round-1 fix. I had scoped the latch per component mount, which I described as "per thread area". It is not: `SplitWorkspaceRoute` keeps the split workspace mounted across thread, compose and plugin navigation, so the component outlives thread changes and one opened diff held up to eight workers for the rest of the session.

The hold is now scoped to the thread that engaged it, keyed on the route's thread id. Within a thread it stays steady, which is where the churn it guards against actually happens — a virtualized timeline scrolling its last diff row out of view. Leaving that thread releases it. I did not take the suggested "reset after demand stays at zero, with a short delay": a timer would reintroduce exactly the transient-zero churn the holder exists to absorb, and the thread id gives the same answer without one.

`ThreadDetailWorkerPoolKeepAlive.test.tsx` covers all three states: the hold survives the last diff surface unmounting, releases when another thread opens without a diff, and never engages for a thread that showed no diff. I checked the middle test is not vacuous — it fails against the previous latch.

The re-review found nothing wrong with the plugin renderer registry from round 1, and called out the shared provider and retry loader as removing duplicate code. Its browser pass could not run — the local launcher could not reach its data directory — so the runtime evidence for this round is mine: the diff panel still renders with syntax highlighting on the fixed build.

I also wrote the ordering invariant that makes the registry safe into `plugin-diff-worker-pool.ts`, since it is the load-bearing claim and was only implied before: `plugin-frontend` is the only writer of slot registrations and sets the renderer at module evaluation, before it can write any, so the value cannot flip between two renders of a mounted slot and remount plugin content.

Final state: 2,675 tests pass across 344 files, typecheck and lint clean, bundle budget passes.


SawyerHood:
Closing without merging. This came out of the 13 Aug 2026 overnight issue sweep. Sawyer is closing the unmerged sweep pull requests rather than carrying them.

The branch is left in place, so this can be reopened or cherry-picked. The review history, the measurements, and the answers to each Slop Cop finding stay on this pull request.

PR #1535, opened 2026-08-13
Stop the installed-plugins page-size measurement from locking the tab

SawyerHood:
## Problem

Changing pages in the installed plugins list froze the whole browser tab. A DevTools trace of the freeze shows no single long task: instead the tab runs a self-feeding render loop at ~120 cycles per second for the entire recording. `measureResourceViewportPageSize` accounts for 3.6s of self time with a forced layout on every call, and DOM nodes grow from 50k to 240k, so each cycle gets slower.

## Root cause

`useResourceViewportPageSize` computed the page size from the tallest row **currently rendered**, but which rows render is decided by that same page size. The measurement read back its own output.

When one page holds a taller row the two values trade places forever: a page size of 5 shows the tall row and measures 2, a page size of 2 hides it and measures 5. The hook's `MutationObserver` (`subtree: true` over the whole viewport) re-runs the measurement on every mutation, so the loop runs as fast as the microtask queue allows and the tab stops responding. Changing pages is the trigger, because a new page swaps in rows whose tallest row differs from the previous page's.

Reproduced in the running app: the list flipped 5 → 2 → 5 rows continuously and a click on **Next** timed out with the page unresponsive.

## Fix

`packages/shared-ui/src/components/ui/resource-pagination.tsx`:

1. The measurement keeps the tallest row height seen for the mounted viewport. The page size is therefore monotonic — it can change at most once per distinct row height, so it always settles.
2. Mutation and resize notifications coalesce into one measurement per animation frame. Each one forces a layout read, and browser extensions and plugin content scripts mutate this subtree too.

The plugins collection and the skills collection both use this hook, so both are fixed.

## Tests

- New regression test in `apps/app/src/components/ui/resource-pagination.test.tsx`. It fails on the old code (7 page-size flips) and passes now. The probe bounds its own renders, so a future regression fails an assertion instead of hanging the runner.
- Live app: **Next** completes in 37ms and the list settles; uniform lists page exactly as before; window resize still re-fits rows.
- `pnpm exec turbo run typecheck --filter=@bb/shared-ui --filter=@bb/app`, 81 related component tests, eslint, and prettier all pass.

## Risk

The page size now only shrinks while one viewport stays mounted. A list whose rows genuinely get shorter keeps the smaller page size until the collection remounts. That is the deliberate trade for a measurement that cannot oscillate; the hook documents it.

## Note, not fixed here

The trace also shows a browser-extension content script mutating inside the list on every cycle, with the foreign-DOM guard's suppression path hot. That is the likely reason old rows stayed attached and node count grew instead of just oscillating. The loop fix stops it from mattering, but it may deserve its own look.

### DIFF
--- a/apps/app/src/components/plugin/PluginsOverview.tsx
+++ b/apps/app/src/components/plugin/PluginsOverview.tsx
@@ -80,12 +80,22 @@ export function PluginsOverview() {
   const [installedQuery, setInstalledQuery] = useState("");
   const [installedViewport, setInstalledViewport] =
     useState<HTMLDivElement | null>(null);
-  const installedPageSize = useResourceViewportPageSize(installedViewport);
   const [installedSortDirection, setInstalledSortDirection] = useState<
     "asc" | "desc"
   >("asc");
   // Empty means unfiltered: the menu has no explicit "All" row.
   const [typeFilters, setTypeFilters] = useState<PluginTypeFilter[]>([]);
+  const normalizedInstalledQuery = installedQuery.trim().toLowerCase();
+  // One projection identity for both the page selection and the row heights
+  // the page size is measured from.
+  const installedResetKey = [
+    normalizedInstalledQuery,
+    installedSortDirection,
+    [...typeFilters].sort().join(","),
+  ].join("\u0000");
+  const installedPageSize = useResourceViewportPageSize(installedViewport, {
+    resetKey: installedResetKey,
+  });
   const [addDialog, setAddDialog] = useState<{
     open: boolean;
     initial: AddPluginInitial | null;
@@ -102,7 +112,6 @@ export function PluginsOverview() {
       }`,
     },
   ];
-  const normalizedInstalledQuery = installedQuery.trim().toLowerCase();
   const visiblePlugins = useMemo(
     () =>
       plugins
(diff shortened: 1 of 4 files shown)


SawyerHood:
Addressed the SlopCop finding in 6fc510fb9.

**Lifetime maximum (Medium)** — fixed. The remembered row height now resets for anything that is not the page itself:

- **New projection.** The hook takes a `resetKey`, and both callers pass the same key they already compute for `useResourcePagination`. A filter that drops the tall row re-measures from the rows that are left.
- **Viewport width change.** `measure()` compares `clientWidth` against the width the remembered height was taken at. A width change is what re-wraps text, so every remembered height is stale.

Neither reset can be caused by the page size, so the loop stays closed: paging never changes the projection, and the Radix viewport hides its native scrollbar, so the row count cannot change the viewport width. The height is still kept across page changes, which is what the original fix needed.

**Test gap** — added. `measures once per frame however many mutations arrive` counts the row heights the hook reads while four separate mutation batches land in one frame, and asserts one pass over the rows. It fails when the observers call `measure` directly.

Also added `measures a new projection from its own rows` and `measures again when the viewport width changes`. Each of the three new tests fails when its own part of the fix is removed, verified individually.

Verified live in the app: baseline 5 rows → tall row settles at 2 → **Next** clicks in 97ms and stays responsive → clearing the filter re-fits to 5 → a width change re-fits to 5.

`turbo typecheck` for `@bb/shared-ui` and `@bb/app`, 518 app component tests, eslint (no new warnings), and prettier all pass.

PR #1560, opened 2026-08-13
Handle retried turn starts in timelines

### LINKED ISSUE #1839: Retried turn starts can make stored timelines unreadable
## Summary

The host-to-server event channel is at-least-once, but daemon ingestion is not idempotent for `turn/started`. A committed batch whose acknowledgement is lost can be retried with the same provider turn ID, persist a duplicate start, and make timeline projection throw; the public timeline route returns 500 even though the thread data remains stored.

Implementation: [PR #1560](https://github.com/get-bb/bb/pull/1560)

## Versions and environment

- Reproduced on parent commit `a83c12946790724383c84b665ac899b9d3bef601`
- Boundary: host daemon event delivery → server ingestion → timeline projection
- Reproduction uses in-memory SQLite and synthetic persisted events

## Steps to reproduce

1. Insert two `turn/started` events for the same `(threadId, turnId)`, simulating a daemon retry after an ambiguous 502/504 response.
2. Request that thread's public timeline.
3. Observe the unfixed route return 500 because projection sees two starts for one provider turn.

The exact red fixture and command are documented in [PR #1560](https://github.com/get-bb/bb/pull/1560).

## Expected vs actual

Expected: re-delivery is idempotent and previously persisted duplicate histories remain readable.

Actual: a duplicate receives a fresh event ID and sequence; projection throws. The regression fails on the parent with `AssertionError: expected 500 to be 200`.

## Evidence

- Ingestion now resolves stored turn identities and skips duplicate starts in [events.ts](https://github.com/get-bb/bb/blob/f4b6cbba7f9862c5ebaa633a4985d76097acb3e5/packages/db/src/data/events.ts#L566-L606).
- Legacy projection compatibility and route-level regressions are included in [PR #1560](https://github.com/get-bb/bb/pull/1560).

## What you ruled out

- Not distinct provider turns: the incident repeated stable turn and item IDs at later server sequences.
- Not general exactly-once delivery: the required boundary is semantic idempotency for `turn/started`.
- Not #1646: that issue is missing/late starts during continuation, not duplicate starts from delivery replay.
- No exact open or closed issue matched the invariant, PR URL, originating thread, or failure output.

High priority, medium effort: affected threads become unreadable; the fix must protect new ingestion and tolerate persisted duplicates.

Originating implementation thread: [thr_tq4bctwps7](https://brsbl.getbb.app/threads/thr_tq4bctwps7)

brsbl:
## Summary

- tolerate legacy timeline histories containing duplicate `turn/started` events after daemon delivery retries
- prevent future retries from persisting the same turn start twice
- cover the ingestion, projection, and public timeline route paths with regressions

## Root cause

The host-to-server event channel is intentionally at-least-once, but daemon event ingestion was not idempotent for `turn/started`:

1. The server writes a daemon event batch inside an immediate database transaction.
2. The transaction can commit before the HTTP request finishes its post-commit effects and returns an acknowledgement.
3. If the daemon receives an ambiguous network failure or retryable 5xx after that commit, its event sink keeps the entire in-memory queue and posts it again on the next flush.
4. Daemon event envelopes have no stable delivery ID. The server assigns a new event ID and sequence to every insert, and the previous ingestion path always appended `turn/started` without checking the semantic identity `(threadId, turnId)`.
5. The retry therefore persisted another start for the same provider turn.
6. Timeline projection requires exactly one start per turn and threw on the duplicate, causing the public timeline route to return 500 even though thread metadata, events, and final output remained stored.

The observed incident matched this mechanism: repeated gateway 502/504 responses were followed by repeated lifecycle rows with the same stable turn and item IDs at later server sequences. That proves the daemon replayed an already-accepted batch; the provider did not create distinct turns.

## Fix boundary

This PR restores the invariant at both relevant boundaries:

- ingestion treats `(threadId, turnId)` as the identity of `turn/started` and skips a replay when that start is already stored
- projection keeps the first start as canonical so histories written before this fix remain readable

This is deliberately not a general exactly-once event-delivery redesign. Other replayed event types retain their existing behavior. General batch idempotency would require a stable daemon delivery identity and a broader persistence contract; it is not necessary to stop duplicate turn starts from making timelines unreadable.

## Reproduce the bug locally

This repro uses only synthetic test data. It does not require access to a production database, user account, or affected thread.

From a checkout of this PR:

```bash
repo_dir=$(git rev-parse --show-toplevel)
repro_dir=/tmp/bb-timeline-pr-repro-1560
base_sha=a83c12946
fixed_sha=f4b6cbba7

git worktree add --detach "${repro_dir}" "${base_sha}"
git diff "${base_sha}" "${fixed_sha}" \
  -- apps/server/test/public/public-thread-data.test.ts |
  git -C "${repro_dir}" apply -

cd "${repro_dir}"
pnpm install --frozen-lockfile
pnpm exec turbo run test \
  --filter=@bb/server \
  --only \
  --force \
  -- \
  --run test/public/public-thread-data.test.ts \
  -t "returns a timeline when persisted daemon retry history has duplicate turn starts"
```

The regression fixture inserts two `turn/started` rows with the same turn ID, then requests the public timeline route. The unfixed parent commit fails with:

```text
AssertionError: expected 500 to be 200
```

Clean up the disposable worktree:

```bash
cd "${repo_dir}"
git worktree remove --force "${repro_dir}"
```

## Verify the fix

On this PR's head:

```bash
pnpm exec turbo run test \
  --filter=@bb/server \
  --only \
  --force \
  -- \
  --run test/public/public-thread-data.test.ts \
  -t "returns a timeline when persisted daemon retry history has duplicate turn starts"
```

Expected result: the route returns 200, the test passes, and the timeline contains one canonical turn.

## Behavior after this PR

- Existing duplicate-start histories remain readable.
- Reposted `turn/started` events are ignored at ingestion, preventing new projection-breaking histories.
- Other replayed event types retain their existing behavior.

## Verification

- Red repro on `a83c12946`: route regression fails with `expected 500 to be 200`
- `@bb/thread-view` timeline test file: 72 passed
- `@bb/db` events test file: 79 passed
- `@bb/server` public thread data test file: 75 passed
- `@bb/server` internal event/tool-call test file: 25 passed
- Turbo typecheck: `@bb/db`, `@bb/thread-view`, and `@bb/server` passed
- CI: all required checks passed
- Independent architecture and code reviews: CORRECT
- Visual QA: not applicable; this changes event ingestion/projection and no UI code

Fixes #1839

BB-Thread-ID: thr_tq4bctwps7

> AGENT GENERATED: by GPT-5

### DIFF
--- a/apps/server/test/internal/internal-events-tool-calls.test.ts
+++ b/apps/server/test/internal/internal-events-tool-calls.test.ts
@@ -1010,13 +1010,18 @@ describe("internal event and tool-call routes", () => {
         harness.db.select().from(threads).where(eq(threads.id, thread.id)).get()
           ?.status,
       ).toBe("idle");
-      expect(
-        harness.db
-          .select()
-          .from(events)
-          .where(eq(events.threadId, thread.id))
-          .all(),
-      ).toHaveLength(4);
+      const storedEventTypes = harness.db
+        .select({ type: events.type })
+        .from(events)
+        .where(eq(events.threadId, thread.id))
+        .orderBy(events.sequence)
+        .all()
+        .map((event) => event.type);
+      expect(storedEventTypes).toEqual([
+        "turn/started",
+        "turn/completed",
+        "turn/completed",
+      ]);
     });
   });
 
--- a/apps/server/test/public/public-thread-data.test.ts
+++ b/apps/server/test/public/public-thread-data.test.ts
@@ -65,7 +65,6 @@ import { withTestHarness } from "../helpers/test-app.js";
 const queuedMessageIdResponseSchema = z.object({
   id: z.string(),
 });
-
 const threadReadResponseSchema = z.object({
   lastReadAt: z.number().nullable(),
 });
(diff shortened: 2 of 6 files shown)


SawyerHood:

PR #1566, opened 2026-08-13
fix(plugins): warn when installing a path plugin from a managed workspace

### LINKED ISSUE #1565: Env destroy deletes registered path-sourced plugins (no plugin-source guard before rm -rf)
## Summary

Destroying (or archiving-into-destroy) a managed environment recursively deletes the environment's directory without checking whether a registered **path-sourced plugin** lives inside it. If a plugin was installed via `bb plugin install <path>` from inside a `personal-workspaces/env_*/` directory, archiving the thread that owns that env silently deletes the plugin's source, leaving it permanently `missing (reinstall)`.

This is a data-loss footgun: the plugin is gone, the registration still points at a path that no longer exists, and the only recovery is reconstructing the source from session transcripts (if they were retained).

## Reproduction

1. Create a thread (managed env `env_A` provisioned under `<dataDir>/personal-workspaces/env_A/`).
2. Build a plugin inside it and install from the local path:
   ```
   cd ~/.bb/personal-workspaces/env_A/bb-plugin-mine
   bb plugin install . --yes
   ```
3. Archive the thread → after the env grace window, `destroyEnvironment` fires and `rm -rf`s `env_A/`.
4. `bb plugin list` now shows:
   ```
   mine@0.1.0  missing  (plugin directory not found: ~/.bb/personal-workspaces/env_A/bb-plugin-mine (reinstall))
   ```

## Real-world impact

I hit this on a single machine and lost **six** plugins this way (`browser-preview`, `automaker`, `file-explorer`, `prime-agent-ui`, `terminal-dock`, `thread-ops`), one of them twice (`terminal-dock` — restored once, then destroyed again by the same mechanism). The env→archive→destroy path is the normal lifecycle for short-lived feature-build threads, and `bb plugin install .` from the just-scaffolded plugin dir is the natural workflow, so the two collide frequently.

## Root cause

The personal-workspace destroy function unconditionally `rm -rf`s the target path with no plugin awareness:

**`packages/host-workspace/src/provision.ts:733-763`**
```ts
async function provisionPersonalWorkspace(opts) {
  ...
  return new ProvisionedHostWorkspace({
    path: targetPath,
    managed: true,
    ...
    destroyFn: () => rm(targetPath, { recursive: true, force: true }),  // line 762 — no plugin check
  });
}
```

Called from the host daemon's environment teardown, also with no plugin awareness:

**`apps/host-daemon/src/runtime-manager.ts:1092-1106`**
```ts
async destroyEnvironment(environmentId: string): Promise<void> {
  ...
  await entry.workspace.destroy();   // line 1104 — invokes the rm above
  ...
}
```

Neither the install path (`bb plugin install <path>`) nor the destroy path consults the other: install happily registers a `source_path` under a managed env, and destroy happily deletes a directory containing a registered plugin's source. The predicate to detect "this path is bb-managed" already exists but is only used for thread path claims:

**`apps/server/src/services/threads/worktree-paths.ts:87-96`**
```ts
export function isBbManagedWorkspacePath(args: { dataDir: string; path: string }): boolean { ... }
```

## Suggested fixes (any one, in rough order of effort)

1. **Warn at install time** (cheapest, highest leverage): when `bb plugin install <path>` resolves a `source_path` for which `isBbManagedWorkspacePath(...)` is true, warn that the source lives inside a managed env and will be deleted when that env is destroyed, and offer to copy it to a stable location first. This alone would have prevented all six losses.

2. **Guard at destroy time**: in `destroyEnvironment` (or in `provisionPersonalWorkspace`'s `destroyFn`), before `rm -rf`, scan `installed_plugins` for any row whose `source LIKE 'path:%'` resolves inside `targetPath`; if found, either (a) block + error with the plugin ids, (b) migrate the source to `<dataDir>/plugins-rescued/<id>` and rewrite the registration, or (c) at minimum log a clear "destroyed plugin source X" warning instead of the current silent deletion.

3. **Refuse managed-path installs by default** unless `--in-managed-env` is passed, pointing the user at `~/.bb/plugins/<name>` or an external repo instead.

Option 1 + 2 together would close the hole end-to-end: prevent new footguns and rescue existing ones.

## Environment

- bb 0.37.0 (`@bb/desktop`)
- macOS
- All six affected plugins were `path:`-sourced from inside `~/.bb/personal-workspaces/env_*/`

Happy to open a PR for option 1 (install-time warning) if a maintainer confirms the desired behavior — it's a small change around the plugin install handler with a reuse of the existing `isBbManagedWorkspacePath` predicate.

PennybagsCX:
## Summary

Cheapest of the three fixes proposed in #1565. When a user runs `bb plugin install <path>` (or `installPath`) and the resolved `source_path` lives inside a bb-managed workspace (`<dataDir>/personal-workspaces/env_*/` or `<dataDir>/worktrees/`), emit a `logger.warn` so they know the source is ephemeral — it will be deleted when that environment is destroyed (e.g. when the owning thread is archived).

This does **not** change install behavior; it only surfaces the hazard. It reuses the existing `isBbManagedWorkspacePath` predicate (already used for thread path-claims in `worktree-paths.ts`), so there's no new concept to maintain.

## Why

I lost six plugins this way on a single machine (`browser-preview`, `automaker`, `file-explorer`, `prime-agent-ui`, `terminal-dock`, `thread-ops`), one of them twice. The natural workflow — scaffold a plugin inside the thread's env via `bb plugin new`, then `bb plugin install .` — collides with the normal archive→env-destroy lifecycle. The plugin source is silently `rm -rf`'d by `provisionPersonalWorkspace`'s `destroyFn` ([`packages/host-workspace/src/provision.ts:762`](https://github.com/get-bb/bb/blob/main/packages/host-workspace/src/provision.ts#L762)), leaving a dangling `missing (reinstall)` registration.

A warning at install time is high-leverage and zero-risk: it teaches the user the model the first time, before any data is lost. The destroy-time guard (option 2 in the issue) is the stronger fix and is worth doing separately, but this PR stands on its own.

## Changes

- `apps/server/src/services/plugins/plugin-registration.ts`: in `installPathSource`, after resolving `rootDir`, check `isBbManagedWorkspacePath({ dataDir: deps.dataDir, path: rootDir })` and `logger.warn(...)` if true.
- `apps/server/test/services/plugins/plugin-service.test.ts`: new test asserts the warning fires for an install under `<dataDir>/personal-workspaces/env_test/...`.

## Test plan

- [x] `pnpm vitest run` in `apps/server` — 19/19 plugin-service tests pass (including the new one)
- [x] `tsc --noEmit` clean
- [x] `eslint` clean on both changed files

Closes #1565 (install-time-warning portion).

### DIFF
--- a/apps/server/src/services/plugins/plugin-registration.ts
+++ b/apps/server/src/services/plugins/plugin-registration.ts
@@ -1,4 +1,5 @@
 import { resolve } from "node:path";
+import { isBbManagedWorkspacePath } from "../threads/worktree-paths.js";
 import {
   getInstalledPlugin,
   getInstalledPluginRegistration,
@@ -334,6 +335,13 @@ export function createPluginRegistration(context: PluginRegistrationContext) {
             pluginRootDir(checkoutDir, subdirectory),
             "plugin subdirectory",
           );
+    if (isBbManagedWorkspacePath({ dataDir: deps.dataDir, path: rootDir })) {
+      logger.warn(
+        `plugin "${rootDir}" is installed from inside a bb-managed workspace; ` +
+          "its source will be deleted when that environment is destroyed (e.g. when the owning thread is archived). " +
+          "Reinstall from a stable path outside the managed workspace to avoid losing it.",
+      );
+    }
     return registerInstalled({
       rootDir,
       source: `path:${rootDir}`,
--- a/apps/server/test/services/plugins/plugin-service.test.ts
+++ b/apps/server/test/services/plugins/plugin-service.test.ts
@@ -1,6 +1,6 @@
-import { mkdtemp, mkdir, rm, symlink, writeFile } from "node:fs/promises";
+import { mkdtemp, mkdir, rename, rm, symlink, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
-import { join } from "node:path";
+import { dirname, join } from "node:path";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { createConnection, migrate, type DbConnection } from "@bb/db";
 import type { SystemChangeKind } from "@bb/domain";
@@ -608,6 +608,31 @@ describe("plugin service", () => {
       status: "running",
     });
     expect(service.getApi("reinstalled")).toBeDefined();
+  });
+
+  it("warns when a path plugin is installed from inside a managed workspace", async () => {
+    const warnSpy = vi.spyOn(logger, "warn");
+    // dataDir is <workDir>/data; install from <dataDir>/personal-workspaces/env_X/...
+    const managedRoot = join(
+      workDir,
+      "data",
+      "personal-workspaces",
+      "env_test",
+      "bb-plugin-managed",
+    );
+    const written = await writePlugin(workDir, {
+      name: "bb-plugin-managed",
+      serverSource: `export default function plugin() {}`,
+    });
+    // Move the written plugin into the managed workspace path so the install
+    // source resolves under <dataDir>/personal-workspaces/.
+    await mkdir(dirname(managedRoot), { recursive: true });
+    await rename(written, managedRoot);
+
+    await service.installPath(managedRoot);
+    expect(warnSpy).toHaveBeenCalledWith(
+      expect.stringContaining("bb-managed workspace"),
+    );
   });
 });
 


SawyerHood:
Ohhh man this is horrible! Let me make sure we do this! TY For the PR

PR #1574, opened 2026-08-14
Rename @bb/plugin-sdk to @get-bb/plugin-sdk

SawyerHood:
Layer 1 of 4 in the plugin-SDK npm publishing stack (#1578). Targets `main`.

Renames the plugin SDK package to `@get-bb/plugin-sdk` so it can publish publicly under the org scope, and adds `publishConfig.access: "public"`. Pure rename plus that one field: all imports and workspace deps (app, server, CLI, templates, 19 in-repo plugins/examples), the bundled-types banner (regenerated by the build, not hand-edited), scaffold tsconfig path maps, docs, and skills.

Deliberately unchanged: `engines.bbPluginSdk`, `sdkMajor`/`sdkVersion`, `PLUGIN_SDK_VERSION`, the `bundled-types/bb-plugin-sdk*.d.ts` file names, and every other `@bb/*` package name. The plugin-build app shim keeps `"@bb/plugin-sdk/app"` as a legacy alias so pre-rename plugin sources still build until the migration window closes.

Known and accepted: the export map's `source` conditions point at `src/`, which the tarball excludes. They stay because `apps/app`'s vite config resolves `conditions: ["source"]` for workspace dev; standard resolvers ignore unknown conditions, so external consumers are unaffected unless they opt into a custom `source` condition.

Verified: project-wide stale-name grep (one deliberate legacy alias remains), typecheck across 7 packages, SDK 82 + templates 36 tests including the external scaffold test that npm-packs the tarball and installs it into a fresh scaffold. Reviewed by GPT-5.6-Sol; both findings addressed.


SawyerHood:
Response to the SlopCop review — addressed in the updated stack:

- **Backend legacy alias (High)**: fixed. The server runtime import map and `PLUGIN_SERVER_EXTERNALS` now carry `@bb/plugin-sdk` alongside the new name (mirroring the existing frontend `/app` alias), with 3 new tests including a build of a pre-rename source plugin. #1577 additionally rewrites old-name imports during `bb plugin migrate`.
- **Test-install command (Medium)**: fixed — the README backend command now includes `cron-parser` and `hono` (a doc gap that predates this stack; the external-scaffold test always installed them explicitly).
- **Trusted-publishing metadata**: `repository`/`homepage`/`bugs` added here (also flagged on #1575).

Declined with rationale:
- **"Add a trusted release path" (High)**: that path is #1575, the next layer of this stack — this layer is deliberately rename-only.
- **`source` export conditions (Medium)**: kept. `apps/app`'s vite config resolves `conditions: ["source"]` for workspace dev, and standard resolvers ignore unknown conditions, so external consumers are unaffected unless they opt into a custom `source` condition. Documented in the PR body.

PR #1575, opened 2026-08-14
Publish @get-bb/plugin-sdk with the release train

### LINKED ISSUE #1134: @bb/plugin-sdk is documented for external tests but unavailable on npm

SawyerHood:
Layer 2 of 4 in the plugin-SDK npm publishing stack (#1578). Requires #1574 (the rename) below it.

Closes #1134.

Adds a `publish-plugin-sdk` job to `publish-bb-app.yml`: build the SDK, then publish only when the current version is absent from npm (idempotent skip otherwise), via OIDC trusted publishing in the `npm-release` environment. The job has no dependents, so a failure cannot block the bb-app or desktop publishes. It honors `RELEASE_DRY_RUN`, and always publishes to `latest` — the SDK has no channel concept.

Adds the pin-to-host CI guard: `check-npm-version-guard.mjs` fails a PR when the committed `bundled-types/` differ from the published content of the current `PLUGIN_SDK_VERSION`, so npm can never silently lag the API. Package/version-not-found passes (pre-bootstrap state); registry outages exit 2 and surface as CI warnings, not failures. Decision logic is a pure function with a unit-tested decision matrix.

Drive-by hardening from review: the main-only guards (both the new SDK job and the pre-existing bb-app job) now compare `github.ref == 'refs/heads/main'` — the old `ref_name` check accepted a tag named `main`.

One-time bootstrap after this lands (documented in `docs/bb-release-process.md`): manual first `npm publish`, then add the trusted publisher entry on npmjs.com pointing at this workflow file and the `npm-release` environment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,25 @@ jobs:
       - name: Build, typecheck, and lint
         run: pnpm exec turbo run build typecheck lint --cache-dir=.turbo/cache --output-logs=new-only
 
+      # Runs after build and typecheck, so the guard's own turbo build is a
+      # cache hit and the package it packs is the one this commit really ships.
+      # A published version whose packed files or consumer-facing manifest
+      # fields changed is a hard failure; an unreachable registry or a failed
+      # local pack (exit 2) is infrastructure, not a finding.
+      - name: Check plugin SDK npm version guard
+        run: |
+          # +e so the exit code can be inspected: GitHub's default shell runs
+          # with -e, which would abort the step before the exit-2 check.
+          set +e
+          node packages/plugin-sdk/scripts/check-npm-version-guard.mjs
+          guard_exit=$?
+          set -e
+          if [[ "$guard_exit" -eq 2 ]]; then
+            echo "::warning::Skipping the plugin SDK npm version guard: the npm registry was unreachable."
+            exit 0
+          fi
+          exit "$guard_exit"
+
       # Guards the JavaScript that blocks first paint in the web app. Heavy
       # editor, diff and math code reaches that path through barrel re-exports,
       # which typecheck and lint cannot see.
--- a/.github/workflows/publish-bb-app.yml
+++ b/.github/workflows/publish-bb-app.yml
@@ -59,9 +59,9 @@ jobs:
 
     steps:
       - name: Require main branch
-        if: ${{ github.ref_name != 'main' }}
+        if: ${{ github.ref != 'refs/heads/main' }}
         run: |
-          echo "::error::Publish bb-app must run from main, got ${GITHUB_REF_NAME}."
+          echo "::error::Publish bb-app must run from the main branch, got ${GITHUB_REF}."
           exit 1
 
       - name: Checkout repository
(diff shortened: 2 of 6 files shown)


SawyerHood:
Response to the SlopCop review — addressed in the updated stack:

- **`repository.url` missing**: fixed in #1574 (repository/homepage/bugs with the `directory` field), so the first OIDC publish can authenticate.
- **Guard compares declarations only**: fixed — the guard now diffs the full packed tarball (`dist/`, `bundled-types/`, README) plus consumer-facing manifest fields (exports, files, peers, publishConfig, repository, types). `dist/` was verified byte-reproducible across forced rebuilds before trusting content equality.
- **Stable/nightly publish race**: fixed — the `publish-plugin-sdk` job has a dedicated serialized concurrency group (`cancel-in-progress: false`), scoped so bb-app jobs are unchanged.
- **`npm-release` environment allows all branches**: real, but it is a GitHub environment setting, not repo code — the bootstrap guide now requires restricting the environment's deployment branches to `main` before the trusted publisher is configured, and it is on the maintainer's bootstrap checklist.

Declined with rationale:
- **Exit-2 leniency in CI**: kept. Registry flakes must not block unrelated PRs; real drift hard-fails (exit 1), and a broken local build fails CI in the earlier build step rather than being masked here. The exit-2 surface is now narrower since extraction/pack failures are explicitly classified.
- **npm@latest pin, composite action, build-once-publish-same-tarball pipeline**: reasonable hardening, deferred — this layer mirrors the existing bb-app job's conventions; restructuring both release jobs belongs in its own change.

PR #1576, opened 2026-08-14
Scaffold plugins against the published SDK instead of vendored types

SawyerHood:
Layer 3 of 4 in the plugin-SDK npm publishing stack (#1578). Requires #1575 (the publish pipeline) below it — a fresh scaffold pins a version that must exist on npm, so this must not merge before the bootstrap publish has run.

`bb plugin new` stops writing `types/` and the tsconfig path maps. The generated package.json instead carries an exact-pinned `@get-bb/plugin-sdk` devDependency equal to the host's `PLUGIN_SDK_VERSION`, alongside the existing `engines.bbPluginSdk` floor. After scaffolding, the CLI probes the registry and warns — never fatally — when the host's SDK version is a positive registry miss (with an `npm pack` + `file:` workaround); an unreachable registry only reports the pin as unverified.

Legacy plugins are untouched: `resolvePluginSdkLayout` classifies a plugin as vendored when either declaration file exists under `types/` or a tsconfig path map (old or new package name; JSONC tolerated via raw-text fallback) is present — ambiguity resolves to vendored, because refreshing too much is recoverable and silently stopping a refresh is not. `bb plugin build`/`dev`/`types` behave byte-for-byte as before on vendored layouts and never write `types/` into package-layout plugins.

Verified: the external scaffold test now installs the packed tarball as the scaffold's own dependency and typechecks with `skipLibCheck: false` and no path maps, proving resolution from `node_modules`. Reviewed by GPT-5.6-Sol; all three findings addressed (detector robustness, README overclaim, probe wording), plus a fourth found in orchestration (old-name path maps).


SawyerHood:
Response to the SlopCop review — both findings addressed, with the fixes riding #1577 (the top layer) to avoid a conflict cascade through code that layer rewrites:

- **Stale package-layout pins**: `bb plugin build`/`dev` now print a one-line warning when the exact pin differs from the running host's SDK version, pointing at `bb plugin types` — which, in #1577, actually rewrites the pin (default) while `--check` stays read-only. That is the "one clear contract" requested: build/dev warn, `types` syncs, `--check` gates CI. The docs were already reworded in this layer to match.
- **Registry probe bypasses npm config**: the probe now shells out to `npm view @get-bb/plugin-sdk@<version> --json` with a timeout, so `.npmrc` registries/proxies/auth apply; E404 is a firm "not published" warning, anything else degrades to a soft "could not verify". Never fatal, no prompt.

PR #1577, opened 2026-08-14
Migrate vendored plugins to the published SDK on request

SawyerHood:
Layer 4 of 4 in the plugin-SDK npm publishing stack (#1578). Requires #1576 (the scaffold switch) below it.

Adds `bb plugin migrate [path] [--yes]`: a consent-gated, idempotent, one-shot conversion of a legacy plugin to the package layout — exact devDependency pin, engines floor raised but never lowered, old-name and new-name path maps removed (other paths like `@/*` kept), vendored declaration files deleted, `types/` removed only when empty. The full plan prints before any write; non-TTY without `--yes` prints the plan and exits nonzero having changed nothing; symlinks and out-of-tree paths abort before the first write.

`bb plugin types` now manages the pin on package-layout plugins: the default invocation repins to the running host's SDK version (reporting old → new, reminding to `npm install`), while `--check` stays read-only with a nonzero exit on mismatch. Vendored behavior is unchanged, and `build`/`dev` print a one-line migrate hint. The guide, `bb-cli`, and `bb-plugin-authoring` surfaces document the new command, including an explicit instruction that agents never migrate a plugin unasked.

Verified: 16 new tests (transform matrix, idempotency, symlink refusal, non-TTY refusal, repin/--check), plus an end-to-end run of the real built CLI against a hand-built legacy plugin in /tmp.

> AGENT GENERATED: by Claude Fable 5 (implementation: Claude Opus 5; review: GPT-5.6-Sol, 9 findings fixed or documented)


SawyerHood:
Response to the SlopCop review — addressed:

- **P1 (legacy imports break after migration)**: fixed on two fronts. `bb plugin migrate` now scans the plugin's sources and rewrites `@bb/plugin-sdk` import specifiers (subpaths preserved, `dist/`/`node_modules/`/symlinks skipped, listed in the consent plan, idempotent, dry-run-safe). Independently, #1574 added the backend legacy alias to the server runtime map and build externals, so even an unmigrated old plugin builds and loads.
- **P2 (rmdir failure reported as success)**: fixed — the apply phase re-checks `types/` after the rmdir, reports `removedTypesDir` from reality, keeps the tsconfig `types` include when the directory survived, and warns naming the leftover directory.

Declined with rationale:
- **P2 (floor parser is not full semver)**: kept as the documented conservative contract — it never lowers a floor and declines to raise ranges it cannot fully parse (`*`, unions, upper bounds are left alone). Sharing the server's `semver.minVersion` policy would add a semver dependency to `@bb/templates`; worth a shared helper later, not load-bearing for migration correctness.
- **P2 (JSONC tsconfigs rejected)**: kept — migrate refuses with a clear error rather than reformatting an author's commented file. Scaffold-generated tsconfigs are pure JSON, so the affected set is authors who added comments, and layout *detection* (unlike migration) already tolerates JSONC.

PR #1591, opened 2026-08-14
Handle late cross-kind command failures

Juns-g:
## Problem

Some providers emit `commandExecution/completed`, then report the same call ID again as a failed generic `toolCall`. The second terminal event currently throws `Cannot merge command with tool-call`, so one record can make the thread timeline return 500.

## Fix

- keep the finalized command row and its command-specific fields
- merge only the late error's output, completion time, and status
- retain strict kind checks for non-finalized calls and existing dedupe for non-error terminal events

## Verification

- `pnpm exec turbo run test --filter=@bb/thread-view --force` (372 tests)
- `pnpm exec turbo run typecheck --filter=@bb/thread-view --force`

> AGENT GENERATED: by GPT-5

### DIFF
--- a/packages/thread-view/src/tool-activity-projection.ts
+++ b/packages/thread-view/src/tool-activity-projection.ts
@@ -300,9 +300,7 @@ function createRunningExecutionBase({
     ...(incoming.parentToolCallId
       ? { parentToolCallId: incoming.parentToolCallId }
       : {}),
-    ...(incoming.presentation
-      ? { presentation: incoming.presentation }
-      : {}),
+    ...(incoming.presentation ? { presentation: incoming.presentation } : {}),
     output: getVisibleTextBufferText(outputBuffer) ?? "",
     completedAt: incoming.completedAt ?? null,
     status: incoming.status ?? "pending",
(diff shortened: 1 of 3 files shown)


Juns-g:
Hi @SawyerHood — could you approve the fork Actions and take a look when you have a chance? This reproduces a `commandExecution/completed` → `toolCall/failed` sequence with the same call ID, which currently makes the timeline return 500. The fix keeps the strict kind checks and only handles the finalized late-error case. Thanks!

Juns-g:
Hi @SawyerHood — quick update: I rebased the branch onto the current main branch and resolved the conflicts. The PR now reports **MERGEABLE / CLEAN**, and all required checks are passing.\n\nI also re-verified the fix against the current provider bridge: the latest ACP translator/assembler can still produce a cross-kind terminal sequence, which still throws "Cannot merge command with tool-call" on main, while this PR handles it successfully. So the fix remains relevant after the provider refactors.\n\nThis should be ready to merge when you have a chance. Thanks!

SawyerHood:

PR #1597, opened 2026-08-14
Fix Fish terminal capability query before xterm attaches

smsunarto:
## Summary

Handle Primary Device Attributes (DA1) queries at the host PTY boundary.

This prevents Fish from waiting 10 seconds and showing:

> fish could not read response to Primary Device Attribute query

## Problem

bb starts the host PTY before the browser terminal attaches. This lets terminal processes start immediately, retain their session, and buffer output while no terminal view is open.

During startup, Fish sends a DA1 query (`ESC[c` or `ESC[0c`) and waits for a terminal emulator to answer. xterm.js supports this query, but it cannot answer before it attaches to the PTY.

Forwarding the query through replay does not provide a safe fix. Replayed output can be processed again after reconnects, which could produce duplicate terminal responses.

## Solution

The host daemon now:

- detects both supported DA1 query forms;
- recognizes queries split across output chunks;
- writes the xterm-compatible `ESC[?1;2c` response directly to the PTY;
- removes handled queries from live and replayed output, which prevents duplicate browser responses;
- preserves near matches and incomplete queries if the PTY exits.

This is intentionally a small protocol-specific handler. Adding a headless terminal emulator package would introduce parser state and a larger dependency for one startup query. If detached terminal emulation needs to support more queries later, the code comments identify `@xterm/headless` as the correct replacement.

The host daemon protocol version increases from 119 to 120 so enrolled machines cannot use old terminal semantics with the updated server.

## Verification

- Host daemon tests: 557 passed
- Host daemon contract tests: 49 passed
- Turbo typechecks passed for both affected packages
- `git diff --check` passed
- Live Fish 4.8.1 session:
  - prompt appeared normally;
  - no warning appeared after 11 seconds;
  - no handled DA1 query remained in terminal replay output
- Electron dev build started successfully

> AGENT GENERATED: by GPT-5.4

### DIFF
--- a/apps/host-daemon/src/terminals/terminal-manager.test.ts
+++ b/apps/host-daemon/src/terminals/terminal-manager.test.ts
@@ -1105,7 +1105,7 @@ describe("TerminalManager", () => {
       tailBytes: 4 * 1024 * 1024,
     });
 
-    expect(pty.writeCalls).toEqual(["\u001b[?1;2c", "\u001b[?1;2c"]);
+    expect(pty.writeCalls).toEqual(["\u001b[?1;2c\u001b[?1;2c"]);
     expect(collectTerminalOutput(harness.messages)).toBe("beforebetweenafter");
     expect(harness.messages).toContainEqual({
       type: "terminal.replay",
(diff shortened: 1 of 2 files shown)


SawyerHood:
Rebased onto `main` and stacked. Merge order: **#1597 → #1599 → #1606**. Each PR branch contains the layers below it, so review the newest commits only.

Changes in the rebase:
- Dropped the 119 → 120 host-daemon protocol bump. The bump now lives once at the top of the stack (#1606, version 122), which covers this terminal change as well.
- SlopCop (High, batch and bound the DA1 replies): the handler now answers at most 8 queries for each PTY output chunk with a single `pty.write`. A flooded 64 KiB chunk can hold more than 20,000 queries, and node-pty copies each reply into an unbounded write queue. Added a flood test.

Host daemon: 562 tests passed. Typechecks passed.

SawyerHood:
Landed in #1606, which rebase-merged the whole stack (#1597 → #1599 → #1606) as six commits. Your two commits are on `main` with their original authorship:

- `92380ed74` fix(terminal): answer fish device queries before attach
- `e7de7a121` fix(terminal): bound automatic DA1 replies to one write (the SlopCop follow-up)

The repo does not allow merge commits, so the rebase rewrote the SHAs and GitHub cannot close this PR by itself. Closing it here. Thanks!

PR #1598, opened 2026-08-14
fix(connect): scope remote sharing guidance to requests

smsunarto:
## Summary

Replace Connect's recent-activity instruction heuristic with a request-scoped plugin configuration signal.

- Add `experimental_connect_isRemote` to `PluginAgentConfigurationContext`.
- Set it from authenticated Connect session requests.
- Select `share-server-links` only when configuration was triggered through Connect.
- Remove remote-client count and five-minute activity checks.
- Reset Connect context for deferred background work.
- Keep app-surface telemetry independent from Connect behavior.

## Problem

Connect currently injects remote-sharing instructions when a remote client is active or Connect recorded activity during the last five minutes. These signals describe tunnel activity, not the request that started or resumed the agent.

This can give a local request remote guidance because another client was recently active, or omit guidance when remote activity state is delayed or unavailable. Pairing and tunnel state also cannot identify how the current interaction reached bb.

## Design

The server adds this experimental field to plugin agent configuration:

```ts
experimental_connect_isRemote: boolean
```

It is true when configuration resolves inside a request marked with `x-bb-gate-auth: session`. Connect removes caller-supplied gate headers and adds this marker after session authentication.

The server carries the value through asynchronous provisioning with `AsyncLocalStorage`. Both `thread.start` and `turn.submit` resolve the field, so it applies to new threads and new messages in existing threads. Work outside an originating request defaults to false, and deferred response work explicitly resets it to false.

The Connect plugin uses the field to select `share-server-links`. It no longer injects a separate dynamic system instruction. The skill is self-contained and includes explicit triggers for development servers, previews, static servers, phone access, and link-sharing requests.

### Why not test localhost reachability?

Localhost reachability is not the product decision. The agent workspace can run on another enrolled host, and browser network arrangements do not reliably indicate which URL the user should receive. Request origin answers the narrower question directly.

### Why not compare host IDs?

Viewer and execution host identity does not determine whether localhost is usable. Host comparison also creates ambiguous desktop-through-Connect, web-on-server, remote-execution, and headless cases without helping Connect choose its sharing guidance.

### Why not reuse app surface?

`AppSurface` distinguishes desktop and web for existing telemetry. It does not identify Connect transport. Combining the concepts would create unnecessary client-mode states when Connect only needs one boolean decision. This change leaves app-surface behavior unchanged.

### Why remove the activity fallback?

Remote-client count, pairing state, and recent activity can describe other tabs or earlier requests. Keeping them as fallbacks would preserve the false positives this change removes.

### Why use a selected skill instead of a dynamic instruction?

The behavior is a reusable workflow. A selected skill provides one source of truth, keeps explicit model trigger phrases, avoids duplicate prompt text, and does not freeze a tunnel URL in provider-session instructions.

### Why is the API experimental?

This is a new public plugin API contract. The audit entry records request propagation, provider-session lifetime, Connect header semantics, and the informational trust boundary that must be reviewed before stabilization.

## Lifetime and trust

The boolean is request-scoped, but the selected configuration follows existing provider-runtime lifetime rules. An active session is not always rebuilt immediately; changes apply when the runtime can apply them safely or when the provider session is next constructed or resumed.

The field is informational only. A direct caller can supply the header, so plugins must not use it for authorization.

## Testing

- Server, plugin SDK, and Connect typechecks
- 21 targeted server tests
- 63 Connect tests
- 76 plugin SDK tests
- 35 Ask User Question tests

> AGENT GENERATED: by GPT-5.4


SawyerHood:

PR #1599, opened 2026-08-14
Support generic ACP session forking

smsunarto:
## Summary

- Enable the existing app, public API, and CLI fork path for generic `acp-*` providers.
- Parse `agentCapabilities.sessionCapabilities.fork` during ACP initialization.
- Send unstable ACP `session/fork` with the source provider session ID, target cwd, and MCP servers.
- Persist the returned child session ID and use it for later turns.
- Reject agents that do not advertise fork support without creating a fresh history-free session.
- Accept the nullable `configOptions` response defined by ACP SDK 1.3.0.

This is provider-generic and has no `acp-amp` special case.

## Implementation notes

The existing server provisioning path already carries the source provider session ID and persists the provider session returned by the host. This change extends the generic ACP adapter and bridge to use that path.

No host-daemon protocol version bump is required because the server-to-daemon fork command already contains the required fork data.

## Validation

- Affected package typechecks passed.
- Host-daemon typecheck and build passed.
- Agent providers: 11 tests passed.
- Agent runtime: 947 tests passed.
- App: 2,706 tests passed.
- Server: 1,499 tests passed.
- ESLint, Prettier, and `git diff --check` passed.

> AGENT GENERATED: by Amp

### DIFF
--- a/apps/app/src/lib/fork-thread-request.test.ts
+++ b/apps/app/src/lib/fork-thread-request.test.ts
@@ -81,7 +81,7 @@ describe("buildForkThreadRequest", () => {
     expect(request).not.toHaveProperty("serviceTier");
   });
 
-  it("returns null when the provider cannot fork sessions", () => {
+  it("builds a fork request for a generic ACP provider", () => {
     expect(
       buildForkThreadRequest({
         environmentId: "env_source",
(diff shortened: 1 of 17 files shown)


SawyerHood:
Rebased and stacked on **#1597**. Merge order: #1597 → #1599 → #1606. This branch contains #1597's commits.

SlopCop findings addressed:

1. **Protocol bump (High)** — deferred by design. The stack bumps `HOST_DAEMON_PROTOCOL_VERSION` once at the top (#1606, version 122), and its contract-test comment names the ACP fork route. Land #1606 with this PR so no enrolled daemon sees the fork route without the bump.
2. **Dropped checkpoint (High)** — fixed. The adapter now forwards `sourceProviderCheckpointId`, and the bridge rejects the fork before `session/fork` because ACP cannot stop at a message checkpoint. A silent fork that kept later source turns is no longer possible.
3. **Provider mismatch (High)** — fixed. `resolveForkDescriptor` now requires the source and the child to use the same provider. A cross-provider request fails with `fork_source_session_unavailable` instead of sending an opaque session ID to another agent.
4. **Fork timeout (Medium)** — not addressed. The outer runtime deadline is shared with every session-creation path, so a fork-only cancel does not fit here. Worth a separate change that applies one deadline to `startAgentSession`.
5. **Duplicate session ID (Medium)** — fixed. The bridge rejects a fork result that returns the source ID or any live session ID.

New tests: checkpoint rejection, session-ID collision, and cross-provider rejection.

Validation: agent-runtime 953 tests passed; server 1657 passed (one pre-existing file-mode failure in `internal-skill-trees.test.ts`, present on `main`); typechecks, ESLint, and Prettier passed.

SawyerHood:
Landed in #1606, which rebase-merged the whole stack (#1597 → #1599 → #1606) as six commits. Your commit is on `main` with its original authorship:

- `53d193144` feat(acp): support generic session forking
- `da2654cf4` feat(acp): guard generic session forks (the SlopCop follow-up: cross-provider rejection, checkpoint rejection, and the session-ID collision guard)

The host-daemon protocol bump that SlopCop asked for landed with the same merge, in `7c33c816d` (version 122), so no enrolled daemon sees the ACP fork route without it.

One SlopCop finding is still open and is not addressed anywhere in the stack: a timed-out ACP fork keeps running in the bridge, so a late result can conflict with a retry. That needs one deadline applied to `startAgentSession` across the start, resume, and fork paths, which did not belong in this PR.

The repo does not allow merge commits, so the rebase rewrote the SHAs and GitHub cannot close this PR by itself. Closing it here. Thanks!

PR #1639, opened 2026-08-15
Merge live timeline windows by raw event range

brsbl:
## Summary

- Merge live timeline refreshes by their raw event-sequence coverage instead of using the first repeated rendered row as a splice point.
- Preserve every already-loaded row older than the new server window, regardless of row kind; update shared/straddling rows in place and treat the server as authoritative inside the new window.
- Reset to the fresh, pageable server response when the raw windows have a gap or their shared rows disagree about order.
- Keep the existing server event-count and byte budgets unchanged. This does not pin prompts, special-case steers, query extra events, or load the events between the retained prefix and bounded suffix again.

Original report and first implementation by @brsbl. The commandeered commit preserves Bersabel Tadesse as the commit author.

## Root cause

Timeline pages are bounded in **raw events**, but the client was merging them using **projected row identity**.

A long-running command can project to one row whose ID and starting sequence stay fixed while new output keeps extending its ending sequence. In a Side chat, the client could initially hold:

```text
prompt · long-running command · accepted steer · later work
```

Once enough command-output events arrived, the next bounded latest response could start after the steer while still containing an updated projection of that same long-running command:

```text
long-running command (updated) · latest work
```

The old merge found the first repeated row ID (`long-running command`) and replaced everything from that row onward. That silently discarded the accepted steer and any other rows after the command but before the new raw window. The events remained in the database and the agent had received the steer; only the loaded UI state lost them. If the initial state had already reached the beginning of history, its older cursor was `null`, so scrolling could not request the discarded interval again.

This was observed in Side chat because an accepted steer commonly lands while a long-running row is still growing and the panel keeps merging live windows for that same turn. The faulty merge was shared timeline code, so the fix is deliberately not Side-chat- or message-specific.

## Why the original draft was insufficient

The original change correctly identified that a budget cut could omit the initiating input and made the server query and inject the first accepted input into every cut active-turn window. That kept one prompt visible, but it treated one symptom as required context rather than repairing the merge:

- it did not preserve accepted steers or unrelated rows lost by the same splice;
- it added extra database work and changed bounded response contents;
- another projected row spanning the budget boundary could still make the client delete loaded history.

## Fix

The loaded client state now records the end sequence of the latest raw server window. On refresh it derives the new raw start from the pagination cursor and:

1. merges only when the two raw windows overlap or abut and time has not moved backward;
2. retains all loaded rows ending before the new raw start;
3. updates shared rows in their existing positions, which keeps older rows around a command that straddles the boundary;
4. removes stale rows covered by the authoritative new window;
5. carries the earliest valid older cursor for the contiguous union; and
6. falls back to the fresh response and its cursor when a gap or ordering conflict makes a safe merge impossible.

The latest fetch remains bounded. For example, with 2,000 intervening events the client retains the prefix it already has and merges the bounded suffix; it does not fetch all 2,000 events.

## Reproduce on base

From this PR branch, temporarily restore only the production merge implementation from `origin/main` while keeping the regression test:

```bash
git restore --source=origin/main --worktree -- apps/app/src/components/thread/timeline/useThreadTimelineController.ts
pnpm exec turbo run test --filter=@bb/app -- --run src/components/thread/timeline/useThreadTimelineController.merge.test.ts -t "retains every loaded row before the latest raw window boundary"
git restore --source=HEAD --worktree -- apps/app/src/components/thread/timeline/useThreadTimelineController.ts
```

The assertion fails because `accepted-steer` and `older-work` disappear. Run the same test with this PR's implementation restored and it passes.

I also reproduced the complete failure with real Codex agents in the dev server: an accepted Side-chat steer was visible, a delayed command emitted enough genuine output events to move the latest raw window past it, and the unfixed client removed the steer even though the server still stored it and the agent continued acting on it. The fixed client retained it through the same refresh shape.

## CI repair after rebase

Current `main` already contains `provider-registrations-changed` in the bundled plugin SDK declarations, but its checked-in generated template omitted that member. That made the unrelated `@bb/templates` check fail once the package shard rebuilt the SDK. The second commit regenerates that artifact with the repository generator; the timeline fix remains isolated in the first, original-author commit.

## Verification

- Base regression: fails with `accepted-steer` and `older-work` missing
- Focused timeline controller tests: 17 passed
- Full timeline component suite: 185 passed
- Full non-app package CI shard: 57 tasks passed
- `@bb/app` Turbo typecheck: passed
- `@bb/app` lint: 0 errors (147 pre-existing warnings)
- `git diff --check`: passed

BB-Thread-ID: thr_2xxfbvsr9r

> AGENT GENERATED: by GPT-5.6 Codex


ymichael:
Added two regression cases and a comment on `timelineWindowsAreContiguous` (16f390b85), from a separate investigation that landed on the same sequence-range conclusion — a large thread whose timeline truncated on follow-up submission and then refilled as auto-load paged the history back.

Both cases already pass on this branch; they pin shapes the previous row-based check got wrong:

- **Follow-up window.** The loaded tail was a `provider/error` row ending at 62634, the turn summary above it reached 62635, and the follow-up prompt opened the next window at 62636. `turn/completed` at 62635 is not a row, so row arithmetic saw a one-event hole and dropped every loaded page.
- **Backfilled first row.** A sequence-cut window's first row can start below its own cursor, since the projection backfills the running turn's `turn/started` row from under the cut. That tripped the old "latest reaches back past the loaded rows" branch on windows that were contiguous.

The comment records the three distinct ways rows mislead here — unprojected events, rows sorting by start rather than end, and backfilled rows starting below the cut — so the sequence-range gate does not get "simplified" back to row comparisons later.

`turbo run test --filter=@bb/app -- timeline`: 188 passed. `typecheck` clean, `lint` 0 errors.

PR #1640, opened 2026-08-15
Agent providers as a first-class plugin surface (provider bridge protocol)

ymichael:
# Agent providers as a first-class plugin surface

Implements [plans/agent-provider-plugin-surface.md](plans/agent-provider-plugin-surface.md) **end to end — phases 1–6 plus graduation**. Providers declare their metadata through the plugin API and run as plugin-registered **provider bridges** speaking one bb-owned, provider-agnostic JSON-RPC protocol. The canonical path is no longer gated: the experiment, the legacy adapters, and the core provider catalog are all deleted, so the bridge protocol is the only way a provider reaches bb.

132 commits, 622 files, **+41,035 / −25,028** (`git diff --stat origin/main..HEAD`, rename-aware — most of the insertions are provider sources *moving* into `plugins/provider-*`).

## What's here, by phase (commits are ordered to match)

1. **Protocol + kit** — `@bb/provider-bridge-protocol`: versioned handshake with session-behavior capabilities, canonical requests/notifications (including end-to-end `skills/configure`), error codes, an event-grammar doc, and a transport-agnostic conformance kit (12 rules, incl. full streaming-delta coverage and zero-work-prompt settlement).
2. **Four conformant bridges** — acp, pi, claude-code rewritten onto the canonical dialect (translation extracted byte-identically into shared modules both paths import); codex is the one new bridge, owning per-thread `codex app-server` children with the #1402 supervision rules. One generic `BridgeProviderAdapter` replaces per-provider adapter glue. Turn-start watchdog (`provider_turn_start_timeout`).
3. **Server registry + SDK** — `ProviderRegistryService` (catalog-equality pinned), `bb.agents.experimental_registerProvider`, and `app.slots.experimental_providerIcon` (plugins ship theme-aware React brand marks; `logoUrl` files remain for static color logos). Both have `docs/api_to_audit.md` entries.
4. **First-party plugins** — `plugins/provider-{codex,claude-code,pi,acp}` take over their core-seed entries in place (position preserved, restored on disable, takeover merge preserves flagship behaviors) and register their brand marks as app components. The app no longer depends on `@bb/agent-providers`.
5. **Third-party artifact delivery** — `bb.providerBridge` manifest key builds a self-contained bridge bundle; server stores/serves it content-addressed; daemon caches by verified hash; a verified `bridgeLaunch` (carrying the declaration's validated execution capabilities) is its own routing authority. `HOST_DAEMON_PROTOCOL_VERSION` 123 → 124 (the PR's only bump; main took 123 for artifact engine ranges). `examples/plugins/echo-provider` proves the path end to end.

## Graduation (waves 1–5)

Each wave ended with a full board (typecheck + all suites) before the next started, and every deletion first scanned the dying adapter's tests for uncovered shared-module invariants and **moved** them.

1. **The pinned gap, then acp + pi** — the codex zero-work-prompt settlement fix landed first (dispatch-ownership seam, no fabricated turns), unpinning the last non-green conformance rule; acp's and pi's legacy adapters were deleted.
2. **claude-code, then codex** — the last two legacy adapters deleted. **No legacy adapter remains anywhere.**
3. **Flag and core-catalog retirement** — the `providerBridge` experiment, the whole provider-bridge-policy endpoint, and the daemon's prefix capture are gone; the registry's core seed is deleted and `@bb/agent-providers` no longer exists. The runtime's codex special cases were audited one by one and consciously kept (one stale comment fixed). v124 was unshipped, so no further bump was needed.
4. **Phase-6 consolidation sweep** — one commit per item, each red-verified: ACP manual compaction as a per-agent declaration; the skill provider id opened and the six per-provider skill scopes collapsed to `provider-user`/`provider-project`; ask-user-question reading `supportsNativeUserQuestion`; edit-message eligibility reading the declared `supportsNativeSessionRewind`; plan mode gating on the declared `plan` composer action (plus deleting the thread-view display-name switch); and the `thread/openWork` notification that finally gives `hasOpenThreadWork` a canonical implementation and stops the reaper killing codex native subagents.
5. **First-party artifact migration** — two moves. First the **bridge kit**: a plugin-shipped bridge cannot import `@bb/agent-runtime`, and nearly every module under `agent-runtime/src/shared/` turned out to have only provider-side consumers, so they moved wholesale (with their tests) to `@bb/provider-bridge-protocol/bridge-kit` — JSON-RPC plumbing, the stdio harness, tool-call/interaction codecs, id scoping, visibility metadata, translation helpers, and the runtime↔bridge structural types; bridge test infrastructure became `@bb/provider-bridge-protocol/testing`. Then the bridges themselves: **codex** (868 KB), **acp** (919 KB) and **claude-code** (2.44 MB) now ship as self-contained, content-addressed plugin artifacts on the same route as `echo-provider`, with every core carve-out removed end to end (daemon bundle target, bb-app `files` entries, launcher assertions, registry branches, catalog capability/session-restore baselines) and the tarball smoke driving the *packed* artifacts.

   **Pi is the one documented exception, and the evidence is in the plan**: a fully-inlined 14.8 MB artifact was built and driven through the real fixtures plus this machine's live pi config. It fails for three reasons, none of them module identity (an identity-divergent control run passed everything): pi-coding-agent's extension loader resolves itself by `__dirname` with no fallback, `import.meta.resolve` for the pi packages has no `node_modules` ancestor in the daemon cache, and pi-ai's OAuth loader is deliberately bundler-hostile. The only shape proven to work is a *directory* artifact (bridge + a real npm tree), which is future design work. Pi's bridge therefore stays daemon-bundled — the only bundled bridge left — while its daemon coupling was still removed (`createConfiguredPiSettingsManager` inlined, the `@bb/agent-runtime` re-export deleted).

After wave 5 every provider-specific line lives in `plugins/provider-*`; `@bb/agent-runtime` keeps only the protocol, the generic adapter, and supervision. The deliberate exceptions still id-switched in core are listed in the plan (skill-root union normalizers, `isAcpProviderId`, the `acp-cursor` launch table, codex error-text classification).

## Verification

- **Final validation gate — live QA, all pass.** Every provider through its graduated route with real CLIs: spawn, multiple turns, steer/stop/resume, fork, archived resume, models, staged bb skills reaching the CLIs, and accept-edits writes landing in thread storage. Plus the paths graduation specifically put at risk: plugin **disable/re-enable**, a **plugin update with a new artifact hash mid-session**, and a **fresh install against an empty data dir**. Process-tree and orphan sweeps clean; zero error events; CI green.
- **Multi-agent adversarial review of the graduation diff**: **49 raw findings → 8 confirmed** (every finding survived only if a dedicated skeptic failed to refute it), and **all eight are fixed with red-verified tests**. The sharpest class was the one that only exists after graduation: `thread.archive` and `thread.unarchive` carried no `bridgeLaunch`, so every graduated provider threw `Unsupported provider` — unarchive always (it runs on a fresh provider-maintenance runtime), archive whenever the thread's process was not already live — and the server's warn-log swallowed it. The launch spec is now threaded exactly like `thread.start` (wire field → server attach → daemon artifact resolve → runtime process key + adapter), and the codex account-restart re-resume carries it on `ThreadRuntimeConfig` too. The other six: codex never retracting `thread/openWork` on child exit (thread never idle-reaped); dynamic ACP agents offered while the ACP plugin is disabled; the boot race where the listener serves before plugins load (now a bounded registry settlement gate awaited by thread create, command building, provider listing and model loads); non-reactive execution-options reads in ThreadDetailView; and custom ACP agents labelled "ACP provider" in the skills library.
- **Earlier live QA, three full passes** (pre-rebase, post-rebase, default-on gate): all four providers through the canonical bridges, verified by process trees and event streams, including a daemon-restart-under-stalled-server policy check.
- **Historical-bug audit**: ~30 past provider fixes classified; every invariant is bridge-tested, shared-by-construction, or has a dedicated bridge-path suite (codex translator, ACP event-translation/interactions, pi `none`→`off`, claude rate-limit classification), each verified red before landing green.
- **Dual-path calibration replay** for all four providers while both paths existed: normalized event streams diffed by LCS, **no unexplained parity divergence on any provider** — only deliberate protocol differences. Post-graduation the suites remain as bridge-only self-consistency checks.
- **Conformance is all green for all five bridges** — the codex zero-work-prompt pin was flipped, not deleted.
- Full-tree typecheck and tests green; CI green.

## Follow-ups (tracked, not in this PR)

- **Pi externals design** — the directory-shaped artifact (bridge `.mjs` plus a pruned npm tree) that would let pi leave the daemon bundle; the empirical findings are recorded in the plan.
- **Auto-mode workspace gap** — surfaced during the gate's live QA, independent of the provider surface.
- **Claude dropped-message-after-stop** — a reproducible upstream-SDK behavior found during QA, written up as its own report rather than patched here.
- **Deferred phase-6 items**, in two groups: (a) the three sessionless daemon→bridge reads — `skills/scanRoots`, `provider/health|install|update`, `provider/usage` — which share one transport and should land as their own wave; and (b) the four surfaces blocked on new plugin API rather than on consolidation — provider-scoped settings toggles, onboarding/settings entries, the provider-retry banner label, and provider icon colors. Both groups want a plugin-facing provider directory, which does not exist on either SDK surface today.
- Smaller: env-var live-change semantics, fork-UI reactivity minors, ACP write-approval presentation.


thesobercoder:
I opened #1668 to track a remaining Pi SDK bridge compatibility gap.

The Pi bridge calls `session.bindExtensions({ mode: "rpc", ... })` without an `uiContext`.

As a result, portable Pi extension UI calls such as `ctx.ui.confirm()`, `select()`, `input()`, and `editor()` cannot reach BB’s pending-interaction interface.

The issue proposes mapping Pi’s public `ExtensionUIContext` onto BB’s provider-independent interactive-request flow.

It does not request support for TUI-only `ctx.ui.custom()` components.

PR #1657, opened 2026-08-15
Keep mid-turn user exchanges visible after a turn completes

### LINKED ISSUE #1656: Replies to mid-turn messages disappear from the timeline when the turn completes
While a turn is running, every assistant message is rendered individually. When `turn/completed` arrives, completed-turn grouping collapses the turn into "Worked for …" summary rows, and three kinds of messages the user already read vanish from the top-level timeline:

- the direct reply to a mid-turn steered user message (only the segment's last assistant message is preserved),
- every assistant message in the final segment except the turn's terminal message (the final flush never preserves a terminal),
- answered `AskUserQuestion` cards and the report that prompted them (answered questions are groupable and are not segment boundaries).

The content is still reachable by expanding the summary row, but visually the final message appears to swallow messages that were on screen seconds earlier.

**Reproduce**

1. Start a turn that produces several assistant messages.
2. While it runs, send a message (default steer). The agent replies, then writes at least one more assistant message before the turn ends.
3. When the turn completes, the reply collapses into the "Worked for …" row.

Deterministic; depends only on the reply not being the last message of its segment.

sholub-dev:
## Summary

Fixes #1656.

Completed-turn grouping buried messages the user had already read once `turn/completed` re-grouped the turn:

- the direct reply to a mid-turn user message (only each segment's last assistant message survived),
- the final segment's last terminal message (the final flush never preserved one),
- answered user questions and the report preceding them (answered questions were groupable and not boundaries).

Changes in `packages/thread-view`:

- Preserve the first assistant/error message after user input (steer, external boundary, or answered question) — the direct reply.
- Final flush preserves its last terminal message, matching boundary flushes.
- Answered `user-question-lifecycle` messages are ungroupable and act as user-input boundaries.

Turns without mid-turn user input keep the existing single-summary collapse. Regression tests in `completed-turn-user-input-visibility.test.ts`; one existing expectation updated (the trailing segment after a legacy user message now keeps its last terminal visible).

### DIFF
--- a/apps/server/test/services/threads/conversation-outline-parity.test.ts
+++ b/apps/server/test/services/threads/conversation-outline-parity.test.ts
@@ -1,0 +1,194 @@
+import { describe, expect, it } from "vitest";
+import { threadScope, turnScope } from "@bb/domain";
+import {
+  createConnection,
+  createProject,
+  createThread,
+  insertEvents,
+  migrate,
+  noopNotifier,
+  upsertHost,
+} from "@bb/db";
+import {
+  buildThreadConversationOutline,
+  buildThreadTimeline,
+} from "../../../src/services/threads/timeline.js";
+
+type EventInput = Parameters<typeof insertEvents>[2][number];
+
+function setup() {
+  const db = createConnection(":memory:");
+  migrate(db);
+  const host = upsertHost(db, noopNotifier, {
+    name: "test-host",
+    type: "persistent",
+  });
+  const { project } = createProject(db, noopNotifier, {
+    name: "test-project",
+    source: { type: "local_path", hostId: host.id, path: "/tmp/test" },
+  });
+  const thread = createThread(db, noopNotifier, {
+    projectId: project.id,
+    providerId: "claude-code",
+  });
+  return { db, thread };
+}
+
+function agentMessage(
+  threadId: string,
+  sequence: number,
+  text: string,
+): EventInput {
+  return {
+    threadId,
+    sequence,
+    type: "item/completed",
+    scope: turnScope("turn-1"),
+    providerThreadId: "provider-thread-1",
+    itemId: `message-${sequence}`,
+    itemKind: "agentMessage",
+    data: JSON.stringify({
+      item: { id: `message-${sequence}`, type: "agentMessage", text },
+    }),
+  };
+}
+
+function userQuestionLifecycle(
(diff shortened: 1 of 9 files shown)


SawyerHood:
Addressed both SlopCop findings (rebased onto `main`, pushed b1061344f):

1. **Mid-turn classification** (`completed-turn-grouping.ts`): the `items.length > 0` position check treated the second initial user row of a grouped `inputGroups` request as mid-turn input and armed the preserve flags for an ordinary turn. Grouping now tracks whether the turn has produced any assistant/tool output and only treats user input that arrives *after* output as mid-turn; consecutive initial rows stay initial. Added a regression test (unsteered turn with two `inputGroups` rows keeps only the terminal assistant row top-level) — it fails on the previous head and passes now.
2. **Question fixture**: replaced the hand-built resolved event (duplicate seq 6, no pending stage) with a typed `userQuestionLifecycle` factory method on the timeline test harness; the test now emits the pending event followed by the resolved event with unique increasing seqs.

`turbo run typecheck` and `turbo run test --force` for `@bb/thread-view` pass (22 files, 382 tests).

SawyerHood:
Pushed 9f825018a addressing the second SlopCop round:

- **First-output answered question**: the grouping now uses one typed helper (`getTimelineUserInputSource` / `isMidTurnUserInputBoundaryMessage` in `timeline-message-helpers.ts`). An answered question is always a boundary regardless of position and also marks the turn as having produced output; the "after output" rule applies only to user requests. Test: `keeps the direct reply after an answered question that is the turn's first output`.
- **Rejected steers**: only `turnRequest.status === "accepted"` user rows classify as user input; rejected/pending steers no longer split the summary. Test: `keeps interim messages folded around a rejected steer`.
- **Outline parity**: `listStoredConversationOutlineEventRows` now loads `system/userQuestion/lifecycle`; new server test `conversation-outline-parity.test.ts` asserts outline items equal the timeline's conversation rows (ids and text) after an answered question. Query-plan param expectation updated.
- Removed the `as never` request-id cast; `build-thread-timeline.test.ts` now reuses the harness `userQuestionLifecycle` builder instead of its own copy.

Verified: `@bb/thread-view` 384 tests, `@bb/db` tests, `@bb/server` thread service tests, and typecheck for thread-view/db/server all pass; the three new tests fail without the source changes.

PR #1665, opened 2026-08-16
fix: harden automation execution

lnittman:
## Summary

- enforce single-flight automation execution per automation
- add exponential retry backoff and auto-pause after three consecutive failures
- terminate the full script process group on timeout
- validate native target threads and permission modes before dispatch
- document native automation delivery via `bb thread tell`

## Verification

- `pnpm --filter bb-plugin-automations test` — 72 tests passed across 6 files
- `pnpm --filter bb-plugin-automations typecheck` — passed
- live Prompt Forge manual canary and subsequent scheduled run both succeeded against a fresh native thread


lnittman:
Superseded by #1675, which carries the same change plus startup recovery for orphaned running rows, a migration that settles duplicate running rows instead of failing, and the split commit series; it tracks #1674. Closing this one to keep a single thread.

PR #1671, opened 2026-08-16
Escape Unicode separators in JSON-RPC lines

georgecollier-nqu:
## Current failure

BB uses one JSON object per line between the host daemon and each provider bridge.

Injected instructions can contain U+2028 or U+2029. These are valid Unicode line separator characters.

1. `JSON.stringify` leaves each character literal inside the JSON string.
2. Node `readline` treats each character as the end of a line.
3. A `thread/start` request becomes two incomplete lines before the bridge parses it.
4. The bridge catches each `JSON.parse` failure and returns without a response.
5. The host waits until `thread/start` times out. Claude never starts.
6. The user can remain on **Preparing workspace** without a useful error.

This failure occurs before BB starts Claude. It is a BB message framing bug, not a Claude Code failure.

## Why BB must fix this

The input text is valid and must remain valid. It can come from these sources:

- `AGENTS.md`
- skills
- system instructions
- prompts
- tool schemas
- provider messages

Sanitizing each source would be fragile and could change user text. The JSON-RPC framing layer owns this requirement.

Escaping these characters keeps each message on one physical line. `JSON.parse` restores the original Unicode text at the receiver.

## Fix

- Escape U+2028 and U+2029 before BB writes line-delimited JSON-RPC.
- Apply the fix to runtime-to-bridge messages.
- Apply the fix to bridge-to-runtime messages.
- Add regression tests that pass both characters through Node `readline`.

## Validation

- `pnpm exec vitest run src/runtime-json-rpc.test.ts --config vitest.config.ts`

### DIFF
--- a/packages/agent-runtime/src/runtime-json-rpc.test.ts
+++ b/packages/agent-runtime/src/runtime-json-rpc.test.ts
@@ -1,7 +1,14 @@
 import { spawn, type ChildProcess } from "node:child_process";
+import { createInterface } from "node:readline";
+import { Readable } from "node:stream";
 import { setTimeout as delay } from "node:timers/promises";
-import { describe, it } from "vitest";
-import { sendJsonRpcResult } from "./runtime-json-rpc.js";
+import { describe, expect, it } from "vitest";
+import {
+  type JsonRpcMessage,
+  sendJsonRpc,
+  sendJsonRpcResult,
+} from "./runtime-json-rpc.js";
+import { createBridgeIo } from "./shared/bridge-harness.js";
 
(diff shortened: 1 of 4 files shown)


georgecollier-nqu:
reviewed manually too

ymichael:
Thanks for tracking this down—the Node 24/26 readline behavior is real. Since then, #1640 replaced both receive paths with a bounded reader that splits only on physical LF bytes, so U+2028/U+2029 can no longer break framing. This PR is now superseded, so I’m closing it. Thanks again!

PR #1673, opened 2026-08-16
Fix iOS touch targets and the two-tap send in the composer

sberan:
Found while using bb daily from an installed iOS PWA on a phone. Three problems in the composer, all touch-only — nothing here changes desktop behaviour.

## 1. Send needs two taps

`handleSubmitPointerDown` skipped `preventDefault()` unless the TipTap editor reported `isFocused`:

```ts
if (!currentEditor || currentEditor.isDestroyed || !currentEditor.isFocused) {
  return;
}
event.preventDefault();
```

On iOS that check reads `false` at `pointerdown` while the keyboard is still up, because WebKit has already begun moving focus by the time the handler runs. That is precisely the case the guard exists to protect against, so it opts out exactly when it is needed: focus leaves the editor, the keyboard dismisses, the composer reflows upward, and the click lands where the button used to be. The first tap appears to do nothing and the message only sends on a second press.

Cancelling the default is safe regardless of focus state — it suppresses the focus transfer for that press only, and the `click` still fires, which is what owns the commit.

Added a regression test. It fails on `main` with `expected true to be false` (the default is not cancelled) and passes with the fix.

## 2. Touch targets under 44px

- The composer's submit / stop / voice slot is **32px** (`size-8`). It is the most-tapped control in the app.
- The inline message editor's cancel is **24px** (`size-6`), and it is the only exit from edit mode.

Both grow on coarse pointers only.

## 3. Queued-message actions are unreachable on touch

Send-now / Edit / Delete are rendered, but the container is `hidden … md:flex` and only revealed on `group-hover`. On a phone neither condition is ever true, so the row falls back to the 28px overflow menu as the only affordance.

They now show outright on coarse pointers. The overflow trigger hides there in the same breath — both are absolutely positioned against the same right edge, so showing both stacks the two sets on top of each other at different offsets.

## Notes

- Uses the existing `max-md:pointer-coarse:` idiom already used in `ProjectList` and `ThreadRow`, so the scoping matches the codebase. Worth noting that a bare `pointer-coarse:` would also cover tablets at `md` and above, where 44px is equally right — happy to widen it if you would prefer that.
- `pnpm exec vitest run apps/app/src/components/promptbox` → 329 passed, 30 files.
- `pnpm exec tsc --noEmit -p apps/app/tsconfig.json` → clean.
- Verified against a real device rather than only emulation; the two-tap behaviour in particular does not reproduce in headless Chromium, since it depends on WebKit's focus timing.

### DIFF
--- a/apps/app/src/components/promptbox/InlineMessageEditorFrame.tsx
+++ b/apps/app/src/components/promptbox/InlineMessageEditorFrame.tsx
@@ -36,11 +36,18 @@ export function InlineMessageEditorFrame({
         type="button"
         size="icon"
         variant="ghost"
-        className="ml-auto size-6 shrink-0 text-subtle-foreground"
+        /* 24px is the smallest target in the app and the only exit from edit
+           mode. Coarse pointers get a real target; precise pointers keep the
+           compact size. */
+        className="ml-auto size-6 shrink-0 text-subtle-foreground max-md:pointer-coarse:size-11"
         onClick={onCancel}
         aria-label={cancelLabel}
       >
-        <Icon name="X" className="size-3" aria-hidden />
+        <Icon
+          name="X"
+          className="size-3 max-md:pointer-coarse:size-4"
+          aria-hidden
+        />
       </Button>
     </div>
   );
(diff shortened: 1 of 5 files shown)


sberan:
Added a follow-up: the first background-activity commit only wrapped the expanded list, but the line most people actually read is the **collapsed summary** — the one rendering as `Running background command: ...` without expanding anything. It truncates too, behind the same hover-only `title`, so on a phone there was no path to the text at all. Both now wrap on coarse pointers.

SawyerHood:

Issue #1676, opened 2026-08-16
Reload flashes: Tasks empty state, plugin route chrome, composer model and reasoning


lnittman:
## 1. Tasks paints list chrome and skeleton rows before it knows there are no projects

Every visit to Tasks (reload or navigation) mounts with the list layout and skeleton rows, then swaps to the "No projects yet" empty state ~450ms later once `listProjects` resolves. Only a resolved empty list selects the empty state, so "projects unknown" renders as the populated layout, and the plugin's query primitive keeps nothing between mounts, so each visit starts from unknown. Populated instances get the same skeleton-then-rows flash on every visit. Switching between the sidebar views has the same shape once loaded: returning from an empty Active to All tasks paints Active's emptiness as "No tasks yet" for the length of the refetch, because the list keeps the previous view's result while the new one loads. Expected: a repeat visit paints the last resolved state (empty or populated) immediately and only changes if the fresh answer differs, and no view ever presents another view's rows or emptiness as its own.

Prototype: an opt-in persisted snapshot in `useTasksQuery` (localStorage, versioned key, validated against the query's own RPC output schema on read, written only after a successful fetch) for folders, projects, and the sidebar summary, seeding the query state on mount; the sidebar skeleton shows only while that data is absent, so a snapshot-hydrated mount never shows it and a refetch keeps the last-known rows on screen. The empty state still waits for a resolved empty list, and the list body treats a changed view (All, Active, a project) as loading until that view's own fetch settles, while filter and sort changes keep painting the rows they already have.

## 2. Reloading a plugin page shows "This plugin panel is not available. The plugin may still be loading, or it has been disabled or removed."

Registrations arrive only after plugin frontends boot, so on a reload or deep link `PluginPanelView` cannot find the panel yet and renders the not-available message for a few hundred milliseconds on every load, then the real page mounts. Expected: a valid deep link stays quiet while registration is pending and shows the unavailable state only once boot has confirmed there is no such panel.

Prototype: track whether plugin frontends have settled (initial boot resolved or failed, with a 15s fallback if system config never resolves); the route renders an empty shell until then, and only a settled boot with no matching panel shows the message.

## 3. The header title and the sidebar's plugin rows paint empty, then pop in

The app header's plugin title + icon, the sidebar's Automations/Tasks rows, and split-pane titles for plugin panels are all derived from the live registrations, so they are blank at first paint and appear when frontends boot. Expected: host-owned plugin chrome is stable from the first paint of a reload.

[...]

PR #1677, opened 2026-08-16
Tasks first paint: last-known projects, folders, and sidebar summary

lnittman:
Part of #1676 (item 1).

Tasks keeps an opt-in last-known snapshot for folders, projects, and the sidebar summary in `useTasksQuery` (localStorage, versioned key, validated against the query's own RPC output schema on read, written only after a successful fetch) and seeds the query state from it on mount, so a repeat visit paints the last resolved state immediately; the sidebar skeleton shows only while that data is absent. The empty state still waits for a resolved empty list. Snapshots from older storage versions are pruned once per load; the shared vitest setup clears storage per test.

Plugin code only (`plugins/tasks`); no SDK or host changes. `isLoading` semantics are untouched (the list view's scroll restoration depends on its flip-on-refetch behavior).

### DIFF
--- a/plugins/tasks/shell/data.ts
+++ b/plugins/tasks/shell/data.ts
@@ -1,9 +1,11 @@
 import { useCallback, useEffect, useRef, useState } from "react";
 import { useRealtime, useRpc } from "@get-bb/plugin-sdk/app";
-import type { TasksRpcContract } from "../shared/contract.js";
+import type { z } from "zod";
+import { tasksRpcContract, type TasksRpcContract } from "../shared/contract.js";
 import type { Task, TaskPriority, TaskStatus } from "../shared/contract.js";
 import { TASKS_PAGE_MAX_LIMIT, type TaskSort } from "../shared/pagination.js";
 import type { MentionItem } from "../editor/extensions.js";
+import { readQuerySnapshot, writeQuerySnapshot } from "./query-snapshot.js";
 import { useTasksRefresh } from "./refresh.js";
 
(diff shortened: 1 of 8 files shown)


SawyerHood:
Pushed 85a4ffc1a addressing the three SlopCop findings (review was against daf1400e; the sidebar gate now lives in `plugins/tasks/shell/navigation-panel.tsx`):

1. **Failed route fetch showed the prior route's rows** — `useTasksQuery` now records which deps key its `data` belongs to. A failed request for a *changed* key drops that data (a failed refetch of the same key still keeps its rows), so `ListView` settles onto the error state instead of the previous route's tasks. Test: `shows the error, not the previous route's rows, when a route change fails`.
2. **Out-of-order snapshot writes across hook instances** — each request claims a per-name revision (`claimQuerySnapshotRevision`) when it starts; `writeQuerySnapshot` ignores any response whose revision is older than one already written, so a slow earlier request can never replace a newer record, including after unmount. Test: `keeps the newer projects snapshot when an older request resolves later`.
3. **Folder/summary failure left a permanent skeleton** — the skeleton now gates on "awaiting a first answer" (`data === undefined && error === null`) per query, so a failed companion request clears it and loaded projects stay navigable. Projects whose folder is not in hand render ungrouped rather than disappearing. Tests: `keeps loaded projects navigable when listFolders / sidebarSummary fails`.

All four new tests fail on the previous head and pass now; `turbo run typecheck` and the full `bb-plugin-tasks` suite (35 files, 350 tests) pass.

PR #1678, opened 2026-08-16
Reload flashes: quiet plugin routes, remembered plugin chrome, last-known composer options

### LINKED ISSUE #1676: Reload flashes: Tasks empty state, plugin route chrome, composer model and reasoning

lnittman:
Fixes #1676 (items 2 to 5; item 1 is the plugin PR).

Host-side last-known state so a reload paints what this profile last saw instead of neutral defaults that the live answers then replace:

- A small last-known localStorage cache primitive (scoped versioned keys, zod-validated reads, reads and writes that never throw, stale versions pruned) shared by everything below.
- Plugin panel routes stay quiet until plugin frontends have settled (initial boot resolved or failed, with a 15s fallback if system config never resolves); the "not available" copy shows only after a settled boot with no matching panel.
- The host-owned chrome of the panels this profile last saw (plugin id, panel id, path, title, icon, never the component) paints first in the app header, the sidebar rows, and split-pane headers; live registrations replace it under the same keys, and the memory is rewritten on every change so a removed plugin does not come back as a ghost.
- The composer's model-catalog placeholder is generalized from Claude Code to every provider (keyed by environment, host, provider, with a provider-level "latest" fallback for a routing never fetched to completion), and the host's provider list is replayed with it so a custom or installed ACP provider paints as itself; when the remembered provider is in no list the placeholder can vouch for, the composer waits as before. The placeholder ceiling is the most restrictive mode, so provisional data fails safe.
- A thread's last server-resolved execution options replay as `placeholderData`; display uses the replay while submission and the permission controls keep waiting for the verified resolution (`isPlaceholderData` counts as loading in both composers).
- The sidebar navigation replays the last bootstrap this profile received (validated against the wire schema on read, written only from a successful fetch), so the rail paints its remembered projects and threads instead of the two-row loading skeleton; the cache primitive's version prune learned to spare a zero-scope entry, which had been deleting exactly that replay on every load.

Client only (`apps/app`); no plugin API surface, no wire changes. The raw boot budget is raised by 4 KB: the sidebar replay adds about a hundred bytes to the boot path and the ratchet had zero raw headroom left (1667.1 KB against 1667.0 KB, brotli still under budget), so this keeps the guard about barrel regressions rather than byte noise.


SawyerHood:
Pushed three commits addressing the SlopCop review (`9c269d5`, `5d62331`, `6aefc2f`):

1. **Model catalog crosses host/environment boundaries (high)** — removed the provider-level "latest" catalog key. The execution-options placeholder now replays a remembered catalog only for the exact `environment/host/provider` routing it was observed on; a routing that was never fetched to completion keeps the built-in provider identity with no model rows and waits for its own probe. Tests: cross-environment and cross-host replays now assert an empty model list (`system-queries.test.tsx`).

2. **Settle floor finishing an active boot (medium)** — `plugin-frontend-boot-state` now tracks `idle | booting | complete`. `bootPluginFrontends` marks the boot started; the 15 s floor only settles a boot that never started and an in-flight boot settles itself when it finishes. Remembered nav-panel chrome is persisted only once the boot has actually completed (`usePluginFrontendBootComplete`), so neither the floor nor a partially mounted boot can overwrite valid remembered chrome. Tests: floor elapsing during an in-flight boot stays unsettled; chrome is not rewritten on floor/partial boot.

3. **Unbounded synchronous sidebar snapshot (medium)** — the stored bootstrap keeps projects whole (consumers resolve the current project by id from the replay) but caps each project's thread list at 30; the write is scheduled via `requestIdleCallback` (with a `setTimeout` fallback) instead of inline in the query function, quota/storage failures are swallowed by the cache primitive, and the replay is read/validated once per page load rather than on every placeholder evaluation. Tests: bounded storage, deferred write, and a rejected `setItem` leaving the fetch successful.

Verified: `turbo typecheck --filter=@bb/app`, the 8 touched/related `@bb/app` test files (82 tests), `turbo build --filter=@bb/app` + `check-bundle-budget.mjs` (OK), eslint/prettier on changed files.

SawyerHood:
Follow-up to the SlopCop fixes: `16e7929` keeps an unknown requested project held (picker loading, submit blocked) while the replayed sidebar bootstrap is placeholder data, instead of demoting it to the personal project. A test covers the replay-then-live path.

PR #1700, opened 2026-08-17
Align new-thread panel opener chrome

brsbl:
## Summary

- Give the closed new-thread right-panel opener the same subtle foreground token used by thread panel controls and the open panel's Hide control.
- Preserve the existing 28px target, disclosure semantics, pinned placement, shortcut, compact drawer behavior, and handoff to panel-owned chrome.
- Add a focused regression assertion for the shared resting and hover foreground classes.

## Rationale and boundary

Thread and new-thread pages already share `SecondaryPanelLayout` and `ThreadSecondaryPanel` for width persistence, responsive drawer behavior, tabs, Browser/Terminal bodies, focus lifecycle, and resize seams. This patch stays at the new-thread wrapper boundary and changes no shared panel internals.

The new-thread page intentionally keeps its own pinned opener while the panel is closed because it has no shared page header. When open, the shared panel owns Hide and New Tab. It also intentionally omits thread-only Info, Diff, and Full Screen actions. Navigation-panel plugins remain page-header-owned and are outside this patch.

This PR is standalone on `main`. It does not depend on the plugin right-panel stack (#1546 → #1553) or the thread split-pane PR (#1601), and it has no file overlap with either.

## Verification

- Focused Vitest: `RootComposeRightPanelToggle.test.tsx` and `RootComposeSecondaryContent.test.tsx` — 10 tests passed.
- `turbo run typecheck --filter=@bb/app` — passed.
- Targeted ESLint and Prettier checks — passed.
- Exact-branch BB desktop app plus browser-driven real interface at 390×844, 768×900, 1440×900, and 3440×1440 — passed.
- 767/768/769 responsive transition and five open/close cycles per representative viewport — passed.

BB-Thread-ID: thr_bwikffmsvp

### DIFF
--- a/apps/app/src/views/RootComposeRightPanelToggle.test.tsx
+++ b/apps/app/src/views/RootComposeRightPanelToggle.test.tsx
@@ -25,7 +25,7 @@ describe("RootComposeRightPanelToggle", () => {
     ).toEqual({ inlinePanelToggle: "button", showPinnedToggle: false });
   });
 
-  it("uses a disclosure state without painting the whole click target as selected", () => {
+  it("matches shared panel chrome without painting the click target as selected", () => {
     const onToggle = vi.fn();
 
     render(<RootComposeRightPanelToggle isOpen onToggle={onToggle} />);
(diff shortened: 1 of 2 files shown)


brsbl:
Closing because this PR addressed only opener styling and did not implement the requested new-thread tabbed-panel consistency work. A correctly scoped replacement will follow.

PR #1703, opened 2026-08-17
Tasks: let folders be deleted from Manage

### LINKED ISSUE #1701: Tasks: folders cannot be deleted from the UI
## What I was trying to do

Remove a folder I no longer need in the Tasks plugin.

## What happens

There is no way to delete a folder anywhere in the UI. The only folder surface is **Manage → Folders**, and its rows offer *Rename* and a parent *Select* only — no delete control (`plugins/tasks/views/manage/manage-panel.tsx`, `FolderRow`). The sidebar renders folders as collapse/expand headers with no menu at all.

The capability itself already exists everywhere else:

- RPC: `deleteFolder` in `plugins/tasks/api/index.ts`
- Contract: `deleteFolder` in `plugins/tasks/shared/contract.ts`
- Store: `deleteFolder` in `plugins/tasks/db/store.ts`
- CLI: `bb tasks` exposes it

So a folder created from the New project dialog can never be removed from the app — only via the CLI.

## What I expected

A delete control on the folder row, like the ones Labels and Presets rows already have, with a confirmation that says what happens to the folder's contents.

Worth noting for the confirmation copy: deleting is not destructive. Both FKs are `ON DELETE SET NULL`, so the folder's projects and subfolders move to the top level and no task is touched.

## Environment

bb 0.38.0, macOS.

I have a fix ready and will open a PR against this issue.

builtui:
Fixes #1701

## Problem

`deleteFolder` ships in the RPC contract, the API handler, the store, and the CLI — but no UI surface calls it. **Manage → Folders** rows offer rename and a parent select; the sidebar renders folders as collapse headers with no menu. A folder created from the New project dialog can only be removed via the CLI.

## Change

Adds the trash control the Labels and Presets rows already have, behind the shared `ConfirmDialog`.

The confirmation names the outcome instead of warning about loss, because deleting a folder is not destructive: both folder foreign keys are `ON DELETE SET NULL`, so its projects and subfolders move to the top level and no task is touched.

- `2 projects and 1 subfolder move to the top level. No tasks are deleted.`
- `The folder is empty.`

A failed delete surfaces in the section's existing `role="alert"` rather than closing silently, matching the labels flow.

## Tests

Two tests in `views/manage/manage.test.tsx`: the delete path (impact copy, then `deleteFolder` called with the right id) and the failure path (error surfaces instead of a silent close).

`turbo run typecheck test --filter=bb-plugin-tasks`: 35 test files pass, typecheck clean.

## Notes

No contract, CLI, or docs change — the surfaces already existed, only the app was missing one.

### DIFF
--- a/plugins/tasks/README.md
+++ b/plugins/tasks/README.md
@@ -65,23 +65,23 @@ resolve on the invoking machine: inside an agent thread that is the thread's
 machine, otherwise the server's machine; pass `--machine <id-or-name>` to
 target another enrolled machine.
 
-| Command                                        | Purpose                                                                                                                             |
-| ---------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------- |
-| `bb tasks status`                              | Show the installed Tasks plugin name and version.                                                                                   |
-| `bb tasks project create\|list\|show\|update`  | Manage tracker projects, folders, colors, prefixes, and bb-project links.                                                           |
-| `bb tasks folder create\|list\|update`         | Organize tracker projects into nested folders.                                                                                      |
-| `bb tasks create`                              | Create a task with description, priority, labels, due date, optional parent, and file attachments (repeatable `--attach <path>`).   |
-| `bb tasks list`                                | Page/filter tasks by project, status, priority, label, active agents, or search text; supports `--sort`, `--limit`, and `--cursor`. |
-| `bb tasks show <key-or-id>`                    | Show the complete task record, including comments, attachments, subtasks, and attached threads.                                     |
-| `bb tasks update <key-or-id>`                  | Update status, priority, title, description, due date, or labels.                                                                   |
-| `bb tasks comment <key-or-id>`                 | Add a Markdown comment from inline text or a file; optionally notify the latest responding task agent.                              |
-| `bb tasks attachment add\|get\|list\|remove`   | Add, fetch, list, or remove attachments. Referenced attachments require `remove --remove-references`.                               |
-| `bb tasks preset list\|create\|update\|delete` | Manage reusable agent execution presets.                                                                                            |
-| `bb tasks delegate <key>`                      | Start and attach a new agent thread using a preset.                                                                                 |
-| `bb tasks attach <key-or-id>`                  | Attach the current bb thread to a task when it was not delegated from Tasks.                                                        |
-| `bb tasks threads <key>`                       | List the bb threads attached to a task.                                                                                             |
-| `bb tasks label create\|list\|delete`          | Manage project-scoped labels.                                                                                                       |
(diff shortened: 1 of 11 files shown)


SawyerHood:
Addressed both SlopCop findings (rebased onto current `main`, pushed as 551d7c5):

1. **CLI path** — added `bb tasks folder delete <id-or-name> [--json]` next to `create|list|update`. Same semantics as the Manage dialog: the folder's projects and subfolders move to the top level, no tasks are deleted. Resolves by id or (unambiguous) name, fails with `folder not found: …`, and reports what moved (`movedProjectIds`/`movedFolderIds` in JSON, counts in the text message). `bb tasks --help`, `bb tasks folder --help`, and the plugin README list it. The SDK surface (`deleteFolder` RPC) already existed. Test covers delete by name and id, unknown folder, missing positional, and that the project/subfolder are unfiled while the task survives.
2. **False impact while projects load** — `FoldersSection` now gates the impact text on both `useFolders()` and `useProjects()` having loaded (`data` present and not in flight). Until then the dialog shows "Checking what the folder contains…" (or "Could not load projects: …" on a failed `listProjects`) and the confirm button is disabled (`ConfirmDialog` gained `confirmDisabled`). Tests cover a delayed and a failed `listProjects`.

`turbo run typecheck` clean; `turbo run test --filter=bb-plugin-tasks --force`: 35 files, 355 tests pass.

SawyerHood:
Pushed 98e95e195 addressing the second-round SlopCop findings:

1. **Stale impact after a failed refresh** — `FoldersSection` now derives `impactError = folders.error ?? projects.error`; `impactReady` requires both errors to be `null`, so a failed refetch that keeps cached rows shows "Could not load the folder's contents: …" and disables confirm. Test: cached data, then a failing `projects:changed` refetch, asserts the error text, no counts, disabled confirm, no RPC call.
2. **`deleted: false` ignored** — UI: the delete action throws (`Folder "x" was already deleted.`), the alert shows, and the panel refetches folders/projects since the server publishes nothing for a no-op delete. CLI: exits 1 with `folder not found: <address> (it was deleted by another client)` and no stdout (JSON mode included). Tests added for both.
3. **Moved IDs computed outside the transaction** — the store's `deleteFolder` now runs in a `db.transaction` that reads the folder's project/subfolder IDs via the indexed `folder_id` / `parent_folder_id` queries and deletes the row, returning `{ deleted, movedProjectIds, movedFolderIds }` (empty arrays when nothing matched). The RPC contract, API handler, CLI, and UI types all use that result; the CLI no longer scans every project. Store test covers the returned IDs and the second no-op delete.
4. **Folder command summary** — the registered `folder` summary now reads "Create, list, update, or delete project folders".

`pnpm exec turbo run typecheck --filter=bb-plugin-tasks` passes; `pnpm exec turbo run test --filter=bb-plugin-tasks --force` is 35 files / 359 tests passing.

PR #1744, opened 2026-08-17
Preserve active plugin host build stages

ymichael:
## Summary

- include the builder process id in each `.host-stage-*` directory name
- remove abandoned legacy or dead-process stages while preserving stages owned by a live process
- cover cleanup of an abandoned stage alongside a live stage

## Why

The cleanup added by #1730 removes every matching staging directory before a build. Concurrent integration workers build the same first-party provider artifacts, so one worker can delete another worker stage and cause `ENOENT` while reading or promoting the artifact.

This change was extracted from #1715 so the New Thread panel PR remains scoped to its UI work.

## Verification

- `pnpm exec turbo run test --filter=@bb/plugin-build --force` — 20 passed, 1 skipped
- `pnpm exec turbo run typecheck --filter=@bb/plugin-build --force` — passed
- `pnpm exec turbo run test --filter=@bb/integration-tests --force` — 55 passed
- `git diff --check` — passed

> AGENT GENERATED: by GPT-5

### DIFF
--- a/packages/plugin-build/src/build-plugin-host.test.ts
+++ b/packages/plugin-build/src/build-plugin-host.test.ts
@@ -123,15 +123,17 @@ describe("plugin host build", () => {
     );
   });
 
-  it("removes host staging directories left by interrupted builds", async () => {
+  it("removes abandoned host stages without touching a live process", async () => {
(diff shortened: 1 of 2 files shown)


ymichael:
Superseded by #1741, which landed the age-gated staging cleanup on main without PID-based stage ownership.

Issue #1558, opened 2026-08-13
bb-app can enroll a host against the wrong server after its server child fails with EADDRINUSE


markasoftware-tc:
(heads up: this whole issue is ai-written but i believe it's a real issue and ran into it myself. I got gpt-5.6-sol to debug and write up this issue) 

 When two `bb-app` launchers use different data directories but the same server and daemon ports, the second
  launcher can mistake the first launcher's server for its own.

  If the second server child fails with `EADDRINUSE`, `bb-app` may still consider startup successful because its
  health check receives a successful response from the existing server. It then requests a host enrollment key from
  that existing server while saving the resulting host credentials in the second data directory.

  After the existing server stops, the second launcher's host daemon connects to the server backed by its own
  database. That database does not contain the enrolled host, resulting in `INVALID_API_KEY` or a 401 response
  saying that the host is not registered.

  ## Environment

  - `bb-app`: 0.35.1
  - Node.js: 26.5.1
  - npm/npx: 11.17.0
  - OS: Linux x86_64

  ## Reproduction

  In one terminal, start bb with a separate data directory:

  ```bash
  BB_DATA_DIR=/tmp/bb-one npx bb-app@latest

  In another terminal, start bb using another data directory but the same default ports:

  BB_DATA_DIR=/tmp/bb-two npx bb-app@latest

  The second server child cannot bind port 38886 and exits with:

  listen EADDRINUSE: address already in use :::38886

  However, the launcher's health request to http://127.0.0.1:38886 can be answered by the first server. The second
  launcher may consequently report:

  ✓ Server listening on http://127.0.0.1:38886

  It then requests an enrollment key from the first server and writes the resulting host identity to /tmp/bb-two/
  auth.json.

  After stopping the first launcher and starting the second one again, the host daemon fails with errors such as:

  Server rejected host credentials — this host is not registered with the server.

  or:

  INVALID_API_KEY

  The host ID from /tmp/bb-two/auth.json is present in /tmp/bb-one/bb.db, but absent from /tmp/bb-two/bb.db.

  ## Expected behavior

  If the managed server child fails to bind its port, the launcher should fail startup and should not create or
  modify host authentication state.

  A health response from another process should not satisfy the managed child’s startup check.

  ## Actual behavior

  The health check accepts a successful response from an unrelated server on the configured port. The launcher then
  enrolls a host against that server while persisting the resulting credentials in a different data directory.

  This leaves valid credentials associated with the wrong database and produces a misleading API-key error on
[...]

PR #1755, opened 2026-08-18
docs: warn that npx installs cannot self-update

### LINKED ISSUE #1747: npx install pins bb to one minor line, so the update button silently does nothing
## Summary

Installing bb with `npx bb-app@latest` produces an install that can never update itself. npm materializes `@latest` into a caret range in the npx cache entry, and that range excludes every future minor. The in-app update affordance then appears to do nothing, with no error surfaced.

My install sat on 0.35.1 for 11 days while 0.37.0 was published, and I had no indication anything was wrong beyond the version number.

## Reproduction

1. `npx bb-app@latest` at a time when latest is `0.35.1`
2. Wait for a new minor to be published (`0.37.0`)
3. Use the update affordance in the bb UI
4. Observe: still running 0.35.1

## Cause

The npx cache entry at `~/.npm/_npx/<hash>/package.json`:

```json
{
  "dependencies": { "bb-app": "^0.35.1" },
  "_npx": { "packages": ["bb-app@latest"] }
}
```

The request was `bb-app@latest`, but npm recorded the resolved version as the caret range `^0.35.1`, meaning `>=0.35.1 <0.36.0`. On subsequent runs npm checks whether the cached tree satisfies that range. 0.35.1 does, so it skips reinstalling and never re-resolves `latest`. **0.37.0 cannot satisfy `^0.35.1`, so this install is permanently pinned to the 0.35.x line.**

This is not a stale cache that heals on its own. It is a hard pin, and it persists across restarts.

## Diagnosis notes

Two things made this hard to identify:

1. **No error is surfaced.** There were no download, install, or version-check log lines in either `server.*.log` or `host-daemon.*.log`, and no partial artifact in `~/.bb/install-cache` (which only ever held the tarball of the already-installed version). The update simply does not happen, silently.

2. **`hostNeedsUpdate` is a different path.** `apps/app/src/lib/host-update-status.ts` covers updating a *remote enrolled daemon* after a protocol-version rejection. It requires `host.status === "disconnected"` and a non-null `lastRejectedProtocolVersion`. That mechanism works, and my two enrolled Macs were fine throughout. It just has nothing to do with updating the machine running the server itself.

As far as I can tell, a web-surface install launched via npx has no self-update path at all. The real auto-update machinery lives in `apps/desktop`, and my `bb-app-runtime.json` reports `"surface": "web"`.

## Impact

Anyone who followed the `npx bb-app@latest` install path is frozen at whatever minor they first resolved and has no signal that updates are failing. The longer the install runs, the further behind it drifts.

## Suggested fixes

Any of these would help, roughly in order of how much:

1. **Do not recommend npx for persistent installs.** A plain `npm install --prefix <dir> bb-app@latest` re-resolves `latest` on every invocation, so a restart is genuinely an update. This is what I moved to and it works.
2. **Detect the pin and warn.** On startup, compare the running version against the registry's `latest` and surface a notice when they differ, rather than failing silently.
3. **Log the update attempt.** Even one line saying what was checked and what was resolved would have made this a five-minute diagnosis instead of an eleven-day one.
4. **Document the trap** in install docs.

## Environment

- bb 0.35.1 at time of discovery, now on 0.38.0 after switching install method
- Ubuntu 22.04.5, Node v24.19.0, npm via nvm
- `"surface": "web"`, launched via `npx bb-app@latest`

Joesirven:
Fixes #1747.

`npx bb-app@latest` writes the resolved version into its cache entry as a caret range:

```json
{ "dependencies": { "bb-app": "^0.35.1" }, "_npx": { "packages": ["bb-app@latest"] } }
```

`^0.35.1` means `>=0.35.1 <0.36.0`. On later runs npm sees the cached tree satisfies that range, skips reinstalling, and never re-resolves `latest`. 0.37.0 cannot satisfy it, so the install is pinned to the 0.35.x line permanently. My install sat on 0.35.1 for 11 days with no error surfaced anywhere.

This adds a short section to the README documenting the pin and pointing at `npm install --prefix` for installs meant to persist, since that re-resolves `latest` on every run.

Docs only, no code change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,25 @@ npx bb-app@nightly
 
 bb uses the provider CLI you already have authenticated.
 
+### Long-running installs: use `npm install`, not `npx`
+
+`npx bb-app@latest` is fine for trying bb out, but it cannot update itself. npm
+records the first resolved version as a caret range in its cache entry, so a
+cache created while `latest` was `0.35.1` pins to `^0.35.1` and will never
+resolve `0.36.0` or newer. The in-app update affordance then appears to do
+nothing, with no error.
+
+For a server or any install you intend to keep, install into a fixed prefix
+instead:
+
+```bash
+npm install --prefix ~/.bb/runtime-install bb-app@latest
+~/.bb/runtime-install/node_modules/.bin/bb-app
+```
+
+`npm install` re-resolves `latest` every time it runs, so restarting is a real
+update.
+
 For install requirements, provider setup, configuration, and package-focused
 docs, start with
 [`packages/bb-app`](./packages/bb-app/README.md).


SawyerHood:
Closing this one. While reproducing #1747 we tested the central claim (that `npx bb-app@latest` materializes a `^0.35.1` caret pin and can never move to a new minor) against the reporter's exact npx cache entry: libnpmexec re-resolves the `latest` tag on every run and upgraded 0.35.1 → 0.38.0. So the warning this PR adds would document a non-bug. The real failure the reporter hit is the launcher printing "bb is ready" while the old server still owns the port, which is #1558.

Full write-up with the repro: https://get-bb.github.io/reports/issues/1747.html

Thanks for digging into it — if you have a case where npx really does stay pinned, please reopen with the cache entry and npm version.

Issue #1748, opened 2026-08-18
Server heap is sized from host RAM, not the cgroup limit, so it fills the memory cap with garbage


Joesirven:
## Summary

The bb server never sets `--max-old-space-size`, so V8 auto-sizes its heap limit from total host RAM and ignores any cgroup memory limit the service runs under. On a memory-capped deployment the server is permitted a heap that is a large fraction of the entire budget, and it fills much of that with garbage it has no incentive to collect.

## Environment

- bb 0.38.0, Ubuntu 22.04.5, Node v24.19.0
- 4 cores / 4.9 GiB RAM, 2.5 GiB swap
- systemd user service with `MemoryHigh=3G`, `MemoryMax=3800M`

## Observed

V8's default heap limit on this host measures **1728 MB**, derived from the 4.9 GiB of physical RAM. The cgroup budget is 3072 MB. So the server alone may grow a heap that is **56% of the entire service budget**, and nothing in bb tells it otherwise.

The result is a large sawtooth. Measured server process:

- trough: 625-670 MB RSS
- peak: 1088 MB RSS
- a **427 MB drop observed within a single 20-second sample**, with no change in workload

That drop is the tell: the excess is collectable garbage, not live set. V8 simply has no reason to run a major GC until it approaches a ceiling that is far above what the cgroup can actually tolerate.

Composition at 1110 MB RSS was **1008 MB anonymous**, with only ~100 MB file-backed (SQLite mmap plus the node binary), so this is genuinely JS heap rather than mapped files. A bare `node` on this host is 40 MB.

Consequence: with bb's own floor at 1.2-1.5 GB and each agent session costing ~320-410 MB, only about 4-6 concurrent agent threads fit before the cap. `MemoryHigh` was breached **2578 times** in one session with only 2-4 agents running, and the service was OOM-killed once.

## Cause

`grep -rn "max-old-space-size"` across `apps/` and `packages/` returns nothing. It is not set in the launcher (`packages/bb-app/src/launcher.ts`), not in any spawn path, and not documented as something an operator should set.

## Suggested fix

Have the launcher pass `--max-old-space-size` in the server's `execArgv`, sized from the cgroup limit when one is detectable (`/sys/fs/cgroup/memory.max`) rather than from host RAM.

[...]

PR #1757, opened 2026-08-18
feat: allow capping the server's V8 heap independently of providers

Joesirven:
Refs #1748.

## Problem

Node sizes the V8 old-space heap from total system memory and ignores any cgroup limit the process runs under. On a 4.9 GiB host the default ceiling measures **1728 MB**, which is 56% of a 3 GiB `MemoryHigh` budget — so the server has no incentive to collect until well past what the unit tolerates.

Measured on that host: the server sawtooths between 625 MB and 1088 MB RSS, and I observed a **427 MB drop in a single 20-second sample with no change in workload**, which shows the excess is collectable garbage rather than live set. The unit was OOM-killed once with only 2-4 agent threads running.

## Why NODE_OPTIONS is not a workaround

`sanitizeInheritedChildProcessEnv` (`packages/process-utils/src/index.ts`) strips only `NODE_ENV` and keys beginning with `BB_`. `NODE_OPTIONS` is therefore inherited by every spawned provider process, so setting it to cap the server also caps each `claude` agent and risks crashing them mid-turn. The limit has to be applied by the launcher, to the server alone.

## Change

Adds an opt-in `BB_SERVER_MAX_OLD_SPACE_MB`, applied at both server spawn sites. Unset by default, so behavior is unchanged for everyone who does not set it.

I deliberately did not make this automatic. Deriving a heap size from `/sys/fs/cgroup/memory.max` is the obvious follow-up, but the right fraction is a judgement call I did not want to bake in from one host's measurements. Happy to add automatic detection if you would prefer that shape.

## Testing

Syntax verified with esbuild. I was **not** able to run the full typecheck or test suite — this is a fresh worktree without `node_modules`, on a 4-core box where a full install is expensive. The change is small and self-contained, but please treat CI as the first real verification.


SawyerHood:
Heads-up: the motivating issue #1748 was closed as not-a-bug (Node already sizes the heap from the cgroup limit; details in https://get-bb.github.io/reports/issues/1748.html). If you want to keep this as a per-process server heap cap feature, it needs the env path fixed first: `resolveServerNodeFlags` reads `process.env`, so `bb-app env set BB_SERVER_MAX_OLD_SPACE_MB` (which the docs row advertises) is silently ignored. Otherwise feel free to close.

PR #1771, opened 2026-08-18
Fix lost first tap on mobile follow-up submit

jshph:
## User-facing problem

On iPhone, submitting a follow-up can feel like the button ignored the first tap:

1. The user taps Submit.
2. The keyboard closes and the composer shrinks.
3. The message remains unsent.
4. The user has to tap Submit again.

This is especially confusing because it resembles network lag, even though the first submission never actually committed.

## What this changes

The submit button now checks which DOM element the browser actually considers focused instead of relying on TipTap's event-derived focus flag.

During the brief iOS focus transition where those states can disagree, bb keeps the editor and keyboard stable long enough for the first tap to produce its click. After submission commits, the existing keyboard dismissal and compact-composer behavior continue normally.

The change is intentionally narrow: genuine focus outside the editor behaves as before, and desktop submission behavior is unchanged.

## Verification

- Added a regression test that models stale TipTap focus while the contenteditable still owns native focus.
- All 98 PromptBoxInternal tests pass.
- Turbo app typecheck passes.
- Prettier and git diff checks pass.

> AGENT GENERATED: by GPT-5

### DIFF
--- a/apps/app/src/components/promptbox/PromptBoxInternal.test.tsx
+++ b/apps/app/src/components/promptbox/PromptBoxInternal.test.tsx
@@ -2174,6 +2174,29 @@ describe("PromptBoxInternal compact layout", () => {
     expect(document.activeElement).toBe(editor);
   });
 
+  it("keeps DOM focus through submit when TipTap focus state is stale", async () => {
+    render(
+      <PromptBoxInternal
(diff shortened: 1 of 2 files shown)


SawyerHood:
Ty for this fix!

PR #1772, opened 2026-08-18
Keep Settings accessible in the iPhone drawer

jshph:
## User-facing problem

On an installed iPhone PWA, Settings can become effectively inaccessible:

- The only entry is an unlabeled gear in the drawer footer.
- The standalone viewport can place that footer below the visible screen or home indicator.
- A user opening the drawer may therefore see neither an obvious Settings action nor the clipped footer control.

This is both a discoverability problem and a viewport-sizing problem.

## What this changes

On compact viewports, the drawer now shows a labeled Settings row beside the primary actions near the top. Desktop keeps the existing icon-only footer action, so its established layout does not change.

The standalone iOS shell also uses the dynamic viewport height plus the exposed top safe-area inset. This covers the physical screen when bb paints under a translucent status bar without extending the drawer below the home indicator when iOS already positions the app beneath an opaque status bar.

Together, these changes give iPhone users a reliable, visible route to Settings while keeping the rest of the footer accessible.

## Verification

- Added regression coverage for the standalone shell-height rule.
- All 14 focused shell and sidebar tests pass.
- Turbo app typecheck passes.
- Prettier and git diff checks pass.

> AGENT GENERATED: by GPT-5

### DIFF
--- a/apps/app/src/app-shell-height.test.ts
+++ b/apps/app/src/app-shell-height.test.ts
@@ -1,0 +1,17 @@
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { describe, expect, it } from "vitest";
+
+const appCss = readFileSync(
+  fileURLToPath(new URL("./app.css", import.meta.url)),
+  "utf8",
+);
+
+describe("app shell height", () => {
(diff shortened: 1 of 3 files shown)


SawyerHood:

PR #1775, opened 2026-08-18
host-daemon: give unanswered user questions a ten-minute deadline

DarrenTsung:
## Problem

A native user-question request — claude-code's `AskUserQuestion` forwarded over `interaction/request` — registers with the server and then resolves only when a client answers or the provider exits. With nobody watching the thread, the `registerAndWait` promise never settles: the turn never ends and the human's typed replies queue behind it (measured: a question pending 2h30m, the reply delivered 13 minutes later as an abort).

The plugin interaction path (`bb.ui.requestInput`) already defaults to a ten-minute timeout; the native path has none (`expiresAt: null`).

## Fix

`InteractiveRequestRegistry.registerAndWait` now arms a ten-minute deadline for `user_question` payloads (approvals keep their current behavior). On expiry the entry is dropped and rejected with `interactive_request_timeout`, which propagates to the provider bridge as a JSON-RPC error and resolves the tool call as a denial telling the model to ask in prose instead — one wasted turn instead of a multi-hour hang. A new `onTimeout` option (wired in `app.ts` to the existing interactive-interrupt push) marks the server row interrupted so the thread is not left stuck "already awaiting user interaction".

## Tests

- `interactive-request-registry.test.ts`: rejects an unanswered user question at the deadline (fails on the old code — the promise never settles), applies no deadline to approvals, clears the deadline when answered in time, and does not time out a question whose registration failed.
- Full `@bb/host-daemon` suite (586 tests) and typecheck pass.

### DIFF
--- a/apps/host-daemon/src/app.ts
+++ b/apps/host-daemon/src/app.ts
@@ -432,6 +432,13 @@ export async function createHostDaemonApp(
         threadIds: [request.threadId],
       });
     },
+    onTimeout: (request) => {
+      enqueueInteractiveInterrupt({
+        providerId: request.providerId,
+        reason: "User question timed out waiting for an answer",
+        threadIds: [request.threadId],
+      });
+    },
   });
 
   let sendServerMessage = (_message: HostDaemonDaemonWsMessage) => false;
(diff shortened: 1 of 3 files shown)


DarrenTsung:
Oops, I didn't even realize my agent would create this. But yeah I've had a few bb threads stuck behind a AskUserQuestion that never resolved and got interrupted for some reason. So I couldn't even answer the question when I realized the thread was waiting for me, needed a manual stop and new message to bump the thread.

SawyerHood:

PR #1781, opened 2026-08-18
Fix Linux AppImage runtime mount lifecycle

salemsayed:
## Summary

- launch the packaged Linux bb runtime through the outer AppImage so it receives an independent FUSE mount
- resolve the bridge script relative to the new runtime mount instead of retaining a path into the desktop shell mount
- make the AppImage supervisor the persistent process-group leader while the bridge and its descendants inherit that anchored group
- preserve the original bridge path in the supervisor command line for stale-PID ownership verification
- exercise the generated AppImage in CI by tearing down its GUI mount and verifying the owned runtime remains healthy

## Root cause

The packaged desktop launched its long-lived bb-app bridge with `process.execPath`. Inside an AppImage, that path points into the desktop shell temporary FUSE mount. If the shell or its AppImage mount helper exited while the bridge server, host daemon, watcher, or plugin workers remained alive, those processes continued executing files from a vanished mount. The observed host daemon and Parcel watcher failures were `SIGBUS` / `BUS_ADRERR` instruction-fetch faults at their respective program counters.

Starting only the bridge from a second AppImage mount was insufficient because the bridge can exit before its descendants. The bootstrap now owns the independent mount and starts as a detached process-group leader. The bridge inherits that group, while the supervisor remains alive until no non-zombie group descendants remain. Because the supervisor itself owns the PGID for that entire interval, bounded group-wide SIGKILL escalation cannot race with PGID reuse.

## Validation

- `pnpm exec turbo run test typecheck --filter=@bb/desktop --force` — 33 files, 217 tests passed; typecheck passed
- `pnpm exec turbo run desktop:build:linux --filter=@bb/desktop` — 9 tasks passed and produced `bb-0.38.0-x86_64.AppImage`
- launched that AppImage with isolated data and random ports, removed the GUI AppImage mount, and confirmed the supervisor/runtime stayed on its independent mount
- confirmed repeated HTTP health probes remained successful after GUI mount teardown and that cleanup left no test mounts, processes, or new core dumps
- retained the earlier stale-runtime relaunch validation: the packaged desktop safely reaped the prior supervisor/process group and started a fresh managed runtime

> AGENT GENERATED: by GPT-5


SawyerHood:

PR #1785, opened 2026-08-18
Add optional frameless Linux desktop windows

salemsayed:
## Summary

- add a Linux-only `--no-window-frame` startup option
- keep the native Electron frame as the default
- document the option for window managers that provide their own controls
- cover argument resolution and BrowserWindow construction

## Why

Tiling Wayland compositors such as Hyprland already provide window movement,
resizing, borders, and close controls. Electron's native Linux menu/title bar
duplicates that chrome and consumes vertical space. This opt-in mode lets those
users rely on their window manager without changing the default experience for
other Linux desktops.

The option is startup-only because Electron chooses `frame` when constructing a
`BrowserWindow`. macOS and Windows ignore the Linux-only flag.

## Screenshots

Before — native Linux frame:

bb with the native Linux frame

After — `--no-window-frame`:

bb without the native Linux frame

## Validation

- `pnpm --filter @bb/desktop test` — 34 files, 217 tests passed
- `pnpm --filter @bb/desktop typecheck`
- focused Prettier and `git diff --check`
- built `bb-0.38.0-x86_64.AppImage`
- launched native-frame and frameless AppImages side by side on Hyprland with isolated data and ports
- verified the frameless native Wayland window, server health, and host-daemon listener

### DIFF
--- a/apps/desktop/README.md
+++ b/apps/desktop/README.md
@@ -101,6 +101,16 @@ Running an AppImage normally requires FUSE and, on some distributions, the
 `libfuse2` compatibility package. If FUSE is unavailable, launch it with
 `--appimage-extract-and-run` instead.
 
+Linux users whose window manager supplies all window controls can remove the
+native Electron title bar with `--no-window-frame`:
+
+```bash
+./bb-x86_64.AppImage --no-window-frame
+```
+
+The native frame remains the default. Changing this startup option requires a
+full desktop app restart.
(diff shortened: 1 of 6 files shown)


SawyerHood (approved):
Rebased on the latest main and reviewed. The desktop typecheck and all 235 desktop tests pass locally. I found no blocking issues.

Issue #1304, opened 2026-08-10
Composer keystroke cost scales with the mounted timeline size


SawyerHood:
Found while profiling the mobile web frontend against a seeded database (`pnpm seed:perf`, #1300). Follow-up to the tap-freeze work in #1269 / #1285: this looks like a remaining re-render amplifier.

## Numbers

With a 2,633-event thread fully loaded (~7.7k DOM nodes mounted), typing 67 characters at a 25 ms cadence took 2,966 ms wall against a 1,675 ms schedule — **~19 ms of main-thread overhead per keystroke**, plus a 94 ms long task mid-sentence (desktop CPU, dev build). At phone speeds this approaches per-keystroke jank, and it compounds with the timeline growth in the timeline-virtualization issue.

## Repro

1. `pnpm seed:perf` on a dev instance.
2. Open a large seeded thread and scroll far up to mount history.
3. Type a long sentence into the composer; watch main-thread work per input event in the performance panel.

## Suggested approach

Profile which ancestors re-render per keystroke; isolate composer draft state from the timeline tree (memo boundary or external store subscription) so typing re-renders only the composer. Timeline virtualization would also shrink the amplified tree.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

SawyerHood:
**Reproduction & root-cause report:** https://get-bb.github.io/reports/issues/1304.html

- **Verdict:** REPRODUCED · root-cause confidence: high
- **Root cause (short):** ThreadDetailViewInternal subscribes to the composer draft store via usePromptDraftStorage (only to get the stable addQuote/storageKey), so every keystroke re-renders the whole 3k-line thread view; each render rebuilds threadOpenContext, which (identity-keyed memo in useLocalOpenTargets) churns openPathInFileTarget -> getLocalFileContextMenuItems -> the MarkdownLocalFileContextMenuContext.Provider value wrapping the timeline, forcing React to scan the entire mounted timeline subtree (1.5k fibers small, 17.9k fibers big) for consumers on every keystroke.
[...]

PR #1818, opened 2026-08-18
Avoid main thread rerenders from prompt draft updates

Roystbeef:
## Summary

typing is slow as hecc. this helps alleviate that by making sure `ThreadDetailView` isn't getting completely re-rendered on every keystroke

- move prompt-draft mutations into the non-subscribing imperative accessor
- compose `usePromptDraftStorage` from that accessor plus its reactive draft snapshot
- use the accessor in `ThreadDetailView`, which needs quote actions and the storage key but does not render draft contents
- verify imperative writes still update subscribed composer state

## Root cause

`ThreadDetailView` called `usePromptDraftStorage` only to access `addQuote` and `storageKey`. Because that hook subscribes with `useSyncExternalStore`, every composer keystroke re-rendered the full thread-detail view even though it did not consume the changing draft value.

The action-only accessor now owns the shared mutations, while components that render draft contents continue using the subscribed hook. This removes the broad thread-detail render from the typing path without duplicating mutation behavior or requiring value consumers to construct the same scope twice.

## Validation

- `pnpm exec turbo run test --filter=@bb/app -- src/hooks/usePromptDraftStorage.test.tsx`
- targeted ESLint and Prettier checks for the changed files
- browser performance traces confirmed `ThreadDetailViewInternal` no longer renders on each keystroke and steady-state input latency improved

> AGENT GENERATED: by GPT-5

### DIFF
--- a/apps/app/src/hooks/usePromptDraftStorage.test.tsx
+++ b/apps/app/src/hooks/usePromptDraftStorage.test.tsx
@@ -3,6 +3,7 @@
 import { act, cleanup, renderHook } from "@testing-library/react";
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import {
+  getPromptDraftAccessor,
   usePromptDraftInputThreadIds,
   usePromptDraftStorage,
 } from "./usePromptDraftStorage";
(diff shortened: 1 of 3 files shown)


Roystbeef:
also related to https://github.com/get-bb/bb/issues/1304

Roystbeef:
lol, i didn't actually read @ymichael's changes in https://github.com/get-bb/bb/pull/1799. Closing since the perf improvements here are already covered in that PR

Issue #1690, opened 2026-08-17
Fail adding remote machine for new install


jeyrb:
* Installed latest bb via get-bb.app download (0.38.0)
* In settings, chose *Add Machine*

Following cmd generated
```
curl -fL --progress-meter --connect-timeout 10 --max-time 60 --retry 2 http://127.0.0.1:38886/install.sh | sh -s -- --join-code bbde_NXKSbuLNhIWuWkTtqBrnZkvnFltREDACTylfKjXsTlnmpfZtRSGuhYLkpHWINFUs --host-id host_ufzfjur8m2 --server http://127.0.0.1:38886
```

Two issues - localhost in URLs provided to run on remote linux box, and secondly the port is only opened on the MacBook on localhost, so isn't accessible if the IP address manually changed to the MacBooks ip

SawyerHood:
**Reproduction & root-cause report:** https://get-bb.github.io/reports/issues/1690.html

- **Verdict:** REPRODUCED · root-cause confidence: high
- **Root cause (short):** AddMachineDialog collapses an unpaired/absent connect plugin to null and pairingCommand() silently falls back to the system-config serverUrl, which on a fresh desktop install is the loopback-bound http://127.0.0.1:38886, so it prints a curl installer no other machine can reach without telling the user to set up Remote access first.
- **Linked PR review:** #1695 → **REQUEST CHANGES (minor)** — Every 503 from the connect RPC is classified as transient 'unavailable'; a disabled connect plugin returns 503 'not running (status: disabled)', so the user gets 'Remote access isn't ready yet.' + a Try again dead end and the 'Waiting for the machine to connect…' spinner, with no link to enable the 
- **Proposed fix:** In AddMachineDialog: distinguish connect unpaired/absent/disabled from transient failure; when no connect machine code exists and the direct serverUrl hostname is loopback/unspecified, render an explanatory notice linking to /settings/plugins/connect and docs instead of a curl command, and hide the 'Waiting for the machine to connect…' box (PR #1695 does this; extend it to treat 503 'status: disabled' as unpaired). Optionally expose the server bind host in GET /system/config so reachability is server-owned rather than a client hostname heuristic.
[...]

PR #1819, opened 2026-08-18
Stop bb connect from truncating the bb-app package download

abdoutelb:
## The bug

"Add a machine" never reaches its join step through a paired bb connect server. The installer reports the package as downloaded, then npm fails on it:

```
✓  Downloaded the server's bb-app package
○  Installing the server's bb-app build
npm warn tar zlib: unexpected end of file
npm error code Z_BUF_ERROR
✗  Could not install bb-app for this machine.
```

No machine is ever added. Reproduced against a live connect-paired server — same URL, two routes:

| route | status | `content-length` | bytes received |
| --- | --- | --- | --- |
| `http://127.0.0.1:38886/install/bb-app.tgz` | 200 | 32167180 | 32167180 ✅ |
| `https://<handle>.getbb.app/install/bb-app.tgz` | 200 | *absent* | 10894137 ❌ |

The truncation point moves between attempts (9318569 bytes on another run), and `curl` exits 0 both times.

## Root cause

`TunnelSession.executeHttp` relayed the origin body with no backpressure — `for await (const chunk of res) … this.send(frame)`. For the ~32 MB package that queues tens of megabytes into `ws`'s send buffer as fast as loopback delivers it. The heartbeat frame then sits *behind* that backlog, `HEARTBEAT_RESPONSE` cannot return inside `HEARTBEAT_DEADLINE_MS` (60 s), and at the next 20 s tick the watchdog calls `tunnel.terminate()` on a perfectly healthy tunnel. Everything still queued is dropped, and `send()` silently no-ops once the socket is not `OPEN`.

The server log shows it on every large download:

```
tunnel heartbeat missed; reconnecting
tunnel closed (code 1006); reconnecting in 1000ms
```

Nothing downstream could catch the short body: workerd drops `content-length` for a relayed stream, so both `curl` and `fetch` treat a truncated response as complete, and `install-machine.sh` only checked the HTTP status before handing the file to npm.

## The fix

**Tunnel client** (`packages/tunnel-client/src/session.ts`)

- Relay bodies against a 1 MiB send high-water mark. `for await` holds the origin stream paused while waiting, so an unkeepable body backs up onto the origin socket rather than into this process or the tunnel queue.
- Treat flush progress as liveness: the watchdog no longer terminates a socket that is still handing bytes to the peer. An idle tunnel has no flush progress, so a genuinely dead peer is still caught by the deadline.

**Package integrity** (`/install/bb-app.tgz`)

- Publish the artifact's size and sha256 as `x-bb-package-bytes` / `x-bb-package-sha256`. Response headers ride in `resp-head`, ahead of the body, so they survive the relay that strips `content-length`.
- `install-machine.sh` verifies both before installing, retries once, and reports `received N of M bytes` instead of letting npm surface a `zlib` error.
- The daemon's protocol self-update had the same unguarded `arrayBuffer()` → `npm install` path and now verifies too, so a protocol bump can't strand enrolled machines the same way.
- Stream the tarball from disk instead of buffering 32 MB per request, and keep it out of the compress middleware so its `content-length` survives.

## Protocol version

Not bumped. The change adds response headers only, and only a server newer than its client ever sends them: the installer is fetched from the same server that serves the package, and the daemon self-updates only toward a higher protocol version. Nothing in session payloads, WebSocket messages, or host RPC changed. Flagging it explicitly since `AGENTS.md` asks for deliberate reasoning here — and bumping *before* this fix ships would force every enrolled daemon through the broken download.

## Verification

- New `packages/tunnel-client/test/session-body-relay.test.ts` covers both halves. Both tests fail on the current code (`expected 8389455 to be 2985350` — the queue kept growing; `expected 3 to be +0` — the watchdog killed a busy tunnel) and pass with the fix.
- New installer cases: truncated package rejected (and not installed), a retried download recovering, and a missing digest refused.
- New self-update cases: truncated and undigested downloads both refused before `npm install`.
- `@bb/server` 1733 tests, `@bb/host-daemon` 585, `@bb/tunnel-client` 14, `@bb/connect` 97, `bb-plugin-connect` 78 — all passing; typecheck clean across the touched packages.
- Digest service checked against a real 3 MB artifact (size, sha256, memoization).

One test-infra note: the `machine install script` describe block now carries a 30 s timeout. Those cases drive the real script through `sh` (spawning curl, npm, node, and a daemon that waits on a socket) and several already ran within a second of the 5 s default when the rest of the suite competed for cores — one timed out mid-review before the bump.

## Not covered here

The dialog emits `--server http://127.0.0.1:38886` when the server is *not* paired to bb connect, which can never work pasted on another machine. Separate defect, left alone.


abdoutelb:
linking #1690

abdoutelb:
@SawyerHood  can you take alook for this PR it can speed up the fixing of solving 'Add a machine' functionality

PR #1824, opened 2026-08-18
Give the ACP plugin its own icon

ymichael:
The ACP plugin's `bb.branding.icon` was `./icons/cursor.svg` — the same asset its `acp-cursor` provider declaration uses — so the protocol-level plugin wore one agent's brand in settings and sidebar chrome.

- Add `plugins/provider-acp/icons/acp.svg`, the official Agent Client Protocol mark (from `zed-industries/agent-client-protocol` `docs/logo/light.svg`), reduced to a single `currentColor` path with a tight viewBox for the mask-based `PluginIcon` rendering. The `acp-cursor` provider keeps `cursor.svg`.
- Reword the description away from the plugin's internal migration state ("starting with Cursor") toward what a user gets.

Note: the ACP mark is an intrinsically wide wordmark (2.6:1), so at a 16px icon box it renders ~16x6px next to the square Cursor/Codex/Claude glyphs. Happy to crop it to a single chevron if that reads badly.

### keep-awake icon fix

`keep-awake` declared `"Coffee"`, which was absent from the shared-ui `ICON_MAP`, so `pluginIconName` silently fell back to `Zap`. Mapped `Coffee` to `Coffee02Icon` and regenerated the derived registry/template copies.

The existing server assertion only compared manifest strings, so it stayed green on a name that doesn't resolve. Added an app-side test that every shipped plugin's named `branding.icon` round-trips through `pluginIconName` — verified it fails on a deliberately bogus name.

### Drive-by

`packages/templates/src/generated/plugin-sdk-dts.generated.ts` was stale on main (missing `provider-registrations-changed`), failing `@bb/templates` typecheck; regenerated.

### Testing

`@bb/app` (2882 tests) + lint, `@bb/shared-ui`, `@bb/plugin-registry`, `@bb/templates`, `bb-plugin-provider-acp` (146 tests), and the `@bb/server` plugin suites pass.


ymichael:
Follow-up pushed: the browse tab (`/extensions/plugins`) showed the generic Zap glyph where the installed tab showed real icons.

**Cause:** a bundled catalog entry shipped `icon: "./icons/acp.svg"` with `iconUrl: null`. A path is not a host glyph name, so `pluginIconName` fell back to `Zap` — this hit every provider plugin (acp, codex, claude-code, pi), not just ACP.

**Fix:**
- Serve a bundled entry's own compact SVG from the existing `/plugin-catalog/icons/:marketplace/:entryId` route. Search hashes the file for the cache-busting URL; `icon()` reads it when the marketplace cache has no fetched bytes.
- Render those bytes as a CSS mask instead of an ``. A plugin-owned compact icon is single-color artwork drawn in `currentColor`, and inside an image document `currentColor` resolves against that document — verified in the running app that the `` path painted the marks black on the dark theme.

Verified in the dev app in both light and dark mode: browse cards now show the ACP, Codex, Claude Code, and Pi marks in the surrounding ink color. `@bb/app` (2884 tests) and `@bb/server` (1726 tests) pass.

PR #1826, opened 2026-08-18
Splice timeline windows on event coverage instead of row arithmetic

ymichael:
## Problem

Submitting a follow-up on a large thread flashes a truncated timeline: the loaded pages vanish, only the newest window renders, and auto-load then pages the history back in.

A thread past the timeline byte budget (`thr_gcuc46ug4j`: 35.5k events, ~61 MB) gets a latest window of a single segment — 4 rows. Everything above it is auto-loaded. Before rendering a fresh latest window, `useThreadTimelineController` asks whether it continues the loaded rows; when it decides it does not, it drops every loaded page and rebuilds from that 4-row window.

## Why the check was wrong

It answered that question with row source-sequence arithmetic, which rows cannot support:

- **Most events never become rows** — `turn/completed`, token-usage and rate-limit updates — so a continuous history routinely shows a non-zero gap between the last loaded row and the next window.
- **Rows sort by where they start, not where they end**, so the last row is not the one that reaches furthest. A turn summary spans its whole turn while shorter rows that begin later sort after it.
- **A window's first row can start below the window**, because the projection backfills a turn's `turn/started` row from under the cut. That tripped the "latest reaches back past the loaded rows" branch on windows that were perfectly contiguous.

On `thr_gcuc46ug4j` the loaded tail was a `provider/error` row ending at 62634, the turn summary above it reached 62635, and the follow-up prompt opened the next window at 62636. One event short — because `turn/completed` at 62635 is not a row.

## Fix

The server already states the interval each page covers: a pagination cursor names the first sequence its page covered (the invariant that makes older pages chain, `apps/server/src/services/threads/timeline.ts:1330`), and a latest window runs to `maxSeq`, the thread head at read time. `LoadedTimelineState` now carries that head as `coverageEndSequence`, and the splice test compares the two intervals instead of inferring them from rows.

The genuinely unspliceable shapes still rebuild, now stated exactly:

- a latest window starting **below** the loaded one — a turn watched mid-flight that collapsed into one summary row; the fresh window is a superset, so nothing is lost;
- a latest window starting **past** the loaded head — history neither response contains, which `olderCursor` could never page back to.

Coverage is also filter-independent, so filtered surfaces (`rowFilter`) stop understating their reach and rebuilding more often than the main timeline.

Client-only: no contract or server change, since `maxSeq` and the cursor were already on the wire.

## Verification

- Two new tests in `useThreadTimelineController.merge.test.ts` — the follow-up shape from the reported thread, and the backfilled-first-row shape. Both fail against the previous logic (verified by running them against the stashed `HEAD` controller) and pass here.
- Existing merge tests re-expressed in coverage terms, including the turn-collapse rebuild, the real-gap rebuild, and stale-cursor recovery.
- `turbo run test --filter=@bb/app -- timeline`: 186 passed. `typecheck` clean, `lint` 0 errors.
- Against live data for `thr_gcuc46ug4j`: latest window is `olderCursor.anchorSeq 62636 … maxSeq 62718` and the loaded head was 62635, so `62636 <= 62636` splices instead of dropping.

### DIFF
--- a/apps/app/src/components/thread/timeline/useThreadTimelineController.merge.test.ts
+++ b/apps/app/src/components/thread/timeline/useThreadTimelineController.merge.test.ts
@@ -22,6 +22,7 @@ interface TimelineTestRowArgs {
 
 interface TimelineTurnTestRowArgs extends TimelineTestRowArgs {
   children?: TimelineRow[];
+  endSequence?: number;
 }
 
 function timelineCursor(args: TimelineTestRowArgs): TimelinePaginationCursor {
@@ -83,7 +84,7 @@ function turnSummaryRow(args: TimelineTurnTestRowArgs): TimelineTurnRow {
     threadId: "thread-1",
     turnId: "turn-1",
     sourceSeqStart: args.sequence,
-    sourceSeqEnd: args.sequence,
+    sourceSeqEnd: args.endSequence ?? args.sequence,
     startedAt: args.sequence,
     createdAt: args.sequence,
     kind: "turn",
@@ -94,9 +95,15 @@ function turnSummaryRow(args: TimelineTurnTestRowArgs): TimelineTurnRow {
   };
 }
 
+/**
+ * `headSequence` is the thread's highest event sequence when the page was read,
+ * which the server reports as `maxSeq`. Together with the cursor naming the
+ * window's first sequence it states the event range the response covers.
+ */
 function makeTimelineResponse(
   rows: TimelineRow[],
   olderCursor: TimelinePaginationCursor | null,
+  headSequence: number,
 ): ThreadTimelineResponse {
   return {
     rows,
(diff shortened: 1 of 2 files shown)


ymichael:
Closing as a duplicate of #1639, which reached the same conclusion first and goes further.

Both changes stop the client from inferring window continuity from projected rows and compare the raw event-sequence ranges the server already states (`timelinePage.olderCursor.anchorSeq` … `maxSeq`). #1639 additionally repairs the splice itself — retaining every loaded row ending before the new window and updating straddling rows in place — which fixes accepted steers disappearing in Side chat, a second bug this PR did not address.

Verified before closing: both regression cases from this PR pass unmodified against #1639's controller. They now live on that branch (16f390b85), along with the comment explaining why row arithmetic cannot answer the continuity question.

PR #1831, opened 2026-08-18
Cut CI wall clock by fixing the macOS runner and Turbo cache handling

ymichael:
## What was wrong

The macOS `package-smoke` job gated every CI run at ~6:37 while every Blacksmith Linux job finished in under 3:30 ([run 32190967208](https://github.com/get-bb/bb/actions/runs/32190967208)). Three compounding causes, none of them individually wrong:

- The job still ran on GitHub-hosted `macos-latest`, even though `blacksmith-6vcpu-macos-15` was already in use by [build-desktop.yml](https://github.com/get-bb/bb/blob/fae5560b3/.github/workflows/build-desktop.yml#L35) and [publish-bb-app.yml](https://github.com/get-bb/bb/blob/fae5560b3/.github/workflows/publish-bb-app.yml#L505). The Blacksmith migration was left half-done, and the leftover job set the length of every run.
- Its Turbo cache returned **0 of 8 task hits** while costing ~75s to restore and ~40s to save a ~2 GB directory. Blacksmith's transparent cache only backs their Linux runners, so on macOS the blob moved over GitHub's own service at ~50 MB/s. `smoke:tarball` is itself `"cache": false`, so only its build dependencies could ever have been reused.
- `.turbo/cache` was restored, appended to, and saved under a new `${{ github.sha }}` key with nothing ever evicting an entry, so it grew without limit. The `checks` cache reached 2.9 GB and took **40.6s** to restore — longer than the 61s of build+typecheck+lint it exists to skip.

The last two multiplied into a quota failure: at ~6 runs/hour × 2.2 GB the six macOS entries put the repository 40% over its 10 GB Actions cache quota, where they evicted each other — which is why that cache never hit. GitHub enforces the quota by silently dropping least-recently-used entries: no failure, no warning, no notification. The only symptom that reached a human was "CI feels slow", which is why it persisted.

## What changed

- **`.github/workflows/ci.yml`** — both macOS smoke matrices move from `macos-latest` to `blacksmith-6vcpu-macos-15` (6 vCPUs vs 3) and set `turbo-cache: "false"`. Superseded pushes to `main` are no longer cancelled: a cancelled run never reaches its post step and so never saves a cache, which let a burst of merges leave pull requests restoring from a stale `main`. The `checks` job gains a cache-usage report and an explicit `contents: read` / `actions: read` permission block.
- **`.github/actions/setup-workspace/action.yml`** — new `turbo-cache` input (default `"true"`) gates cache restore and save. The cache key now reads across *every* job prefix while still writing to its own: Turbo entries are content-addressed by task hash, so another job's entry is either a hit or absent, never wrong, and the test shards can now reuse builds the `checks` job already did. No two jobs race on one key.
- **`scripts/prune-turbo-cache.mjs`** (new) — caps `.turbo/cache` at 1 GB, newest first, immediately after restore, so the pruned directory is what the job saves back. This bounds restore latency, not the quota; the quota is fixed by the macOS change above.
- **`scripts/check-actions-cache-usage.mjs`** (new) — reports Actions cache usage every `checks` run and warns past 70%. It never fails the build; a missing token or unreachable API is infrastructure, not a finding. Its only job is to make a silent leak visible.

No release path is touched. The `blacksmith-6vcpu-macos-15` runners in `build-desktop.yml` and `publish-bb-app.yml` are desktop build + Apple signing jobs, and the npm publish jobs remain on GitHub-hosted `ubuntu-latest`. The runners changed here only run `smoke:tarball`.

Out of band, I deleted the 9.05 GB of stale macOS Turbo cache entries; the repository went from 40% over quota to 744 MB of 10 GB. They will refill until this merges.

## How you verified

`prune-turbo-cache.mjs` against a real Turbo cache rather than only synthetic fixtures — pruning to 1 MB evicted the older entry and the survivor still replayed as a genuine hit:

```
Turbo cache prune: removed 1 of 2 entries (1.2 MB); 0.2 MB kept of 1 MB budget.
@get-bb/plugin-sdk:build: cache hit, replaying logs 0e6257b71a27ec3f
@bb/scripts:build: cache miss, executing 129c10bb8c07a245
```

The local hash `0e6257b71a27ec3f` matches the one in CI's logs, confirming entries are portable across machines. Separately verified that the prune keeps the `<hash>.tar.zst` / `<hash>-meta.json` pair together (a stranded meta file reads as a corrupt hit rather than a miss), leaves unrelated files untouched, and no-ops cleanly on a missing directory.

`check-actions-cache-usage.mjs` against the live API on all four paths — normal, warn, unset env, HTTP 401 — each exiting 0.

Both YAML files parse and the matrices expand to the intended values. `prettier --check` passes on all four files.

**Not verified:** end-to-end timing, which needs a real run on the changed workflow — this PR's own CI is that run. Expect the critical path to drop from ~8 min to ~3.5 min, with macOS smoke and `Tests (app)` roughly tied. Worth confirming against the job timings here before merging.

## Known limitation

The prune evicts by mtime, which records when an entry was *created*, not last *used* — Turbo reads an entry on a hit without touching it. So it can evict a stable package's long-lived entry that hits every run. It self-corrects (the entry misses once, is recomputed, and is re-saved with a fresh mtime) and only bites if the live set approaches the cap, currently ~400–500 MB against 1 GB. The precise alternative is pruning after the run from Turbo's `--summarize` output, which needs a second composite action invoked at the end of every job; happy to build it if the trade isn't worth it.

## Follow-up found, not fixed

`packages/plugin-sdk/bundled-types/*.d.ts` is git-tracked but regenerates non-deterministically — zod enum keys come out in a different order each build. Since it is a tracked input to `@get-bb/plugin-sdk#build`, every commit carrying a regenerated copy invalidates that build's cache for everyone, and it is a `dependsOn` of the server, templates, plugin-build, agent-runtime and integration test tasks, so the miss cascades. Intermittent rather than per-run, and a separate piece of work.

No issue filed; this came out of a direct investigation request.

### DIFF
--- a/.github/actions/setup-workspace/action.yml
+++ b/.github/actions/setup-workspace/action.yml
@@ -9,8 +9,12 @@ inputs:
     description: pnpm version to install
     required: true
   cache-prefix:
-    description: Distinguishes Turbo cache entries per job type (e.g. checks, test-server)
+    description: Names this job's Turbo cache entry (e.g. checks, test-server). Every job reads every prefix; this only decides which entry the job writes back.
     required: true
+  turbo-cache:
+    description: Restore and save the Turbo cache. Set to "false" on runners where the cache costs more to move than the tasks cost to run.
+    required: false
+    default: "true"
 
 runs:
   using: composite
@@ -27,14 +31,28 @@ runs:
         node-version: ${{ inputs.node-version }}
         cache: pnpm
 
+    # Writes to a prefix-scoped key but reads from every prefix: Turbo cache
+    # entries are content-addressed by task hash, so an entry another job saved
+    # is either a hit or absent, never wrong. Without the shared read the test
+    # shards could not see the builds the checks job had already done.
     - name: Cache Turbo task outputs
+      if: inputs.turbo-cache == 'true'
       uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
       with:
         path: .turbo/cache
-        key: ${{ runner.os }}-node-${{ inputs.node-version }}-${{ inputs.cache-prefix }}-turbo-${{ hashFiles('pnpm-lock.yaml') }}-${{ github.sha }}
+        key: ${{ runner.os }}-node-${{ inputs.node-version }}-turbo-${{ hashFiles('pnpm-lock.yaml') }}-${{ inputs.cache-prefix }}-${{ github.sha }}
         restore-keys: |
-          ${{ runner.os }}-node-${{ inputs.node-version }}-${{ inputs.cache-prefix }}-turbo-${{ hashFiles('pnpm-lock.yaml') }}-
-          ${{ runner.os }}-node-${{ inputs.node-version }}-${{ inputs.cache-prefix }}-turbo-
+          ${{ runner.os }}-node-${{ inputs.node-version }}-turbo-${{ hashFiles('pnpm-lock.yaml') }}-${{ inputs.cache-prefix }}-
+          ${{ runner.os }}-node-${{ inputs.node-version }}-turbo-${{ hashFiles('pnpm-lock.yaml') }}-
+          ${{ runner.os }}-node-${{ inputs.node-version }}-turbo-
+
+    # The restored directory is what this job saves back, so pruning here is
+    # what bounds it. Nothing else evicts entries, and an unbounded cache costs
+    # more to restore than the tasks it skips.
+    - name: Prune the Turbo cache
+      if: inputs.turbo-cache == 'true'
+      shell: bash
+      run: node scripts/prune-turbo-cache.mjs --dir .turbo/cache --max-size-mb 1024
 
     - name: Install dependencies
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,11 @@ env:
 
 concurrency:
   group: ci-${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  # Superseded pull request pushes are wasted work, but superseded main pushes
+  # are not: a cancelled run never reaches its post step, so it never saves a
+  # Turbo cache. Cancelling them let a burst of merges leave every pull request
+  # restoring from a main commit well behind HEAD.
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
   checks:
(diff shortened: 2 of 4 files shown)


ymichael:
## First CI run: green, and one caveat worth flagging before merge

All checks pass on [run 32193242148](https://github.com/get-bb/bb/actions/runs/32193242148). Full run wall clock **4:56, down from ~8:00** — and that is on a *cold* cache, so it understates the steady state.

### The macOS job, which is what this PR is about

This comparison is trustworthy even cold, because the old job got 0 of 8 cache hits — it never had a warm state to lose.

| Step | Before | After | Δ |
|---|---|---|---|
| Setup workspace | 2:19 | 0:38 | −101s |
| Smoke bb-app tarball | 3:37 | 2:24 | −73s |
| Post / cleanup | 0:31 | 0:23 | −8s |
| **Total** | **6:37** | **3:32** | **−185s** |

Both changes contributed about as predicted: dropping the useless 2 GB cache accounts for the setup win, and 6 vCPUs vs 3 for the build win.

### Caveat: the first run after merge will be a one-time cold start

Changing the cache key layout from `...-<prefix>-turbo-<lock>-` to `...-turbo-<lock>-<prefix>-` means no existing restore key matches. Confirmed in the logs:

```
Cache not found for input keys:
  Linux-node-22.x-turbo-<lock>-checks-<sha>,
  Linux-node-22.x-turbo-<lock>-checks-,
  Linux-node-22.x-turbo-<lock>-,
  Linux-node-22.x-turbo-
Turbo cache prune: .turbo/cache does not exist yet; nothing to do.
```

So on this run every Linux job rebuilt from scratch and the task steps look *slower* than baseline (`checks` build/typecheck/lint 143s vs 61s warm; `Tests (server)` 138s vs ~30s warm). That is the cache being rebuilt, not a regression — `Setup workspace` correspondingly dropped from 67s to 32-36s because there was nothing to download. It self-resolves on the second run. **Please don't read the first post-merge run as a regression.**

I have kicked off a re-run against the now-warm caches to measure the steady state and will post those numbers here.

### Also verified from the logs

- The prune step runs and handles a missing directory cleanly.
- The cache-usage check works: `Actions cache usage: 5.28 GB of 10.00 GB (53%) across 5 entries.`
- The test shards save near-empty Turbo caches of their own (`Tests (server)` saved 210 KB, since test tasks have no outputs). That is precisely why the cross-prefix restore matters — their reuse has to come from the `checks` entry, which they previously could not see.

### The purge does not stick until this merges

I deleted 9.05 GB of stale macOS entries earlier, taking the repo from 40% over quota to 744 MB. Because `main` still runs the old workflow, it refilled to **5.67 GB within four minutes**:

```
2.21 GB  22:29:55  macOS-...-package-smoke-turbo-...
2.21 GB  22:31:34  macOS-...-package-smoke-turbo-...
0.21 GB  22:31:07  macOS-...-compat-smoke-turbo-...
0.21 GB  22:31:28  macOS-...-compat-smoke-turbo-...
```

That is the ratchet described in the PR body, observed live. It stops on merge.

### Next bottleneck

With macOS fixed, `Tests (app)` becomes the long pole (282s cold, 3:13 warm at baseline, ~2m40 of which is the vitest suite itself). No caching change touches that; sharding or speeding up that suite is separate follow-up work.

PR #1853, opened 2026-08-18
Speed up large composer paste and first thread open

jjcm:
A 1 MB paste into the composer now paints in ~0.5 s instead of ~2.5 s, and the first thread open of a session stays at ~242 ms instead of regressing to ~469 ms behind a lazy boundary.

| Scenario | Before | After |
| --- | --- | --- |
| 1 MB composer paste, rich text on, until painted | 2453 ms | 530 ms |
| &nbsp;&nbsp;isolated Markdown parse of that paste | 2362 ms | 11.9 ms |
| First thread open in a session (lazy pane vs static) | 469 ms | 242 ms (−48.5%) |

**1 MB paste, before/after** ([mp4](https://github.com/jjcm/bb/blob/cursor/composer-paste-and-thread-open-perf-cf7d/perf-clips/paste-1mb-before-after.mp4)):

1 MB paste before/after

**First thread open, before/after** ([mp4](https://github.com/jjcm/bb/blob/cursor/composer-paste-and-thread-open-perf-cf7d/perf-clips/thread-open-before-after.mp4)) — production build, real timescale slowed 8×, counter freezes at each side's painted time (399 ms lazy vs 126 ms static on the recording machine; medians over 5 runs were 405 ms vs 146 ms):

First thread open before/after

## Changes

- **Markdown parse linearization** (`prompt-editor-serialization.ts`): parsing a large single-line paste was superlinear; it is now linear, taking the isolated parse from 2362 ms to 11.9 ms on a 1 MB minified-JS paste. This is the headline paste win.
- **Typeahead trigger window** (`find-active-trigger.ts`): the `@`/`/` trigger scan looks at the last 256 characters before the caret instead of rescanning the whole text on every keystroke.
- **Structural value compare** (`PromptBoxInternal.tsx`): the controlled-value sync compares text and mention ranges structurally instead of `JSON.stringify`ing the entire prompt twice per keystroke.
- **Draft flush serialization** (`usePromptDraftStorage.ts`): the localStorage draft serializes on the existing 250 ms flush, not on every keystroke.
- **Deferred decoration rebuild** (`prompt-decoration-extension.ts`): above 100k positions the decoration rebuild is deferred. Tradeoff: decorations can lag typing by up to ~200 ms on very large drafts.
- **Thread pane stays a static import** (`SplitThreadArea.tsx`): a `React.lazy` boundary here makes the first thread open 469 ms instead of 242 ms (the Suspense retry mounts the pane at transition priority — prefetching the chunk recovers only ~7 ms). This PR keeps the static import and pins the measurement in a comment. Tradeoff: the thread view (~899 KB raw) ships in the workspace route chunk instead of a click-time round trip.

## Notes

- Rich-text Markdown editing is off by default; the paste headline numbers need it on. With it off, paste was already fast.
- Typing latency after a very large paste is still bad (ProseMirror `getClientRects` on a huge text node) and is not addressed in this PR.
- The thread view's own per-keystroke re-render fix landed upstream independently in #1799; this branch is merged up to current `main`, so that part dropped out of the diff here.
- Fork PR with the same content: https://github.com/jjcm/bb/pull/5


SawyerHood:
Thanks @jjcm — moved into #1857 (an in-repo branch so CI runs, with you as co-author on the ported commit) plus the cleanups from review: the deferred decoration rebuild no longer re-runs draft observers, initial editor content is built once per editor instance, drafts survive a localStorage quota failure, and the perf clips are left out of the tree. Closing this one in favor of #1857.

PR #1904, opened 2026-08-19
Push the composer draft subscription below ThreadDetailPromptArea (D1)

SawyerHood:
## What was wrong

`ThreadDetailPromptArea` (mobile-perf D1) called `useActiveComposerDraft`, a `useSyncExternalStore` subscription to the prompt-draft store. Every keystroke re-ran its ~1600-line body and ~70 hooks (thread creation options, typeahead, queued-message actions, attachment uploads, three queries, five mutations) and rebuilt the composer, attachment and plugin-host objects. Only the controlled composer and the plugin composer host need the live draft.

## What changed

- `usePromptDraftStorage.ts`: `createPromptDraftAccessor` builds the full imperative draft API once per storage key. `getPromptDraftAccessor` returns it; new `usePromptDraftAccessor` memoizes it per key. `usePromptDraftStorage` layers the reactive draft on the same accessor (same return shape). New `usePromptDraftHasSubmittableInput` flips only when `promptDraftToInput(draft).length > 0` changes. New `hasPromptDraftSubmittableInput` in `prompt-draft.ts`.
- `useActiveComposerDraftWriters` split out of `useActiveComposerDraft` so the inline-or-stored writers work without a subscription.
- `ThreadDetailPromptArea` holds the accessor and the submittable bit. Send and modifier-submit read `promptDraft.getCurrent()` at event time. The queued-message inline editor uses the inline edit draft directly.
- New `ThreadDetailFollowUpComposer.tsx`: `ThreadDetailFollowUpComposer` (wraps the memoized `FollowUpPromptBox`) and `ThreadDetailPendingInteractionComposer` own the draft subscription, build the controlled composer config, attachment items and plugin host `draft`, and publish the host (or the inline editor host while one is open) to plugin hooks. `FollowUpPromptBox`'s contract is unchanged.
- New `ThreadDetailPromptStack.tsx`: memoized stack (workflow, background-command, todo cards, context banner, model fallback card, queued list) with its expansion state and the unarchive mutation. Goal and prompt-mode card state stays in the area because both branches show them.
- babel-plugin-react-compiler@1.0.0: `ThreadDetailPromptArea` CompileSuccess (400 slots); all new components/hooks compile.

## How you verified

[...]

PR #1905, opened 2026-08-19
Window the top-level timeline on phones and evict far pages (C1 step 2)

SawyerHood:
## What was wrong

Every loaded timeline page stayed mounted. `TimelineRowsList` rendered one wrapper per top-level row, `prependOlderTimelineRows` concatenated pages without a cap, and auto-load pulled the next page 600px before the top. On a phone, reading back through a long thread left hundreds of markdown bodies and action bars in the DOM. Each style/layout pass (keyboard, orientation, streaming growth) and the memory footprint grew with how far the user had scrolled. Step 1 (`content-visibility: auto`, already on this branch) skips layout and paint of off-screen rows; the DOM itself was still unbounded.

## What changed

- New `timeline-windowing.ts` (pure geometry): entry layout with measured heights or per-kind estimates plus the flex gap, binary search for the entry at an offset, range = viewport +/- overscan (2 viewports, >= 1600px), hysteresis (`rangeCoversViewport` with half-overscan slack), id-based range/anchor resolution that survives prepends, and segment building (mounted entries + spacers sized to the rows they replace).
- New `useTimelineWindow.ts`: samples the bottom-anchored scroll body on rAF-throttled scroll and on scroll-area resize; measures mounted wrappers with one ResizeObserver; keeps a viewport anchor by row id (null = bottom); re-ranges only when the viewport nears an edge; corrects `scrollTop` in the same layout pass by keeping the first visible mounted row in place (idempotent under Chromium scroll anchoring, does the job on WebKit); snaps the `AutoHeightContainer` on a swap; pins the live turn, unread divider, search-target ancestors and the focused row; skips swaps while a text selection is inside the list; seeds the first window from the saved per-thread scroll anchor; renders everything until sampled when windowing switches on later; re-anchors from the DOM when the anchored row is replaced.
- `ThreadTimelineRows.tsx`: the top-level list renders windowed segments (spacers carry `data-timeline-row-spacer`, `overflow-anchor: none`) when compact and inside a scroll body; nested lists, desktop and surfaces without a scroll body render exactly as before. Unread divider gets a measured wrapper. Manual expand/collapse choices are remembered per row for the timeline instance so evicted rows re-mount in the state the user left them.
- `ExpandableTimelineRow`: optional `initialManualExpansionOverride` / `onManualExpansionOverrideChange` seed props (state stays local).
- `height-transition.tsx`: `useAutoHeightSnap()` context so the window can snap the wrapper after a swap.
- `timeline-row-containment.ts`: `estimateSkippedTimelineRowBlockSizePx` shared by containment and windowing so spacer and skipped rows agree.

## How you verified

- New `timeline-windowing.test.ts` (13 tests): layout with gaps and measured heights, offset lookup, range with overscan and clamping, coverage/hysteresis, anchor resolution across a prepend, id round-trips, spacer heights add up to the un-windowed list height.
- New `ThreadTimelineRows.windowing.test.tsx` (14 tests, jsdom with a simulated flex layout, ResizeObserver and rAF): all rows render when not compact or without a scroll body; initial window is the trailing block plus one estimated spacer; scrolling to the top mounts the head, evicts the middle and keeps the last row; no re-render inside the slack; measured heights size the spacer and rows re-mount in place; scrollTop is corrected when re-mounted rows are taller than their spacer share; a prepended page keeps the same rows mounted and grows the spacer by exactly the page's estimate; the anchored row being replaced keeps the window and re-anchors instead of jumping; running turn rows stay mounted far below; the search target's top-level row stays mounted; the saved scroll anchor seeds the first window; windowing that switches on after mount does not evict the current view; the unread divider is wrapped and pinned; a manually expanded row is still expanded after eviction and re-mount.
- `pnpm exec turbo run typecheck --filter=@bb/app`: pass. `pnpm exec turbo run lint --filter=@bb/app`: 0 errors (6 new `react-hooks/refs` warnings in the hook for the deliberate controller reads, same category as existing timeline code). Vitest over `src/components/thread`, `src/components/ui`, `src/components/plugin`, `src/views`: 126 files / 1086 tests pass, including every existing timeline, scroll-body and containment test.

## Stack context

Layer 26 of 28 in the `bb/mobile-perf/*` stack (bottom → top: quick wins first, big rocks last).
- Prerequisite (layer below): `bb/mobile-perf/composer-draft-pushdown` (#1904).
- Next layer: `bb/mobile-perf/persisted-query-cache` (#1906).
- Reviewer notes / follow-ups: Not fixed (timing-dependent, self-healing): when a thread with a saved non-bottom scroll anchor mounts with rows arriving in a later commit than the scroll body (cache miss after g | Pre-existing/inherited from C1 step 1, noted by the implementer: rows re-mounted above the viewport carry the content-visibility estimate until realized, so WebKit (no scroll ancho | Design trade-off, not a bug: on compact viewports AutoHeightContainer is snapped on every window swap, so the timeline height glide is lost on phones (aligned with C2 direction).
- Wire/contract: None. No server/daemon contract, WebSocket message, host RPC, route, CLI flag or config knob changed; HOST_DAEMON_PROTOCOL_VERSION untouched. Only apps/app client code.

### DIFF
--- a/apps/app/src/components/thread/timeline/ExpandableTimelineRow.tsx
+++ b/apps/app/src/components/thread/timeline/ExpandableTimelineRow.tsx
@@ -36,6 +36,14 @@ export interface ExpandableTimelineRowProps {
    * state until the user toggles the row or the row unmounts.
    */
   terminalAutoExpanded?: boolean;
+  /**
+   * Seed for the user's manual expand/collapse choice, and where to report
+   * changes to it. Lets a host that unmounts far-away rows (the windowed
+   * timeline) restore the choice when the row mounts again; the state itself
+   * stays local to the row.
+   */
+  initialManualExpansionOverride?: boolean | null;
+  onManualExpansionOverrideChange?: (override: boolean) => void;
   onBeforeExpand?: () => void;
   renderBody: () => ReactNode;
   title: TimelineTitle;
@@ -86,8 +94,10 @@ function ExpandableTimelineRowComponent({
   expandable = true,
   forceExpanded = false,
   horizontalPadding = "default",
+  initialManualExpansionOverride = null,
   leadingIcon,
   onBeforeExpand,
+  onManualExpansionOverrideChange,
   onTitleAction,
   renderBody,
   resolveSegmentLinkHref,
@@ -97,7 +107,7 @@ function ExpandableTimelineRowComponent({
   titleContent,
 }: ExpandableTimelineRowProps) {
   const [manualExpansionOverride, setManualExpansionOverride] =
-    useState<ManualExpansionOverride>(null);
+    useState<ManualExpansionOverride>(initialManualExpansionOverride);
   const [terminalAutoExpandedLatch, setTerminalAutoExpandedLatch] =
     useState(terminalAutoExpanded);
   const [collapsedPreviewActive, setCollapsedPreviewActive] = useState(false);
@@ -123,7 +133,8 @@ function ExpandableTimelineRowComponent({
       onBeforeExpand?.();
     }
     setManualExpansionOverride(!isExpanded);
-  }, [isExpanded, onBeforeExpand]);
+    onManualExpansionOverrideChange?.(!isExpanded);
+  }, [isExpanded, onBeforeExpand, onManualExpansionOverrideChange]);
   const handleCollapsedPreviewClick = useCallback(
     (event: CollapsedPreviewClickEvent): void => {
       if (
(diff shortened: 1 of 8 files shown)


SawyerHood:
Closing: the maintainer decided to land only layers 1–22 of the mobile-perf stack (#1880–#1901). The branch stays on the remote for reference; it can be reopened and rebased later if the remaining big rocks (sidebar-bootstrap patching, thread bootstrap include=, composer draft push-down, timeline windowing, persisted cache, service worker) are picked up again.

PR #1906, opened 2026-08-19
Persist an allowlisted query cache slice to IndexedDB behind an experiment (J2)

SawyerHood:
## What was wrong

Mobile perf sweep J2 (`apps/app/src/lib/query-client.ts`): the TanStack query client starts empty on every cold launch. On a phone through bb connect the first paint of the sidebar, palette and the current thread waits for four network round-trips (sidebar bootstrap, /system/config, thread bootstrap, timeline) over a 100-300 ms tunnel, after the JS has already downloaded and parsed.

## What changed

- New `persistedQueryCache` experiment (default off) in `@bb/domain`, exposed in Settings -> Experiments ("Persisted app cache"), `bb settings experiment persistedQueryCache <true|false>`, docs/configuration.md, the CLI guide template and the bb-cli skill.
- `apps/app/src/lib/persisted-query-cache/persisted-query-cache.ts`: closed allowlist (`classifyPersistedQueryKey`: sidebarNavigation, systemConfig, threadDetailBootstrap/<id>, threadTimeline/<id>, exact key shapes only), zod-validated envelope, `selectPersistedQueryCacheEntries` (dedupe by key, 24 h age limit, 5 most recently updated threads, 4 MB byte budget with config/sidebar/bootstrap prioritized over timelines), `restorePersistedQueryCache` (hydrate with original `dataUpdatedAt`, 400 ms timeout, invalid blob is cleared), `restorePersistedQueryCacheIfEnabled` (never touches storage when the local flag is off), and `startPersistedQueryCachePersister` (debounced 1.5 s, flush on pagehide/hidden, read-modify-write so entries GC'd from memory survive, quota error clears the store and disables for the session).
- `persisted-query-cache-store.ts`: IndexedDB backend (single string blob, connection opened per call) plus a memory backend for tests. `persisted-query-cache-flag.ts`: localStorage mirror of the experiment.
- `main.tsx` awaits the gated restore before `createRoot().render()`. `App.tsx` mounts `usePersistedQueryCacheSync`, which mirrors the server value locally, runs the persister while on, and clears the store when turned off.
- Staleness reconciliation reuses existing mechanisms: hydrated entries keep their original `dataUpdatedAt`, so `staleTime`-bounded queries (systemConfig, timeline) refetch on mount, the initial WebSocket connect invalidates realtime-owned queries with `dataUpdatedAt < connectedAt` (sidebar, thread bootstrap), and a hydrated timeline window lets `fetchThreadTimeline` request a delta.

Boot payload: 1655.6 KB raw / 448.8 KB br before, 1663.3 KB / 451.4 KB after (budget 1671 / 456.4).

## How you verified

- `apps/app/src/lib/persisted-query-cache/persisted-query-cache.test.ts` (16 tests): allowlist accepts exactly the four shapes and rejects near misses (extra key parts, empty/non-string thread ids, threads, plugin-settings, host file previews); selection keeps config + sidebar + N most recent threads (timeline recency counts), dedupes by key keeping newest, drops expired, skips entries over the byte budget while keeping earlier ones; serialize/parse round-trip drops disallowed and expired entries, rejects malformed blobs and other format versions; restore hydrates with original dataUpdatedAt/success/idle, does not overwrite fresher live data, gives up after the timeout and ignores a late read, clears an invalid blob; gated restore never touches storage when the flag is off and swallows store failures; persister writes only allowlisted successful queries and merges the previous snapshot, debounces into one write and stops after stop(), clears the store and disables itself on QuotaExceededError.
- `SettingsView.experiments.test.tsx` gains a toggle test for the new switch.
- `pnpm exec turbo run typecheck` for @bb/app, @bb/server, @bb/domain, @bb/db, @bb/desktop, @bb/cli, @bb/templates, @get-bb/plugin-sdk: pass. `pnpm exec turbo run lint --filter=@bb/app --filter=@bb/server --filter=@bb/domain`: 0 errors (pre-existing warnings only).
- Tests: @bb/domain, @bb/db, @bb/templates full suites; @bb/server test/system/experiments.test.ts + test/threads/thread-runtime-config.test.ts; @bb/app persisted-query-cache, SettingsView.experiments, AppLayout.plugin-panel-header, AppLayout.root-compose-project, system-cache-effects, system-config-atoms; @bb/desktop preload-build; @bb/cli settings command output. All pass.
- `pnpm exec turbo run build --filter=@bb/app` + `check-bundle-budget.mjs`: OK.

Fixes mobile-perf J2.

## Stack context

Layer 27 of 28 in the `bb/mobile-perf/*` stack (bottom → top: quick wins first, big rocks last).
- Prerequisite (layer below): `bb/mobile-perf/timeline-windowing` (#1905).
- Next layer: `bb/mobile-perf/service-worker-shell` (#1907).
- Reviewer notes / follow-ups: Persisted payloads are still opaque (z.unknown) rather than parsed with the server-contract zod schemas; the new crash-time store clear bounds the blast radius (a bad blob cannot c | The pre-render IndexedDB read (up to 400 ms) and the hydrate/ingest cost have not been measured on a real phone or iOS Simulator; the report also says not device-verified. | systemConfig is not in the WS-connect invalidation list, so a relaunch within 60 s (its staleTime) can show config changed elsewhere in that window; same window as the in-memory ca | docs say the localStorage mirror key `bb.persistedQueryCache` and 4 MB / 24 h / 5-thread caps; those are constants in code with no test pinning the documented numbers (low value, l
- Wire/contract: None on the server<->daemon wire. `Experiments` gains the `persistedQueryCache` key (server-side setting exposed in /system/config and PUT /settings/experiments; z.record over the enum makes it required in the PUT body, same as every other key). Only individual experiment booleans (providerSessionReaping) cross the daemon protocol, so HOST_DAEMON_PROTOCOL_VERSION was not bumped. Regenerated packag


SawyerHood:
Closing: the maintainer decided to land only layers 1–22 of the mobile-perf stack (#1880–#1901). The branch stays on the remote for reference; it can be reopened and rebased later if the remaining big rocks (sidebar-bootstrap patching, thread bootstrap include=, composer draft push-down, timeline windowing, persisted cache, service worker) are picked up again.

PR #1909, opened 2026-08-19
Connect: parallel gate lookups, zombie-socket freshness, gate fonts, per-message compression (B29-B32)

SawyerHood:
## What was wrong

- The connect gate awaited the label lookup and then the session lookup, two single-region D1 round trips in series on every uncached visitor request. Its per-isolate caches stored only settled values, so a page load's burst of asset requests all missed together and each ran its own query.
- TunnelDO treated any tunnel socket with `readyState === OPEN` as live. A client whose network died without a close frame left a zombie that still accepted `send()`. Every visitor request was proxied into it and hung 30 s before a 504, for as long as the zombie lingered (up to ~80 s).
- The 401/503 gate pages loaded a Google Fonts stylesheet, a render-blocking cross-origin fetch before first paint on a phone.
- The tunnel client sent every relayed frame through permessage-deflate, including body chunks that the origin had already brotli/gzip encoded.

## What changed

- `apps/connect/src/worker.ts`: visitor cookie verification starts before `await resolveLabel` and overlaps it (skipped on tunnel dials). `apps/connect/src/session.ts`: label and session caches also share the in-flight promise; a rejected lookup is dropped so the next request retries.
- `apps/connect/src/tunnel-do.ts`: liveness now requires `max(acceptedAt, getWebSocketAutoResponseTimestamp) <= TUNNEL_STALE_MS (50 s)`. `acceptedAt` is stored in the socket attachment (survives hibernation). Stale sockets are closed; if no live socket remains, in-flight streams fail at once and the presence alarm stops. Sockets accepted before this change are stamped on first sight and get one grace window. `packages/tunnel-client`: heartbeat deadline 60 s -> 45 s (detects a dead link at the third 20 s tick, 60 s, instead of 80 s).
- Gate pages use the system font stack; no third-party stylesheet.
- `TunnelSession.send` takes a `compress` option; body chunks for responses with a non-identity `Content-Encoding` are sent with `compress: false`. Identity bodies and control frames keep deflate (deviation from "all body-chunk frames": small uncompressed responses still benefit). `apps/host-daemon/src/ws.d.ts` gains the `send(data, options)` overload.

## How you verified

- New tests (fail before, pass after): `apps/connect/src/worker.test.ts` (gate lookup overlap: session verified while label pending, no session check on tunnel dial, no unhandled rejection on 404; gate page has no font link; TunnelDO zombie sockets: 503 + close for stale, live with recent heartbeat, live just after accept, legacy grace window, in-flight streams fail immediately, route around zombie to fresh replacement without abandoning streams, alarm drops presence). `apps/connect/src/session.test.ts` (5 concurrent resolves -> 1 query; fresh not shared; rejected lookup retried; 5 concurrent session verifications -> 1 query; real in-memory SQLite). `packages/tunnel-client/test/session-compress.test.ts` (precompressed body chunks `compress:false`, identity `compress:true`, control frames compress; `isPrecompressedResponse` parsing).
- `pnpm exec turbo run typecheck --filter=@bb/connect --filter=@bb/tunnel-client --filter=bb-plugin-connect --filter=@bb/host-daemon`: 4/4 successful.
- `pnpm exec turbo run test --filter=@bb/connect --filter=@bb/tunnel-client --filter=bb-plugin-connect`: 112 / 15 / 78 passed. `pnpm exec turbo run test --filter=@bb/host-daemon -- src/connect-tunnel`: 8 passed.
- `pnpm exec eslint` and `prettier --check` on changed files: clean.

## Stack context

Standalone PR (no overlap with the `bb/mobile-perf/*` stack); targets `main`.
- Deliberately not done here: B29 optional 'D1 read replication' - infra/wrangler configuration, not a code change; left as a follow-up.
- Reviewer notes / follow-ups: B30 relies on the runtime preserving getWebSocketAutoResponseTimestamp across hibernation (documented Cloudflare behaviour, but unverifiable in unit tests). Failure mode is a visib | The relay window (TUNNEL_STALE_MS=50s in apps/connect) and client heartbeat interval/deadline (20s/45s in @bb/tunnel-client) are coupled only by comments; no cross-package test gua
- Wire/contract: No server<->host-daemon wire change (HOST_DAEMON_PROTOCOL_VERSION untouched). Tunnel wire (relay<->connect client) frame format unchanged; only ws-level per-message compression flag and heartbeat timings changed. TunnelDO tunnel-socket attachment gains `{acceptedAt}` (previously none); sockets without it are handled (stamped on first sight). No CLI/config/plugin-API surface changes.

### DIFF
--- a/apps/connect/src/session.test.ts
+++ b/apps/connect/src/session.test.ts
@@ -4,13 +4,14 @@ import { fileURLToPath } from "node:url";
 import Database from "better-sqlite3";
 import { eq } from "drizzle-orm";
 import { drizzle } from "drizzle-orm/better-sqlite3";
-import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import {
   labelClaim,
   machine,
   profile,
   schema,
   server,
+  session,
   user,
 } from "@bb/connect-db";
 
@@ -19,6 +20,7 @@ import {
   markMachineSeen,
   resolveLabel,
   verifyMachineCredentialDetails,
+  verifySessionCookie,
 } from "./session.js";
 import { assignMachineLabel } from "./machine-label.js";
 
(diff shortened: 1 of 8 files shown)


SawyerHood:
Closing this draft as stale: 253 commits behind main with conflicts. The branch stays available if any of it is picked up again.

> AGENT GENERATED

PR #1964, opened 2026-08-19
Remember new-thread provider and model per machine

MateoCerquetella:
## Summary

- Remember the new-thread provider for each directly selected machine.
- Remember model and reasoning for each machine and provider.
- Use the existing browser-wide preference as the fallback for machines without a saved selection.

This lets a local machine default to Codex while an SSH or VPS machine defaults to Cursor. Reused-environment routing keeps its current behavior and is intentionally outside this compact change.

## Scope

Three files: two production files and one focused regression test. No picker, fork, draft, server, database, daemon protocol, or public API changes.

## Testing

- Focused selection tests: 27 passed
- App typecheck
- App lint: 0 errors, existing warning baseline only


SawyerHood:

Issue #1159, opened 2026-08-07
Select multiple threads and archive them together


arunsathiya:
Archive works on one thread at a time. Only a parent cascade covers more (`POST /threads/:id/archive-all`).

Add multi-select in the sidebar:

1. Shift-click selects a range.
2. Cmd-click adds one thread.
3. The context menu archives the selection.

Related: #1118 (Ctrl+Z to undo archives).

SawyerHood:
Love this! Was thinking of making archiving just single click anyway which might reduce the need for this

SawyerHood:
Closing in favor of #1958 (one-click Archive button on hover for each sidebar thread). As the comment above says, single-click archive reduces the need for multi-select. If one-click is still not enough after #1958 ships, reopen this issue with the concrete workflow.

> AGENT GENERATED: by Claude Opus 5

PR #1965, opened 2026-08-19
Add a one-click Archive button to hovered sidebar thread rows

### LINKED ISSUE #1958: One-click Archive button on hover for each thread in the sidebar
### The workflow

I finish a thread and I want to archive it from the sidebar with one click. I do this many times a day while I clean the thread list.

### What happens today

The sidebar thread row has no archive control. I must open the thread context menu, or open the thread, to archive it. This is two or three actions for each thread.

### What you would expect

When I hover a thread row in the sidebar, an Archive icon button appears at the right edge of the row. One click archives the thread. The button does not change the row height. Cursor GUI has this interaction; it is the reference.

### Context and alternatives

- Workaround today: the thread context menu and the `bb thread archive` CLI.
- The button should follow the same behavior as the existing archive action (same permission and same undo path if one exists).
- Source: user feedback collected on 2026-08-19.

### Checks

- [x] If an agent wrote this, the body ends with `> AGENT GENERATED: by <model>`.

SawyerHood:
## What was wrong

Archiving a thread from the sidebar took two or three actions: open the `…` menu or the right-click menu, then pick Archive. Users who clean up many threads a day asked for a one-click hover button like Cursor has (#1958, which supersedes #1159; the keyboard shortcut from #1172 is opt-in and not discoverable from the list).

## What changed

- `apps/app/src/components/thread/ThreadActionsMenu.tsx`: new `ThreadArchiveQuickAction`, an icon button with a tooltip that calls the same `archiveThreadAndChildren` / `unarchiveThread` actions as the menu entry. It uses the existing `Archive` / `ArchiveRestore` icons and stops propagation so the row link does not navigate.
- `apps/app/src/components/sidebar/ThreadRow.tsx`: the hover overlay is anchored `inset-y-0 right-0` so it can hold two actions; the archive button sits left of the `MoreHorizontal` trigger with a 2px gap.
- `apps/app/src/components/ui/theme.css` + `sidebar-hover-actions.ts`: new `bb-sidebar-hover-actions-inset` class reserves one action width on the title under the same row states that reveal the overlay (hover, focus-visible, open menu), so long titles truncate instead of running under the button. Reset to 0 on coarse phones where the overlay is hidden. Not applied when the row shows a shortcut pill instead of hover actions.

No wire, CLI, or doc changes; no new icon registered.

## How you verified

- New `ThreadArchiveQuickAction.test.tsx`: one click archives without bubbling to the row; an archived thread unarchives instead.
- Updated `ThreadRow.test.tsx` mock and wrapped `ProjectRow.interactions.test.tsx` renders in `TooltipProvider`.
- `pnpm exec turbo run typecheck lint test --filter=@bb/app`: 404 files, 3100 tests pass.
- Manual: dev app in headless Chromium — hover shows the button left of `…`, click removed the row and set `archived_at` in the dev DB; long title truncates before the button.

Fixes #1958

### DIFF
--- a/apps/app/src/components/sidebar/ProjectRow.interactions.test.tsx
+++ b/apps/app/src/components/sidebar/ProjectRow.interactions.test.tsx
@@ -11,6 +11,7 @@ import type { ThreadListEntry } from "@bb/domain";
 import type { ProjectResponse } from "@bb/server-contract";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import { MemoryRouter } from "react-router-dom";
+import { TooltipProvider } from "@bb/shared-ui/tooltip";
 import { Provider, createStore } from "jotai";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import {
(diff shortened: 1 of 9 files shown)


SawyerHood:
Addressed both SlopCop findings in 569c22f854e72b35de93a6352c8c3ba2b5a632c8:

1. **Rapid archives lose first callbacks** — `archiveThreadAndChildren` now uses `mutateAsync` and handles the per-call promise, so each archive keeps its own Undo toast, pane closure, navigation, and error toast even when a second archive starts before the first settles.
2. **Overlapping click targets** — added `SIDEBAR_PAIRED_ACTION_LEADING/TRAILING_TARGET_CLASS` that re-anchor each button's pseudo-target to stop 1px short of the shared edge. Verified in headless Chromium with `elementFromPoint` across the pair: Archive owns x 261–285, the menu owns 286–310, with no shared pixel. The repo has no browser pointer test harness in CI, so I did not add one.

PR #1979, opened 2026-08-19
Render PDF file previews instead of "Preview not available"

### LINKED ISSUE #1985: PDF file previews show "Preview not available for application/pdf"
## Summary

Opening any PDF in the file preview panel shows an error message instead of
the document. The content routes already serve the bytes with the correct
mime type; the app has no branch for `application/pdf`, so every PDF falls
through to the unsupported state. Images, video, CSV, and markdown all
render.

## Versions and environment

- bb 0.39.0, desktop app, macOS 26
- Code path confirmed unchanged on `main` at `b61ec88`

To be precise about the repro: I observed the message in the packaged 0.39.0
desktop app, and confirmed by reading `main` that the same code path still
produces it. I did not click through a from-source build to see the message
a second time.

## Steps to reproduce

1. Open a thread.
2. Open any PDF in the preview panel, either from a file link in the
   timeline or from a file tab.

## Expected vs actual

Expected: the document renders, the way images, video, CSV, and markdown do.

Actual, verbatim:

```
Preview not available for application/pdf.
```

The path header and open-in-editor action render normally. Only the body
shows the error.

## Evidence

Image

`buildFilePreview` tests the mime type in this order: image types, known
text types, a UTF-8 fallback, then video types. `application/pdf` matches
none of them, so the function returns kind `unsupported`:

https://github.com/get-bb/bb/blob/b61ec88de2f2634d23b1448c24719cefa8d0f18d/apps/app/src/lib/file-preview.ts#L236-L286

`SecondaryPanelFilePreview` then renders the message above:

https://github.com/get-bb/bb/blob/b61ec88de2f2634d23b1448c24719cefa8d0f18d/apps/app/src/components/secondary-panel/ThreadStorageFilePreview.tsx#L256-L270

The server side already works. Against a dev server built from this
checkout, a GET to `/api/v1/threads/:id/host-files/content` for a PDF
returns `200`, `content-type: application/pdf`, and the raw bytes.

## What you ruled out

- Not a server or route problem. The content route returns the correct mime
  type and bytes; see above.
- Not a duplicate. No open or closed issue mentions PDF previews. #1670 is
  the same class of gap (HEIC does not render) but a different format, and
  HEIC needs decoding rather than a new branch.
- Not configuration. No plugin provides PDF rendering, and nothing in
  settings changes this path.

Small. Chromium renders a PDF with its built-in viewer when a frame loads an
`application/pdf` response, so the fix is one new preview kind plus an
iframe, with no PDF library and no new dependency. Same shape and files as
#520, which added CSV previews.

PR open at #1979.

## Checks

> AGENT GENERATED: by Claude Opus 5 and Claude Fable 5

jsilets:
## What was wrong

`buildFilePreview` tests the mime type in this order: image types, known
text types, a UTF-8 fallback, then video types. `application/pdf` matches
none of them, so every PDF returns kind `unsupported` and the panel shows
`Preview not available for application/pdf.` The content routes already
serve PDFs as `application/pdf` with the raw bytes; the gap was frontend
only.

## What changed

A `pdf` preview kind, rendered by framing the preview URL. Chromium hands an
`application/pdf` response to its built-in viewer, so this needs no PDF
library and no new dependency.

- `apps/app/src/lib/file-preview.ts`: a `pdf` kind for `application/pdf`,
  tested before the UTF-8 fallback. A PDF with no compressed streams is all
  printable ASCII, and the fallback would render its markers as source text.
- `apps/app/src/components/secondary-panel/FilePreview.tsx`: a `pdf` state
  and a `FilePreviewPdf` iframe. The frame sets no `sandbox`: a sandboxed
  frame blocks the viewer's own scripts and the load fails with
  `ERR_BLOCKED_BY_CLIENT` (verified in Electron 41 with this app's exact
  `webPreferences`). This is not the exposure the HTML preview sandbox
  guards against; Chromium gives the bytes to the viewer instead of parsing
  them as a document, so the file never executes in the app origin.
- `apps/app/src/components/secondary-panel/ThreadStorageFilePreview.tsx`:
  routes the new kind, same shape as the image and video branches.
- `apps/app/src/hooks/queries/environment-queries.ts`: PDFs join images and
  video in the `data:` URL rewrite, since the workspace surface reads
  previews from a JSON route rather than a byte stream.

All four preview surfaces render PDFs: host files, project files, thread
storage, and workspace files. No wire changes
(`HOST_DAEMON_PROTOCOL_VERSION` unchanged), no CLI or doc surfaces.

## How you verified

Tests (fail before, pass after):

- `file-preview.test.ts`: builds a `pdf` preview for `application/pdf`.
- `file-preview.test.ts`: keeps an all-ASCII PDF a PDF instead of falling
  back to text.
- `FilePreview.test.tsx`: frames the preview with no `sandbox` attribute.

`turbo typecheck` and `turbo lint --filter=@bb/app` clean (lint warnings
unchanged from base). Full `--filter=@bb/app` suite: 405 files, 3107 passed.

Manually rendered the new Ladle story in Electron 41: the viewer loads with
its toolbar, page rail, zoom, download, and print controls inside the normal
preview header.

image

Fixes #1985

> AGENT GENERATED: by Claude Opus 5 and Claude Fable 5

### DIFF
--- a/apps/app/src/components/plugin/file-opener-tabs.test.ts
+++ b/apps/app/src/components/plugin/file-opener-tabs.test.ts
@@ -2,7 +2,10 @@ import { describe, expect, it } from "vitest";
 import { threadTabsSchema } from "@bb/server-contract";
 import type { PluginFileOpenerSlot } from "@/lib/plugin-slots";
 import type { OpenSecondaryPanelTabRequest } from "@/components/secondary-panel/useThreadFileTabs";
-import { createFileOpenerTabForRequest } from "./file-opener-tabs";
+import {
+  createFileOpenerTabForRequest,
+  parseFileOpenerParams,
+} from "./file-opener-tabs";
 
 const MARKDOWN_OPENER = {
   component: () => null,
@@ -91,6 +94,34 @@ describe("createFileOpenerTabForRequest thread-tabs contract", () => {
       kind: "workspace-file-preview",
       projectId: null,
       threadId: null,
+    });
+    expect(() => threadTabsSchema.parse([tab])).not.toThrow();
+  });
+
+  it("preserves the selected host for a project-backed opener", () => {
+    const tab = createFileOpenerTabForRequest({
+      fileOpeners: [MARKDOWN_OPENER],
+      preference: {},
+      projectHostId: "host_remote",
+      projectId: "proj_1",
+      request: {
+        kind: "workspace-file-preview",
+        tab: {
+          lineRange: null,
+          path: "docs/readme.md",
+          source: { kind: "working-tree" },
+          statusLabel: null,
+        },
+      },
+      resolvedEnvironmentId: null,
+      threadId: null,
+    });
+
+    const params = parseFileOpenerParams(tab?.paramsJson ?? null);
+    expect(params?.source).toMatchObject({
+      kind: "workspace",
+      projectId: "proj_1",
+      experimental_hostId: "host_remote",
     });
     expect(() => threadTabsSchema.parse([tab])).not.toThrow();
   });
--- a/apps/app/src/components/plugin/file-opener-tabs.ts
+++ b/apps/app/src/components/plugin/file-opener-tabs.ts
@@ -65,12 +65,14 @@ export function parseFileOpenerParams(
   const { path, source } = parsed as { path?: unknown; source?: unknown };
   if (typeof path !== "string" || path.length === 0) return null;
   if (typeof source !== "object" || source === null) return null;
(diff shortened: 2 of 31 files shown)


jsilets:
Rebased onto main and addressed the review feedback in https://github.com/get-bb/bb/issues/1985#issuecomment-5357425988: the classifier moved to `packages/client-core`, `apps/mobile` maps the new `pdf` kind to `unsupported`, and the `data:` URL branch of `buildEnvironmentFilePreview` now has a test. Full-repo typecheck passes. Details in the issue thread.

PR #2043, opened 2026-08-20
Refuse queued messages for threads whose environment is gone

### LINKED ISSUE #1789: A thread whose environment no longer exists still reports idle, and the message queue accepts sends into it that can never be delivered
A thread whose backing worktree has been removed continues to report `status: idle` with its `environmentId` still attached. The environment is gone from disk; nothing on the thread record says so.

Queueing a message to that thread succeeds. The CLI returns a queued-message id and the sender has every indication of success. The message will never be delivered, because there is no environment in which the thread can run. Messages accumulate indefinitely — we found four — and nothing surfaces the condition to the sender, to the recipient, or to anything reading the thread list.

Attempting to interact with the thread's environment does surface a 409, but only on that path, and only if you already suspect something is wrong. The thread record itself reports health.

## Why this is worse than a lost message

A send that fails tells you to retry or escalate. A send that succeeds and is undeliverable tells you the recipient has your instruction and is working on it. In our case an agent thread had been given work orders it could never receive, while the record showed an idle, healthy worker that had simply not replied yet. The natural reading of that state — worker is fine, worker is thinking — is exactly wrong, and the longer it holds the more confident it looks.

The sender cannot distinguish this from a slow recipient on any current surface. That is what makes it indefinite rather than merely wrong.

The general form: delivery confirmation without delivery is worse than refusal, because it converts a detectable failure into an indefinite silent one.

## What would resolve it

In rough order of how much it helps:

1. Reflect missing-environment state on the thread record, so `idle` cannot be reported by a thread that cannot run.
2. Reject queue sends to a thread whose environment is unavailable, rather than accepting them.
3. Surface a queue depth that is not draining.

Related: execution-profile observability, #1787.

SawyerHood:
## What was wrong

`POST /threads/:id/queued-messages` (`createQueuedMessageForThread` in `apps/server/src/routes/threads/actions.ts`) only checked archived/stopping/deleted. It never applied the gone-environment rule (`goneThreadEnvironmentDetails`) that the direct send path applies through `requireThreadCommandEnvironment`. A thread whose managed worktree was destroyed (archive → grace window → destroy → unarchive) kept `status: idle`, answered `201` with a queued-message id, and the message could never drain: the auto-send hit the same `409 thread_environment_unavailable` internally and the 10 s sweep retried forever. The CLI also labelled the destroyed worktree "Provisioning".

Issue: #1789. Report: https://get-bb.github.io/reports/issues/1789.html

## What changed

- `apps/server/src/routes/threads/actions.ts`: `createQueuedMessageForThread` admits the message inside the same `BEGIN IMMEDIATE` transaction as the insert, on the freshly loaded thread row (`admitQueuedMessage`): writable check, environment check, and one provider-thread-id read. A `destroying`/`destroyed` environment returns the same `409 thread_environment_unavailable` as the send path. A thread with `environmentId === null` that already has a provider thread id (the row was pruned after destroy) returns `409` with reason `never_attached`, again the same as send. A thread that has not run yet and has no environment still accepts queued messages, because that is how messages wait for provisioning.
- `packages/db/src/data/queued-thread-messages.ts`: `createQueuedThreadMessageInTransaction` for caller-owned transactions; `listIdleThreadsWithQueuedMessages` joins `environments` and skips `destroying`/`destroyed`, so queued rows that survived archive → destroy → unarchive no longer fail the sweep every 10 s.
- `packages/core-ui/src/environment-display.ts`: `formatEnvironmentDisplay` labels a `destroying` environment "Destroying" and a `destroyed` one "Destroyed" instead of "Provisioning". This changes `bb thread show` and app metadata labels.
- No wire shape changed, so `HOST_DAEMON_PROTOCOL_VERSION` is unchanged. No new routes or CLI flags.

Not done here: the report also suggests a `runtime.environmentStatus`/`canRun` field on the thread response. That is an API contract addition and is left as a follow-up.

## How you verified

- `apps/server/test/public/public-thread-queue-gone-environment.test.ts` (from the report, extended): 3 of 5 tests fail before the route fix (`expected 201 to be 409`); the sweep test (real `archiveThread` → `retire.requested` → `destroy.started` → `destroy.completed` → `POST /unarchive`) fails before the query change; all 5 pass after. One test is a control that a never-run, environment-less thread still gets `201`.
- `packages/core-ui/test/environment-display.test.ts`: new case for `destroying`/`destroyed` labels.
- `pnpm exec turbo run test typecheck --filter=@bb/db --filter=@bb/server --filter=@bb/core-ui --filter=@bb/cli`: typecheck clean; db 406/406; core-ui 17/17; cli 452/452; server 1798/1799. The one failure is `test/internal/internal-skill-trees.test.ts` (file mode `420` vs `436`), a umask difference on this machine, unrelated to this change.

Fixes #1789

### DIFF
--- a/apps/server/src/routes/threads/actions.ts
+++ b/apps/server/src/routes/threads/actions.ts
@@ -1,8 +1,9 @@
 import {
-  createQueuedThreadMessage,
+  createQueuedThreadMessageInTransaction,
   deleteQueuedThreadMessage,
   getEnvironment,
   getQueuedThreadMessage,
+  getThread,
   listActiveVisiblePinnedThreadRootsWithPendingInteractionState,
   pinThread,
   reorderPinnedThread,
@@ -15,6 +16,7 @@ import {
   type ReorderPinnedThreadResult,
   type ReorderQueuedThreadMessageResult,
   type SetQueuedThreadMessageGroupBoundaryResult,
+  type DbQueryConnection,
 } from "@bb/db";
 import {
   publicApiRoutes,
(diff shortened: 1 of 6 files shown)


SawyerHood:
Addressed the SlopCop findings in c4c6df9:

1. **Existing queue rows still loop in the sweep** — fixed. `listIdleThreadsWithQueuedMessages` now joins `environments` and skips `destroying`/`destroyed` rows, so a row that survived archive → destroy → unarchive is no longer a sweep candidate. Regression test: `drops a thread from the auto-send sweep once its environment is gone` drives the real lifecycle (`archiveThread`, `retire.requested`, `destroy.started`, `destroy.completed`, `POST /unarchive`) and asserts the thread leaves the candidate list while the row stays visible. It fails without the query change.
2. **Check before two awaits** — fixed structurally. The environment check no longer runs before the awaits at all. `admitQueuedMessage` (writable + environment + provider-thread-id, one event read) runs on the freshly loaded thread row inside the same `BEGIN IMMEDIATE` transaction as the insert (`createQueuedThreadMessageInTransaction`). Any state the checks see is the state the insert commits against. There is no deterministic injection point in the route to interleave an archive between the awaits, so the admission path is covered by the 409 tests, which now go through it.
3. **`destroying` displayed as `Destroyed`** — fixed: `Destroying` vs `Destroyed`, with a test.
4. **Custom name hides `Destroyed`** — not changed, on purpose. `formatEnvironmentDisplay` has always let a custom name win over every generated lifecycle label, including `Provisioning`; that is its documented contract and the app shows the gone state with a separate banner. Changing name priority is a product decision outside this fix.
5. **Double provider-history read** — fixed; the id is read once in the transaction and reused for the auto-send decision.

PR #2067, opened 2026-08-20
fix(server): retain one timeline revision per request shape

### LINKED ISSUE #2066: Timeline cache retains obsolete revisions for the same request shape
### Summary

The thread timeline response cache keys entries by both request parameters and the thread's current max sequence. Every appended event therefore creates a new key while the obsolete revision remains resident until global LRU eviction, so active threads can retain many superseded response payloads instead of one current revision per request shape.

### Versions and environment

- bb 0.39.0 desktop app
- macOS on Apple Silicon, 8 GB RAM
- Node v22.19.0 for the source-level reproduction
- Source behavior inspected on main; the affected cache implementation was unchanged at `6be4505`

### Steps to reproduce

1. Create a `ThreadTimelineCache` with a capacity greater than one.
2. Call `getOrBuild` for a cacheable response using one request shape and revision/max-sequence A.
3. Append an event, then call it again with the same request shape and revision/max-sequence B.
4. Inspect the cache entry count.
5. A regression assertion expecting one retained entry fails before the fix with `2 !== 1`.

A concrete regression test and fix are available in commit https://github.com/Yazington/bb/commit/206096e4bdcb7e4a69ccf6161d27369afc20ffa0.

### Expected vs actual

```
Actual:   cache size is 2 after two revisions of the same request shape
Expected: cache size is 1, containing only the newest revision
```

The same request shape should replace its obsolete revision immediately. Distinct request shapes should continue to coexist and follow the normal LRU limit.

### Evidence

- A live active thread had 5,749 events and about 11.3 MiB of event data.
- Server logs contained 54 slow timeline builds, with windows around 1,400 events, response payloads around 1.1–4.8 MiB, and build times up to 1.34 seconds.
- The desktop app used roughly 1.5–1.7 GB total memory during the incident; the renderer was about 670 MiB and the server about 300–650 MiB.
- The cache allowed 128 entries. Because its key included `maxSeq`, each append made the prior response unreachable by lookup but still strongly retained by the map.
- The active Codex thread remained below the cache's projected-row cutoff, so those revision payloads were eligible for caching.

### What you ruled out

- Not a duplicate of #1749: that issue focuses on timeline event-budget calibration; this issue is specifically obsolete revision retention.
- The direct cache regression reproduces without Playwright or browser automation.
- The fix does not change response data, request parameters, or cacheability thresholds.

Medium — affects long-running active threads and can amplify server heap use; the fix is localized and low effort.

### Checks

> AGENT GENERATED: by OpenAI Codex (GPT-5)

Yazington:
## What was wrong

The timeline response cache included the thread max sequence in its map key. Each appended event therefore made the prior entry unreachable by future lookups while still retaining its full payload until global LRU eviction. Active threads could accumulate many obsolete revisions for the same request shape, contributing avoidable server heap pressure. See [#2066](https://github.com/get-bb/bb/issues/2066).

## What changed

- Key the cache map by normalized request shape and store the current revision alongside the value.
- Return a hit only when the stored revision matches the requested revision.
- Replace the previous revision immediately when a new cacheable response is built.
- Remove the previous revision when the replacement exceeds the cache size limit, preventing stale data from remaining strongly retained.
- Preserve independent entries for distinct request shapes and the existing global LRU bound.
- Add regression coverage for revision replacement, independent shapes, oversized replacements, repeat hits, and LRU eviction.

## How you verified

- Focused `timeline-cache.test.ts` Vitest suite: 7/7 passed; the revision-retention assertion failed before the implementation change.
- `@bb/server` Turbo typecheck passed.
- ESLint and Prettier checks passed for all three changed files.
- `git diff --check origin/main...HEAD` passed after rebasing onto current `main`.
- No Playwright or browser automation was used.

Fixes #2066

> AGENT GENERATED: by OpenAI Codex (GPT-5)

### DIFF
--- a/apps/server/src/routes/threads/data.ts
+++ b/apps/server/src/routes/threads/data.ts
@@ -358,8 +358,12 @@ export function registerThreadDataRoutes(app: Hono, deps: AppDeps): void {
       summaryOnly,
       includeProviderUnhandledOperations,
     };
+    const paramsKey = buildThreadTimelineParamsKey(keyArgs);
     const full = timelineCache.getOrBuild(
-      buildThreadTimelineCacheKey({ ...keyArgs, maxSeq }),
+      {
+        paramsKey,
+        revisionKey: buildThreadTimelineCacheKey({ ...keyArgs, maxSeq }),
+      },
       () => {
         const { profile, response } = buildThreadTimelineWithProfile(
           deps.db,
@@ -402,7 +406,6 @@ export function registerThreadDataRoutes(app: Hono, deps: AppDeps): void {
       query.afterSequence,
       "afterSequence",
     );
-    const paramsKey = buildThreadTimelineParamsKey(keyArgs);
     const previous =
       afterSequence === undefined
         ? undefined
--- a/apps/server/src/services/threads/timeline-cache.ts
+++ b/apps/server/src/services/threads/timeline-cache.ts
@@ -13,20 +13,20 @@ import type { ThreadTimelinePageRequest } from "./timeline-pagination.js";
  * (detail view + side-chat tabs), debounced realtime invalidations that fire
  * after the tail already settled, and re-opening a thread.
  *
- * Keying on the thread high-water `maxSeq` makes invalidation implicit: any
- * appended event bumps `maxSeq`, producing a new key and a cold rebuild. The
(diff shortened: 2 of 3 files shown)


SawyerHood:

PR #2074, opened 2026-08-20
Add EAS Update (OTA) to the mobile app

SawyerHood:
## What was wrong

A JS-only fix for the mobile app needed a full EAS build and an App Store Connect upload. That round trip is acceptable for internal testers, who install every nightly. It is expensive for external testers, who wait for processing and, for a real release, for review.

`expo-updates` was not installed, and the build profiles in `eas.json` carried no channels.

## What changed

- **`expo-updates`**, with `updates.url` and `runtimeVersion.policy: "fingerprint"` in `app.json`, plus a channel per build profile in `eas.json`.
- **`fingerprint.config.js`** skips `ExpoConfigVersions`. `mobile-ios-eas.yml` rewrites `app.json` `version` with the npm version on every nightly, and the fingerprint hashes the whole Expo config by default. Without the skip, each nightly forked the runtime version and no update could ever match an installed build.
- **`app.config.js`** turns the update client off when `BB_DISABLE_UPDATES=1`, and **`mobile-e2e.yml`** sets it. The E2E job builds in Release. A Release binary with updates enabled asks the production channel for a bundle at launch, and that bundle would replace the embedded E2E bundle in the middle of a Maestro flow. The failures would look random.
- **`mobile-update.yml`** publishes an update. It is manual: an update reaches every tester in minutes with no review in between. The job prints the runtime fingerprint next to the recent builds' fingerprints, because an update that reaches nobody otherwise looks like a successful publish.
- **README**: a new "Over-the-air updates (EAS Update)" section replaces the deferral note.

## How verified

Fingerprints, generated with `expo-updates fingerprint:generate --platform ios`:

| case | fingerprint |
| --- | --- |
| `version: 0.0.1`, no skip | `7f9c90b3…` |
| `version: 0.39.0`, no skip | `0f650d4d…` |
| `version: 0.0.1`, with skip | `b162a732…` |
| `version: 0.39.0`, with skip | `b162a732…` |
| with skip, plus one extra Info.plist key | `f9b4c845…` |

So the marketing version no longer forks the runtime version, and a native change still does.

`expo prebuild --platform ios`, reading the generated `Expo.plist`: `EXUpdatesEnabled` is `false` with `BB_DISABLE_UPDATES=1` and `true` without it.

`pnpm exec turbo run typecheck lint test --filter=@bb/mobile --force`: green, 834 tests. The new `app.config.test.ts` covers both silent failure modes; its E2E case fails when the guard in `app.config.js` is removed and passes with it.

## Note

Build `0.0.1 (2)`, already processed in TestFlight, carries no update client. The first build made after this merges is the earliest one that can receive an update.


SawyerHood:

PR #2079, opened 2026-08-20
Surface provider/error on thread.failed instead of a generic workflow failure

### LINKED ISSUE #1914: A rate-limited workflow call reports only 'Workflow worker failed' — the 429 is hidden in a child thread
A workflow call that dies on a provider **429** surfaces to every observer as the string `Workflow worker failed`, with no rate-limit detail anywhere in `bb workflows status` or `bb workflows history`. The real cause is present — but only inside the call's hidden child thread, which those two commands do not expose.

## Measured

bb 0.39.0, macOS. Run `wfr_e17de78c-4084-40bc-b9ea-07b0ca8b20d3` (a scheduled workflow, phase `Collect`):

```
bb workflows status <run>   ->  status failed, error "Workflow worker failed", calls 1/1 failed
bb workflows history <run>  ->  same string, no provider detail
```

The failing call was `callIndex 0`, dead 3.5s in. Its child thread `thr_3pm5bt4jh8` holds what actually happened:

```
provider/rateLimits/updated   five_hour status "blocked", reachedReason "five_hour",
                              overageStatus "rejected" / org_level_disabled
provider/error                "You've hit your session limit - resets 2:40pm"
                              errorInfo { category: rate-limit, providerCode: rate_limit_event,
                                          httpStatusCode: 429 }
```

Not a one-off: run `wfr_e86c4356` failed **13 of 29** calls with the identical opaque string and the identical 429 underneath.

## Why it matters more than a cosmetic message

`Workflow worker failed` is indistinguishable from a bb-side crash, so the reasonable first move is to go looking for a bug in the workflow script. On this machine that cost a full investigation of a 3,000-line workflow before the child thread was read — the script was never implicated.

It also **collapses two failures with opposite responses**:

| surfaced string | actual cause | correct response |
|---|---|---|
| `Workflow worker failed` | provider 429 | wait for the window, or switch account |
| `fetch failed` | bb server stall | investigate bb / the machine |

A rate limit is not a defect and needs no code change; a stall is and does. Told apart only by opening a hidden child thread, they get treated the same.

## Suggested fix

Propagate the child thread's `provider/error` — at minimum `category`, `providerCode` and `httpStatusCode` — into the call's own error, so `bb workflows status` and `history` can say "rate limited (429), resets HH:MM" instead of a generic worker failure. The information already exists and is already stored; it simply does not reach the surface anyone reads.

Filed from a machine that runs four scheduled bb workflows; happy to supply the full run records.

MPIsaac-Per:
## What was wrong

`getLastThreadErrorMessage()` only read `system/error` rows. Providers persist failures as `provider/error` (with optional `detail` and `errorInfo` for category, HTTP status, and provider code). On a workflow child that died on a Claude 429, `emitThreadFailed()` therefore published `error: null`, and the workflows plugin stored `error ?? "Workflow worker failed"`. `bb workflows status` and `history` showed that fallback. The 429 was only in the hidden child thread.

Root cause and independent repro: #1914, [report](https://get-bb.github.io/reports/issues/1914.html).

## What changed

TIER-2. No `HOST_DAEMON_PROTOCOL_VERSION` bump. `thread.failed` stays `{ thread, error: string | null }`. No plugin SDK change.

- `packages/db/src/data/events.ts` — `getLatestThreadErrorEventRow()` returns the newest `system/error` or `provider/error` for one thread, ordered by `sequence` descending. A later terminal failure outranks an earlier `willRetry: true` row. Filtering `willRetry` out would hide a 429 that is itself the newest event.
- `apps/server/src/services/threads/thread-data.ts` — `getLastThreadErrorMessage()` uses that query. `system/error` still returns `message`. `provider/error` formats `detail` (else `message`), then `category`, `http`, and `providerCode` when present.
- Workflows already persist a non-null `thread.failed.error` string; no plugin change.

Deviation from the issue comment's optional structured `errorInfo` / `resetsAt` field: this slice keeps the existing string payload.

## How you verified

Red before green in `apps/server/test/services/plugins/plugin-thread-events.test.ts`:

- `delivers thread.failed with a provider/error detail and metadata` — failed before the query/format change; now asserts the exact string `429 rate limit: too many requests | category: rate-limit | http: 429 | providerCode: rate_limit_event`.
- `surfaces a terminal provider/error over an earlier willRetry provider/error` — sequence 1 `willRetry: true` does not beat sequence 2 terminal.
- Existing `delivers thread.failed with the latest system/error message` still passes.

Query contract: `packages/db/test/data/events.test.ts` `returns the newest system/error or provider/error by sequence`.

```sh
pnpm exec turbo run test --filter=@bb/db -- --run test/data/events.test.ts
pnpm exec turbo run test --filter=@bb/server -- --run test/services/plugins/plugin-thread-events.test.ts
pnpm exec turbo run typecheck --filter=@bb/db --filter=@bb/server
```

db: 84 passed. server plugin-thread-events: 13 passed. typecheck: 0 errors. eslint on the five changed files: 0 errors.

Fixes #1914


SawyerHood:

PR #2090, opened 2026-08-20
Graduate idle provider session reaping

### LINKED ISSUE #1604: Idle agent processes are never reclaimed for non-Codex providers
> Synced from [ariofrio/bb#2](https://github.com/ariofrio/bb/issues/2). Original author: @ariofrio.

## Summary

The host daemon's idle provider-session reaper only reaps **Codex** processes. Claude Code and every ACP provider retain one agent OS process per idle thread indefinitely, and there is currently no non-destructive way to release one.

On a 16 GB machine, retained idle agents were the single largest consumer of memory (~5.2 GB), and the resulting pressure turns [#1131](https://github.com/get-bb/bb/issues/1131)-style event-loop stalls from milliseconds into seconds.

This is the automatic counterpart to [#1573](https://github.com/get-bb/bb/issues/1573) / [#1584](https://github.com/get-bb/bb/pull/1584), which add a *manual* release verb. Nothing calls that verb for an ordinary thread that finishes a turn and goes idle.

## Root cause

`apps/host-daemon/src/app.ts:69-70` sweeps every 5 minutes for sessions idle 30 minutes via `reapIdleProviderSessions({ idleForMs, nowMs })` — provider-agnostic and already parameterized. But `packages/agent-runtime/src/runtime.ts:2205` skips everything else:

```ts
if (!isThreadScopedCodexProcess(proc)) continue;
```

`isThreadScopedCodexProcess` (`runtime.ts:419-424`) requires `providerId === CODEX_PROVIDER_ID`, which can never hold for another provider, because only Codex gets a thread-scoped process key (`runtime.ts:392-401`).

Both landed in `2a84ecfdc` ([#130](https://github.com/get-bb/bb/pull/130), "Run Codex app-server per thread"), whose scope was per-thread Codex auth/rate-limit isolation and which states: *"Codex process keys now include the bb thread id for live thread work. Other providers keep their existing provider-scoped process key."* Other providers did not opt out of idle reclamation — they were never eligible for it.

## Evidence

Measured on `0.37.1-nightly.31693508850.1` (commit `596aab429`), macOS 26.5.1, 16 GB, mixed Claude Code and Codex threads.

- **Every reap event is Codex.** Across all 11 rotated `~/.bb/logs/host-daemon.*.log` files: 88 `Reaped idle provider sessions` entries — **88 `codex`, 0 `claude-code`.**
- **Live census** (processes mapped to threads via `BB_THREAD_ID`, idle time from the thread's last event in `bb.db`): **10 of 27 processes past the 30-minute threshold — all `claude`, no `codex`.** No Codex process is ever observed past the threshold. A second census by thread status: 24 of 32 processes belonged to `idle` threads, only 3 to `active`; 20 of the 24 were ordinary visible threads, not hidden plugin workers.
- **Per-agent cost: ~215 MB physical footprint** (`vmmap`: 209.9 / 226.2 / 211.7 / 217.0 MB), far more consistent than RSS, which read 109–326 MB for those same processes. The 24 idle-thread agents therefore account for **~5.2 GB** — about a third of installed RAM held by processes doing nothing, and close to the 5.5 GB the memory compressor held at the same moment.
- Agent CLIs were **35% of resident memory** (3.51 GB of 10.1 GB), the largest category on the machine.

## Why stop and archive don't cover this

- **`thread.stop`** is a no-op for an idle thread: `requestThreadStopForCurrentState` (`apps/server/src/services/threads/thread-lifecycle.ts`) matches neither branch and returns without dispatching. #1584 fixes this, but it is a manual call.
- **`thread.archive`** does not release an idle runtime — *"Archive only stops active runtime work"* (`thread-archive.ts:55-61`), *"Archive works on stored provider state, not on the live session"* (`command-dispatch.ts:483-484`). The release users observe minutes later comes from `requestEnvironmentCleanup`, gated by `wouldCleanupEnvironment` (`environment-cleanup-internal.ts:286-305`), which requires zero live threads and then **destroys the workspace**.

So for anyone keeping uncommitted work in a managed workspace, there is currently no non-destructive way to release an idle Claude Code process.

## Impact

With the host in this state, the bb server's own writable memory was **28 MB resident (2%), 447 MB swapped out**, so synchronous work on the serving loop becomes page-fault-bound. Stack samples put ~10% of main-thread samples in `pcache1TruncateUnsafe`, SQLite's per-commit page-cache walk via `pager_end_transaction` — **~30 ms per commit** at the measured ~3 commits/sec, against **~0.06 ms** for the same operation benchmarked on a copy of the same 229 MB `bb.db` with memory available. Over 18 minutes, **108 of ~217 five-second windows exceeded the stall threshold**, worst block **13.4 s**.

An *idle* Node process on the same machine at the same moment saw only 24–53 ms of loop delay, so this is not general CPU starvation. Stalls scale with concurrent thread activity and fall to near zero once the swarm goes quiet, memory pressure unchanged — pressure sets the cost of each synchronous operation, concurrency sets how often it is paid.

## Proposed change

No spawn-semantics change and no admission queue required.

**1. Generalize the reaper to providers whose sessions can be restored.** There are *two* Codex gates: `findReapableIdleProviderSession` (`runtime.ts:707`) rejects non-Codex threads during candidate selection, before the check at `runtime.ts:2205` is reached. The idle tracking underneath is already provider-agnostic — `observeProviderSessionIdleState` (`runtime.ts:676-693`) keys off generic `turn/started` / `turn/completed` / `provider/error` events — so the timestamps this needs are already recorded for every provider today.

Two constraints make this more than deleting the conditions. Both look tractable:

- **Restorability is not universal.** Claude Code resumes via the SDK's `resume` option (`claude-code/bridge/sdk-session.ts:211-257`), but ACP resume runs only if the agent advertises `agentCapabilities.loadSession`; otherwise the bridge falls back to `session/new`, warning *"could not restore the previous session; continuing in a fresh session without in-agent history"* (`acp/bridge/bridge.ts:1524-1573`). Reaping must be gated on restorability. `supportsLoadSession` is currently a local (`bridge.ts:1517`); persisting it on the session follows the `supportsImageInput` pattern already in that file (declared line 127, assigned line 1515). Agents without the capability keep today's behavior.
- **Non-Codex process keys are shared.** `proc.identity.threadIds` is a `Set`, so one process can host several threads in an environment. The reaper calls `forgetThreadRuntimeState` *before* `shutdownThreadScopedCodexProcessIfIdle`, which is safe only because Codex processes are 1:1 with threads; on a shared process still serving others it would clear a thread's state and free nothing. Fix: require the thread to be the last on its process before forgetting — reclaiming per environment rather than per thread — or extend #130's thread-scoped keys to other providers.

The rest of the lifecycle is already generic: `shutdownProvider` (`runtime-provider-process.ts:289`) takes only a process key and provider id, and `handleProviderProcessExit` (line 578) works off `identity.threadIds` and already separates expected shutdowns from crashes via `consumeExpectedProviderProcessShutdown`. bb must already survive a provider process dying and returning, so reaping is a controlled instance of a path every provider supports.

**Verified empirically** on the build above, using the harshest version of that path. I spawned a Claude Code thread, gave it a codeword, let it go idle, then `SIGKILL`ed its agent process and sent another turn:

- The thread went to `error` immediately after the kill — expected for an *unexpected* exit; a reaper using `shutdownProvider` marks the shutdown expected and should not produce this.
- The next turn transparently started a replacement process (new pid) and answered the codeword correctly.
- The `events` table shows a **single** `provider_thread_id` (`cb32df02-…`) spanning sequences 6–36, i.e. across both turns — so it was a real session resume, not bb replaying history into a fresh session.

Claude Code sessions therefore survive losing their process, with in-agent history intact, even without the graceful path a reaper would use.

**2. Make eviction demand-driven, not only scheduled.** `idleForMs` is already a parameter, so `idleForMs: 0` is immediate reaping; the 5-minute sweep is just one policy over it. Every acquisition funnels through one admission point — `RuntimeManager.ensureEnvironment` (`runtime-manager.ts:905`, single creation site at line 957) — which is the moment new memory is about to be committed, and so the natural moment to reclaim an idle runtime instead of leaving it until the next sweep. It is also what keeps the bound in (3) from ever blocking a start: a starting agent reclaims a slot on the spot rather than waiting for a sweep to free one. Victim selection needs no new bookkeeping: `getLiveThreadIds()` (`runtime.ts:2226`), `hasOpenBackgroundWork()` (`runtime.ts:2235`), and `idleProviderSessionSinceMsByThreadId` (`runtime.ts:332`) to pick the runtime idle longest.

**3. Bound the retained-*idle* pool proactively** — but not concurrency. (1) acts only after 30 minutes and (2) only under contention, so a burst of parallel subagents that all go idle is retained in full until the TTL expires, even on a quiet host. Capping *actives* instead would require an admission queue and would block `spawn`, throttling exactly the burst workloads that should keep working.

Only (3) needs tuning. **Host memory pressure is the better trigger where available** — Linux PSI at `/proc/pressure/memory`, macOS memory pressure — because it self-tunes across a 16 GB laptop and a 31 GB headless box. It needs hysteresis, or it evicts in a loop when the pressure originates elsewhere, and it is reactive, firing once the host is already swapping. A static count is therefore worth keeping as a portable, preventive fallback: at ~215 MB/agent, roughly **one idle agent per 2 GB of RAM, floor 2** — 8 on the machine above, versus the 24 observed.

## Open questions

- Was excluding non-Codex providers deliberate, or incidental to #130's scope? Claude Code looks straightforwardly reapable; the two constraints above are the only ones I found. Are there others?
- Should non-Codex providers get thread-scoped process keys (extending #130's model)? That would remove the shared-process ordering problem and make reclamation per-thread rather than per-environment.
- Is host memory pressure acceptable as an input at this layer, or should the proactive bound stay a static count for predictability and portability?

## Environment

- bb `0.37.1-nightly.31693508850.1` (desktop nightly, commit `596aab429`), macOS 26.5.1 arm64, 16 GB
- Providers in use: Claude Code and Codex; `bb.db` 229 MB, ~96k events, ~362 threads

ScaleLeanChris:
## What was wrong

Idle release for restorable non-Codex providers remained behind the default-off `providerSessionReaping` experiment added after #1604. Existing installations could also retain a persisted `false` value. As a result, the host daemon skipped otherwise safe release for idle Claude Code, Pi, and ACP sessions under normal settings. The reproduced root cause and alternatives are documented in [issue #1604](https://github.com/get-bb/bb/issues/1604#issuecomment-5324681080).

## What changed

- Graduate idle release into standard behavior for every provider session that reports restore support.
- Preserve the legacy thread-scoped Codex fallback and apply open-turn, command, agent, workflow, and monitor protections to every provider.
- Remove the experiment setting and the server-owned `/internal/runtime-policy` endpoint that existed only to distribute it.
- Bump `HOST_DAEMON_PROTOCOL_VERSION` from 146 to 147 because older daemons still request that removed endpoint and skip non-Codex release when it is unavailable.
- Update web, mobile, desktop, database, contract, documentation, and generated-guide source fixtures that referenced the graduated experiment.

## How you verified

- Added a lifecycle regression that failed before the runtime change because a restorable Claude Code session remained loaded, then passed after the change.
- `pnpm exec turbo run typecheck` for the nine affected packages: 12 tasks passed.
- Agent runtime lifecycle: 36 tests passed.
- Host daemon focused suite: 85 tests passed.
- Host-daemon contract: 38 tests passed.
- Web settings/layout: 9 tests passed. Server experiments: 4 tests passed. Database experiments: 1 test passed.
- Mobile full suite: 123 files and 830 tests passed.
- Templates: 41 tests passed across the local and packed external scaffold suites after building the Plugin SDK prerequisite.
- `git diff --check` passed and no removed experiment or runtime-policy symbols remain.
- Mobile typecheck is not included because the current package script invokes `tsc`, while its TypeScript 6 alias exposes `tsc6`; this is unchanged by this PR.

Controlling BAC work order: [BAC-117](https://linear.app/scale-lean/issue/BAC-117)

Fixes #1604

> AGENT GENERATED: by GPT-5.6

### DIFF
--- a/apps/app/src/components/layout/AppLayout.plugin-panel-header.test.tsx
+++ b/apps/app/src/components/layout/AppLayout.plugin-panel-header.test.tsx
@@ -26,7 +26,6 @@ vi.mock("@/hooks/queries/system-queries", () => ({
         changelogPreview: false,
         editMessages: false,
         mobileApp: false,
-        providerSessionReaping: false,
         timelineWindowing: false,
       },
     },
--- a/apps/app/src/components/layout/AppLayout.root-compose-project.test.tsx
+++ b/apps/app/src/components/layout/AppLayout.root-compose-project.test.tsx
@@ -31,7 +31,6 @@ vi.mock("@/hooks/queries/system-queries", () => ({
         changelogPreview: false,
         editMessages: false,
         mobileApp: false,
-        providerSessionReaping: false,
         timelineWindowing: false,
       },
     },
--- a/apps/app/src/components/layout/AppLayout.sidebar-resize.test.tsx
+++ b/apps/app/src/components/layout/AppLayout.sidebar-resize.test.tsx
@@ -37,7 +37,6 @@ vi.mock("@/hooks/queries/system-queries", () => ({
     data: {
       experiments: {
         editMessages: false,
-        providerSessionReaping: false,
       },
     },
   }),
--- a/apps/app/src/lib/system-config-atoms.ts
+++ b/apps/app/src/lib/system-config-atoms.ts
@@ -24,7 +24,6 @@ const unavailableSystemConfig: SystemConfigResponse = {
     changelogPreview: false,
     editMessages: false,
     mobileApp: false,
-    providerSessionReaping: false,
     timelineWindowing: false,
   },
   appearance: defaultAppTheme,
--- a/apps/app/src/views/SettingsView.experiments.test.tsx
+++ b/apps/app/src/views/SettingsView.experiments.test.tsx
@@ -8,7 +8,6 @@ afterEach(cleanup);
 function renderSection(overrides?: {
   onChangelogPreviewEnabledChange?: (enabled: boolean) => void;
   onMobileAppEnabledChange?: (enabled: boolean) => void;
-  onProviderSessionReapingEnabledChange?: (enabled: boolean) => void;
   onTimelineWindowingEnabledChange?: (enabled: boolean) => void;
 }) {
   return render(
@@ -17,16 +16,12 @@ function renderSection(overrides?: {
       disabled={false}
       editMessagesEnabled={false}
       mobileAppEnabled={false}
-      providerSessionReapingEnabled={false}
       timelineWindowingEnabled={false}
       onChangelogPreviewEnabledChange={
         overrides?.onChangelogPreviewEnabledChange ?? vi.fn()
       }
       onEditMessagesEnabledChange={vi.fn()}
       onMobileAppEnabledChange={overrides?.onMobileAppEnabledChange ?? vi.fn()}
-      onProviderSessionReapingEnabledChange={
-        overrides?.onProviderSessionReapingEnabledChange ?? vi.fn()
-      }
       onTimelineWindowingEnabledChange={
         overrides?.onTimelineWindowingEnabledChange ?? vi.fn()
       }
@@ -46,13 +41,6 @@ describe("ExperimentsSettingsSection", () => {
     const onChange = vi.fn();
     renderSection({ onMobileAppEnabledChange: onChange });
     fireEvent.click(screen.getByLabelText("Mobile app"));
-    expect(onChange).toHaveBeenCalledWith(true);
-  });
-
-  it("reports idle provider session release changes", () => {
-    const onChange = vi.fn();
-    renderSection({ onProviderSessionReapingEnabledChange: onChange });
-    fireEvent.click(screen.getByLabelText("Idle provider session release"));
     expect(onChange).toHaveBeenCalledWith(true);
   });
 
--- a/apps/app/src/views/SettingsView.stories.tsx
+++ b/apps/app/src/views/SettingsView.stories.tsx
@@ -356,7 +356,6 @@ function ExperimentsStory() {
       disabled={false}
       editMessagesEnabled={state.experiments.editMessages}
       mobileAppEnabled={state.experiments.mobileApp}
-      providerSessionReapingEnabled={state.experiments.providerSessionReaping}
       timelineWindowingEnabled={state.experiments.timelineWindowing}
       onChangelogPreviewEnabledChange={(enabled) =>
         state.setExperiments((current) => ({
@@ -374,12 +373,6 @@ function ExperimentsStory() {
         state.setExperiments((current) => ({
           ...current,
           mobileApp: enabled,
-        }))
-      }
-      onProviderSessionReapingEnabledChange={(enabled) =>
-        state.setExperiments((current) => ({
-          ...current,
-          providerSessionReaping: enabled,
         }))
       }
       onTimelineWindowingEnabledChange={(enabled) =>
--- a/apps/app/src/views/SettingsView.tsx
+++ b/apps/app/src/views/SettingsView.tsx
@@ -186,12 +186,10 @@ interface ExperimentsSettingsSectionProps {
   changelogPreviewEnabled: boolean;
   editMessagesEnabled: boolean;
   mobileAppEnabled: boolean;
-  providerSessionReapingEnabled: boolean;
   timelineWindowingEnabled: boolean;
   onChangelogPreviewEnabledChange: (enabled: boolean) => void;
   onEditMessagesEnabledChange: (enabled: boolean) => void;
   onMobileAppEnabledChange: (enabled: boolean) => void;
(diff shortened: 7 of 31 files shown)


celeroncoder:
Related context from the workflows plugin side: opened #2284 where `bb-plugin-workflows` omits `stopChild(threadId)` on unstructured subagent completion.

While the session reaper in this PR provides a solid daemon-level safety net for orphaned threads, burst workflows running dozens of subagent tasks exhaust RAM within minutes without immediate worker thread stops upon task completion.

SawyerHood:

PR #2103, opened 2026-08-20
Include project dot paths in file listings

### LINKED ISSUE #2093: Path listing hides every dotfile and node_modules unconditionally, so file search cannot find .github/workflows/ci.yml
### Human comments

This will be useful for file tree plugins to be more customizable

------

### Summary

`host.list_paths` skips any entry whose name starts with `.`, plus `node_modules` and symlinks, with no way to opt in. Because BB's own file search runs through that command, quick-open cannot find `.github/workflows/ci.yml` — a file BB reads and previews perfectly well through `host.read_file`. The rule is hardcoded in the daemon's walker, not derived from gitignore, and callers (first-party or plugin) cannot override it.

I hit this building a file tree in a `fileOpener` plugin, but the plugin is incidental: the same gap is visible in BB's own UI, and that is what the repro below uses.

### Versions and environment

- bb dev build from source at `6be45053be92f964895038fc374a45e759087479` (`main`), plus bb 0.38.0 desktop (same behavior).
- macOS 14 (Darwin 23.3.0), Node v26.7.0.
- Repository under test: this one (`.github/workflows/ci.yml` is present).

### Steps to reproduce

1. Open a thread whose environment is a checkout of this repo.
2. Open the secondary panel's file search (`+` → file search) and type `ci.yml`.

Or against the API directly, which is what I did:

```bash
curl -s "http://127.0.0.1:<port>/api/v1/environments/<envId>/paths?query=ci.yml&limit=5&includeFiles=true&includeDirectories=false"
```

### Expected vs actual

Actual — `.github/workflows/ci.yml` is absent, and the four fuzzy matches returned are all unrelated:

```
query ci.yml     -> 4 results: apps/mobile/e2e/flows/phase4a-conversation-rows.yaml,
                               apps/mobile/e2e/flows/phase4b-actions.yaml,
                               apps/mobile/e2e/flows/phase4b-interactions.yaml
query workflows  -> 5 results: plugins/workflows/README.md, plugins/workflows/package.json, ...
query AGENTS.md  -> 1 results: AGENTS.md            # control: the endpoint works
```

Expected: `.github/workflows/ci.yml` appears for `ci.yml`. It is an ordinary YAML file, and BB is otherwise perfectly happy with it — reading the very same path through the host file API succeeds:

```
read .github/workflows/ci.yml -> OK, 7794 chars, sha d9d35944bf59
```

So one host command serves the file and another refuses to admit it exists.

### Evidence

The exclusions are three unconditional `continue`s in the recursive walker, [`listPathsRecursively`](https://github.com/get-bb/bb/blob/6be45053be92f964895038fc374a45e759087479/apps/host-daemon/src/command-handlers/file-list.ts#L137-L152):

```ts
if (entry.name.startsWith(".")) continue;
if (entry.name === "node_modules") continue;
if (entry.isSymbolicLink()) continue;
```

- [`host.list_paths` command params](https://github.com/get-bb/bb/blob/6be45053be92f964895038fc374a45e759087479/packages/plugin-sdk/src/host-contract.ts) carry `path`, `query`, `limit`, `includeFiles`, `includeDirectories` — nothing to relax the above.
- [`browseHostDirectory`](https://github.com/get-bb/bb/blob/6be45053be92f964895038fc374a45e759087479/apps/host-daemon/src/command-handlers/host-files.ts#L152-L157) applies the same two name rules, so the interactive path browser cannot reach them either. (It does follow symlinks, deliberately — single-level browsing has no loop risk.)
- Consumers that inherit the behavior: `environments.paths` ([routes/environments.ts#L625](https://github.com/get-bb/bb/blob/6be45053be92f964895038fc374a45e759087479/apps/server/src/routes/environments.ts#L625)), `projects.paths` ([routes/projects.ts#L680](https://github.com/get-bb/bb/blob/6be45053be92f964895038fc374a45e759087479/apps/server/src/routes/projects.ts#L680)), and `files.listPaths` ([routes/files.ts#L292](https://github.com/get-bb/bb/blob/6be45053be92f964895038fc374a45e759087479/apps/server/src/routes/files.ts#L292)).
- `host.read_file` has no equivalent exclusion, which is why the read above succeeds.

Note the dotfile rule is much broader than the `node_modules` one. It hides `.github/`, `.claude/`, `.bb/`, `.vscode/` and everything under them — directories users edit routinely — while `node_modules` at least matches most people's intent most of the time.

### What you ruled out

- Not gitignore: nothing reads an ignore file; the names are literals in the walker. `.github/workflows/` is committed and not ignored.
- Not the fuzzy ranker dropping it: raising `limit` and querying `workflows` directly still returns nothing under `.github/`. The `AGENTS.md` control proves the endpoint and environment are right.
- Not a permissions or symlink artifact: `.github/workflows/ci.yml` is a regular file and reads fine through `host.read_file` on the same host and root.
- Not a duplicate: searched open and closed for `list_paths dotfiles`, `file search hidden files`, `node_modules exclusion`, `listPathsRecursively` — no matches.
- Not plugin-specific: the repro above is entirely first-party.

Medium priority, small-to-medium effort. It silently makes a class of real project files unreachable from quick-open, and the failure is invisible — the search returns *other* results, so it reads as "no such file" rather than "this file is filtered".

Suggested shape: make the exclusions caller-configurable on `host.list_paths` rather than baked into the walker — for example `includeHidden` and `includeNodeModules` (or a general `exclude` list), defaulting to today's behavior so nothing changes for existing callers. That keeps quick-open's defaults sane while letting a file tree offer "show hidden files" and letting a plugin decide for itself. Note this is a wire change and would need a `HOST_DAEMON_PROTOCOL_VERSION` bump per AGENTS.md.

If the appetite is smaller, narrowing the dotfile rule alone — skipping `.git/` rather than every dot-leading name — would fix the reported symptom without a protocol change.

Found in bb thread `thr_nnk8m7d8gw` while building a Monaco `fileOpener` plugin; see also #2026 and #2072 from the same work.

ScaleLeanChris:
## What was wrong

The host daemon's shared recursive file walker skipped every entry whose name started with `.`. Both `host.list_paths` and `host.list_files` use that walker, so ordinary project files such as `.github/workflows/ci.yml` never reached file search even though the file read API could read them. See #2093.

## What changed

- Replace the broad dot-name exclusion with explicit exclusions for `.git` and `node_modules`.
- Continue to skip recursive symlinks.
- Add a regression test that includes `.github/workflows/ci.yml` and `.env` while excluding `.git/config` and a dependency file under `node_modules`.
- Increase `HOST_DAEMON_PROTOCOL_VERSION` from 146 to 147 because the returned RPC path set changes.
- Leave `host.browse_directory` unchanged. That single-level remote folder picker starts at the host home directory and should continue hiding home dot entries.

This deliberately does not add the issue's proposed caller-configurable wire fields. The narrower policy fixes normal project discovery without expanding the public API. No CLI, guide, SDK, or route shape changed.

## How you verified

- Before implementation, the new regression test failed with no dot paths returned. After implementation, `pnpm exec turbo run test --filter=@bb/host-daemon -- --run src/command-handlers/file-list.test.ts` passed 15 tests.
- `pnpm exec turbo run test --filter=@bb/host-daemon-contract -- --run test/contract.test.ts` passed 38 tests.
- `pnpm exec turbo run typecheck build --filter=@bb/host-daemon --filter=@bb/host-daemon-contract --force` passed all 7 Turbo tasks.
- A manual post-fix walker probe returned `.github/workflows/ci.yml` and omitted `.git/config` and `node_modules/pkg/index.js`.
- `git diff --check` passed.

Two unrelated full-suite assertions remain environment-sensitive in this managed worktree. The host-daemon suite passed 556 tests and failed the detached-process cleanup test, which also failed alone. The host-daemon-contract suite passed 51 tests and failed an exact gzip-size fixture by 1 to 4 bytes under the local zlib runtime. Neither failure imports or exercises the changed listing behavior.

Fixes #2093

> AGENT GENERATED: by GPT-5

### DIFF
--- a/apps/host-daemon/src/command-handlers/file-list.test.ts
+++ b/apps/host-daemon/src/command-handlers/file-list.test.ts
@@ -236,6 +236,42 @@ describe("listPathsRecursively", () => {
     }
   });
 
+  it("includes project dot paths without traversing Git internals or dependencies", async () => {
+    const root = await fs.mkdtemp(path.join(os.tmpdir(), "bb-file-list-"));
+    try {
+      await fs.mkdir(path.join(root, ".github", "workflows"), {
+        recursive: true,
+      });
+      await fs.writeFile(path.join(root, ".github", "workflows", "ci.yml"), "");
+      await fs.writeFile(path.join(root, ".env"), "");
+      await fs.mkdir(path.join(root, ".git"));
+      await fs.writeFile(path.join(root, ".git", "config"), "");
+      await fs.mkdir(path.join(root, "node_modules", "dependency"), {
+        recursive: true,
+      });
+      await fs.writeFile(
+        path.join(root, "node_modules", "dependency", "index.js"),
+        "",
+      );
+
+      const result = await listPathsRecursively({
+        dir: root,
+        root,
+        includeFiles: true,
+        includeDirectories: true,
+      });
+
+      expect(result.map((entry) => entry.path).sort()).toEqual([
+        ".env",
+        ".github",
+        ".github/workflows",
+        ".github/workflows/ci.yml",
+      ]);
+    } finally {
+      await fs.rm(root, { recursive: true, force: true });
+    }
+  });
+
   it("does not return symlinked files as regular path entries", async () => {
     const root = await fs.mkdtemp(path.join(os.tmpdir(), "bb-file-list-"));
     try {
--- a/apps/host-daemon/src/command-handlers/file-list.ts
+++ b/apps/host-daemon/src/command-handlers/file-list.ts
@@ -49,6 +49,8 @@ export interface ListPathsRecursivelyArgs extends PathListInclusion {
   root: string;
 }
 
+const RECURSIVE_PATH_SKIP_NAMES = new Set([".git", "node_modules"]);
+
 function shouldIncludePath(
   pathKind: HostPathEntryKind,
   inclusion: PathListInclusion,
@@ -140,8 +142,7 @@ export async function listPathsRecursively(
   const entries = await fs.readdir(args.dir, { withFileTypes: true });
   const results: ListedPath[] = [];
   for (const entry of entries) {
-    if (entry.name.startsWith(".")) continue;
-    if (entry.name === "node_modules") continue;
+    if (RECURSIVE_PATH_SKIP_NAMES.has(entry.name)) continue;
     if (entry.isSymbolicLink()) continue;
 
     const fullPath = path.join(args.dir, entry.name);
--- a/packages/host-daemon-contract/src/protocol.ts
+++ b/packages/host-daemon-contract/src/protocol.ts
@@ -1,3 +1,7 @@
+// Version 147 includes ordinary project dot paths in recursive file listings
+// while continuing to exclude `.git`, `node_modules`, and symlinks. Older
+// daemons silently omit paths such as `.github/workflows/ci.yml`.
+//
 // Version 146 adds the lightweight `host.list_branch_options` RPC so branch
 // pickers can read cached refs while the daemon refreshes remotes in the
 // background. Older daemons cannot parse or serve that command.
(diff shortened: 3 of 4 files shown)


SawyerHood:

PR #2107, opened 2026-08-20
Add streamer mode to hide custom models from model lists

SawyerHood:
## What was wrong

bb had no way to hide the `customModels` entries from `~/.bb/config.json`. A user with a private or early-access model id showed it in every model picker during a screen share or stream.

## What changed

- `packages/domain/src/app-settings.ts`: new `streamerMode` general setting, default `false`. The key/value settings table needs no migration, and `bb settings general streamerMode <true|false>` works with no CLI change.
- `apps/server/src/services/system/execution-options.ts`: new `listVisibleCustomModels(deps)` returns `[]` when streamer mode is on. All three `appendCustomModels` call sites use it, so the web and mobile pickers, plugin pickers, `bb provider models`, and `sdk.providers.models` all hide the entries. The filter runs after the memoized daemon probe, so a toggle does not re-probe the host. An explicit thread model request bypasses the catalog, so a hidden model still runs when a caller names it.
- `apps/app/src/views/SettingsView.tsx`: "Streamer mode" switch in Settings → General, plus story wiring.
- `apps/app/src/hooks/cache-owners/system-cache-effects.ts`: a General settings write now also invalidates the execution options queries, so open pickers refetch at once.
- `apps/mobile/src/screens/settings/GeneralSettingsScreen.tsx`: the same switch in a Privacy section.
- Docs: `docs/configuration.md`, the `bb guide` customization template, and the bb-cli skill and its settings reference.

No `HOST_DAEMON_PROTOCOL_VERSION` bump. Nothing on the server ↔ daemon wire changed.

## How you verified

- New server test in `apps/server/test/system/execution-options.test.ts`: the catalog shows the custom model, streamer mode on hides it, streamer mode off restores it, and the daemon receives one `provider.list_models` call. With the filter removed it fails with `expected [ 'claude-opus-5', …(1) ] to deeply equal [ 'claude-opus-5' ]`.
- `apps/app/src/hooks/mutations/settings-mutations.test.tsx` now also asserts the execution options query is invalidated.
- `pnpm exec turbo run typecheck --filter=@bb/domain --filter=@bb/db --filter=@bb/server --filter=@bb/app --filter=@bb/mobile` passes.
- `pnpm exec turbo run test` for the settings, migration, CLI, SettingsView, and execution-options test files passes.
- Manual end-to-end check in the dev app: see the follow-up comment.

Fixes #

### DIFF
--- a/apps/app/src/hooks/cache-owners/system-cache-effects.ts
+++ b/apps/app/src/hooks/cache-owners/system-cache-effects.ts
@@ -153,7 +153,10 @@ export function invalidateSystemExecutionOptions({
   });
 }
 
-/** Refresh settings and timeline projections after a General settings write. */
+/**
+ * Refresh settings, timeline projections, and model catalogs after a General
+ * settings write. Streamer mode changes which custom models the server lists.
+ */
 export function invalidateGeneralSettingsDependencies({
   queryClient,
 }: QueryClientArg): void {
@@ -163,6 +166,7 @@ export function invalidateGeneralSettingsDependencies({
       systemConfigQueryKey(),
       allThreadTimelineQueryKeyPrefix(),
       allThreadTimelineTurnSummaryDetailsQueryKeyPrefix(),
+      allSystemExecutionOptionsQueryKeyPrefix(),
     ],
   });
 }
--- a/apps/app/src/hooks/cache-owners/system-cache-effects.ts
+++ b/apps/app/src/hooks/cache-owners/system-cache-effects.ts
@@ -33,6 +33,7 @@ import {
 } from "../queries/query-keys";
 import { allThreadDefaultExecutionOptionsQueryKeyPrefix } from "../queries/thread-default-execution-options-query";
 import type { QueryClientArg } from "../cache-effect-types";
+import { clearCachedModelCatalogs } from "@/lib/model-catalog-cache";
 import { bumpAllDiffPatchEvictionGenerations } from "./environment-diff-patch-cache-owner";
 import { invalidateSystemVersion } from "./system-version-cache-owner";
 import {
@@ -153,10 +154,7 @@ export function invalidateSystemExecutionOptions({
   });
 }
 
-/**
- * Refresh settings, timeline projections, and model catalogs after a General
- * settings write. Streamer mode changes which custom models the server lists.
- */
+/** Refresh settings and timeline projections after a General settings write. */
 export function invalidateGeneralSettingsDependencies({
   queryClient,
 }: QueryClientArg): void {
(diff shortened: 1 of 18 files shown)


SawyerHood:
## Manual end-to-end check

Dev app from this branch, with this `config.json` in the dev data dir:

```json
{ "customModels": [ { "providerId": "claude-code", "model": "claude-secret-early-access", "displayName": "Secret Early Access" } ] }
```

CLI against the live server:

```
$ bb provider models claude-code --json        # streamerMode=false
[..., 'claude-horchata-eap[1m]', 'claude-secret-early-access']
$ bb settings general streamerMode true
$ bb provider models claude-code --json
[..., 'claude-horchata-eap[1m]']
$ bb settings general streamerMode false
$ bb provider models claude-code --json
[..., 'claude-horchata-eap[1m]', 'claude-secret-early-access']
```

Browser (headless Chrome):

1. The "Streamer mode" switch renders in Settings → General with the default off.
2. New thread composer → Claude tab: the picker lists "Secret Early Access" last.
3. Turned the switch on in Settings, went back to the composer in the same tab with no reload.
4. The picker no longer lists "Secret Early Access", and "More models" does not list it either.

Note: `claude-horchata-eap[1m]` stays in the list because the Claude Code CLI reports it in its own catalog. Streamer mode hides only `customModels` entries from `config.json`.

SawyerHood:
## SlopCop follow-up (e2e1a55d9)

1. **High — execution policy.** Fixed in part, and one part is a stated trade-off.
   - `resolveSystemProviderModels` (the thread-create default path) now keeps the full custom model list. A thread created without an explicit model resolves the same default with streamer mode on, and a provider whose only models are custom can still start. New test: "keeps custom models in the thread-create default catalog while streamer mode is on".
   - The composer fallback stays as is. When the stored selection is a hidden model, the client treats it as unavailable, selects the provider default, and the next send records that default. The alternative is to keep the hidden selection and mask its label in both clients, but the picker is the only surface that shows a model name, and a masked "Hidden model" row would still signal a secret on stream. Hiding the model and moving to the default is the leak-free choice. The docs and the skill reference now state this behavior and tell the user to re-select the model after they turn streamer mode off.
2. **Medium — stale client catalog.** Fixed. A streamer mode flip now clears the localStorage model catalog preload and resets the execution-options queries instead of invalidating them, so open pickers drop the old list and refetch. Test: "drops cached model catalogs when streamer mode flips".
3. **Low — duplicate refetch.** Reduced. Other General settings writes no longer touch model catalogs; the `config-changed` broadcast covers them. The streamer flip itself still resets locally and then receives the broadcast, the same pattern the existing settings writes use for `/system/config`. The server memoizes the model probe, so the second request is cheap.
4. **Low — provider guide.** Fixed. The providers chapter and the bb-cli skill's `customModels` entry now point at `streamerMode`.

PR #2109, opened 2026-08-20
Fix T3 Sidebar long child-menu scrolling

### LINKED ISSUE #2105: T3 Sidebar child-thread menu cannot scroll long lists
## Summary

With T3 Sidebar selected, a parent thread with enough children to exceed the viewport opens a child-thread menu whose list has no bounded scroll region. Lower children are clipped and unreachable: wheel scrolling fails on desktop, and touch scrolling fails on mobile/tablet.

## Versions and environment

- bb 0.39.0
- T3 Sidebar example at public commit `e171ceba4ef00b6729ca2d6b579fe9af88be4c9c`
- Desktop reproduction: Chrome, 1728 × 813 viewport
- Compact reproduction: iPhone 17 / iOS 26.4 Simulator Safari; also observed by the reporter on mobile and tablet

## Steps to reproduce

1. Enable T3 Sidebar under **Settings → Appearance → Sidebar**.
2. Open a parent thread with enough unarchived children to exceed the viewport; the measured parent had 26.
3. Click the header chip labelled `26 child threads`.
4. On desktop, place the pointer over the menu and use the mouse wheel. On mobile/tablet, swipe vertically inside the menu.

## Expected vs actual

```text
Expected:
- The menu is capped to the available viewport height.
- Its child list scrolls, so every child remains reachable.

Actual desktop measurements before the fix:
- viewport height: 813px
- menu height: 1,157px (bottom: 1,203px)
- list clientHeight: 1,125px
- list scrollHeight: 1,125px
- list maxScroll: 0px
- list scrollTop after a 500px wheel input: 0px

Actual compact behavior:
- The visible portion of the list does not move on a vertical touch gesture.
- Children below the viewport cannot be opened from this menu.
```

## Evidence

The released example renders all children into an uncapped list with no overflow behavior: https://github.com/get-bb/bb/blob/e171ceba4ef00b6729ca2d6b579fe9af88be4c9c/examples/plugins/t3sidebar/src/SubagentsChip.tsx#L62-L75

After adding a capped flex menu plus an internal `overflow-y-auto` list, the desktop menu measures 512px tall, list maxScroll is 645px, and the same 500px wheel input moves list scrollTop to 500px.

## What you ruled out

- Searched open and closed issues in both `get-bb/bb` and `SawyerHood/bb-plugin-t3sidebar`; no duplicate found.
- This is not the core composer “Active child thread” card; the affected control is T3 Sidebar's `experimental_threadHeaderAction`.
- Desktop is confirmed; this is not mobile-only.
- The separate compact horizontal collision/positioning problem is intentionally tracked independently.

Medium / small. No data is lost, but a parent with many children cannot navigate to clipped children through T3 Sidebar's intended header control.

> AGENT GENERATED: by GPT-5.6

Willhong:
## What was wrong

T3 Sidebar rendered every child in an uncapped menu and an ordinary `ul`. With enough children, the menu extended beyond the thread viewport, but the list had no scroll range, so lower children were unreachable with both a desktop wheel and compact touch input. Fixes #2105. Standalone counterpart: https://github.com/SawyerHood/bb-plugin-t3sidebar/issues/4.

## What changed

- Cap the child menu at the smaller of `32rem` and the available dynamic viewport height.
- Make the menu a flex column and the child list its `min-h-0`, `overflow-y-auto` scroll region.
- Contain scroll chaining and preserve mobile WebKit touch/momentum scrolling.
- Add a 26-child regression test that pins the bounded-menu and scroll-container contract.
- No host-daemon wire, public plugin API, CLI, configuration, guide, or documentation changes.

## How you verified

- `pnpm exec turbo run test typecheck --filter=bb-plugin-t3sidebar` — 81 tests passed in the scroll-only iteration; typecheck passed.
- Desktop Chrome at 1728 × 813, 26 children:
  - before: menu 1,157px; list maxScroll 0; wheel 500px left `scrollTop = 0`
  - after: menu 512px; list maxScroll 645px; wheel 500px moved `scrollTop = 500`
- The regression test fails on the old classes and passes with this patch.

Fixes #2105

> AGENT GENERATED: by GPT-5.6

### DIFF
--- a/examples/plugins/t3sidebar/src/SubagentsChip.test.tsx
+++ b/examples/plugins/t3sidebar/src/SubagentsChip.test.tsx
@@ -1,0 +1,95 @@
+// @vitest-environment jsdom
+import { afterEach, describe, expect, it } from "vitest";
+import { cleanup, fireEvent, screen } from "@testing-library/react";
+import { loadPluginApp, renderSlot } from "@get-bb/plugin-sdk/testing/app";
+import type { PluginSidebarThread } from "@get-bb/plugin-sdk";
+
+const app = await loadPluginApp(() => import("../app"));
+const childrenChip = app.threadHeaderActions.find(
+  (slot) => slot.id === "children",
+)!;
+
+function thread(
+  overrides: Partial<PluginSidebarThread> = {},
+): PluginSidebarThread {
+  return {
+    id: "thr_1",
+    projectId: "proj_1",
+    title: "A thread",
+    titleFallback: null,
+    parentThreadId: null,
+    sectionId: null,
+    originKind: null,
+    originPluginId: null,
+    providerId: "codex",
+    hasPendingInteraction: false,
+    activity: {
+      workflows: 0,
+      backgroundAgents: 0,
+      backgroundCommands: 0,
+      planMode: 0,
+      goals: 0,
+    },
+    indicator: "none",
+    indicatorLabel: null,
+    isUnread: false,
+    isPinned: false,
+    isArchived: false,
+    environment: null,
+    host: null,
+    createdAt: 100,
+    updatedAt: 100,
+    lastReadAt: 100,
+    latestAttentionAt: 100,
+    ...overrides,
+  };
+}
+
+afterEach(cleanup);
+
+describe("SubagentsChip child list", () => {
+  it("caps a long menu and makes the child list scrollable", () => {
+    renderSlot(
+      childrenChip,
+      {
+        threadId: "parent",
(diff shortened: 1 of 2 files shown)


Willhong:
Standalone implementation counterpart: https://github.com/SawyerHood/bb-plugin-t3sidebar/pull/6

ymichael:
the plugin now lives exclusively in [SawyerHood/bb-plugin-t3sidebar](https://github.com/SawyerHood/bb-plugin-t3sidebar)

PR #2114, opened 2026-08-21
Add a keyboard-first quick palette

andrewkchan:
## Human comments

Adds a new quick palette menu that can be opened via CMD+SHIFT+P on MacOS and navigatable via arrow keys. 

Screenshot 2026-08-21 at 1 32 42 PM

We already have a command system in bb, this makes a subset of the commands accessible via quick palette.

----

## What was wrong

Every app command was reachable only through its own chord. A command whose
default binding ships unassigned (`thread.rename`, `thread.archive`,
`pane.focus.previous/next`) had no way to be run at all, and any command a user
had not memorized was effectively unreachable — Settings → Keyboard was the only
surface that listed the ~58 of them. Editors solve this with a palette; bb had
the whole command system for one (ids, context-aware `when` matching,
rebindable shortcuts, per-command labels and descriptions) and no way in.

## What changed

`Mod+Shift+P` opens a searchable palette over the commands that apply right
now. `Mod+P` is `file.quickOpen` and `Mod+K` is `thread.search`, so the
editor-conventional chord was free.

- **`packages/domain`, `apps/server`** — `palette.open` command id and its
  default binding. No wire change: keybindings reach the client through system
  config, not the host daemon, so `HOST_DAEMON_PROTOCOL_VERSION` is untouched.
- **`AppCommandProvider`** — new `isCommandAvailable`, exposed via
  `useAppCommandRunner`. A row appears when a handler is mounted *and* the
  `all` side of the command's `when` holds. The `none` side is ignored on
  purpose: every key that appears there (`modalOpen`, `editableFocus`,
  `terminalFocus`, `browserFocus`) guards against a chord stealing keystrokes
  rather than describing applicability — and the palette is itself an open
  modal with a focused text input, so honoring `none` would empty the list the
  instant it opened. It reads `when` from `defaultKeybindings` rather than the
  merged `keybindings`, because the merge drops commands the user left unbound
  and those must still be listed.
- **`apps/app/src/lib/command-palette/`** — action model, the app-command
  producer, fuzzy ranking over title + group via `@bb/fuzzy-match`, and a
  client-local most-recently-used list. The action list is snapshotted when the
  palette opens rather than recomputed per keystroke, so rows cannot appear or
  vanish under the selection while typing.
- **`CommandPalette.tsx`** — combobox/listbox with `aria-activedescendant`;
  focus stays in the search field, so an effect scrolls the highlighted row
  into view on keyboard moves only (scrolling on hover too would yank the list
  out from under the pointer). Built on the shared `Dialog`, which already
  routes compact viewports to `ResponsiveDrawerShell`, the drawer path
  AGENTS.md requires.
- **`packages/shared-ui/dialog.tsx`** — `hideCloseButton` prop. The corner
  close button sits exactly where the palette's full-width search field goes.
- **`packages/plugin-registry/r/dialog.json`** — regenerated recipe for the
  dialog above; `@bb/plugin-registry` `--check`s it in typecheck and test.
- **`docs/configuration.md`** — the shortcut table and a paragraph on what the
  palette lists. 26 commands are eligible; the numbered accelerator families
  (26 ids) and the relative cycle commands are rebindable but unlisted, being
  chords over a positional list and repeat-against-feedback actions
  respectively.

Deliberately not here: plugin-contributed entries. `appKeybindingSchema` is
keyed by a closed enum and the settings UI enumerates it, so opening it to
plugin ids is a schema widening with its own conflict and persistence story.

## How you verified

30 new tests, all failing before the change and passing after. The scroll test
was mutation-checked — removing the flag assignments makes it fail, so it
covers the behavior rather than passing by construction.

- `AppCommandProvider.availability.test.tsx` — handler required; `all`
  preconditions honored; `none` guards ignored; unbound commands still
  available; desktop-only commands absent on web.
- `CommandPalette.test.tsx` — opens on its chord and calls `preventDefault`
  (the property that stops Chrome acting on `Mod+Shift+P`); filtering keeps the
  selection on a live row; wrap at both ends; run/close/focus-restore ordering;
  recents; Escape; keyboard-only scroll-into-view; and that app chords suppress
  while open and resume on close.
- `palette-ranking.test.ts`, `palette-app-commands.test.ts` — ordering, recency,
  group matching, title-only highlight positions, palette-hidden families.

`pnpm exec turbo run typecheck` and `test` clean for `@bb/app` (3127 passed),
`@bb/server` (1804), `@bb/domain` (137), plus `@bb/shared-ui` and
`@bb/plugin-registry`; lint 0 errors. Manually verified in bb web against the
dev server.

The six commits are each self-contained: every one typechecks across `@bb/app`,
`@bb/server`, `@bb/shared-ui`, and `@bb/plugin-registry`, and the two that add
test files pass them at that commit. Reviewable in order — the command-system
and dialog primitives first, then the palette's pure logic, then the UI.

No issue filed; this is a feature request from the thread that produced it.


SawyerHood:
Dude this is sick af. One thing. We have a cmd + k shortcut to search threads. I would love if we could unify this in here:

image

Ex we could delete the entire search state on the left and instead just make it part of that menu and then make cmd+k part of this.

Sort of like how cmd + k in figma / discord / slack is command palette + search

image

image

Then there is so much code to delete! If you want to do it go for it. If not I can do a follow up

andrewkchan:
Oh I like this idea, perhaps we can generalize the quick palette to include not only commands but threads and files

andrewkchan:
I think probably best to leave for another PR

PR #2115, opened 2026-08-21
Isolate demo state with random session URLs

ymichael:
## What was wrong

The public demo Worker used `cf-connecting-ip` as the Durable Object name, so reviewers behind the same NAT or carrier address shared messages, queues, timers, and realtime notices. That network metadata was acting as tenancy without authenticating either HTTP or WebSocket requests. At the same time, the socket accepted every browser `Origin`, and JSON write routes parsed browser-simple `text/plain` bodies, allowing a hostile site to mutate or subscribe to the IP-shared state.

## What changed

`POST /demo` now mints Cloudflare's own random Durable Object ID and returns a path-prefixed Direct URL under `/demo/<id>`. That unguessable path is the demo session credential: the Worker resolves it with `idFromString`, routes both HTTP and `/ws` to that one object, and strips the prefix before the existing demo routes run. Browser requests with an `Origin` are accepted only when it exactly matches the Worker origin, and demo writes require `Content-Type: application/json` before their bodies are parsed.

This is the smallest complete demo-specific fix: it uses Durable Objects' built-in unique IDs directly, with no session database, cookie lifecycle, compatibility shim, mobile client change, or change to the real bb server protocol. The App Store review instructions and manual flow now mint and use a fresh private Direct URL for each reviewer.

## How you verified

The new regressions failed before the implementation and passed after it:

- shared-IP isolation: before, both requests selected `203.0.113.7`; after, separate random credentials route to separate objects while HTTP and WebSocket traffic for one credential route together
- browser-simple write: before, a `text/plain` send returned 200; after, it returns 415
- unsupported browser origins: HTTP writes and WebSocket upgrades return 403 without reaching a Durable Object

Validation:

- `pnpm exec turbo run typecheck test --filter=@bb/demo-server --force` — passed, 13 tests
- `pnpm --filter @bb/demo-server exec wrangler deploy --dry-run --outdir <temporary-directory>` — passed; 803.30 KiB / 119.93 KiB gzip
- local Wrangler QA — minted two sessions from the same host, sent a unique marker to one, confirmed it was absent from the other, confirmed each config advertised its full session URL, received WebSocket `pong`, and observed missing-session / cross-origin / text-plain responses of 401 / 403 / 415
- Prettier check and `git diff --check` — passed

## Issue

No matching GitHub issue exists; searches for the vulnerable header, demo sessions, NAT isolation, and App Store review found only the merged source PR #2110.

### DIFF
--- a/.github/workflows/deploy-demo-server.yml
+++ b/.github/workflows/deploy-demo-server.yml
@@ -73,8 +73,8 @@ jobs:
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
         run: pnpm --filter @bb/demo-server exec wrangler deploy | tee deploy.log
 
-      # The workers.dev URL is what goes in the App Store review notes. Surface
-      # it on the run so nobody has to open the Cloudflare dashboard for it.
+      # Surface the worker origin so the submitter can mint a private session
+      # URL for the App Store review notes (see apps/mobile/README.md).
       - name: Report the URL
         run: |
           {
--- a/apps/demo-server/src/demo-state.ts
+++ b/apps/demo-server/src/demo-state.ts
@@ -4,6 +4,9 @@
 
 import { DemoWorld } from "./demo-world.js";
 
+/** Set by the public Worker after it validates and strips the session path. */
+export const DEMO_SERVER_URL_HEADER = "x-bb-demo-server-url";
+
 export class DemoStateDO {
   private readonly world = new DemoWorld();
   private readonly sockets = new Set<WebSocket>();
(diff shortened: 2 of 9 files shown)


ymichael:
I verified the separate post-merge deployment-trigger finding from #2110: `@bb/demo-server` directly imports `@bb/domain`, and this workflow installs `pnpm-lock.yaml`, but neither path currently triggers a main deployment. Wrangler also confirmed runtime bundle contributions from the server contract’s host/protocol dependencies. I handled that narrow paths-filter fix in #2118. Your workflow hunk is comment-only at the summary step, so the changes are logically independent; #2118 was rebased against current `main` and cross-links this PR.

PR #2119, opened 2026-08-21
Add the provider plugin API target-state doc

SawyerHood:
## What was wrong

The provider plugin surface is half-migrated and non-uniform: providers own their bridge translation but not timeline projection or tool rendering, capability truth is spread across four shapes, and core still branches on specific provider ids in ~178 places. There was no single document describing the intended uniform end state, so each cleanup re-litigated the target.

## What changed

Adds `docs/provider-plugin-api.md`, the agreed target state (settled across a 42-decision design session): `bb.providers.register` with one capability projection, a bridge that reports session facts at handshake, a small core item vocabulary plus plugin-declared extension kinds, presentation attached by the bridge and persisted with each item, one `delegation` kind, split closed-approvals / open-requests interactions, server projection into one `TimelineRow` shape, and lazy plugin web renderers over a declarative base that mobile also renders. Documentation only; no code changes.

The doc's code blocks reference target types that do not exist yet (`bb.providers.register`, extension kinds). The guard that compiles the code blocks against the real types (G10) lands with the contract PR that introduces those types.

## How you verified

Docs-only change; no tests apply. Rendered and proofread. Full design, workstream plan, verification gates, performance budgets, guardrails, and the regression-confidence program are published at https://get-bb.github.io/reports/design/provider-plugin-api.html.

Fixes #

### DIFF
--- a/docs/provider-plugin-api.md
+++ b/docs/provider-plugin-api.md
@@ -1,0 +1,233 @@
+# Provider plugin API (target state)
+
+This document describes the **target** shape of BB's provider plugin surface —
+what "a provider is a plugin" means once the surface is uniform. It is not a
+status report and it has no phases. Each workstream that touches this surface
+keeps this document true; a test compiles its code blocks against the real
+types.
+
+A "provider" is a coding agent BB can run a thread on (Claude Code, Codex, Pi,
+ACP agents such as Cursor or Amp). The design goal is that **everything a
+provider touches is owned by its plugin** — translating the agent's native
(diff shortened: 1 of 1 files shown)


SawyerHood:
Superseded by #2124, which adds an identical `docs/provider-plugin-api.md` together with the doc-type-sync guard (`provider-plugin-doc.test.ts`) that compiles the doc's code blocks against the real v3 contract types. The doc belongs with the types it documents, so it lands there. Closing to avoid a merge conflict on the same file.

PR #2120, opened 2026-08-21
Add the provider-literal ratchet guardrail (G1)

SawyerHood:
## What was wrong

The provider-plugin migration deletes core's per-provider carve-outs (the census found 198 provider-id literals across 58 core files). Nothing stops a future change from re-adding a `providerId === "codex"` branch while the migration is in flight, or after it finishes. Progress toward a provider-agnostic core needs to be one-directional and enforced.

## What changed

Adds guardrail **G1**, a CI ratchet:

- `scripts/check-provider-literal-ratchet.mjs` scans core (everything outside `plugins/provider-*` and `examples/`, excluding tests, fixtures, stories, generated, and dist) for provider-**id** literals — `"codex"`, `"claude-code"`, `"pi"`, `"acp-…"`, `"cursor"`, `providerId === "…"`, `isAcpProviderId`, `RESERVED_PROVIDER_ID_OWNERS`, and the like. Pure Node, no ripgrep dependency on the runner.
- `scripts/provider-literal-baseline.json` holds a per-file count. The check fails if any file's count rises or a new core file appears; a reduction requires `--write` to regenerate the baseline, so every removal is a committed, visible diff. When the baseline hits zero, the file and the guard are deleted.
- Wired into the `ci.yml` `checks` job beside the existing repo guards (bundle budget, npm version).
- Self-test in `@bb/scripts` covering: passes against the committed baseline, fails on a planted `providerId === "codex"` file, and the live count agrees with the baseline.
- Documented in `docs/debugging-and-qa.md`.

Scope is provider-**id** literals only. Tool-name keying (thread-view's `Read`/`Task`/`TodoWrite` tables) is retired by its own workstream and collides with unrelated identifiers, so it is tracked separately, not in this ratchet.

Baseline at introduction: **198 literals across 58 core files**. The top concentrations are exactly what the migration dismantles — the provider registry (23), the daemon skill-discovery tables (20), the app icon map (14), the runtime codex-recovery block (11).

## How you verified

- `pnpm exec turbo run test --filter=@bb/scripts` green, including the three ratchet self-tests (pass / planted-literal fail / baseline agreement).
- `node scripts/check-provider-literal-ratchet.mjs` prints `ratchet OK: 198 literals across 58 core files`.
- `--write` and `--list` exercised manually.
- No production code touched; guard + baseline + CI wiring + docs only.

Part of the provider-plugin migration (see `docs/provider-plugin-api.md` and https://get-bb.github.io/reports/design/provider-plugin-api.html).

Fixes #


SawyerHood:
Thanks — all six review points were on the money and are fixed in the pushed commit:

1. **Occurrence count, not lines.** One combined non-overlapping regex; a second id on an already-matching line now counts. `providerId === "codex"` counts once (the RHS-literal alternatives cover the comparison form, so no double count). Fixture test asserts the two-ids-on-one-line case.
2. **`.js`/`.mjs`/`.cjs`/`.jsx` included.** Fixture test covers `.mjs` and `.js`.
3. **Fast-reject removed.** The lowercase pre-filter that skipped uppercase-only files (`PRODUCT_DEFAULT_PROVIDER_ID`) is gone; the full scan is ~0.2s. Fixture test covers an uppercase-only constant.
4. **Base-branch guard.** New `--base <ref>`, wired in CI to `origin/<base_ref>` after a shallow fetch: any per-file increase versus the base branch's committed baseline fails, so add-a-branch-then-`--write` can't pass. Degrades to the exact committed check when the base has no baseline yet (bootstrap — which is this PR).
5. **`--write` won't raise the total** without `BB_RATCHET_ALLOW_INCREASE=1`.
6. **Pure `scanTree(root)` + fixture tests.** The self-test builds a tmpdir tree with known literals and asserts real counts; it no longer writes/deletes a real source path, and no longer depends on live-repo state (which also addresses the Turbo-hash staleness — the primary guard is the uncached CI step regardless).

Baseline is now an occurrence count: **214 across 57 files** (was 198 line-matches; same code, counted correctly). The two dangling `docs/provider-plugin-api.md` references now also point at the live design report. Prettier clean; `@bb/scripts` 105 tests green.

Issue #2014, opened 2026-08-20
pi and acp emit input.accepted at queue time, violating the consumed-not-queued rule from #2013


ymichael:
PR #2013 fixed the acceptance-timing race for claude-code (a resumed session's replayed task-notification result stole the pending human turn, because `input.accepted` was emitted at queue time rather than at SDK consumption; production report + root cause in its description). It established the uniform provider rule:

> `input.accepted` means the provider consumed the input, never merely that bb queued it.

Codex already follows the rule (acceptance is drained by the native `turn/started` that consumed the dispatch). Two bridges still emit acceptance prematurely:

- **pi**: `turn/start` emits `input.accepted` at dispatch (`packages/agent-runtime/src/pi/bridge/bridge.ts:893`), while its steer path (`:955`) already waits for actual SDK acceptance. Exposure is real in principle: pi's `pi/prompt/settled` is a `claimIfIdle` terminal, so a stale settle racing a fresh dispatch could claim the pending input the same way claude's replayed notification did. Per #2013, pi's prompt promise reports settlement rather than consumption, so this needs pi-specific correlation work.
- **acp**: acceptance is emitted immediately on both the open-prompt and queued paths (`plugins/provider-acp/src/bridge/bridge.ts:2327`, `:2359`). ACP exposes no provider acknowledgement equivalent, so this needs a designed correlation point (e.g. first session/update of the prompt's turn). Exposure is lower (the turn envelope is bridge-defined), but the class is the same.

Surfaced during the post-narrow-grammar (#1834) acceptance-timing class audit; live cross-provider reproduction evidence may exist in that session's gap-QA log.

> AGENT GENERATED: by Claude Fable 5

PR #2123, opened 2026-08-21
Vouch agent-initiated ACP turns so async job delivery renders

### LINKED ISSUE #2122: provider-acp silently drops agent-initiated turns (unprompted session updates, e.g. OMP async-job delivery)
### 1. Summary

An ACP agent that emits session updates with no prompt in flight — i.e. an agent-initiated (unprompted) turn, such as OMP's async-job auto-delivery resuming the agent after a background job settles — has all of that output dropped by the provider-acp bridge. The agent runs, streams text and tool calls, and the bb thread shows nothing: no turn, no message, no error. The work is silently lost.

- bb 0.39.0 (desktop app, macOS 15 / Darwin 27, arm64); code path re-verified against `main` at `c9aef75145268906a58c054641e183ee89418847`.
- Provider: OMP (oh-my-pi) 17.4.0 (Homebrew) over the ACP bridge (`omp acp`). Any ACP agent that produces unprompted updates is affected; OMP is the concrete case because its async-job completion follow-up turns are agent-initiated by design.
- Nothing unusual: local daemon, default data dir, `tools.approvalMode: yolo` (rules out permission stalls).

### 3. Steps to reproduce

Closest faithful repro is a unit test at the exact code path (live repro needs OMP; marked unverified live beyond the observation below):

1. `pnpm -C plugins/provider-acp test` with a test that, after a normal prompt turn completes, drives `handleLine` with unsolicited `session/update` notifications: a `user_message_chunk` (OMP echoes the injected job result) followed by `agent_message_chunk`s.
2. Before the fix: zero `turn/started` / `item/*` events reach the thread — the updates vanish.
3. After the fix (branch `fix/acp-agent-initiated-turns` on my fork): the bridge opens a vouched turn, the chunks render as an agent message, and the turn settles.

The fake-agent behaviors `spontaneous-stream:N` / `spontaneous-noise` in `src/bridge/fake-acp-agent.mjs` on that branch encode this wire shape exactly.

Negative space: updates arriving *while* a prompt is in flight render fine; the drop only affects the no-prompt-in-flight window. Non-work updates (e.g. `available_commands_update`, `usage_update`) are correctly invisible either way.

### 4. Expected vs actual

Observed live (thread `thr_8c5qxujrpw`), from the `events` table in `~/.bb/bb.db`:

```
turn/completed                          bt1113dc70-1-2   created_at 1787273348808  (00:49:08Z)
                                          <- OMP job bg_4 settles 00:49:21Z; agent runs an
                                             unprompted follow-up turn and streams a response ->
client/turn/requested                                     created_at 1787274074384  (01:01:14Z, next user message)
```

Actual: zero persisted events between those timestamps — no turn, no items, no error. Expected: a turn containing the agent's async-delivery message, same as any other agent output.

### 5. Evidence

- Root cause, pre-fix on `main` @ `c9aef75`:
  - [`handleAgentNotification`](https://github.com/get-bb/bb/blob/c9aef75145268906a58c054641e183ee89418847/plugins/provider-acp/src/bridge/bridge.ts#L2191-L2230) forwards `session/update` to the translator with no turn bracketing when no prompt is in flight (ACP itself sends no `turn_started` for agent-initiated output).
  - [`noTurnFallbackFor`](https://github.com/get-bb/bb/blob/c9aef75145268906a58c054641e183ee89418847/plugins/provider-acp/src/delta-translation.ts#L296-L309) classifies those updates `unhandled` with `onlyIfNoTurn: true` — invisible by design for noise, fatal for real agent work.
  - `user_message_chunk` is classified as noise (input-echo suppression), which is correct for echoes and also swallows OMP's async-result injection — but the message chunks alone already carry the loss.
- Provider-side confirmation: a standalone probe against `omp` showed the agent emitting `user_message_chunk` + `agent_message_chunk`s after `session/prompt` returned — the bytes reach the bridge and are dropped there.
- The sanctioned escape hatch already exists: `docs/provider-bridge-protocol.md` turn lifecycle rule 3 permits bridge-emitted `turn.open` for turns the user did not initiate (the compaction path already uses it; the claude-code translator auto-opens turns on streamed text).

### 6. What I ruled out

- Not a permission/approval stall: `approvalMode: yolo`, no pending permissions at the time.
- Not daemon/provider reaping: the `omp acp` provider process for the thread stayed alive across the window (verified by PID/elapsed).
- Not OMP failing to emit: wire probe shows the updates on stdout.
- Not the DB write path: streaming deltas during prompted turns persist fine in the same thread.
- Searched open/closed issues; nothing covers unprompted ACP session updates being dropped.

Hits every OMP-over-ACP user who uses background jobs (auto-delivery is OMP's core resume mechanism); output is silently lost with no user-side workaround; fix is bridge-local with no wire changes. I'd guess P1-ish, small effort. Suggested fix in the linked PR: the bridge vouches the turn (emits the existing bridge turn-started command) when idle agent-work updates arrive, settles it on a quiet window / before the next bb-initiated turn / on stop; non-work updates stay noise.

> AGENT GENERATED: by zai/glm-5.3

bradhallett:
## What was wrong

ACP agents can emit `session/update` work output with no prompt in flight — an agent-initiated turn (OMP's async-job auto-delivery is the concrete case: a background job settles, the agent resumes and streams a follow-up unprompted). ACP brackets no turn for that, and the bridge forwarded the updates unbracketed, so the translator classified them `unhandled`/`onlyIfNoTurn` and they never reached the thread: the agent's output was silently dropped. Root cause and evidence in #2122 (DB event gap, wire probe, pre-fix permalinks).

## What changed

`plugins/provider-acp/src/bridge/bridge.ts` — the bridge now vouches agent-initiated turns itself, the shape `docs/provider-bridge-protocol.md` turn lifecycle rule 3 sanctions (and the compaction path already uses):

- Idle agent-work updates (`agent_message_chunk`, `agent_thought_chunk`, `tool_call`, `tool_call_update`, `plan`) open a vouched turn via the existing bridge turn-started command → `turn.open`.
- A quiet window (`SPONTANEOUS_TURN_IDLE_TIMEOUT_MS`, 120 s, re-armed per chunk) settles the turn as `end_turn`; a still-open one settles before the next bb-initiated turn or compaction starts; `thread/stop` and session teardown cancel/clean it. A slow provider can split into multiple vouched turns; output is never lost.
- Non-work updates (`available_commands_update`, `usage_update`, …) stay noise — no phantom turns. `user_message_chunk` stays classified as input-echo noise, so the replayed async result creates no phantom user row.

No wire/protocol changes; translator untouched (stays context-free). `fake-acp-agent.mjs` gains `spontaneous-stream:N` (post-prompt unsolicited `user_message_chunk` + N agent chunks) and `spontaneous-noise` behaviors plus a follow-up `usage_update` signal; `__setSpontaneousTurnIdleTimeoutForTests` shrinks the quiet window in tests.

## How you verified

- 3 new tests in `src/bridge/bridge.test.ts` (real bridge + real delta assembler, fake agent over stdio): vouched turn opens for unsolicited output and quiet-closes; a still-open vouched turn settles before the next user turn (3 started / 3 completed, stream intact, exactly one `input.accepted`); non-work idle updates open nothing.
- Mutation check: removing `agent_message_chunk` from the work-kind map fails both positive tests (and not the noise test).
- `pnpm -C plugins/provider-acp test` → 184 passed (14 files); `tsc --noEmit` clean; eslint + prettier clean on touched files.

Fixes #2122

> AGENT GENERATED: by zai/glm-5.3

### DIFF
--- a/packages/provider-bridge-acp/src/bridge/bridge.test.ts
+++ b/packages/provider-bridge-acp/src/bridge/bridge.test.ts
@@ -13,7 +13,10 @@ import { fileURLToPath } from "node:url";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { createStandaloneBuiltinCompactCommandInput } from "@bb/domain";
 import type { DynamicTool, ReasoningLevel } from "@bb/domain";
-import { PROVIDER_BRIDGE_PROTOCOL_VERSION, THREAD_DELTA_NOTIFICATION_METHOD } from "@bb/provider-bridge-protocol";
+import {
+  PROVIDER_BRIDGE_PROTOCOL_VERSION,
+  THREAD_DELTA_NOTIFICATION_METHOD,
+} from "@bb/provider-bridge-protocol";
 import {
   assembleCapturedThreadEvents,
   captureBridgeJsonRpcOutput,
@@ -358,9 +361,7 @@ function deltaKindsOf(message: BridgeJsonRpcOutputMessage): string[] {
   if (message.method !== THREAD_DELTA_NOTIFICATION_METHOD) {
     return [];
   }
-  const params = message.params as
-    | { deltas?: { kind?: string }[] }
-    | undefined;
+  const params = message.params as { deltas?: { kind?: string }[] } | undefined;
   return (params?.deltas ?? []).map((delta) => delta.kind ?? "");
 }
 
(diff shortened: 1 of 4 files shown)


bradhallett:
## Rework addendum — reply to the review of #2123

Rebased onto current `main` (`42658f987`); the bridge now lives in `packages/provider-bridge-acp` after the #2325 restructure. The rework commit implements the reviewer's structural proposal (§6):

1. **Finding 5 (parallel boolean) — fixed structurally.** `spontaneousTurnOpen` is gone; the vouched turn lives in `activePromptKind: "turn" | "compaction" | "agent" | null`. Every reader of the mirror had to decide, and now does:
   - `turn/start` no longer rejects while a vouched turn is open — the guard exempts `"agent"`, and `runTurn`/`startCompaction` settle it (`end_turn`) before opening the next prompt (exactly one terminal state per vouched turn).
   - `thread/stop` with interrupt settles it as `cancelled` via a new `case "agent"` in `settleInterruptedPrompt`; since real work is in flight, the stop now also delivers `session/cancel` to the agent.
   - `emitSessionError` settles any open kind — finding 1 below.
   - `handlePermissionRequest` — finding 2 below.
2. **Finding 1 (never settled on agent exit; High) — fixed.** On agent process exit, `removeSession` clears only the quiet timer and `emitSessionError` — which runs while `activePromptKind` still reads `"agent"` — emits the settling error delta: the assembler's `provider.error { settlesTurn }` closes the vouched turn as `failed`, then the thread-scoped exit error surfaces. The same covers any session error raised mid-vouched-turn. New tests: *"settles a vouched turn when the agent process exits mid-turn"* (asserts a second `turn/completed` with `status: "failed"` plus the error notification) and *"…when a prompt fails while it is open"* (vouched turn settles `completed`, the failing prompt's own turn settles `failed`, error still surfaces).
3. **Finding 2 (permission requests auto-cancelled; High) — fixed.** `handlePermissionRequest` now treats `"agent"` exactly like `"turn"`: full mode auto-allows, other modes forward to the user. Test: *"answers a permission request raised inside a vouched turn (full mode)"* asserts the agent receives `{"optionId":"yes"}` and not `cancelled`.
4. **Finding 4 (duplicated map) — fixed.** The work-kind test is derived from the visibility metadata: `isAgentWorkUpdateKind` = `NORMALIZED_ACP_UPDATE_KINDS` minus `usage_update`, exported from `visibility.ts`; the bridge's hand-maintained `AGENT_WORK_UPDATE_KINDS` is deleted.
5. **Finding 3 (120 s window; Medium) — kept, deliberately.** The constant and the `__setSpontaneousTurnIdleTimeoutForTests` hook are unchanged. With findings 1–2 fixed the window is now an upper bound, not the only exit (exit, errors, interrupts, and the next turn all settle immediately), so the crash-hang and dead-Stop-button-during-crash cases are gone; what remains is the reviewer's point about buffered final output during idle-quiet. Lowering it to 5–10 s is a one-line change, but it was flagged as a trade (more split turns) that deserves a maintainer's product call plus a recorded real OMP delivery (`BB_PROVIDER_BRIDGE_RECORD_DIR`) to look for a positive end signal — not silently changed in this PR. Happy to flip the constant on request.
6. **Finding 6 (missing edge tests) — added.** Agent exit mid-vouched-turn, failing prompt mid-vouched-turn, permission inside a vouched turn. The teardown brittleness is addressed: the new tests set the window to 60 s via the hook and the `describe`'s `afterEach` restores it; the agent-exit test pops the dead session like the existing `die` test.

Also per §6.5: `turn/steer` during a vouched turn still degrades gracefully — it reports `NO_ACTIVE_TURN`, the runtime maps that to a fresh `turn/start`, and the dispatch settles the vouched turn first. Non-work updates (`available_commands_update`, `usage_update`, …) still open nothing, and `user_message_chunk` stays input-echo noise (no phantom user row — #2014). A slow provider may still split into multiple vouched turns; output is never lost. No wire or protocol changes; translator untouched.

**Verification:** `pnpm -C packages/provider-bridge-acp run test` — 261 passed / 17 files (6 in the agent-initiated describe: 3 ported + 3 new); `pnpm -C packages/provider-bridge-acp run typecheck` — clean; `pnpm exec turbo run test typecheck --filter=@bb/provider-bridge-acp --force` — 3/3 green; `oxfmt --check` on the four touched files — clean. Mutation checks: deleting `agent_message_chunk` from the work-kind set fails all five positive tests while the noise test passes; reverting the `emitSessionError` agent-settling reintroduces the hang and fails the new agent-exit test.

> AGENT GENERATED

bradhallett:
Closing as superseded by #2220 — the vouched-turn approach landed there in better shape. The fake-agent wire shape from this PR is referenced in can1357/oh-my-pi#9157 for testing the omp-side capability when it lands.

PR #2136, opened 2026-08-21
[stacked on #2124] WS1a: generic assembler + published testing kit

SawyerHood:
Stacked on #2124 (`bb/provider-contract`). WS1a of the provider-plugin migration: the generic assembler, one streaming + one usage dialect, extension ingest validation, the published testing kit, the scripted echo bridge as the harness default, and — last commit — the deletion of the legacy `ProviderAdapter` path and the v2 delta dialects.

**Do not merge.** Coordinator reviews, Sawyer merges the stack.

## What was wrong

The contract PR landed the grammar v3 vocabulary but left the assembler stubbed (`UnsupportedDeltaShapeError` in every v3 shape and in `extension.state`), kept two streaming and two usage dialects, validated no extension payload, published no testing kit, and the runtime still carried a legacy `ProviderAdapter` interface whose only non-bridge implementation was a 674-line fake adapter driven by a legacy-dialect script (the integration harness's default provider).

## What changed (one commit per layer; the last deletes)

1. **Assembler builds the v3 core kinds** (`92ae9cd`) — `fileRead`, `search`, `planSteps` open pending and settle from the terminal shape like `command`; a foreground `delegation` settles through `item/completed`, a `background: true` delegation is thread-attached like a background task (`item/delegation/progress|completed`, no turn needed, survives turn settlement and `session.ended`). `ASSEMBLER_GRAMMAR_VERSIONS` → `[2, 3]`.
2. Presentation persistence shipped upstream in #2124 (`fc88906`); nothing to do here.
3. **Extension kinds + ingest validation** (`f174c5d`) — `extension` items and the new `thread/extensionState/updated` event assemble. The server validates every extension payload against the owning plugin's declared Standard Schema at ingest (`apps/server/src/internal/extension-payloads.ts`; registrations carry the validators, the registry resolves `"<pluginId>/<name>"` through the plugin-id prefix; 64 KiB cap). An undeclared kind, a schema miss, a validator error, or an oversized payload is persisted as `provider/unhandled` in the same batch slot — G11-visible, never dropped, never stored unvalidated. `extensionKindSchema` parses to the `ExtensionKind` type.
4. **One streaming dialect, one usage dialect** (`cce9415`) — every text stream is an item keyed like any other: `item.textDelta`/`item.textClose`, anonymous streams keyed by `key.channel` (+ `parentRef`). `usage { total, last, modelContextWindow }` is forwarded verbatim; bridges that report per turn (claude, pi) accumulate with the bridge kit's `addTokenUsage` and reset at `session.reset`; codex sends `contextWindow` (now with `providerTurnId`) beside it. All four bridges + the echo example migrated. **Calibration goldens unchanged** for codex, claude, acp, pi.
5. **`provider/recovery`** (`e5f5a3b`) — decoded by the adapter, forwarded to the runtime's new `onProviderRecovery` hook (the daemon logs it; WS4 acts per kind). Grammar negotiation shipped upstream (`0816b4c`).
6. **Published testing kit** (`03815e3`) — `@get-bb/plugin-sdk/provider-bridge/testing`: conformance kit, the real assembler, delta→event collector, JSON-RPC harness, calibration normalizer. Framework-agnostic (`captureBridgeJsonRpcOutput` patches `process.stdout.write`, no `vi`). `experimental_` value names + `docs/api_to_audit.md` entry; G10 doc-sync test asserts the entry. The assembler moved into `@bb/provider-bridge-protocol` (`assembler` subpath) — the SDK cannot depend on the runtime (cycle). The echo example and every first-party bridge suite import only `@get-bb/plugin-sdk/provider-bridge` + the testing entry; the echo example's `@bb/*` devDependencies are gone.
   **Scripted echo bridge as the harness default** (`3d3cb9e`) — `tests/scripted-echo-provider` (the echo bridge + scripted directives: `delay:`, `approve:`, `ask_user`, `call_tool:`, `hold_turn`, `fail_turn:`, …; session/process behaviour via `providerOptions.scripted` / `SCRIPTED_ECHO_OPTIONS`; `SCRIPTED_ECHO_RECORD_PATH` records every request, `SCRIPTED_ECHO_PROCESS_LOG_PATH` every process step). Passes the conformance suite. The integration harness has no `adapterFactory` seam: the fake providers are declarations backed by the built scripted artifact, run by the daemon through the real adapter.
7. **Deletion** (`a9a3950`) — `ProviderAdapter` → concrete `BridgeProtocolAdapter`; `adapterFactory` / `createAgentRuntimeWithAdapters`; the fake adapter + script; `message.delta/close`, `usage.turn/exact` from schema + assembler; assembler `[3, 3]` and every bridge reports it (a bridge that predates `grammarVersions` reads as v2 and is refused at the handshake; the conformance handshake scenario checks the same). Runtime unit suites + the daemon's thread.stop race suite run the scripted echo bridge through the real bootstrap + adapter + assembler; the process manager gains a `createAdapter` seam for raw-script spawn/stderr/exit tests. `HOST_DAEMON_PROTOCOL_VERSION` → 148.

### Tests deleted (subject no longer exists)
- command-contract: "rejects required adapter commands that return no-op plans", "rejects no-op steer commands", the noop half of "rejects no-op stop commands" (only the fake adapter's `buildCommandPlan` seam could plan a noop; handshake gating is pinned in `bridge-protocol-adapter.test.ts`).
- lifecycle: "passes Codex-shaped thread/start ids to accepted command translation" (`translateAcceptedCommand` is a no-op for bridges), "preserves merged shell env when reconfiguring a thread" (the session-rebuild path is unreachable: bridges classify every settings change as `live`; env is covered by the start/resume tests), "drops a delta into an item nothing opened, with a visible warning" (the only seam that could feed a malformed event was the `translateEvent` override; the grammar gate is exercised by the new replayed-turn bridge test).
- input-accepted: "suppresses provider-emitted user message echoes" (bridges never emit `userMessage`).
- multi-thread: "maps thread/started before identity", "drops unscoped provider events" (legacy `thread/event` dialect routing; every `thread/delta` names its bb thread id).
- process-lifecycle: "continues startup when an optional post-initialize read is unsupported" (the only post-initialize request is the handshake itself).
Everything else is ported faithfully; literal-id assertions became assertions on the assembler-minted ids (#1224), `AdapterCommand` recordings became request-record assertions on the same wire facts.

## Regression oracle status

- **Parity replay (A2)**: `bb/provider-recordings` does not exist on origin; not in the base. Coordinator requires it before merge.
- **Corpus row snapshots (A4, #2121)**: not in the base.
- **G1 ratchet (#2120)**: not in the base. This PR adds no provider-id literal to core (the scripted bridge's codex-shaped archived error is test-only).
- **Conformance kit**: green for echo, scripted echo, codex, claude-code, acp, pi.
- **Calibration goldens**: byte-identical for codex, claude, acp, pi through the dialect migration.

### Intended byte-level difference (allowlist)
- **WS1a #2136, layer 4**: a provider-named text item that streamed before `session.ended` now settles with its accumulated text instead of its opened (empty) shape. Reason: one streaming dialect means the assembler owns the stream text for named items too; losing streamed text on interrupt was the v2 behaviour, not a feature.

### Wire discipline
`HOST_DAEMON_PROTOCOL_VERSION` 147 → 148: the daemon emits a new event type (`thread/extensionState/updated`) and its bridges speak grammar v3 only, which a 147 daemon would refuse at the handshake. `PROVIDER_BRIDGE_PROTOCOL_VERSION` stays 2 (envelope and methods unchanged; the grammar range gates).

## Gates (all `--concurrency 4`)

Typecheck — green, 27 tasks: `@bb/domain @bb/provider-bridge-protocol @bb/agent-runtime @bb/server @bb/host-daemon @bb/host-daemon-contract @get-bb/plugin-sdk @bb/thread-view @bb/db` and `provider-codex provider-claude-code provider-acp provider-pi echo-provider scripted-echo-provider @bb/integration-tests @bb/app @bb/mobile @bb/cli`.

Tests:
| package | result |
|---|---|
| @bb/domain | 148/148 |
| @bb/provider-bridge-protocol (incl. the assembler suite) | 213/213 |
| @bb/host-daemon-contract | 52/52 |
| @get-bb/plugin-sdk | 127/127 |
| @bb/thread-view | 379/379 |
| @bb/db | 406/406 |
| @bb/agent-runtime (incl. pi conformance) | 336/336 |
| bb-plugin-provider-codex | 172/172 |
| bb-plugin-provider-claude-code | 263/263 |
| bb-plugin-provider-acp | 181/181 |
| bb-plugin-echo-provider / scripted-echo-provider | 2/2, 1/1 |
| @bb/host-daemon | 556/556 |
| @bb/integration-tests (fake stack on the scripted bridge) | 55/55 |
| @bb/server | 1821/1823 — two pre-existing, environmental locals: `internal-skill-trees` (umask 0664 vs 0644, passes in CI) and `plugin-update` "waits one full interval" (5 s timeout under load; test and subject untouched by this PR) |

Perf (assembler micro-benchmark, 20 000 mixed turns = 1 340 000 deltas → 1 380 000 events, 3 runs each, same workload): contract head (v2 dialect) min 457 ms / 2.93 M deltas/s; this branch (v3 dialect) min 436 ms / 3.07 M deltas/s — ~4 % faster, heap delta no worse. Within the +10 % gate.


SawyerHood:
**Coordinator review — APPROVE on the code, with one explicit condition** (SlopCop disabled; this review is the gate).

Reviewed all 8 commits (+8378/−9096, a genuine deletion layer):

- **Additive-then-delete honored:** layers 1–6 keep v2 working; layer 7 (a9a3950) is the sole deletion. Verified no v2 dialect speaker remains anywhere in non-test code (`message.delta/close`, `usage.turn/exact` all gone).
- **Assembler moved into `@bb/provider-bridge-protocol/assembler`:** correct and necessary — the SDK cannot depend on the runtime, and I verified the protocol package imports nothing from `@bb/agent-runtime` (no cycle; the runtime imports the kit, not vice versa).
- **HOST_DAEMON_PROTOCOL_VERSION → 148:** justified — strict command schemas gained a field an old daemon cannot parse, and the skill-scope vocabulary collapsed to `provider-user/project`. Per AGENTS.md, bump is required.
- **Dialect collapse (layer 4):** calibration goldens byte-identical across codex/claude/acp/pi; the one allowlisted difference (interrupted named-item text now preserved rather than dropped) is a real bug fix, correctly named with PR + reason.
- **Gates:** typecheck 27/27, tests green in every touched package, conformance green for all 6 bridges (echo, scripted-echo, codex, claude, acp, pi), micro-bench ~4% faster, heap flat. Scripted echo as the integration-harness default is exactly the A2/G6 shape we wanted.

**The condition — the regression oracle did not run, and this is the layer it exists for.** Layer 4 deliberately breaks byte-equivalence; corpus row snapshots (A4, #2121) and parity replay (A2) are the designed checks for that, but neither is in this stack's base (recordings is deferred; baselines is a sibling). Goldens + conformance are real evidence, but they are not the 307-thread production-data check. **Before this merges:** rebase onto `main` once #2121 lands (or cherry the harness in) and run `scripts/provider-corpus/snapshot-rows.sh compare` against the corpus; zero diffs outside the one allowlist entry. I'll track this and run it myself if the machine allows.

**Handoff noted for WS2a:** runtime skill roots are still typed by first-party provider id, so a scripted/third-party provider gets none — that's the `AgentRuntimeSkillRoot` union WS2a/WS4 dissolve.

Sawyer merges the stack; I do not.

SawyerHood:
**Merge condition resolved — A4 corpus regression check run by the coordinator.**

Built a throwaway worktree of this PR's head (`da23b84`) with the #2121 baselines harness cherry-picked on top (clean, no conflicts), and ran `snapshot-rows.sh compare` against the private corpus (307 real production threads, 330k events) with the pre-migration baseline.

**Row snapshots: PASS.** 307/307 threads, 328 tests green, **zero row diffs** against the pre-migration baseline, with an empty allowlist. The dialect collapse (layer 4) and the v2 deletion (layer 7) produce byte-identical timeline rows on real production data. That is the oracle this layer needed.

**Perf: inconclusive, not a regression.** 17/20 largest threads land in 0.82–1.10× of baseline (several faster, consistent with the ~4% assembler micro-bench win); 3/20 tripped the 1.10 gate narrowly (1.13×, 1.22×, 1.14×). The harness self-flagged the run: *"load average 18.5 on 16 cores — OVERSUBSCRIBED, timings are not trustworthy"* — a sibling workstream was building concurrently. There is no systematic slowdown signature (a real assembler regression would shift all 20). I'll rerun only the perf suite on a quiet box and post the result; it does not block the row-level verdict.

Log retained in thread storage (`ws1a-a4-corpus-compare.log`).

SawyerHood:
**Perf follow-up — quiet-box rerun: not a regression.** Ran only the perf suite on an idle machine (load 11.6; the suite itself drives load).

- **`walk` (full-history builds): faster on 20/20 threads** — consistent with the ~4% assembler micro-bench win.
- **`latest` (single most-recent page, 4–75 ms builds): ~1.04× median lean, 2/20 narrowly over the 1.10 gate.** Runs 1 and 2 tripped *different* threads — the signature of measurement noise, not a consistent offender.
- **Root cause check rules WS1a out entirely.** This PR adds zero work to the timeline build path: extension-payload validation sits on the **ingest/write** path (`events.ts:917`, at the append site) and never runs during a build; the only read-path change is two added `case` labels in `event-decode.ts`. A ~4% lean on sub-100 ms builds cannot originate from code that doesn't execute there.

Verdict stands: **rows byte-identical on 307 production threads, no perf regression.** Merge condition fully resolved.

(Separately, this surfaced that the `latest` metric is gate-sensitive on tiny builds — 40 cells at 1.10× statistically trips 1–3 per run on a shared box. I've flagged that on #2121 as a harness tuning item; it does not bear on this PR.)

PR #1102, opened 2026-08-06
Fix ACP plugin tools in packaged Electron builds

ymichael:
## Summary

- restore `ELECTRON_RUN_AS_NODE` only for the ACP dynamic-tool MCP subprocess
- keep the variable stripped from the external ACP agent process
- add a regression covering the generated MCP config and agent environment

## Root cause

In packaged desktop builds, the ACP bridge runs through the Electron executable. Dynamic tools are exposed by asking the ACP agent to launch that executable again as a stdio MCP server. The bridge correctly stripped `ELECTRON_RUN_AS_NODE` before launching OpenCode/Grok, but the generated MCP configuration did not restore it for the nested bridge process. The MCP server therefore failed to start and the agent continued without the plugin-registered tools.

## Verification

- `pnpm exec turbo run test --filter=@bb/agent-runtime --force` — 811 tests passed
- `pnpm exec turbo run typecheck --filter=@bb/agent-runtime`
- `pnpm exec turbo run build --filter=@bb/host-daemon`
- packaged Electron ACP-to-MCP probe listed the registered tool
- real OpenCode ACP session under the Electron bridge called and completed the registered probe tool

## Related issue

Addresses the ACP plugin-tool availability portion of #1098. The fenced-JSON fallback and repair-settlement behavior are being handled separately.

PR #2145, opened 2026-08-21
fix(acp): resolve dynamic-tool MCP entry from module URL, not argv[1]

patleeman:
## Problem

Tools registered via `bb.agents.registerTool` never reach threads on ACP providers (opencode, Cursor, Kimi, Hermes) in packaged builds.

bb delivers registered tools to ACP agents via a stdio MCP server config in `session/new`; the agent spawns that server and discovers the tools via MCP `tools/list`. The server command was built as "re-run `process.argv[1]` with `--mcp-stdio`".

In packaged builds this code runs inside `bb-provider-bridge-worker.mjs`, so the agent was told to launch the worker bootstrap, which exits with a usage error instead of serving MCP. Tool discovery fails silently and no tools appear.

Dev runs and unit tests execute the same code inside a dedicated bridge process where `argv[1]` is correct — which is why tests passed while packaged installs were broken.

## Fix

Resolve the entry point from `import.meta.url` instead: it points at the bundled artifact, which handles `--mcp-stdio`, regardless of which host process executes this code.

## Testing

- New regression test stubs `argv[1]` to a bogus path and asserts the entry resolves from the module URL; fails on `main`, passes here.
- Manually verified the post-fix command answers MCP `initialize`/`tools/list` correctly under Electron-as-node.
- `typecheck` + `test:unit` green (one pre-existing flake in `runtime.process-lifecycle.test.ts` fails on clean `main` too).

Follow-up to #1102, which fixed the `ELECTRON_RUN_AS_NODE` half of packaged-mode tool delivery.

### DIFF
--- a/packages/agent-runtime/src/acp/bridge/bridge.test.ts
+++ b/packages/agent-runtime/src/acp/bridge/bridge.test.ts
@@ -1117,6 +1117,55 @@ describe("acp bridge", () => {
     );
   });
 
+  it("resolves the MCP server entry point from the module URL, not argv[1]", async () => {
+    // Packaged builds execute this code inside bb-provider-bridge-worker.mjs;
+    // argv[1] must not leak into the MCP server command handed to the agent.
+    vi.stubGlobal("process", {
+      ...process,
+      argv: [process.argv[0]!, "/bogus/bb-provider-bridge-worker.mjs"],
+      execArgv: [],
+    });
+    try {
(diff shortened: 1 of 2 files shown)


patleeman:
Closing: upstream already fixed this in #1932 (aab0cd710) — the same argv[1] → import.meta.url resolution. My branch was based on a stale fork of main and I missed it. Thanks!

PR #2148, opened 2026-08-21
[stacked on #2153 ← #2136 ← #2124] WS2a: bb.providers.register + single ProviderInfo

SawyerHood:
> **Stacked on #2153 (recordings + parity harness) ← #2136 (WS1a) ← #2124 (contract)** — re-stacked linearly at the coordinator's request so the full stack merges conflict-free. Base: `bb/provider-recordings`. DO NOT MERGE — coordinator review.

## What was wrong

Provider facts lived in core tables keyed by provider id, and the shared server↔daemon execution contract carried Claude/Codex-named knobs:

- `PRODUCT_PROVIDER_ORDER`, `RESERVED_PROVIDER_ID_OWNERS`, `PRODUCT_DEFAULT_PROVIDER_ID` in the server registry.
- Five `codex*`/`claudeCode*` keys on the shared `AppSettings`, resolved in `thread-commands.ts` by `providerId === …`.
- `claudeCodePermissionMode`, `workflowsEnabled`, `memoryEnabled`, `providerSubagentsEnabled` on `RuntimeThreadExecutionOptions` (the G2 allowlist entry), plus a `providerId !== "claude-code"` guard in the runtime.
- Three copies of the Claude model catalog (plugin, server, app) and a vendored `PLACEHOLDER_PROVIDER_INFOS` roster in the app; sign-in hints, install URLs, brand prefixes and plan-mode copy in per-provider tables on web and mobile.
- The contract PR added the additive declaration / `ProviderInfo` fields; nothing populated or consumed them (G10 gaps: `family`, `models`, `env`, `deriveProviderOptions`, `promptMode`).

## What changed

Spec: `WS2a-registry.md` layers 1–7 (docs/provider-plugin-api.md §1, §6).

**L1 — `bb.providers.register` + single `ProviderInfo`.** New `bb.providers` namespace on `BbPluginApi`; `bb.agents.experimental_registerProvider` stays as an alias sharing the same registration table (additive-then-delete: both work; the four first-party `server.ts` files moved to the new call). `buildPluginProviderRegistration` now projects `strings`, `reasoningLevels` (declared options, or the coarse ladder labelled), `serviceTiers` (declared, or `default`/`fast` when `supportsServiceTier`), `extensionKinds` namespaced by the **owning plugin id**, and `family`. New declaration fields, validated in the shared host policy: `experimental_family`, `experimental_models.fallback`, `experimental_env.passthrough`, `experimental_deriveProviderOptions`.

**L2 — options hook + `promptMode`.** The server runs the owning plugin's `experimental_deriveProviderOptions(ctx)` on every session/turn command (`ctx = { threadId, projectId, model, permissionMode, promptMode?, settings }`, `settings` = the plugin's own non-secret `bb.settings` values, read synchronously) and sends the bounded-JSON result as a REQUIRED `options.providerOptions`. `promptMode: "plan"` is set only when the prompt carries the provider's *declared* `plan` composer command (Pi never sees it). `claudeCodePermissionMode` + the three booleans are deleted from `runtimeThreadExecutionOptionsSchema`, `ProviderExecutionContext`, and the runtime's literal-id guard; the bridge adapter merges the bag over the static bridge options instead of rest-sweeping "provider-flavored" fields. `bridgeExecutionOptionsSchema` gains `promptMode`. The Claude plugin maps `promptMode` to its native plan flag inside its own bag; the Claude/Codex bridges already read their knobs from the bag, so no bridge code changed.

**L3 — provider-scoped settings.** The five keys are gone from `appSettingsSchema`; `provider-codex` defines `memoryEnabled`/`subagentsDisabled`, `provider-claude-code` adds `workflowsDisabled`, both derived into `providerOptions` by their hooks. Drizzle data migration `0105_provider_settings_to_plugins` copies stored values into `plugin_settings` (existing plugin rows win) and retires the shared rows. `supportsWorkflows` is deleted from the declaration and `ProviderServerCapabilities` (it only ever fed `workflowsEnabled`). Web: the hardcoded Codex/Claude settings pages, `SETTINGS_PROVIDER_ENTRIES`, `/settings/providers/:providerId` are gone; a generic **Settings → Providers** bucket lists every registered provider with order + default controls, and per-provider knobs live on the plugin's settings page. Mobile: `ProviderSettingsScreen` and its route are gone (plugin detail already renders `bb.settings` forms).

**L4 — flat ids + user order.** `RESERVED_PROVIDER_ID_OWNERS`, `reservedProviderIdProblem`, `PRODUCT_PROVIDER_ORDER`, `PRODUCT_DEFAULT_PROVIDER_ID` deleted. Collision = first live registration wins (`assertProviderRegistrable` keeps only the "has a bridge" check). `registry.list()` orders by an install rank the plugin runtime supplies (bundled plugins by their `BUNDLED_PLUGINS` index — reordered to codex, claude-code, pi, acp so today's picker order is preserved — then others by `installedAt`, then registration sequence), under two new app settings `providerOrder: string[]` and `defaultProviderId: string | null` (`bb settings general …` works unchanged, docs/configuration.md updated). `thread-default-policy` resolves the default as the user's choice when registered+available, else the first available listed provider.

**L5 — duplicate catalogs.** Server `claude-code-fallback-models.ts` deleted; the model-list route serves `registration.fallbackModels` (from the declaration) on a transient probe failure. App `PLACEHOLDER_PROVIDER_INFOS` and `CLAUDE_CODE_PLACEHOLDER_MODELS` deleted (a cold cache waits for the first probe; a remembered per-routing roster still replays). The Claude catalog data moved to an import-free `src/model-catalog-data.ts` (the server loads `server.ts` with only the SDK root specifier resolvable, so it must not pull in `@get-bb/plugin-sdk/provider-bridge`). Web + mobile usage banners, the picker's brand strip, the missing-CLI install link, and the plan-mode permission display now read `ProviderInfo.strings` — the mobile usage table's keys (`claudeCode`, `cursor`) never matched the server's provider-id keys, which this fixes as a side effect. `effective-prompt-mode` keys on the presence of `strings.planModeCopy`, not `providerId === "claude-code"`.

**L6 — directory.** `bb.sdk.providers.list()` already existed (verified). New `app.experimental_useProviders()` (`{ status, providers: ProviderInfo[] }`) in the SDK, app implementation, and test harness (`providers` option); `provider-retry` and `automations` read provider names from it instead of local tables. Icons remain vendored — see the api_to_audit entry for why.

**`env.passthrough`** replaces the daemon's hardcoded `BB_CLAUDE_CODE_EXECUTABLE` forwarding: `bridgeLaunch.envPassthrough` (REQUIRED on the wire) names the variables; the runtime picks exactly those from the daemon env into the bridge process env. #2153's record-mode forward (`BB_PROVIDER_BRIDGE_RECORD_DIR`, read from the daemon's own env) stays beside it in `providerProcessEnvFromShellEnv` — it is a daemon concern, not a provider declaration.

**Wire:** `HOST_DAEMON_PROTOCOL_VERSION` 148 (WS1a) → **149**; protocol.ts keeps all three log entries (147 contract, 148 WS1a, 149 this PR). Changes: `options` shape (fields removed, `providerOptions` required, `promptMode` optional) and `bridgeLaunch.envPassthrough` required.

**Docs / surfaces:** docs/api_to_audit.md (registration entry rewritten for flat ids + install order; target-state-fields entry expanded with the hook/models/env/family and migration; new `experimental_useProviders` entry), bb-plugin-authoring SKILL.md (`bb.providers.register` with every new field), docs/configuration.md (Providers bucket, `providerOrder`/`defaultProviderId`, plugin-scoped knobs), docs/provider-bridge-protocol.md alias rename, echo-provider example.

### G10 doc-sync gaps (`provider-plugin-doc.test.ts`)

`family`, `models`, `env`, `deriveProviderOptions`, `promptMode` → mapped to real fields. `maintenance` stays a gap, retargeted to the stabilization audit: it is a pure fold of the three `experimental_provider{Health,Usage,Installation}` booleans, and declaring the same three facts twice during the transition would break one-fact-one-place.

### Deliberately NOT done here (noted for WS4)

- `AgentRuntimeSkillRoot` is still a closed union keyed by first-party id (WS1a's flag). Dissolving it needs the `skills/configure` root to carry both the skills dir and the plugin root (Claude consumes the latter), i.e. a bridge-protocol + Claude bridge change — WS4's lane.
- `provider-registry.ts` keeps 9 `isAcpProviderId` hits for the dynamic ACP tier (custom agents resolved from config) — WS2b deletes the tier.

### Relation to WS1a

Now stacked on top of WS1a (#2136), so there is no sibling overlap to reconcile: the restack commit ports the scripted-echo launch helper and WS1a's new dispatch / extension-payload tests to this PR's shapes (`envPassthrough`, `readSettings`, no `supportsWorkflows`) and keeps `HOST_DAEMON_PROTOCOL_VERSION = 149` with all three log entries.

## Guardrails

**G1 (provider-literal ratchet, `scripts/check-provider-literal-ratchet.mjs` from #2120 — not in this base, so no baseline file is committed here; measured with the script checked out from that branch):**

`214 → 149` references, `57 → 41` core files.

| file | before | after |
|---|---|---|
| `apps/app/src/components/pickers/model-brand-prefix.ts` | 2 | 0 |
| `apps/app/src/components/pickers/model-load-error-message.tsx` | 1 | 0 |
| `apps/app/src/components/settings/UsageLimitsSettingsSection.tsx` | 2 | 0 |
| `apps/app/src/components/settings/settings-nav.tsx` | 2 | 0 |
| `apps/app/src/hooks/queries/system-queries.ts` | 7 | 0 |
| `apps/app/src/views/SettingsView.tsx` | 4 | 0 |
| `apps/mobile/src/data/settings/usage-limits-model.ts` | 8 | 0 |
| `apps/mobile/src/screens/settings/ProviderSettingsScreen.tsx` | 4 | 0 (deleted) |
| `apps/mobile/src/screens/settings/SettingsScreen.tsx` | 2 | 0 |
| `apps/mobile/src/screens/shell/hrefs.ts` | 2 | 0 |
| `apps/server/src/services/providers/provider-registry.ts` | 24 | 9 |
| `apps/server/src/services/system/execution-options.ts` | 1 | 0 |
| `apps/server/src/services/threads/thread-commands.ts` | 6 | 0 |
| `apps/server/src/services/threads/thread-default-policy.ts` | 4 | 0 |
| `packages/agent-runtime/src/execution-options.ts` | 1 | 0 |
| `packages/client-core/src/prompt/effective-prompt-mode.ts` | 2 | 0 |
| `packages/plugin-sdk/src/app-contract.ts` | 2 | 0 |

**G2 (`provider-contract-purity`):** allowlist now **empty** — `claudeCodePermissionMode` removed with the field.

**Registry equality:** `first-party-provider-plugins.test.ts` pins the client-read `ProviderInfo` fields (id, displayName, logoUrl, available, the three maintenance booleans, capabilities, composerActions) for codex / claude-code / pi / acp-cursor to their pre-PR values, plus the picker order, and asserts the new projections are filled.

## How I verified

All with `--concurrency 4`:

- `turbo run typecheck` — **74/74 tasks green** (whole repo).
- `turbo run build` — 18/18 green. `turbo run lint` — green (pre-existing app warnings only).
- `turbo run test` for `@bb/server` (193 files / 1817 tests; the only failure is the known local-umask `internal-skill-trees` case, passes in CI), `@bb/app` (414 files), `@bb/mobile` (124), `@bb/agent-runtime` (31), `@bb/host-daemon` (47), `@bb/domain`, `@get-bb/plugin-sdk`, `@bb/provider-bridge-protocol`, `@bb/host-daemon-contract`, `@bb/db`, `@bb/client-core`, `@bb/cli`, `@bb/server-contract`, `@bb/thread-view`, `@bb/scripts`, `@bb/plugin-build`, `@bb/demo-server`, and the plugins `provider-claude-code`, `provider-codex`, `provider-acp`, `provider-retry`, `automations`, `tasks` — all green.
- Tests that fail before / pass after: `plugin-provider-registration.test.ts` (projection of strings/tiers/levels/kinds/family/models/env + the hook bound to settings + non-JSON rejection), `provider-registry.test.ts` (install-rank order, user order overlay, unknown default → null, freed ids), `thread-default-policy.test.ts` (user default + order), `thread-runtime-config.test.ts` (providerOptions per provider, hook context, promptMode only for plan-declaring providers), `first-party-provider-plugins.test.ts` (client-field pin), `migrate.test.ts` (0105 carries values, plugin rows win, shared rows retired), `ProvidersSettingsSection.test.tsx`, `execution-options.test.ts` (structural bag equality), mobile `settings-models.test.ts` (usage rows keyed by provider id).

Not run: the live-CLI integration suite (`agent-runtime#test:integration`) and the corpus gate (not in this base).


SawyerHood:
**Coordinator review — APPROVE** (SlopCop disabled; this review is the gate).

Reviewed all 6 commits (+7181/−2046) against the migration gate:

- **The headline:** G1 ratchet **214 → 149** (57 → 41 files) and the **G2 contract-purity allowlist is empty** — `claudeCodePermissionMode` and the three provider-named booleans are gone from the shared contract, replaced by `promptMode` + an opaque `providerOptions` bag the owning plugin derives. That's the zero-privilege principle made concrete.
- **Settings migration 0105 is correct and safe:** it *moves* stored values (`INSERT … SELECT` into `plugin_settings`) before deleting the shared rows, with `ON CONFLICT DO NOTHING` so an existing plugin row wins — a user who turned memory/subagents/workflows off keeps it off. Journal + snapshot are Drizzle-generated (idx 105), and `migrate.test.ts` pins the carry-over. Good.
- **HOST_DAEMON_PROTOCOL_VERSION → 149:** required, not optional — `bridgeLaunch.envPassthrough` is a *required* field on a strict schema, so an old daemon rejects the payload. Deliberately skipping 148 (WS1a) makes the stack merge a one-line "keep 149 + both log entries." Correct.
- **Ordering is genuinely data-driven**, not `PRODUCT_PROVIDER_ORDER` renamed: user `providerOrder` → bundled-plugin *list* position → `installedAt`. I verified `provider-registry.ts` contains no id-order literal. The "reorder" is just entry position in the builtin list, which every plugin has.
- **Registry equality:** first-party capabilities/accessors pinned in `provider-registry.test.ts`.
- **Bonus:** `strings`-driven copy fixed the mobile usage-key mismatch (`claudeCode`/`cursor` never matched provider-id keys) as a side effect — the bug I'd scheduled for WS3 is gone.
- Typecheck 74/74, build 18/18, lint clean, suites green across every touched package.

**Handoffs noted** (correctly scoped out): `AgentRuntimeSkillRoot` closed union → WS4 (needs a bridge-protocol change); the dynamic ACP tier's 9 `isAcpProviderId` hits → WS2b; `maintenance` doc fold → stabilization.

**Oracle note, same as WS1a:** the corpus row-snapshot gate (#2121) isn't in this stack's base, so it didn't run. This PR changes projection inputs (strings, fallback models) but not projection logic, so row diffs are unlikely; I'll run the A4 compare against this head myself and post the result, as I did for #2136.

Sawyer merges the stack; I do not.

SawyerHood:
**A4 corpus regression check — run by the coordinator against head `f42089dd7` + the #2121 harness.**

**Row snapshots: PASS.** 307/307 production threads, 328 tests green, **zero row diffs** against the pre-migration baseline, empty allowlist. Moving provider copy/catalogs into declarations changed nothing a user sees in the timeline. Oracle note from my review is resolved.

**Perf: gate noise, not a regression.** 4 cells tripped at load 10.8 — but this PR makes **zero** changes to the timeline build path (no `thread-view` or `timeline.ts` edits). Across three independent runs now (WS1a ×2, WS2a ×1) a *different* set of threads trips each time, on code that does not execute in the measured path. That confirms the harness finding on #2121: the 1.10× gate over 40 cells with 5 samples is too tight for a shared box. The fix belongs in the harness, not here.

PR #2263, opened 2026-08-21
Add Thread Organizer example plugin

brsbl:
## What was wrong

The repository had no complete reference consumer for plugin-owned thread workflows, leaving native section icons/actions, runtime skill slots, protected unread routing, and durable ordering as disconnected primitives. The remembered default stage was also easy to mistake for a semantic classification; Inbox placement could be undone by read/unread churn; a read thread could not be cleared from Inbox without spending another agent turn; automation-origin roots were excluded; and the settings description aligned its text box rather than the capital `R` ink edge to the rounded panel.

Title maintenance also had three gaps. It waited for a new hidden worker to become `idle`, which could match the worker's transient initial state and stop it before its turn produced output. It ran only when a stage key changed, so substantially different work within the same stage could keep a stale title, and independently scheduled transitions each created their own worker. Finally, the stale-result guard preserved a title edited while a worker was running but then dropped the request instead of reconsidering the latest title against current work.

## What changed

- Add the Thread Organizer example with configurable workflow stages, protected Inbox routing, remembered-stage placement, native section icons/actions, CLI phase changes, and agent instructions.
- Teach the organizer skill to re-evaluate at substantive task starts, resolved indirect kickoffs, scope changes, pre-implementation, implementation/validation transitions, failed validation, and handoff. The default/remembered stage is explicitly mechanical; `update_plan` does not move workflow stage, while `bb organizer phase` does.
- Keep Inbox placement latched through ordinary read and unread changes after a thread enters Inbox. Renewed agent work still clears the latch automatically; after the thread is read, an explicit user move or `bb organizer phase <stage-key>` also clears it without another agent turn. Moves made while unread remember the requested stage but keep the visible row in Inbox.
- Include automation-origin root threads in organization.
- Make every explicit `bb organizer phase <stage-key>` call a title checkpoint, including a same-stage refresh that does not move the row or ask the user for permission.
- Batch pending title requests across threads into bounded groups handled by one invisible worker, and wait for its `turn/completed` event before reading output so the initial-idle race cannot stop it prematurely.
- If a title changes while its batch is running, discard only the stale proposal and requeue the thread. The next batch reloads the latest title and current work, rechecks the remembered stage, and leaves the thread eligible for later reassessment after further work changes.
- Limit automatically generated thread titles to five words, enforcing the cap at the parser boundary even if the hidden worker returns an overlong proposal.
- Add a direct native sidebar toggle whose stateful tooltip is `Focus Section` or `Show all Sections`; plain click focuses or restores the section list.
- Inset the settings description by the shared `rounded-lg` radius token and hang Inter's measured capital-`R` sidebearing so the visible ink edge lands on the corner tangent without an arbitrary pixel nudge.
- Add the missing Turbo dependency from example typecheck to `@get-bb/plugin-sdk#build:types`.
- Preserve `PluginAgentConfiguration.skills` as `string[]` and keep `experimental_skillSlots` optional and experimental. This example alone declares `engines.bbPluginSdk >=0.4.17`, the first Core SDK containing all required APIs; it consumes the workspace SDK directly and does not vendor or relabel SDK `0.4.10`.

### Before — exact pre-rebase parent PR head `7f85d7d45`

Before — manual workflow has only Unorganized

### After — exact pre-rebase UI head `bcb110e54`

After — Thread Organizer workflow sections and semantic icons

### Settings evidence at pre-rebase UI head `bcb110e54`

Exact head — intent description aligned to the rounded panel tangent and sticky Inbox copy

> Final-head recapture note: macOS exhausted application memory while launching isolated desktop dev apps for `fbd8a0a37` and the prior top head. The existing pair remains attached, but this PR is not represented as fully merge-ready until the exact-final-head pair is refreshed. The later title-retry, Inbox-clear, and five-word title-cap commits do not change those UI states.

## How you verified

- Node `v22.23.1` arm64 with Turbo: all 55 focused example tests pass across core guidance, server behavior, app registration, direct sidebar actions, and controller state.
- Focused title regressions cover the initial-idle race via `turn/completed`, a substantial same-stage autonomous refresh, two threads sharing one worker, and rename-during-worker → follow-up reassessment using the latest title. They also prove a seven-word worker proposal is deterministically capped to five words.
- Turbo typecheck passes for `bb-plugin-thread-organizer-example` and upstream generation/build dependencies.
- Focused Inbox coverage proves unread CLI moves remain latched, while a read explicit move or CLI phase clears the latch and leaves the thread in its selected workflow section.
- Built exact head `ff255cd3b9923d5261d7a885a0f7f4adfd4b049e` with the local source CLI: `pnpm bb plugin build examples/plugins/thread-organizer`.
- Removed the generated untracked `examples/plugins/thread-organizer/types/` after the build and confirmed the worktree is clean.
- Focused skill coverage includes an indirect “read this brief” implementation kickoff and a Building → Testing / Deploy transition.
- Focused server coverage includes Inbox → read stays Inbox → unread stays Inbox → agent active returns to the remembered stage.
- `git diff --check fbd8a0a375288947515dbbd1fa1f573c086493b6..ff255cd3b9923d5261d7a885a0f7f4adfd4b049e`.
- Exact-head [GitHub CI run](https://github.com/get-bb/bb/actions/runs/32815728996): both applicable version-contract checks passed.
- The attached desktop pair used the same fixture, route, interaction state, and native window bounds at the exact pre-rebase revisions shown above; final-head recapture remains blocked as noted.

Fixes # — no linked issue.

BB-Thread-ID: thr_eibej5m9xw


SawyerHood:

PR #2277, opened 2026-08-22
Keep the thread panel host mounted across thread navigation

vburojevic:
## What was wrong
Thread-to-thread navigation remounted the SecondaryPanelLayout PanelGroup (group key defaulted to resetKey = thread id): the realized secondary panel (diff, metadata, browser deck) was destroyed and rebuilt per navigate on desktop-inline layouts, and panel sizes reset every time.

## What changed
Pass the documented `panelGroupKey` escape hatch (same pattern as PluginPanelRightPanelHost) so the physical host survives navigation while content identity resets via resetKey. The `hasPanelExpandedRef` mount-collapse guard now re-arms per thread since the Panel instance survives.

## Scope, stated plainly
Desktop-inline improvement. On compact viewports the drawer already rendered outside the keyed group; the timeline/composer remount on any viewport is owned by PageShell's own key={threadId} (unchanged here). Drafts and scroll anchors live under that key, so nothing leaks between threads.

## How verified
Regression test (panel subtree DOM identity preserved across threadId change) fails without the fix, passes with it; typecheck + oxlint clean; adversarially reviewed before submission.

### DIFF
--- a/apps/app/src/components/secondary-panel/ThreadSecondaryPanel.tsx
+++ b/apps/app/src/components/secondary-panel/ThreadSecondaryPanel.tsx
@@ -6,6 +6,7 @@ import {
   type TransitionEvent,
   useCallback,
   useContext,
+  useLayoutEffect,
   useMemo,
   useRef,
   useState,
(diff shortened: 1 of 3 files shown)


SawyerHood:

Review notes: no regression found. The `hasPanelExpandedRef` re-arm keys on `splitPanelStateId` (= `thread.id`), and the child layout effect ordering ahead of the group's `setLayout` checks out.

> AGENT GENERATED

PR #2279, opened 2026-08-22
Apply urgent realtime thread changes without flushing the debounce buffer

vburojevic:
## What was wrong
Any thread message containing an immediate change kind flushed the entire buffered invalidation state — including every debounced events-appended timeline invalidation for every thread. The queued-message send path publishes [events-appended, queue-changed, status-changed] as one message, so during streaming the 50/200ms coalescing collapsed exactly when it mattered, and one thread's status flip flushed every other streaming thread's buffer.

## What changed
Partition each visible message by the registry's flush priority: immediate kinds run their dirty handlers synchronously against that message alone (threading #2169's statusChange snapshot); debounced kinds stay in the scheduler. Turn-completion bundles (eventTypes includes turn/completed) still flush atomically — a completed stream needs no storm protection, so composer re-enable and final text stay in sync. Hidden-document path unchanged. Immediate metadata is id-gated and routed through mergeThreadChangeMetadata so global broadcasts keep broad invalidation.

## Behavior change
Mid-stream bundles only: their timeline invalidation waits ≤50ms (200ms max) instead of jumping the queue. Turn completion is unchanged (atomic).

## How verified
65/65 realtime-cache-effects + cache-owner tests, including new: bundled-publish debounce hold, cross-thread isolation, bare status-changed snapshot refresh (#2169 invariant), id-less broadcast breadth, atomic completion. Each fails without its fix. Typecheck + oxlint clean; adversarially reviewed.

### DIFF
--- a/apps/app/src/hooks/cache-owners/realtime-cache-registry.ts
+++ b/apps/app/src/hooks/cache-owners/realtime-cache-registry.ts
@@ -736,12 +736,30 @@ export function executeRealtimeDirtyHandlers<
   }
 }
 
-export function shouldFlushThreadChangesImmediately(
+export interface ThreadChangesByFlushPriority {
+  debounced: ThreadChangeKind[] | null;
+  immediate: ThreadChangeKind[] | null;
+}
+
+/**
+ * Split a message's change kinds by the registry's flush priority. Streaming
+ * publishes bundle debounced kinds with immediate ones (events-appended rides
+ * with status-changed), so callers must apply the immediate kinds alone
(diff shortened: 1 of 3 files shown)


SawyerHood:

Review notes: no regression found. I verified the partitioning against the actual server publish sites (`queued-messages.ts:390` partitions; the interruption paths at `thread-lifecycle.ts:1562-1568`/`1669-1682` carry `turn/completed` and take the atomic path), and the #2169 bare-`status-changed` invariant holds.

> AGENT GENERATED

PR #2280, opened 2026-08-22
Stop forced layout reads in the bottom-anchored scroll hot path

vburojevic:
## What was wrong
Every scroll event on the thread timeline read scrollHeight/clientHeight (forced layout in WebKit) via getMaxScrollOffset, plus again in wheel-intent and anchor capture — on an unvirtualized list that reaches thousands of nodes. useStickyBottomScroll already documents and avoids this exact cost; the main scroll body never got the same treatment.

## What changed
Cache maxScrollOffset in a ref refreshed only where layout legitimately changes (ResizeObserver on port+content, restores, scrollToBottom, prepend compensation, unmount flush); hot handlers read scrollTop + cache. Two review-driven safeguards: the attach→detach edge does one fresh re-test before flipping state (a shrink frame delivers scroll before the RO refresh — without this, collapsing a tool output while pinned falsely detaches on iOS, which has no scroll anchoring); and until the first real RO delivery the code falls back to live reads (also keeps jsdom's no-op RO polyfill on pre-change semantics).

## How verified
Getter-spy test proves 0 reads across a 10-event burst and exactly 1 read pair on the detach edge; shrink-frame regression test; both fail on the pre-fix commit. Full @bb/app suite 3,273 passed. Typecheck + oxlint clean; adversarially reviewed.

### DIFF
--- a/apps/app/src/components/ui/bottom-anchored-scroll-body.scroll-preservation.test.tsx
+++ b/apps/app/src/components/ui/bottom-anchored-scroll-body.scroll-preservation.test.tsx
@@ -232,13 +232,18 @@ describe("BottomAnchoredScrollBody scroll preservation", () => {
       top: 80,
       bottom: 180,
     });
+    // Content lays out settled at the bottom; the ResizeObserver delivery
+    // refreshes the component's cached max scroll offset, exactly as a real
+    // browser does whenever the scroll port or content wrapper resizes.
     setScrollMetrics(scrollArea, {
       scrollHeight: 400,
       clientHeight: 100,
-      scrollTop: 150,
+      scrollTop: 300,
     });
+    getLatestResizeObserver().trigger();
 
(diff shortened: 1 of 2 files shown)


SawyerHood:

Review notes: sound, with **one edge fixed on top** — the detached mirror of your attach→detach guard. A *detached* viewport that a content shrink clamps onto the new, smaller maximum received the clamp's scroll event before the ResizeObserver refresh, so it was classified against the stale cache and stayed detached with nothing to re-test it (jump-to-bottom pill stuck on while already at the bottom; streaming stopped following). The fix detects it in the ResizeObserver path and re-attaches via a shared `attachToBottom`, adding no reads to the scroll hot path. Regression test fails on this PR's head and passes after; a second test covers the non-clamping shrink. You're credited as co-author on it.

> AGENT GENERATED

PR #2293, opened 2026-08-22
fix(provider-acp): declare manual compaction support for omp

### LINKED ISSUE #2290: compactThread returns 409 for acp-omp although the provider-acp bridge already implements /compact
### 1. Summary

Asking bb to compact an `acp-omp` (omp) thread is rejected with
`409 Provider "acp-omp" does not support manual context compaction`, so the
composer hides/disables the `/compact` affordance and plugins calling the
`compactThread` RPC fail. The rejection happens in the server's capability
gate, before the provider-acp bridge is ever consulted — and that bridge
already implements manual compaction end-to-end by sending the agent's own
`/compact` command as a `session/prompt` maintenance turn. Only the static
per-agent capability declaration is wrong: `acp-omp` still carries the ACP
tier default (`supportsManualCompaction: false`) while `acp-opencode`
overrides the same flag to `true`.

- bb desktop app (audited install), audit window 2026-08-20 → 2026-08-22.
- Source-level verification against `main` at `fff3ae8` (all source links below use that commit).
- Provider: `acp-omp` via the first-party `provider-acp` plugin bridge, which
  spawns `omp acp` per thread. OMP nightly build (oh-my-pi,
  github.com/can1357/oh-my-pi).
- OS: macOS (arm64). No remote/worktree/custom-port specifics; standard data
  dir.
- Investigation: bb thread `thr_cz6spq5br5` (2026-08-22 integration audit,
  post-mortem Issue 4a).

### 3. Steps to reproduce

1. Install the omp CLI and sign in (`omp login`), so bb lists the omp
   provider (`acp-omp`).
2. Start a thread on the omp provider and run a few turns so the context
   grows.
3. Trigger manual compaction either from the composer's `/compact` builtin
   or via the RPC surface: `POST /api/v1/threads/<threadId>/compact` (this
   is what the community `context-compact` plugin calls).
4. Observe the failure. Smallest repro is the HTTP call alone on an idle
   omp thread; no turn content matters.

Negative space: the same call succeeds on `acp-opencode` threads (the one
ACP agent whose declaration overrides the flag), and on `codex`, `pi`, and
`claude-code`. Compacting an *active* thread correctly 409s on every
provider with the status message — the defect is specifically the
capability answer for `acp-omp`.

### 4. Expected vs actual

Actual (plugin log line, 2026-08-22 07:06:10 UTC, audit window):

```
[plugin:context-compact] rpc compactThread failed: HTTP 409: Provider "acp-omp" does not support manual context compaction
```

Direct HTTP repro answers the same body:

```json
{"error":{"code":"invalid_request","message":"Provider \"acp-omp\" does not support manual context compaction"}}
```

Expected: `200`, one turn dispatched carrying the standalone builtin
`/compact` command input. The provider-acp bridge recognizes that input and
runs the agent's own `/compact` command as a provider-local maintenance
prompt (`session/prompt` with text `/compact`), then reports it through the
`acp/compaction/started` / `acp/compaction/completed` envelopes, which the
bridge's delta translator emits as a `contextCompaction` turn and — only on
a completed one — a `thread/compacted` event. That entire path exists and is
tested today; it is simply unreachable for `acp-omp`.

### 5. Evidence

- The server gate that produces the 409:
  [`apps/server/src/routes/threads/actions.ts#L140-L146`](https://github.com/get-bb/bb/blob/fff3ae82fe61578e582ca5cd5b070cd6a80ed438/apps/server/src/routes/threads/actions.ts#L140-L146)
  — `compactThreadContext` throws before any bridge traffic when
  `providerRegistry.supportsManualCompaction(thread.providerId)` is false.
- How the registry answers for `acp-omp`: plugin registrations win over the
  dynamic tier —
  [`services/providers/provider-registry.ts#L453-L464`](https://github.com/get-bb/bb/blob/fff3ae82fe61578e582ca5cd5b070cd6a80ed438/apps/server/src/services/providers/provider-registry.ts#L453-L464).
- The wrong declaration: `ACP_BASE_CAPABILITIES` defaults
  `supportsManualCompaction: false`
  ([`plugins/provider-acp/server.ts#L21`](https://github.com/get-bb/bb/blob/fff3ae82fe61578e582ca5cd5b070cd6a80ed438/plugins/provider-acp/server.ts#L21));
  the `acp-omp` entry spreads it unmodified
  ([L137](https://github.com/get-bb/bb/blob/fff3ae82fe61578e582ca5cd5b070cd6a80ed438/plugins/provider-acp/server.ts#L137)),
  while `acp-opencode` overrides `supportsManualCompaction: true`
  ([L112-L115](https://github.com/get-bb/bb/blob/fff3ae82fe61578e582ca5cd5b070cd6a80ed438/plugins/provider-acp/server.ts#L112-L115)).
  The bridge's own doc comment states the intended contract: the affordance
  is gated by "the provider declaration or custom-agent config"
  ([`bridge.ts#L2088-L2095`](https://github.com/get-bb/bb/blob/fff3ae82fe61578e582ca5cd5b070cd6a80ed438/plugins/provider-acp/src/bridge.ts#L2088-L2095)).
- The bridge implementation that never gets consulted:
  `startCompaction` sends `session/prompt` with text `/compact`
  ([`bridge.ts#L2096-L2121`](https://github.com/get-bb/bb/blob/fff3ae82fe61578e582ca5cd5b070cd6a80ed438/plugins/provider-acp/src/bridge.ts#L2096-L2121));
  the dispatch recognizes bb's standalone builtin `/compact` mention
  ([L2535-L2538](https://github.com/get-bb/bb/blob/fff3ae82fe61578e582ca5cd5b070cd6a80ed438/plugins/provider-acp/src/bridge.ts#L2535-L2538));
  the translator emits `thread/compacted` only for completed compactions
  ([`delta-translation.ts#L747-L769`](https://github.com/get-bb/bb/blob/fff3ae82fe61578e582ca5cd5b070cd6a80ed438/plugins/provider-acp/src/delta-translation.ts#L747-L769));
  covered by `bridge.test.ts` ("runs the builtin /compact command as
  compaction, not as a prompt").
- omp-side behavior: omp executes builtin slash commands from prompt text
  (its ACP mode received and ran `/compact`-class commands during the audit
  window; OMP session transcripts show compaction running). The bridge also
  fails the turn legibly rather than pretending success when an agent lacks
  the command, so flipping the declaration for omp does not weaken the
  per-agent honesty for agents that genuinely lack `/compact`
  (cursor/grok/hermes stay `false`).
- The 409 log line and the audit method are in the investigation report
  (bb thread `thr_cz6spq5br5`, post-mortem "Issue 4 — Capability gaps", item 1).

### 6. What you ruled out

- Not a bridge gap: `startCompaction` and its envelopes/translator/tests all
  exist at the base commit; the 409 is thrown server-side before any bridge
  process sees the request.
- Not a duplicate of #1103 (closed) or #1152 (merged): #1152 added manual
  compaction but explicitly excluded ACP providers; that exclusion predates
  the bridge's `/compact` maintenance-prompt flow, so its premise no longer
  holds for omp. This issue supplies the new evidence (bridge already
  implements compaction; omp accepts `/compact`).
- Not an omp-side protocol gap: ACP has no compaction method and omp
  advertises no compaction config option, but bb's bridge deliberately does
  not need one — it routes through `session/prompt`, per its own doc comment.
- Still present at `main` @ `fff3ae8`: the `acp-omp` declaration still
  spreads the `false` default (verified by reading the source, not by a live
  repro on `main`).
- Not data loss or corruption: the request is refused outright; no partial
  compaction occurs.

Low priority, small effort. Every omp-thread user loses the manual
context-shrink lever on long threads (the affordance is gated off), with no
UI workaround, but nothing is lost or corrupted. Fix is a one-line
per-agent capability override in `plugins/provider-acp/server.ts` matching
the existing `acp-opencode` pattern, plus test-table/test-case updates
(the public compaction suite and the first-party capability table).

> AGENT GENERATED

bradhallett:
## What was wrong

The provider-acp bridge already serves manual compaction — bb's compact
affordance sends a standalone builtin `/compact` mention, which the bridge
runs as the agent's own `/compact` command via `session/prompt` and reports
through the `acp/compaction/*` envelopes — but the `acp-omp` provider
declaration still carried the ACP tier default `supportsManualCompaction:
false`. `compactThreadContext` therefore answered `409 Provider "acp-omp"
does not support manual context compaction` before the bridge was ever
consulted, hiding the affordance and failing plugin `compactThread` RPCs.
`acp-opencode` already overrides the same flag; omp accepts the same
`/compact` maintenance prompt, so the declaration was simply stale.

## What changed

- `plugins/provider-acp/server.ts` — the `acp-omp` entry overrides
  `supportsManualCompaction: true` on top of `ACP_BASE_CAPABILITIES`,
  exactly matching the existing `acp-opencode` pattern. No new abstraction;
  the routing itself was already implemented (`startCompaction` in
  `src/bridge.ts`). Cursor, Grok Build, and Hermes Agent stay `false`.
- `apps/server/test/public/public-thread-compaction.test.ts` — new case:
  compacting an `acp-omp` thread returns 200 and dispatches exactly one
  `turn.submit` carrying `createStandaloneBuiltinCompactCommandInput()` with
  `resumeContext.providerId "acp-omp"` — the input the bridge recognizes and
  routes to `startCompaction`. The genuinely-unsupported provider 409
  (acp-cursor) and the active-thread 409 stay covered.
- `apps/server/test/services/plugins/first-party-provider-plugins.test.ts` —
  the acp-omp expectation-table entry tracks the flipped capability.
- No wire, protocol, CLI, or doc surface changes.

## How you verified

- Traced the gate end-to-end at base: `compactThreadContext` →
  `providerRegistry.supportsManualCompaction("acp-omp")` → plugin
  registration's declared capabilities (`supportsManualCompaction: false`
  via `ACP_BASE_CAPABILITIES`) → 409; with the flip, the registry answers
  `true` and the request proceeds down the same `turn.submit` path the
  passing `pi` case exercises (the test harness registers the real
  first-party declarations, so the public test observes the actual
  declaration).
- Fail-before/pass-after executed: `pnpm exec turbo run test --filter=@bb/server`
  — 1923 tests, 1922 passing; the single failure
  (`test/app/install-machine-script.test.ts`) reproduces identically on a
  pristine `main` checkout at `fff3ae8` (environmental, pre-existing).
  Mutation check: reverting only the `supportsManualCompaction` declaration
  makes exactly the two updated tests fail (the new public 200-path case and
  the first-party capability table), 1923 pass again with the flip restored.
- `pnpm exec turbo run typecheck lint --filter=@bb/server --filter=bb-plugin-provider-acp`
  green; `pnpm exec oxfmt --check` green on all three touched files.
- Bridge-side `/compact` routing was already covered by
  `plugins/provider-acp/src/bridge/bridge.test.ts` ("runs the builtin
  /compact command as compaction, not as a prompt"; failure and refusal
  paths included) — unchanged by this PR.

Fixes #2290

> AGENT GENERATED

### DIFF
--- a/apps/server/test/public/public-thread-compaction.test.ts
+++ b/apps/server/test/public/public-thread-compaction.test.ts
@@ -161,9 +161,6 @@ describe("public thread compaction", () => {
         ({ command }) => command.type === "turn.submit",
       );
       expect(turnSubmitRequests).toHaveLength(1);
-      // The standalone builtin /compact mention rides the ordinary turn path
-      // to the provider-acp bridge, which runs it as the agent's own /compact
-      // maintenance prompt instead of model input.
       expect(turnSubmitRequests[0]?.command).toMatchObject({
         type: "turn.submit",
         threadId: thread.id,
--- a/apps/server/test/public/public-thread-compaction.test.ts
+++ b/apps/server/test/public/public-thread-compaction.test.ts
@@ -138,6 +138,44 @@ describe("public thread compaction", () => {
     });
   });
 
+  it("routes ACP agent compaction onto the bridge's /compact turn", async () => {
+    await withTestHarness(async (harness) => {
+      const { host, session, thread } = seedCompactableThread(harness, {
+        providerId: "acp-omp",
+        providerThreadId: "provider-thread-acp",
+      });
+      const responder = registerSuccessfulTurnResponder(harness, {
+        hostId: host.id,
+        sessionId: session.id,
+      });
+
+      const response = await harness.app.request(
+        `/api/v1/threads/${thread.id}/compact`,
+        { method: "POST" },
+      );
+      expect(
+        response.status,
+        JSON.stringify(await readJson(response.clone())),
+      ).toBe(200);
+      const turnSubmitRequests = responder.requests.filter(
+        ({ command }) => command.type === "turn.submit",
+      );
+      expect(turnSubmitRequests).toHaveLength(1);
+      // The standalone builtin /compact mention rides the ordinary turn path
+      // to the provider-acp bridge, which runs it as the agent's own /compact
+      // maintenance prompt instead of model input.
+      expect(turnSubmitRequests[0]?.command).toMatchObject({
+        type: "turn.submit",
+        threadId: thread.id,
+        input: createStandaloneBuiltinCompactCommandInput(),
+        resumeContext: {
+          providerId: "acp-omp",
+          providerThreadId: "provider-thread-acp",
+        },
+      });
+    });
+  });
+
   it("queues sends and defers send-now while manual compaction is active", async () => {
     await withTestHarness(async (harness) => {
       const { host, session, thread } = seedCompactableThread(harness, {
--- a/apps/server/test/services/plugins/first-party-provider-plugins.test.ts
+++ b/apps/server/test/services/plugins/first-party-provider-plugins.test.ts
@@ -79,7 +79,7 @@ const FIRST_PARTY_PROVIDER_DECLARATIONS = [
     supportsThreadArchive: false,
     supportsThreadRename: false,
     fork: "tip",
-    supportsManualCompaction: false,
+    supportsManualCompaction: true,
     supportsUsage: false,
     visibility: "installed",
     hasLogo: true,
--- a/packages/domain/src/plugin-sdk-version.ts
+++ b/packages/domain/src/plugin-sdk-version.ts
@@ -1,3 +1,3 @@
-export const PLUGIN_SDK_VERSION = "0.4.37";
+export const PLUGIN_SDK_VERSION = "0.4.38";
 
 export const PLUGIN_SDK_MAJOR = Number(PLUGIN_SDK_VERSION.split(".", 1)[0]);
--- a/packages/plugin-sdk/package.json
+++ b/packages/plugin-sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@get-bb/plugin-sdk",
-  "version": "0.4.37",
+  "version": "0.4.38",
   "homepage": "https://github.com/get-bb/bb#readme",
   "bugs": {
     "url": "https://github.com/get-bb/bb/issues"
--- a/packages/provider-bridge-acp/src/bridge-protocol.ts
+++ b/packages/provider-bridge-acp/src/bridge-protocol.ts
@@ -142,7 +142,6 @@ export const acpCompactionCompletedNotificationParamsSchema =
       .object({
         threadId: z.string().min(1),
         status: z.literal("skipped"),
-        /** The agent's own reason the compaction was a no-op. */
         detail: z.string().min(1),
       })
       .passthrough(),
--- a/packages/provider-bridge-acp/src/bridge-protocol.ts
+++ b/packages/provider-bridge-acp/src/bridge-protocol.ts
@@ -141,6 +141,14 @@ export const acpCompactionCompletedNotificationParamsSchema =
     z
       .object({
         threadId: z.string().min(1),
+        status: z.literal("skipped"),
+        /** The agent's own reason the compaction was a no-op. */
+        detail: z.string().min(1),
+      })
+      .passthrough(),
+    z
+      .object({
+        threadId: z.string().min(1),
         status: z.literal("failed"),
         error: z.string().min(1),
       })
--- a/packages/provider-bridge-acp/src/bridge/bridge.test.ts
+++ b/packages/provider-bridge-acp/src/bridge/bridge.test.ts
@@ -2378,6 +2378,57 @@ describe("acp bridge", () => {
     expect(threadEventsOfType("thread/compacted")).toEqual([]);
   });
 
+  it("fails the compaction turn when the agent reports the failure in an end-turn message", async () => {
+    const { providerThreadId } = await startThread({
+      envVars: {
+        FAKE_ACP_COMPACT_AGENT_MESSAGE:
+          "Compaction failed: summary model rejected the request",
+      },
+    });
+
+    const turnId = sendTurnRequest("turn/start", providerThreadId, {
+      input: compactCommandInput(),
+    });
+    expect((await waitForResponse(turnId)).error).toBeUndefined();
+
+    // omp answers end_turn even when its /compact handler failed and said so
+    // in an ordinary agent message; that text must fail the turn instead of
+    // the end_turn being read as a shrunk context (#2290).
+    const completed = await waitForTurnCompleted();
+    expect(completed).toMatchObject({
+      status: "failed",
+      error: {
+        message: "Compaction failed: summary model rejected the request",
+      },
+    });
+    expect(threadEventsOfType("thread/compacted")).toEqual([]);
+  });
+
+  it("completes a no-op compaction turn without reporting a compacted context", async () => {
+    const { providerThreadId } = await startThread({
+      envVars: {
+        FAKE_ACP_COMPACT_AGENT_MESSAGE:
+          "Compaction failed: Nothing to compact (session too small)",
+      },
+    });
+
+    const turnId = sendTurnRequest("turn/start", providerThreadId, {
+      input: compactCommandInput(),
+    });
+    expect((await waitForResponse(turnId)).error).toBeUndefined();
+
+    // A small session has nothing to compact: the turn ends cleanly with the
+    // agent's reason surfaced as a warning, and no `thread/compacted`.
+    const completed = await waitForTurnCompleted();
+    expect(completed).toMatchObject({ status: "completed" });
+    expect(threadEventsOfType("thread/compacted")).toEqual([]);
+    expect(threadEventsOfType("provider/warning").at(-1)).toMatchObject({
+      category: "compaction-skipped",
+      summary: "Context compaction skipped",
+      details: "Compaction failed: Nothing to compact (session too small)",
+    });
+  });
+
   it("accepts turn input only after the prompt carrying it goes out", async () => {
     const { providerThreadId } = await startThread();
     const turnId = sendTurnRequest("turn/start", providerThreadId, {
--- a/packages/provider-bridge-acp/src/bridge/bridge.test.ts
+++ b/packages/provider-bridge-acp/src/bridge/bridge.test.ts
@@ -2429,6 +2429,60 @@ describe("acp bridge", () => {
     });
   });
 
+  it("keeps classifying a no-op compaction when the agent rewords its prose", async () => {
+    const { providerThreadId } = await startThread({
+      envVars: {
+        FAKE_ACP_COMPACT_AGENT_MESSAGE:
+          "compaction failed: nothing to compact — the session is still small",
+      },
+    });
+
+    const turnId = sendTurnRequest("turn/start", providerThreadId, {
+      input: compactCommandInput(),
+    });
+    expect((await waitForResponse(turnId)).error).toBeUndefined();
+
+    // The no-op phrasing is the agent's own prose, not a contract
+    // (can1357/oh-my-pi#9786): a reworded, lowercased reason must still
+    // complete the turn as a skip instead of failing it as an error.
+    const completed = await waitForTurnCompleted();
+    expect(completed).toMatchObject({ status: "completed" });
+    expect(threadEventsOfType("thread/compacted")).toEqual([]);
+    expect(threadEventsOfType("provider/warning").at(-1)).toMatchObject({
+      category: "compaction-skipped",
+      summary: "Context compaction skipped",
+      details:
+        "compaction failed: nothing to compact — the session is still small",
+    });
+  });
+
+  it("fails the compaction turn when the failure report is reworded or preceded by other text", async () => {
+    const { providerThreadId } = await startThread({
+      envVars: {
+        FAKE_ACP_COMPACT_AGENT_MESSAGE:
+          "Tried shrinking the context.\nCompaction failed: session is locked by another compaction",
+      },
+    });
+
+    const turnId = sendTurnRequest("turn/start", providerThreadId, {
+      input: compactCommandInput(),
+    });
+    expect((await waitForResponse(turnId)).error).toBeUndefined();
+
+    // A failure sentence buried after other streamed text must still fail
+    // the turn; reading `end_turn` as success here would report a compacted
+    // context that never compacted.
+    const completed = await waitForTurnCompleted();
+    expect(completed).toMatchObject({
+      status: "failed",
+      error: {
+        message:
+          "Tried shrinking the context.\nCompaction failed: session is locked by another compaction",
+      },
+    });
+    expect(threadEventsOfType("thread/compacted")).toEqual([]);
+  });
+
   it("accepts turn input only after the prompt carrying it goes out", async () => {
     const { providerThreadId } = await startThread();
     const turnId = sendTurnRequest("turn/start", providerThreadId, {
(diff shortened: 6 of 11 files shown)


bradhallett:
Live production evidence that the bridge compaction path works end-to-end for acp-omp — gathered today (2026-08-22) on a real thread, independent of this PR's tests.

A local auto-compaction plugin (context-compact) queues `/compact` for busy threads instead of calling the direct RPC. On thread `thr_4tm7g4ie78` (provider `acp-omp`, OMP 17.4.1) the queued command drained at the turn boundary and executed a real compaction through the provider:

```
14:31:39Z  plugin log: auto-compact thr_4tm7g4ie78: 73111 tok (7%) → queue
           (idle acp-omp threads in the same sweep: "provider acp-omp has no
            manual compaction; skipping its threads" — the 409 gate)
14:36:50Z  turn/completed (prior turn ends)
14:36:50Z  client/turn/requested: "/compact" (builtin command mention)
14:36:50Z  turn/started → item/started contextCompaction → turn/input/accepted
14:38:17Z  thread/compacted → turn/completed
```

Sequence numbers on the thread's event log: 19648 (prior turn/completed), 19649–19652 (the `/compact` turn: requested/started/contextCompaction/input-accepted), 19656 (`thread/compacted`), 19657 (turn/completed). Provider session id `01a027f5-2537-7000-98be-c1612765f2f8` is stable across all four.

Two conclusions:

1. **The turn-input compaction path already works for acp-omp in production** — the same path this PR's new public-route test exercises (`createStandaloneBuiltinCompactCommandInput` → bridge → `session/prompt "/compact"`). The compaction above completed successfully at 14:38:17Z.
2. **The 409 is purely the declaration-level gate** (`supportsManualCompaction: false` in `ACP_BASE_CAPABILITIES`) blocking the idle/direct `compactThread` RPC. Busy threads are servable today only because the queue path never consults that gate.

So the declaration flip in this PR doesn't enable an untested path — it un-gates the direct RPC onto a path production already exercises successfully.

> AGENT GENERATED

bradhallett:
## Rework addendum (reply to review of #2293)

**Finding 1 (blocking — obsolete server.ts hunk):** Addressed by the rebase. The branch now sits on current main (`42658f987`); the conflict in `plugins/provider-acp/server.ts` was resolved in main's favor and the one-line capability is re-expressed where #2325 put it: `supportsManualCompaction: true` on the acp-omp entry of `plugins/provider-acp/src/known-agents.ts`, exactly as the review's §6 step 1 specifies. The first-party capability table test and the acp-omp 200-path public-route case are ported onto main's layout (rebased commit `6287a8ec0`).

**Finding 2 (major — false `thread/compacted` on omp):** Addressed in the same PR (new commit `f8620a8ea`), following the pi/#1721 precedent the review pointed at. While the compaction maintenance prompt is in flight (`activePromptKind === "compaction"` only), the bridge accumulates `agent_message_chunk` text; when the prompt ends with `end_turn` the completed text is classified:
- `"Compaction failed: Nothing to compact (session too small)"` / `"Compaction failed: Already compacted"` (the exact live strings from §5b) → new `skipped` compaction outcome → translator emits `provider.warning` with `category: "compaction-skipped"` plus a **completed** turn boundary and **no** `thread/compacted` — identical treatment to pi's delta translation.
- Any other `"Compaction failed: …"` text → turn **fails** with the agent's own message; no `thread/compacted`, no success banner.
- Anything else → unchanged `completed` + `thread/compacted`.

Detection is scoped narrowly to the compaction turn and never inspects ordinary agent traffic; the message strings are matched verbatim from the report. Covered by three new tests (two bridge-level against the fake ACP agent — which gained a `FAKE_ACP_COMPACT_AGENT_MESSAGE` knob to replay omp's message-then-end_turn behavior — and one delta-translation unit test), each verified fail-before/pass-after by mutation:
- early-return `{status:'completed'}` before the failure check → both bridge tests fail, including the exact false `thread/compacted` the report predicted;
- disabling the skipped→warning mapping → the translation test fails expecting the `provider/warning` row;
- reverting the known-agents.ts declaration → the public 200-path test fails with the 409 body and the capability table fails `expected false to be true`.

**Finding 3 (minor — bridge behavior untested):** The two bridge-level tests close this gap.

**Finding 4 (minor — unrelated reformatting):** Dropped during the rebase, but the repo-pinned oxfmt 0.64.0 re-introduces both hunks on main's current copy of the file (main's own blob fails `oxfmt --check`), so they are kept to stay format-clean; they are pure formatter output, no semantic change.

**Finding 5/6 (ok):** No wire/daemon/CLI changes; the `skipped` status is a bridge-internal envelope value (the wire still only carries existing delta kinds), so `HOST_DAEMON_PROTOCOL_VERSION` stays untouched.

**Gates:** `@bb/provider-bridge-acp` + `bb-plugin-provider-acp` test/typecheck/lint 8/8 green (258 + 76 tests); `@bb/server` typecheck/lint green, full suite 2005/2006 — the single failure (`install-machine-script.test.ts`, `Abort trap: 6` in a nohup'd host-daemon join) reproduces identically with these changes stashed on clean main; scoped `test/public/` + `test/services/plugins/` 793/793.

> AGENT GENERATED

bradhallett:
## Rebase + SDK-version fix (CI failure root-caused)

The last push failed CI with `check-npm-version-guard`: the compaction-honesty change alters `dist/provider-bridge-acp.js`, and `@get-bb/plugin-sdk@0.4.16` was already published, so the guard demanded a version bump (`node scripts/bump-plugin-sdk.mjs --patch`) — the same requirement `packages/domain/src/plugin-sdk-version.ts` + `packages/plugin-sdk/package.json` exist to enforce.

Done, plus a refresh onto current `main` (`31d66d9d4` — the rework base `42658f987` had drifted):

- Rebased via cherry-pick; both rework commit messages preserved, one new commit bumps `@get-bb/plugin-sdk` 0.4.22 → 0.4.23 using the repo script (both files moved together).
- Feature-content parity vs the previous head verified marker-for-marker (declaration, `skipped` outcome, failure/skip classification, fake-agent knob) before running gates.

Gates:

- `node packages/plugin-sdk/scripts/check-npm-version-guard.mjs` → **PASS** (`0.4.23` not on npm yet).
- `turbo run test typecheck lint --filter=@bb/provider-bridge-acp --filter=bb-plugin-provider-acp --force` — 8/8 green, 291 tests (17 files).
- `turbo run test --filter=@bb/server --force` — 2048/2049; the single failure (`install-machine-script.test.ts`, per-server data-dir default) is the documented pre-existing environmental one (reproduces on pristine main in this class of environment).
- `oxfmt --check` clean on all touched files.

@SawyerHood — re-review when you get a chance; the two rework commits are unchanged in content, only rebased, plus the mechanical version bump.

bradhallett:
## Hardened: compaction classification no longer pins exact prose

Follow-up to the review thread: the no-op detection matched omp's failure strings verbatim, so an upstream reword of the `Compaction failed:` opener would have made a failed or no-op compaction indistinguishable from a shrunk context (false `thread/compacted`). omp's phrasing is TUI presentation code rather than a contract — asked for a structured signal in can1357/oh-my-pi#9786 — so until one exists, classification is now semantic:

- failure detection: `/\bcompaction failed\b/i` anywhere in the streamed message (case-insensitive) — a failure sentence can no longer masquerade as success regardless of position, case, or surrounding text
- no-op detection: `/\b(?:nothing to compact|already compacted)\b/i` — a reworded, lowercased reason still completes as a skip with the agent's text surfaced verbatim
- every other `Compaction failed: …` message stays a failed turn

Two new bridge tests cover the drift cases (reworded no-op prose → skip; failure sentence preceded by other streamed text → failed), each mutation-verified: reverting no-op matching alone fails the first, reverting to the old prefix-only check fails the second with a false `completed`.

Gates: bridge 293/293 (17 files), typecheck, lint, oxfmt, `check-npm-version-guard` PASS (0.4.23 still unpublished).

bradhallett:
Refreshed onto current main — main took `0.4.23` (mention aliases) and `0.4.24` (lost connection reason) while this branch carried its own bridge dist changes, so the branch re-bases its SDK claim to **0.4.25**. No code changes beyond the merge: the classifier and drift tests are untouched, and main's new ACP fixes (inbound-dispatch-after-kill #2594, topology marker race #2571, intentional-stop write race #2552) merged clean.

Gates: bridge 304/304 (17 files — includes the 11 tests main added), typecheck, lint, oxfmt, `check-npm-version-guard` PASS (0.4.25 unpublished).

bradhallett:
Refreshed onto current main (SDK 0.4.27 → branch claims 0.4.28). This merge also adopts main's two newer conventions where they collided with this branch's regions: the comment-free style from #2624 (the classifier's rationale lives in the PR description) and main's independently-landed `supportsManualCompaction: true` on the omp entry (duplicate removed; the compaction turn/classifier themselves are not on main). Gates: bridge 305/305, plugin 76/76, typecheck, lint, oxfmt, version-guard PASS.

bradhallett:
## CI status on head `9a026a4`: pre-existing load-marginal tests, not this diff

The `Tests (packages)` job has failed 4× in a row on this head, each time in a different package this PR doesn't touch, always as a timeout, and each preceded by a runner-wide 30s `ACP-native model discovery … timed out` warning from the bridge suite (that discovery test still passes — it's a victim of the same CPU starvation, not a cause).

| Run | Failing test | Budget |
|---|---|---|
| 33185878477 | `bb-plugin-provider-pi` `catalog.test.ts` › "re-reads model scope after the catalog child restarts" | 5000ms |
| 33186437294 | `@bb/host-workspace` › "does not overflow the call stack merging a large subdirectory" (hook) | 10000ms |
| 33187231762 | pi `catalog.test.ts` (same test) | 5000ms |

### The diff is exonerated

Running the exact CI shard at full turbo parallelism on a **pristine `origin/main` checkout** reproduces the identical failure set:

```
pnpm exec turbo run test --continue \
  --filter='!@bb/server' --filter='!@bb/app' --filter='!@bb/integration-tests' --force
```

→ `@bb/host-workspace` "Hook timed out in 10000ms" (2), `bb-plugin-provider-pi` (catalog child restart), `bb-plugin-tasks`, `@bb/plugin-build` — same set as on this branch. Every one of these tests passes when its package runs alone (verified locally: bridge 305/305, provider-acp 76/76, the pi catalog test in isolation ×2, server 2072/2073 with the known env failure). Main's and #2626's green packages-job logs contain zero discovery-timeout lines; this branch's slightly longer bridge suite appears to be what tips the marginal scheduling over on the CI runner.

Filed separately with the deterministic main-branch repro and fix options: #2653.

### Ask

Fork PRs can't re-run their own checks (`Must have admin rights to Repository`). If a maintainer re-runs the packages job — or #2653 lands a budget/concurrency fix — this should go green. All local gates on this head pass (typecheck, lint, oxfmt, version guard, full suites above).

bradhallett:
cc @SawyerHood — evidence above that the four red `Tests (packages)` runs on this head are pre-existing load-marginal tests (deterministic repro on pristine `main`: #2653). All local gates pass; a re-run of the packages job is the only thing between this and green. Re-review welcome when you get a chance.

bradhallett:
CI is green on this head after get-bb/bb#2655 landed the `--concurrency=4` turbo cap (#2653 confirmed + closed). All checks pass; mergeable state CLEAN. Ready for re-review.

bradhallett:
Refreshed against main: absorbed the #2701 SDK bump via merge + re-bump to `0.4.30` (merge `03dde13a2`, conflicts were the two version files only). Classifier and tests untouched; bridge 305/305, typecheck/lint/version-guard green locally; CI green on the new head.

bradhallett:
Follow-up for the two review findings on this merged commit: #3013. Status: the Major finding (compaction text rule applying to all ACP agents) was already fixed on main by #2957; #3013 closes the Medium finding — compaction prose accumulation is now gated to the omp dialect, clamped at 64 KiB head-preserved, and cleared in `finishCompaction`, with mutation-checked regression tests for the cap and the cleanup.

PR #2294, opened 2026-08-22
Treat auth-required probe failures as expected online RPC failures

### LINKED ISSUE #2291: provider.list_models startup probe logs "ACP agent is not authenticated." at warn on every daemon start
### 1. Summary

Every time the host daemon starts (or restarts), bb's first model-catalog probe for an ACP provider fires before the ACP agent has finished its auth handshake, so the daemon logs a warn-level `online host RPC failed` for `provider.list_models` with `ACP agent is not authenticated.` The probe failure is benign in practice — the server already logs it separately and the model picker falls back to declared/synthetic models, and the next probe succeeds once the agent's auth settles — but the daemon line fires on every single start and reads like a bug.

- bb desktop app (installed builds under `/Applications/bb.app` and `/Applications/bb Nightly.app`, per stack traces in `~/.bb/logs/host-daemon.*.log`); source audited at `fff3ae8` (main).
- macOS (Apple Silicon, darwin 27.0.0). Daemon connecting to local server at `http://127.0.0.1:38886`.
- Providers affected: `acp-cursor` (observed in server logs on this host) and `acp-omp` (OMP ACP agent; same mechanism per the integration post-mortem). The mechanism is generic to any ACP agent whose model-list command reports an auth-required failure.
- Setup context: OMP↔bb integration audit (2026-08-20 → 2026-08-22), 45 threads; this is post-mortem Issue 4b.

### 3. Steps to reproduce

1. Install bb desktop with at least one ACP provider configured (e.g. custom ACP agent for `omp acp`, or Cursor's `cursor-agent`).
2. Restart the bb server/daemon (restart the desktop app, or kill and relaunch the host daemon).
3. Watch `~/.bb/logs/host-daemon.<n>.log` for the first ~2 seconds after `Host daemon connecting`.

Observed every time; the warn lands ~0.6–0.9 s after the daemon's startup banner. No user interaction with the provider is needed. Negative space: the line appears once per daemon session (the first probe), not on later picker opens — later probes succeed once the agent's auth state settles, because probe failures are deliberately not memoized (`apps/server/src/services/system/execution-options.ts`, `listProviderModelsMemoized`: "failures are not memoized").

### 4. Expected vs actual

Actual, `~/.bb/logs/host-daemon.1.log` line 59 (trimmed; 33 such lines across `host-daemon.1.log`/`host-daemon.2.log`, one per daemon start; each is mirrored by a server-side `Failed to resolve provider models` warn — 12 + 21):

```json
{"level":40,"time":1787261024711,"component":"host-daemon","serverUrl":"http://127.0.0.1:38886","type":"provider.list_models","err":{"type":"hB","message":"ACP agent is not authenticated.","stack":"JsonRpcResponseError: ACP agent is not authenticated.\n    at MZr (.../host-daemon/dist/daemon-bundle.mjs:49:3384)\n    at gtt (.../daemon-bundle.mjs:49:…)"},"msg":"online host RPC failed"}
```

Server-side counterpart, `~/.bb/logs/server.1.log` line 310:

```json
{"level":40,"time":1787261024715,"component":"server","errorCode":"auth_required","errorMessage":"ACP agent is not authenticated.","errorStatus":502,"errorRetryable":false,"hostId":"host_jgwgbjvi9c","providerId":"acp-cursor","msg":"Failed to resolve provider models"}
```

Expected: the daemon should not warn for an auth-required probe outcome. The condition is a user state, not a daemon bug; the failure already propagates to the server, which logs it with full context (`errorCode`, `hostId`, `providerId`) and surfaces a model-load error in the UI. A genuinely unauthenticated agent (persistent failure) must still surface — via that server-side path, which this report does not propose to touch.

### 5. Evidence

Chain, each hop source-cited at `fff3ae8`:

- Probe origin (server): [`apps/server/src/services/system/execution-options.ts` L679-L700](https://github.com/get-bb/bb/blob/fff3ae8/apps/server/src/services/system/execution-options.ts#L679-L700) — the memo key includes the daemon session id ("a reconnected daemon may have a new CLI or account, so its first probe is fresh"), so every fresh daemon session forces a new `provider.list_models` probe; failures are not memoized.
- Probe execution (daemon → bridge): [`apps/host-daemon/src/command-router.ts` L152-L196](https://github.com/get-bb/bb/blob/fff3ae8/apps/host-daemon/src/command-router.ts#L152-L196) runs the RPC and, on failure, warns `online host RPC failed` unless `isExpectedOnlineRpcFailureError` is true. That is the logging site of the noise: `this.logger.warn({ type: message.command.type, err: error }, "online host RPC failed")`.
- Auth classification (daemon): [`apps/host-daemon/src/command-dispatch-support.ts` L314-L335](https://github.com/get-bb/bb/blob/fff3ae8/apps/host-daemon/src/command-dispatch-support.ts#L314-L335) — `getErrorCode` maps the message to `auth_required` (pattern at L149-L150). At `fff3ae8` `auth_required` is **not** in `EXPECTED_ONLINE_RPC_FAILURE_CODES` (L135-L138: only `file_too_large`, `provision_cancelled`), so the warn fires.
- Bridge origin of the message: [`plugins/provider-acp/src/bridge/bridge.ts` L520-L521](https://github.com/get-bb/bb/blob/fff3ae8/plugins/provider-acp/src/bridge/bridge.ts#L520-L521) defines `"ACP agent is not authenticated."`; [`L774-L822`](https://github.com/get-bb/bb/blob/fff3ae8/plugins/provider-acp/src/bridge/bridge.ts#L774-L822) runs the agent's model-list CLI command and rejects with `AcpModelListAuthRequiredError` ([L1015-L1022](https://github.com/get-bb/bb/blob/fff3ae8/plugins/provider-acp/src/bridge/bridge.ts#L1015-L1022)) when the CLI output looks like an auth failure ([L1024-L1043](https://github.com/get-bb/bb/blob/fff3ae8/plugins/provider-acp/src/bridge/bridge.ts#L1024-L1043)); `runBridgeRequest` ([`packages/provider-bridge-protocol/src/bridge-kit/bridge-harness.ts` L59-L70](https://github.com/get-bb/bb/blob/fff3ae8/packages/provider-bridge-protocol/src/bridge-kit/bridge-harness.ts#L59-L70)) serializes the throw as a JSON-RPC error, which the daemon's client re-raises as `JsonRpcResponseError` — matching the log stack.

Race window (measured from log timestamps, daemon ready banner → probe warn):

- `host-daemon.1.log` L58→L59: `Host plugin worker ready` @1787261024061 → warn @1787261024711 (**+650 ms**)
- `host-daemon.1.log` L114→L115: **+663 ms**; L123→L124: **+640 ms**
- `host-daemon.2.log` L27→L28: **+706 ms**; L44→L45: **+731 ms**; L9→L10: **+888 ms**

So the probe lands well under a second into the daemon session, while the ACP agent's auth state (login-state/token availability for its model-list command, and the agent-side handshake that settles after the agent process comes up) is not yet ready. The timing was measured by differencing `time` fields of adjacent pino lines in `~/.bb/logs/`.

Counts measured by parsing the logs with a JSON-lines script: 12 matching daemon lines in `host-daemon.1.log`, 21 in `host-daemon.2.log`; identical counts for the server-side `Failed to resolve provider models` lines in `server.1.log`/`server.2.log` (1:1 with the daemon lines by timestamp).

Investigation report: OMP↔bb integration post-mortem, Issue 4b ("Auth-probe noise: every host-daemon start logs `ACP agent is not authenticated.` from `provider.list_models` — the probe fires before the OMP agent completes auth handshake. Benign but recurring.").

### 6. What you ruled out

- Not a duplicate: bb issue search for the log string / "auth probe" found nothing open or closed (cross-reference table of the fix map lists this as unfiled, minor).
- Still on `main` at `fff3ae8`: `EXPECTED_ONLINE_RPC_FAILURE_CODES` contains only `file_too_large` and `provision_cancelled`.
- Not a bridge misclassification: the bridge must keep failing the probe when the agent CLI genuinely reports auth-required (persistent logged-out state) — that behavior is unchanged by the proposed fix, and the server-side surfacing (`modelLoadError`, `errorCode: auth_required`) is untouched.
- Not a daemon-side retry problem either: the daemon has no notion of a "startup probe", and the server already re-probes (failures are not memoized), so the transient case self-heals; only the daemon-side duplicate warn is wrong.

Minor. Log noise only — no data or work loss, no functional damage (picker falls back; next probe succeeds). Workaround: ignore the line. Effort: small (one classification-set entry plus a regression test).

> AGENT GENERATED

bradhallett:
## What was wrong

On every host-daemon start, the server's first execution-options resolution probes `provider.list_models` on the fresh daemon session (the probe memo is keyed by daemon session id, so a reconnected daemon always re-probes; failures are not memoized). The probe lands ~0.6–0.9 s into the daemon session — before the ACP agent's auth handshake/login state has settled — so the provider-acp bridge maps the agent CLI's auth-required failure to `"ACP agent is not authenticated."` ([`plugins/provider-acp/src/bridge/bridge.ts` L520-L521, L798-L801](https://github.com/get-bb/bb/blob/fff3ae8/plugins/provider-acp/src/bridge/bridge.ts#L798-L801)). The daemon classifies that as `auth_required`, which at base `fff3ae8` is not an expected online-RPC failure code, so [`CommandRouter.handleOnlineRpcRequest`](https://github.com/get-bb/bb/blob/fff3ae8/apps/host-daemon/src/command-router.ts#L170-L180) logs a warn `online host RPC failed` on every start — 33 occurrences across two rotated daemon logs on the reporting host, each benign and each already mirrored by the server's own contextual warn + UI model-load error. Issue: #2291 (post-mortem Issue 4b of the OMP↔bb integration audit).

## What changed

`apps/host-daemon/src/command-dispatch-support.ts` — added `auth_required` to `EXPECTED_ONLINE_RPC_FAILURE_CODES`, following the existing `file_too_large` / `provision_cancelled` pattern. An unauthenticated agent is a user-state outcome, not a daemon bug: the failure's `errorCode` still travels to the server in the RPC response, and the server still logs `Failed to resolve provider models` and surfaces the model-load error in the UI — persistent (genuinely logged-out) failures still surface exactly as before. Only the daemon-side duplicate warn is silenced; the failure-counting debug line in `logOnlineRpc` (which records `ok: false` with the `errorCode`) still fires. No wire changes; no behavior change for any other error class.

Deviation from the issue's alternatives: the issue offered retry-with-backoff or log-level demotion; demotion via the existing expected-failure set is the smaller change and the established repo pattern — the daemon has no notion of a "startup probe", and the transient case already self-heals because the server re-probes (failures are not memoized).

## How you verified

- Regression test in `apps/host-daemon/src/command-dispatch-support.test.ts` (`treats ACP model-probe auth failures as expected RPC failures`): asserts `isExpectedOnlineRpcFailureError` is true for both message shapes the router can see (`"ACP agent is not authenticated."` and the Cursor CLI's `Authentication required … CURSOR_API_KEY/…` text). Fails before the change by construction — the sibling test at the same site already pins `getErrorCode(…)` to `auth_required` for both messages, and `auth_required` was absent from the set — and passes after.
- Log forensics in the linked issue: the warn's race window (+640…+888 ms after the daemon ready banner) and its 1:1 pairing with the server-side warn were measured from `~/.bb/logs/host-daemon.*.log` / `server.*.log` timestamps.
- `pnpm exec turbo run test --filter=@bb/host-daemon` — 561/561 passing.
  Mutation check: with `command-dispatch-support.ts` reverted to base, the new
  test fails (1 failed / 5 passed) and passes again with the entry restored.
- `pnpm exec turbo run typecheck lint --filter=@bb/host-daemon` green;
  `pnpm exec oxfmt --check` green on both touched files.

Fixes #2291

> AGENT GENERATED

### DIFF
--- a/apps/host-daemon/src/command-dispatch-support.test.ts
+++ b/apps/host-daemon/src/command-dispatch-support.test.ts
@@ -1,8 +1,32 @@
+import { AgentRuntimeRecoveryError } from "@bb/agent-runtime";
 import { describe, expect, it } from "vitest";
 import {
   CommandDispatchError,
   isExpectedOnlineRpcFailureError,
 } from "./command-dispatch-support.js";
+
+const ACP_MODEL_LIST_AUTH_MESSAGE = "ACP agent is not authenticated.";
+
+/**
+ * The error `provider.list_models` surfaces when an ACP bridge rejects the
+ * model probe with a typed `authRequired` recovery hint: the agent runtime
+ * turns that into an `AgentRuntimeRecoveryError` whose string `code` the
+ * daemon's `getErrorCode` reads. Constructed exactly as the runtime does, so
+ * this test fails if the classifier ever returns to message-shape matching.
+ */
+function createAcpAuthRequiredError(): AgentRuntimeRecoveryError {
+  return new AgentRuntimeRecoveryError({
+    code: "auth_required",
+    message: ACP_MODEL_LIST_AUTH_MESSAGE,
+    recovery: {
+      kind: "authRequired",
+      message: ACP_MODEL_LIST_AUTH_MESSAGE,
+      providerId: "acp-cursor",
+      retryable: false,
+    },
+    cause: new Error("Authentication required."),
+  });
+}
 
 describe("command dispatch support", () => {
   it("classifies oversized file reads as expected RPC failures", () => {
@@ -11,5 +35,23 @@ describe("command dispatch support", () => {
         new CommandDispatchError("file_too_large", "File exceeds the limit"),
       ),
     ).toBe(true);
+  });
+
+  it("classifies typed auth_required recovery failures as expected RPC failures", () => {
+    expect(isExpectedOnlineRpcFailureError(createAcpAuthRequiredError())).toBe(
+      true,
+    );
+  });
+
+  it("keeps unclassified failures unexpected", () => {
+    expect(isExpectedOnlineRpcFailureError(new Error("boom"))).toBe(false);
+    expect(
+      isExpectedOnlineRpcFailureError(
+        new CommandDispatchError(
+          "provider_bridge_unavailable",
+          "No plugin host artifact fetcher configured",
+        ),
+      ),
+    ).toBe(false);
   });
 });
--- a/apps/host-daemon/src/command-dispatch-support.ts
+++ b/apps/host-daemon/src/command-dispatch-support.ts
@@ -133,6 +133,12 @@ export function isExpectedCommandDispatchError(
 const EXPECTED_ONLINE_RPC_FAILURE_CODES = new Set([
   "file_too_large",
   "provision_cancelled",
+  // An unauthenticated agent is a user-state outcome, not a daemon bug: the
+  // errorCode travels to the server, which logs and surfaces it (model load
+  // error, turn error). The daemon-side "online host RPC failed" warn for it
+  // is duplicate noise — the startup provider.list_models probe races the
+  // ACP agent's auth handshake and hits this on every daemon start.
+  "auth_required",
 ]);
 
 export function isExpectedOnlineRpcFailureError(error: unknown): boolean {
--- a/apps/host-daemon/src/command-dispatch-support.ts
+++ b/apps/host-daemon/src/command-dispatch-support.ts
@@ -133,11 +133,12 @@ export function isExpectedCommandDispatchError(
 const EXPECTED_ONLINE_RPC_FAILURE_CODES = new Set([
   "file_too_large",
   "provision_cancelled",
-  // An unauthenticated agent is a user-state outcome, not a daemon bug: the
-  // errorCode travels to the server, which logs and surfaces it (model load
-  // error, turn error). The daemon-side "online host RPC failed" warn for it
-  // is duplicate noise — the startup provider.list_models probe races the
-  // ACP agent's auth handshake and hits this on every daemon start.
+  // An unauthenticated agent is a user-state outcome, not a daemon bug. The
+  // typed code still reaches the server, which logs and surfaces it (model
+  // load error, turn error), so the daemon-side "online host RPC failed"
+  // warn would only duplicate it with two nested stacks. Scoped to the
+  // auth_required code, so it covers every online RPC that can surface it,
+  // not just the provider.list_models probe.
   "auth_required",
 ]);
 
--- a/apps/host-daemon/test/command/command-router.test.ts
+++ b/apps/host-daemon/test/command/command-router.test.ts
@@ -3,6 +3,7 @@ import type {
   HostDaemonOnlineRpcRequestMessage,
   HostDaemonOnlineRpcResponseMessage,
 } from "@bb/host-daemon-contract";
+import { AgentRuntimeRecoveryError } from "@bb/agent-runtime";
 import { WorkspaceError } from "@bb/host-workspace";
 import {
   encodeClientTurnRequestIdNumber,
@@ -45,7 +46,7 @@ type ThreadStartCommand = Extract<HostDaemonCommand, { type: "thread.start" }>;
 type TurnSubmitCommand = Extract<HostDaemonCommand, { type: "turn.submit" }>;
 
 interface RunRouterCommandArgs {
-  command: HostDaemonCommand;
+  command: HostDaemonOnlineRpcRequestMessage["command"];
   requestId: string;
   router: CommandRouter;
 }
@@ -61,6 +62,7 @@ interface CreateTurnSubmitCommandArgs {
 
 interface CreateRouterArgs {
   logger?: CommandRouterOptions["logger"];
+  listModels?: CommandRouterOptions["listModels"];
   resolveInteractiveRequest?: CommandRouterOptions["resolveInteractiveRequest"];
   runtimeManager?: RuntimeManager;
 }
@@ -85,6 +87,7 @@ function createRouter(
     fetchProjectAttachment: unexpectedProjectAttachmentFetch,
     fetchPluginHostArtifact: fetchDispatchTestArtifact,
     ...unexpectedProviderMaintenance,
+    ...(args.listModels === undefined ? {} : { listModels: args.listModels }),
     logger: {
       debug: () => undefined,
       warn: () => undefined,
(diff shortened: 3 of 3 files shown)


bradhallett:
## Rework addendum (reply to review of #2291)

Thanks for the thorough report — reworked per the REQUEST CHANGES verdict. Rebased onto current `main` (`42658f987`) and replaced the regression test as requested.

**Test is dead on arrival / Merge conflict (Blockers)** — Fixed. The rebase hit exactly the predicted conflict: `e42a4ef48` (#2325) removed `ACP_AUTH_REQUIRED_PATTERN` and reduced `command-dispatch-support.test.ts` from ~150 to 15 lines. Resolved to main's structure and dropped the bare-`new Error(...)` test entirely. The new tests construct the error the way the runtime actually does — `new AgentRuntimeRecoveryError({ code: "auth_required", message, recovery: { kind: "authRequired", providerId, retryable: false }, cause })` — so they fail if the classifier ever returns to message-shape matching.

**Test exercises the wrong unit (Medium)** — Fixed. Mirrored your §4a verification as router-level tests in `apps/host-daemon/test/command/command-router.test.ts` (placed beside the existing provision-cancellation precedent): `provider.list_models` through `CommandRouter.handleOnlineRpcRequest` with `listModels` throwing the typed error asserts (a) the typed `{ ok: false, errorCode: "auth_required", errorMessage }` response reaches the server intact, (b) `logger.warn` is called zero times, and (c) the debug accounting line (`Online host RPC`, with commandType/errorCode/ok:false) still fires. Added the contrast case: an unclassified failure still warns exactly once with `online host RPC failed`.

**Blast radius not stated (Low)** — Fixed in the code comment: it now says the entry covers *every* online RPC that can surface the code (provider.usage, provider.installation.*, any plugin.host.call error carrying code `auth_required`), not just the probe. On the two alternatives you offered: I kept the scope tight and did **not** add `rate_limited` — there's no report of that warn being noise, and each code can be added independently when warranted — and did **not** switch to `instanceof AgentRuntimeRecoveryError`, because the string-code allow-list matches the existing pattern beside it (`file_too_large`, `provision_cancelled` from 4c7d2a97d) and stays decoupled from the daemon importing the runtime's error class. Happy to switch if maintainers prefer the type-based form.

**"Auth handshake race" wording (Low)** — Dropped. The new comment states the real mechanism (user-state outcome; typed code reaches the server, which logs and surfaces it; the daemon warn duplicates it with two nested stacks). Note the original commit message is preserved unsquashed per repo convention, so its stale race wording remains in that commit's message only; this addendum and the rework commit message correct the record.

**Verification** — mutation-checked: deleting the `"auth_required"` entry makes both new tests fail (unit: `expected false to be true`; router: warn called 1 time), restoring makes them pass. Scoped gates: `tsc --noEmit` clean, package lint clean, `oxfmt --check` clean on touched files, host-daemon suite 553/557 — the 4 failures (`thread-stop-races.test.ts`) reproduce identically on pristine main sources in this environment (esbuild cannot resolve `@get-bb/plugin-sdk/provider-bridge` because `packages/plugin-sdk` imports `@bb/provider-bridge-acp/launch-spec` without declaring that dependency, so its dist never builds under pnpm strict isolation) — unrelated to this PR and worth a look separately.

> AGENT GENERATED

bradhallett:
## Refreshed onto current main

Six `apps/host-daemon` commits landed since the rework base `42658f987`, including `787eb14d9` (retry provider probes after invalidation, #2381) — directly adjacent to this change's probe path. Cherry-picked both commits onto current `main` (`31d66d9d4`): no conflicts, no content changes, commit messages preserved.

Gates on the refreshed branch:

- `turbo run test typecheck --filter=@bb/host-daemon --force` — 563/563 passing, 45 files, typecheck clean (the `thread-stop-races.test.ts` environmental failures from the last addendum do not reproduce in this environment).
- No plugin-sdk version bump needed — `apps/host-daemon` is not part of the published SDK package.

@SawyerHood — re-review when convenient.

bradhallett:
cc @SawyerHood — this one stayed clean through the recent main drift (host-daemon 563/563 locally). Re-review welcome when you get a chance. (Neighbor PR #2293's red CI is a pre-existing shard flake, analyzed in its thread and filed as #2653 — not from either diff.)

bradhallett:
cc @brsbl @ymichael — re-review ping with fresh context: sibling get-bb/bb#2293 landed yesterday, and this is the remaining piece of the pair. Scope: typed error for daemon auth-probe failures (replaces bare `Error` objects, regression test included). Clean against main, CI green, host-daemon 563/563 locally, small diff. Happy to split or rebase further if that helps review.

SawyerHood:

PR #2296, opened 2026-08-22
Resume daemon-restarted threads with persisted provider sessions

### LINKED ISSUE #1592: Auto-restart thread after "host daemon restarted"
I use bb against an external server I run and I frequently run into "host daemon restarted". I'm not sure why it would be restarting, I'm not touching it at all, any guidance on how to debug that would be appreciated!

But one thing which would be nice is if the daemon picked up the thread agains automatically after it had restarted. I run a lot of work over night and it's a bit annoying to see it didn't get all the way there. :)

bradhallett:
## What was wrong

When a daemon re-registers with a new instance id, `handleHostSessionOpened` unconditionally failed every active thread on that host before reconciliation could look at it: a synthetic `turn/completed` (interrupted), a `system/error` ("Thread interrupted because the host daemon disconnected" / "Please retry the thread to continue."), and `run.failed` landing the thread in `error` (base fff3ae8: [`session-owner-side-effects.ts`](https://github.com/get-bb/bb/blob/fff3ae8/apps/server/src/internal/session-owner-side-effects.ts#L115-L125) calling [`thread-lifecycle.ts`](https://github.com/get-bb/bb/blob/fff3ae8/apps/server/src/services/threads/thread-lifecycle.ts#L374-L375)). That is a lie for any run whose provider session survives the restart: ACP and Codex sessions persist on the host and re-attach through `thread/resume` on the next dispatch. #1592 asks for exactly this — overnight work dying on every daemon restart with a "please retry" prompt instead of continuing.

## What changed

- `apps/server/src/services/threads/background-task-reconciliation.ts`: new `settleDaemonRestartedThreadRunsForResume`, the daemon-restart counterpart of the dangling-task backstop. Active threads on the host that have a persisted provider thread id settle like a user stop — `turn/completed` (interrupted) plus `system/thread/interrupted` (reason `host-daemon-restarted`), **no synthetic `system/error`** — and land `idle` via `stop.requested`/`stop.settled`. Runs with no provider session (e.g. a `thread.start` that never established one) are left active and keep the existing failure interruption; there is nothing to re-attach.
- `apps/server/src/internal/session-owner-side-effects.ts`: the replacement-instance branch calls the resume settlement before the interruption sweep, which then only sees the non-resumable runs. `settleDanglingBackgroundTasks` and the same-instance reconnect / disconnect-grace / host-removed paths are unchanged.
- No wire changes (`HOST_DAEMON_PROTOCOL_VERSION` untouched); no CLI or doc surface affected.

Deviations from the issue's ask, flagged for maintainers:

1. #1592 asks that the daemon "picked up the thread agains automatically". This PR makes restarted threads cleanly *resumable* rather than auto-replaying them: an idle thread re-attaches its persisted provider session on the next `turn.submit`, and the existing queued-message auto-send sweep continues threads that had work queued — no manual retry, no misleading error. It does **not** automatically re-submit the interrupted turn's stored input: that duplicates the user message in the provider transcript and is a product decision (happy to implement if you want it).
2. The interrupted turn still gets a canonical `turn/completed` (interrupted) record, same as a user stop. Leaving the turn open would misroute subsequent sends as steers to a turn no live runtime owns.

## How you verified

Extended `apps/server/test/internal/background-task-reconciliation.test.ts`:

- "resumes active turns whose provider session survives a daemon restart" — drives `handleDaemonSocketClosed` then `/internal/session/open` with a new instance id and asserts the thread lands `idle` with exactly `[turn/completed, system/thread/interrupted]` and no `system/error`. On base fff3ae8 the same scenario produces `error` plus the synthetic `system/error`, so this test fails before the change.
- "interrupts active runs with no provider session to re-attach when a different daemon instance registers" — guards the nothing-to-re-attach case: still `error` with the failure interruption.
- The existing same-instance-reconnect and disconnect-grace tests are untouched and still cover their paths.

- Locally: `pnpm exec turbo run test --filter=@bb/server` — 1923 tests, 1922
  passing; the single failure (`test/app/install-machine-script.test.ts`)
  reproduces identically on a pristine `main` checkout at `fff3ae8`
  (environmental, pre-existing). Mutation check: reverting only the
  `session-owner-side-effects.ts` wiring makes the resume test fail (1
  failed / 11 passed) and pass with the call restored.
- `pnpm exec turbo run typecheck lint --filter=@bb/server` green; `pnpm exec
  oxfmt --check` green on all three touched files.

Fixes #1592

> AGENT GENERATED

### DIFF
--- a/apps/server/src/internal/session-owner-side-effects.ts
+++ b/apps/server/src/internal/session-owner-side-effects.ts
@@ -20,7 +20,10 @@ import {
   reconcileDaemonReportedThreads,
 } from "../services/threads/thread-lifecycle.js";
 import { buildThreadStatusChangeMetadataByThreadId } from "../services/threads/thread-runtime-display.js";
-import { settleDanglingBackgroundTasks } from "../services/threads/background-task-reconciliation.js";
+import {
+  settleDanglingBackgroundTasks,
+  settleDaemonRestartedThreadRunsForResume,
+} from "../services/threads/background-task-reconciliation.js";
 
 const DAEMON_RESTARTED_PENDING_INTERACTION_REASON =
   "Host daemon restarted while awaiting user interaction; retry the thread to continue";
@@ -109,6 +112,7 @@ export async function handleHostSessionOpened(
         hostId: args.hostId,
         reason: DAEMON_RESTARTED_PENDING_INTERACTION_REASON,
       });
+      settleDaemonRestartedThreadRunsForResume(deps, { hostId: args.hostId });
       interruptActiveThreadsForHost(deps, {
         hostId: args.hostId,
         reason: "host-daemon-restarted",
--- a/apps/server/src/services/threads/background-task-reconciliation.ts
+++ b/apps/server/src/services/threads/background-task-reconciliation.ts
@@ -1,7 +1,11 @@
+import { and, eq, isNull } from "drizzle-orm";
 import { z } from "zod";
 import {
+  environments,
   listOpenBackgroundTaskItemRowsForHost,
   listOpenBackgroundTaskItemRowsForThread,
+  listThreadTurnInterruptionEventStates,
+  threads,
   type DbNotifier,
   type DbTransaction,
   type OpenBackgroundTaskItemRow,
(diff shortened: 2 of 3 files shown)


bradhallett:
cc @ymichael — re-review ping; no feedback since this opened.

The one product decision flagged in the description, restated so it's easy to rule on: #1592 asks that a daemon restart "pick the thread up again automatically". This PR makes restarted threads cleanly **resumable** — the persisted provider session re-attaches on the next `turn.submit` and queued-message sweeps continue — but does not auto-replay the interrupted turn's stored input, because that duplicates the user message in the provider transcript. Auto-replay is a small addition on top if maintainers want the full behavior; ruling either way unblocks the PR.

Gates: server suite green (single documented pre-existing env failure), mutation-checked, no wire changes.

bradhallett:
## Rebase

Rebased onto `main` @ `f6868ad0c` (was 12 days behind). Single commit, message unchanged; `git range-diff` confirms every hunk present.

Conflicts in all 3 files, resolved as:

- `session-owner-side-effects.ts` — #2797 moved `interruptPendingInteractionsForHostThreads` inside the `!sameDaemonInstance` branch; the resume settlement now sits after it and before `interruptActiveThreadsForHost`, preserving the PR's ordering (settle-for-resume precedes the interruption sweep).
- `background-task-reconciliation.ts` — `settleDaemonRestartedThreadRunsForResume` replayed verbatim; the PR's doc/inline comments were dropped since #2624 (`bb/no-comments`) now bans them repo-wide (rationale stays in this description).
- test file — #2514/#2797 renamed the restarted-instance test on `main`; the PR's rewrite (resume expectations + the new no-provider-session failure case) wins, with `main`'s current event vocabulary intact.

No behavioral change vs the previous head; no duplicate logic on `main` to drop (#2797 is pending-interaction handling, not run settlement).

## Gates

- `pnpm exec turbo run test typecheck lint --filter=@bb/server --force` — typecheck ✅, lint ✅, tests **2194/2195**: the sole failure is the documented pre-existing env failure (`install-machine-script.test.ts`, `Abort trap: 6` in the spawned host-daemon, reproduces on pristine `main` here). This file's own suite: **12/12** ✅.
- `pnpm exec oxfmt --check` on the 3 touched files — clean.
- `node packages/plugin-sdk/scripts/check-npm-version-guard.mjs` — PASS (`@get-bb/plugin-sdk@0.4.40` matches; apps/server-only change, no bump needed).

cc @ymichael — re-review ping; the diff-vs-main is now just the settlement logic + its tests.

> AGENT GENERATED

SawyerHood:

PR #2309, opened 2026-08-22
Fall back to another provider when the default's model catalog fails

### LINKED ISSUE #2306: Spawn falls back to Codex unconditionally when a project has no remembered provider, even if Codex has no models configured
Per `bb guide providers`: "If the project has no remembered choice, bb uses the explicitly requested provider or Codex, then resolves the model marked default by that provider on the target machine."

That fallback is unconditional. On a host where Codex is registered as a provider but has zero models available (`bb provider models codex` → "No models available"), any `threads.spawn` into a project with no remembered default fails outright:

```
HTTP 503: Unable to load codex models to resolve the default. Try again once the host is connected and the provider is ready.
```

This happens even when other providers (Claude Code, pi) are fully configured and working on the same host.

Expected: if Codex isn't usable (no models, not authenticated, whatever), the default-provider resolution should fall through to another configured/working provider instead of hard-failing. Codex being *the* unconditional default seems wrong for a host that never set it up in the first place.

**Repro:**
1. Have a provider like Codex registered but unconfigured (no models available on the host)
2. Have at least one other provider (e.g. Claude Code) fully working
3. `threads.spawn` (directly or via a plugin) into a project with no remembered provider/model default

**Environment:** bb `0.39.1-nightly.32567325585.1`, macOS 26.5.2

## Context: discovered via bb-plugin-dispatch

Found this while filing slogsdon/bb-plugin-dispatch#2. That plugin's destination-thread spawn (`spawnThread` in server.ts) intentionally omits `providerId`/`model` and just passes `projectId`, `prompt`, and `environment` to `threads.spawn`:

```ts
async function spawnThread(bb: BbPluginApi, projectId: string, prompt: string): Promise<string> {
  const spawned: any = await bb.sdk.threads.spawn({
    projectId,
    prompt,
    environment: { type: "host", hostId: await defaultHostId(bb), workspace: { type: "unmanaged", path: null } },
  } as any);
  ...
}
```

So the plugin itself has no opinion on provider for the destination thread — it relies entirely on bb's own default-provider resolution. The plugin's separate *intake* lane (used for classify/expand, a different hidden thread) works fine because it explicitly spreads a saved `providerId`/`model`/`executionInputSources` into its own spawn call. It's only the unconfigured default path, hit when dispatching into a project bb has no remembered choice for, that surfaces this.

Filing this here since the unconditional Codex fallback is bb's behavior, not something the plugin can work around without hardcoding a provider of its own (which would defeat the point of "inherit the host's defaults").

technicalpickles:
## What was wrong

Thread creation picked the default provider by "first available entry of the registry listing" (`requireDefaultProviderId` in `thread-default-policy.ts`), where "available" only meant the provider's plugin loaded — not that it had a usable model catalog. Codex is the first bundled provider plugin, so it always won that tiebreak on a host with no `defaultProviderId` set and no project-remembered provider. Once chosen, `resolveCatalogExecutionDefaults` in `thread-create.ts` loaded that provider's model catalog and hard-failed with a 503 if it errored or was empty, with no fallback to any other provider — even a fully working one. See #2306 for the repro.

## What changed

- `thread-default-policy.ts`: `resolveCreateThreadExecutionDefaults` now also returns `providerFallbackCandidates`, the other available providers in order — populated only when the provider came from the product default rather than an explicit request or a stored project default, so an explicit choice is never silently overridden.
- `thread-create.ts`: `resolveCatalogExecutionDefaults` tries the default provider, then each fallback candidate in turn, returning the first one with a usable model catalog. If the winner differs from the initial guess, `request.providerId` is synced so the thread record, fork resolution, and telemetry all agree on the provider actually used.
- `project-execution-defaults.ts` threads the new field through.
- No wire/protocol changes; this is server-side default-resolution policy only.

## How you verified

- Added unit coverage in `thread-default-policy.test.ts` for the new candidate ordering and for when fallback is (and isn't) eligible.
- Added an integration test in `public-threads.defaults.test.ts` reproducing the issue: Codex registered but its model catalog fails or is empty, Claude Code works — thread creation now succeeds on Claude Code instead of 503ing.
- Adjusted the existing "catalog cannot be loaded" integration test to request Codex explicitly, since an explicit provider choice is still expected to hard-fail rather than fall through.
- On Node 22.23.2, `pnpm exec turbo run typecheck --filter=@bb/server` passes.
- Turbo-backed tests pass: 79 tests across `thread-default-policy.test.ts`, `public-threads.defaults.test.ts`, `thread-execution-plan.test.ts`, and `plugin-provider-registration.test.ts`. These cover failed and empty catalogs, fallback provider persistence and dispatch, provider ordering, explicit and remembered choices, and source-aware execution defaults.
- GitHub CI passes on `283978a0ba`: build/typecheck/lint, all server/app/package/integration test jobs, Linux and macOS package smoke tests, contributor approval, and version checks.
- Live UI verification was not run.

Fixes #2306

> AGENT GENERATED


SawyerHood:

PR #2331, opened 2026-08-24
Sync a Tasks project across bb instances through git

kr3t3n:
## What was wrong

bb Tasks keeps tasks in each instance's own SQLite database and nothing syncs them, so two hosts running bb hold two unrelated boards under the same project prefix. Observed on 2026-08-23: one host held AGT-1 to AGT-7 with 28 notes while the other had no AGT project at all. A task closed on one host stays open on the other, and neither host can tell. The root cause is that the board has no shared record; a script bolted on beside the plugin proved the shape of the fix but had to identify the plugin's own audit comments by matching a body prefix, because nothing marked them structurally.

## What changed

- `plugins/tasks/sync/store-file.ts` defines the on-disk store: a versioned JSON file plus a rendered `<KEY>.md` per task. Comment identity is a 16-hex hash of the trimmed body, never an id, because ids are per instance.
- `plugins/tasks/sync/plan.ts` holds the pure logic: build the store from a board, plan an import, classify drift.
- `plugins/tasks/cli/sync.ts` adds `bb tasks sync export|import|check`. `import` is a dry run until `--apply`. It creates missing tasks at the store's task number so a key means the same task on every instance, updates changed fields, and appends notes the local board has not seen. It refuses a field update when the local task changed after the export and reports it; `--force` overwrites. `check` reports `MISSING`, `DRIFT`, `AHEAD`, `BEHIND`, and `UNTRACKED`, and exits `2` on drift. `export` refuses to write an empty board over an existing record. Store files are read and written on the invoking machine through `bb.sdk.files`, so `--machine` works like the other file-taking commands.
- Comments gain a `systemEvent` column and contract field, set on every audit comment the plugin writes (`status_changed`, `priority_changed`, `due_changed`, `labels_changed`, `dispatched`, `thread_finished`). Sync excludes those rows structurally. Propagating them made both boards drift permanently, because each import generated fresh ones on the far side. `kind: "system"` already narrowed this, but only by convention; the field is the predicate sync now depends on.
- Projects gain `syncStorePath` and `syncRole`, set with `bb tasks project create|update --sync-store <path> [--no-sync-store] --sync-role source|mirror` and shown by `project show`. A `mirror` is a supported configuration, not an error: it is the host holding a read-only deploy key on purpose, so `export` refuses there and `check` names the local edits that cannot travel back.
- `db/store.ts` accepts an explicit task `number` on create (sync only) and moves `next_task_number` forward past it.
- Docs updated in the same change: `plugins/tasks/skills/tasks/SKILL.md`, `packages/templates/src/templates/bb-guide-plugins.md`, the `bb-cli` builtin skill, and `docs/configuration.md`.

No host-daemon wire change: this touches the plugin's own database and RPC contract, not `packages/host-daemon-contract`. `HOST_DAEMON_PROTOCOL_VERSION` is unchanged.

Out of scope by design: automatic background sync, conflict resolution beyond refuse-and-report, and syncing labels, attachments, or sub-task hierarchy.

## How you verified

`plugins/tasks/sync/sync.test.ts` runs two independent plugin instances, each with its own database, over one shared directory that stands in for the git checkout. It covers convergence, key preservation, import idempotence, audit-comment exclusion in both directions, the mirror role (export refused, unsyncable local edits reported, exit 2), the empty-board refusal, and the newer-local-edit refusal with `--force`. A unit test over `checkDrift` and `planImport` pins all five drift classes.

```
pnpm exec turbo run typecheck            # 75 successful
pnpm exec turbo run build --filter=bb-plugin-tasks
cd plugins/tasks && pnpm exec vitest run --config vitest.config.ts --project bb-plugin-tasks
                                         # 15 files, 136 tests passed
```

Also exercised live against a real dev server, host daemon, and on-disk database: created a project and two tasks, changed a status, exported (the audit comment was excluded), hand-edited the store to stand in for a far instance, then `check` reported `BEHIND`/`DRIFT`/`MISSING` and exited 2, `import` dry-ran, `import --apply` applied three changes, `check` returned `drift: 0`, and a second `--apply` reported `0 change(s) applied`. Switching the project to `--sync-role mirror` made `export` exit 1 and `check` report the mirror-only note as unsyncable. A fresh project pointed at the same store refused to overwrite it.

Not verified: the two-real-host test across the `bb-agents` VM. That host runs released bb `0.39.0`, whose Tasks plugin has no `sync` command (`bb tasks sync --help` returns `unknown command: sync`), so a genuine cross-host run needs this change released or a source build deployed there.

The plugin's jsdom test files fail in this environment before any test code runs, at `vitest.setup.ts:41` with `SecurityError: Cannot initialize local storage without a --localstorage-file path`. That is pre-existing here and unrelated to this change; their project fixtures were still updated for the two new fields.

Refs TP-15.

> AGENT GENERATED

### DIFF
--- a/GAP.md
+++ b/GAP.md
@@ -1,0 +1,30 @@
+# gap-bb
+
+Private fork of [get-bb/bb](https://github.com/get-bb/bb). Product name in this checkout: **gap-bb**.
+
+## Remotes
+
+| Remote | URL | Role |
+| --- | --- | --- |
+| `origin` | `https://github.com/kr3t3n/gap-bb.git` | Private gap-bb |
+| `upstream` | `https://github.com/get-bb/bb.git` | Public bb |
+
+Keep package and CLI names as `bb` / `bb-app` so upstream merges stay clean. Brand as gap-bb in docs and companion apps only until a deliberate rename.
+
+## Sync from upstream
+
+```bash
+git fetch upstream
+git merge upstream/main
+# or: git rebase upstream/main
+git push origin main
+```
+
+## Companion iOS app
+
+See [kr3t3n/gap-bb-ios](https://github.com/kr3t3n/gap-bb-ios) — Expo client against this server's HTTP API and bb connect.
+
+## Local layout
+
+- Checkout: `~/Developer/gap-bb`
+- iOS app: `~/Developer/gap-bb-ios`
--- a/GAP.md
+++ b/GAP.md
@@ -1,13 +1,13 @@
 # gap-bb
 
-Private fork of [get-bb/bb](https://github.com/get-bb/bb). Product name in this checkout: **gap-bb**.
+Public GitHub fork of [get-bb/bb](https://github.com/get-bb/bb). Product name in this checkout: **gap-bb**.
 
 ## Remotes
 
 | Remote | URL | Role |
 | --- | --- | --- |
-| `origin` | `https://github.com/kr3t3n/gap-bb.git` | Private gap-bb |
-| `upstream` | `https://github.com/get-bb/bb.git` | Public bb |
+| `origin` | `https://github.com/kr3t3n/gap-bb.git` | Public fork (`kr3t3n/gap-bb`) |
+| `upstream` | `https://github.com/get-bb/bb.git` | Upstream bb |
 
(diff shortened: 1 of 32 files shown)


kr3t3n:
Closing: opened against the wrong repository by mistake. This belongs on the kr3t3n/gap-bb fork, not upstream.

PR #2335, opened 2026-08-24
Fall back to a resolved @get-bb/plugin-sdk entry for the legacy SDK alias in source dev mode

### LINKED ISSUE #2334: Legacy @bb/plugin-sdk alias not registered in source dev mode (pnpm dev / dev:desktop), breaking plugins on the pre-rename specifier
### Summary

Running the server from source in dev mode (`pnpm dev`, `pnpm dev:desktop`), any installed plugin that still imports the legacy `@bb/plugin-sdk` specifier (pre-rename name, see #1574) fails to load with `Cannot find module '@bb/plugin-sdk'`. The same plugin loads fine in a packaged/production build (e.g. the nightly desktop app), so this is a dev-mode-only regression in the legacy-SDK-alias compatibility shim.

### Versions and environment

- bb from source, branch `fix/provider-default-catalog-fallback` at `9375a5495464c16fd9c3d03176add0a1dfca5172` (same code path confirmed present on `main` at the same commit for `apps/server/src/services/plugins/plugin-runtime.ts`)
- Ran via `pnpm dev:desktop` (source dev server + Electron desktop shell)
- macOS (Darwin 25.5.0), Node 22.19 (required for Electron dev; confirmed same bug also applies to `pnpm dev`, which uses the same server dev entrypoint)
- Plugin: `bb-plugin-dispatch@0.1.3` (BB Community, installed from npm), which still imports `@bb/plugin-sdk` (pre-rename specifier)

### Steps to reproduce

1. `pnpm dev:desktop` (or `pnpm dev`) from a source checkout.
2. Install/have installed a plugin whose `server` entry imports the legacy specifier, e.g. `import { defineRpcContract, type BbPluginApi } from "@bb/plugin-sdk";` (this is what `bb-plugin-dispatch@0.1.3`'s `server.ts` does).
3. Open Extensions → the plugin's detail page in the dev app.

Did not reproduce with: the packaged nightly desktop app running the same plugin version — the legacy alias works there.

### Expected vs actual

Expected: per the code comment in `plugin-runtime.ts` ("Legacy alias for `@get-bb/plugin-sdk`, kept so plugin server artifacts built before the rename — and pre-rename plugin sources — still resolve the SDK"), the plugin should load successfully using the legacy alias.

Actual: the plugin's Extensions page shows:

```
Failed
The plugin couldn't start. Reload the plugin. If it still fails, remove it and install it again.
```

Server log (`~/.bb-dev/launchers/github.com-get-bb-bb/dev.log`):

```
WARN: [server] plugin dispatch: ignoring prebuilt dist/server.js (built with SDK 0.4.6, running SDK is 0.4.14) — loading from source
WARN: [server] plugin dispatch failed to load: Cannot find module '@bb/plugin-sdk'
Require stack:
- /Users/technicalpickles/.bb-dev/github.com-get-bb-bb-ec1120be8084/plugins/cache/npm/bb-plugin-dispatch/0.1.3/node_modules/bb-plugin-dispatch/server.ts
```

### Evidence

- https://github.com/get-bb/bb/blob/9375a5495464c16fd9c3d03176add0a1dfca5172/apps/server/src/services/plugins/plugin-runtime.ts#L88-L126

  The legacy alias is only registered when a build artifact exists at `dist/plugin-sdk-runtime.js` next to the running module:

  ```ts
  const pluginSdkRuntimePath = join(
    dirname(fileURLToPath(import.meta.url)),
    "plugin-sdk-runtime.js",
  );
  ...
  const pluginSdkAlias: Record<string, string> | undefined = existsSync(
    pluginSdkRuntimePath,
  )
    ? pluginSdkAliasFor(pluginSdkRuntimePath)
    : undefined;
  ```

  That file is produced only by the server's `build` script:
  https://github.com/get-bb/bb/blob/9375a5495464c16fd9c3d03176add0a1dfca5172/apps/server/package.json#L10

  ```
  "build": "... && node ../../scripts/build-node-entry.mjs ../../packages/plugin-sdk/src/index.ts dist/plugin-sdk-runtime.js",
  "dev": "node --conditions=source --import tsx scripts/dev-supervisor.mjs",
  ```

  `pnpm dev` / `pnpm dev:desktop` run the server via `tsx` straight from `src/`, never running the `build` script. `import.meta.url` for `plugin-runtime.ts` therefore resolves under `apps/server/src/services/plugins/`, where `dist/plugin-sdk-runtime.js` does not exist (confirmed: only `apps/server/dist/plugin-sdk-runtime.js` exists, nothing under `src/`). `existsSync` returns false, `pluginSdkAlias` is silently `undefined`, and the legacy specifier is never registered with jiti for the plugin's source-fallback load — even though the same jiti instance is otherwise correctly configured to use the alias when it's present:
  https://github.com/get-bb/bb/blob/9375a5495464c16fd9c3d03176add0a1dfca5172/apps/server/src/services/plugins/plugin-runtime.ts#L1584-L1592

- Confirmed no leftover real `@bb/plugin-sdk` package exists anywhere in `node_modules` — only `@get-bb/plugin-sdk` (current name, renamed in #1574 on 2026-08-13). This is not a stale namespace issue on the plugin author's end interacting with something changed recently in bb; the rename is old and unrelated to this dev-mode gap.

### What you ruled out

- Not a stale/duplicate namespace on the plugin's side causing a *new* incompatibility: `@bb/plugin-sdk` → `@get-bb/plugin-sdk` was renamed on 2026-08-13 (#1574) and is an ancestor of both `main` and this branch; nothing about the namespace changed between the nightly build the user was previously running and this checkout.
- Not specific to this feature branch: the same `plugin-runtime.ts` code (the `existsSync`-gated alias) is present unchanged on `main` at the same commit.
- Confirmed the prebuilt `dist/server.js` mismatch triggering the source-fallback path is expected/working as designed (SDK 0.4.6 vs running 0.4.14) — the bug is specifically in what happens *after* falling back to source.

Medium — blocks any plugin still on the legacy `@bb/plugin-sdk` specifier from loading at all in every source dev workflow (`pnpm dev`, `pnpm dev:desktop`), with no workaround short of forking the plugin to update its import. Likely low effort: either generate/copy `plugin-sdk-runtime.js` as part of the dev startup path, or have `resolveServerEntry`/the jiti alias fall back to resolving `@get-bb/plugin-sdk` from the workspace directly when the prebuilt runtime bundle isn't present (mirroring how source-checkout servers already resolve the workspace package per the doc comment on `PLUGIN_SDK_SPECIFIER`).

> AGENT GENERATED

technicalpickles:
## What was wrong

`pluginSdkAlias` in `apps/server/src/services/plugins/plugin-runtime.ts` only registered the legacy `@bb/plugin-sdk` specifier alias (kept since the #1574 rename to `@get-bb/plugin-sdk`) when a prebuilt `dist/plugin-sdk-runtime.js` bundle existed next to the running module. That bundle is produced only by the server's `build` script. `pnpm dev` and `pnpm dev:desktop` run the server from `src/` via `tsx` and never run `build`, so `pluginSdkAlias` silently resolved to `undefined` in every source dev workflow. Any installed plugin still importing the pre-rename `@bb/plugin-sdk` specifier (e.g. `bb-plugin-dispatch@0.1.3`) then failed to load with `Cannot find module '@bb/plugin-sdk'`, even though the same plugin loads fine in a packaged/production build (nightly, `npx bb-app`). See #2334 for the full repro and log evidence.

## What changed

- `apps/server/src/services/plugins/plugin-runtime.ts`: extracted `resolvePluginSdkAliasTarget()`. When the prebuilt runtime bundle is missing, it falls back to `import.meta.resolve("@get-bb/plugin-sdk")` (the package already resolves naturally in source dev mode per the existing doc comment) and aliases the legacy specifier to that resolved path instead of leaving it unaliased.
- `apps/server/test/services/plugins/plugin-sdk-alias.test.ts`: added a test asserting the fallback resolves to the plugin-sdk source entry when no prebuilt bundle is present (this is how the suite always runs, so it exercises the exact dev-mode gap).

No wire/protocol changes, no CLI/doc surface changes — this only affects how the server resolves an internal module alias during plugin loading.

## How you verified

- `pnpm exec turbo run typecheck --filter=@bb/server` — passes.
- `pnpm exec turbo run test --filter=@bb/server -- --run test/services/plugins/plugin-sdk-alias.test.ts` — both tests pass (the new one fails without the fix, since `resolvePluginSdkAliasTarget()` would return `undefined`).
- Manual repro: ran `pnpm dev:desktop` before the fix and confirmed `bb-plugin-dispatch@0.1.3` failed with `Cannot find module '@bb/plugin-sdk'` in the server log and showed "Failed" in the Extensions UI. Applied the fix, restarted `pnpm dev:desktop`, confirmed the same plugin now loads (`plugin dispatch@0.1.3 loaded` in the log) and shows healthy/enabled in the Extensions UI with no error banner.
- Verified against a clean install in a fresh worktree off `main` (not just the branch where I found the bug), so this isn't an artifact of leftover state.

Fixes #2334

> AGENT GENERATED

### DIFF
--- a/apps/server/src/services/plugins/plugin-runtime.ts
+++ b/apps/server/src/services/plugins/plugin-runtime.ts
@@ -119,11 +119,30 @@ export function pluginSdkAliasFor(runtimePath: string): Record<string, string> {
   };
 }
 
-const pluginSdkAlias: Record<string, string> | undefined = existsSync(
-  pluginSdkRuntimePath,
-)
-  ? pluginSdkAliasFor(pluginSdkRuntimePath)
-  : undefined;
+/**
+ * The packaged runtime bundle only exists after the server `build` script
+ * runs. Source dev servers (`pnpm dev`, `pnpm dev:desktop`) run straight from
+ * `src/` via tsx and never produce it, which left the legacy alias
+ * unregistered there and broke any installed plugin still on the pre-rename
+ * specifier (see #2334). `@get-bb/plugin-sdk` itself resolves naturally in
+ * that mode (doc comment above), so fall back to its resolved entry point and
+ * alias the legacy specifier to that instead of leaving it unaliased.
+ */
+export function resolvePluginSdkAliasTarget(): string | undefined {
+  if (existsSync(pluginSdkRuntimePath)) {
+    return pluginSdkRuntimePath;
+  }
+  try {
+    return fileURLToPath(import.meta.resolve(PLUGIN_SDK_SPECIFIER));
+  } catch {
+    return undefined;
+  }
+}
+
+const pluginSdkAlias: Record<string, string> | undefined = (() => {
+  const target = resolvePluginSdkAliasTarget();
+  return target === undefined ? undefined : pluginSdkAliasFor(target);
+})();
 
 /**
  * Per-root reload generation for mutable (path:/source-builtin) plugin trees.
--- a/apps/server/test/services/plugins/plugin-sdk-alias.test.ts
+++ b/apps/server/test/services/plugins/plugin-sdk-alias.test.ts
@@ -1,5 +1,8 @@
 import { describe, expect, it } from "vitest";
-import { pluginSdkAliasFor } from "../../../src/services/plugins/plugin-runtime.js";
+import {
+  pluginSdkAliasFor,
+  resolvePluginSdkAliasTarget,
+} from "../../../src/services/plugins/plugin-runtime.js";
 
 describe("pluginSdkAliasFor", () => {
   it("resolves the pre-rename specifier to the same SDK runtime bundle", () => {
@@ -10,5 +13,20 @@ describe("pluginSdkAliasFor", () => {
 
     expect(alias["@get-bb/plugin-sdk"]).toBe("/srv/plugin-sdk-runtime.js");
     expect(alias["@bb/plugin-sdk"]).toBe("/srv/plugin-sdk-runtime.js");
+  });
+});
+
+describe("resolvePluginSdkAliasTarget", () => {
+  it("falls back to the resolved @get-bb/plugin-sdk entry when no prebuilt runtime bundle exists (#2334)", () => {
+    // This test runs the server from source (no `dist/plugin-sdk-runtime.js`
+    // next to plugin-runtime.ts), the same way `pnpm dev`/`pnpm dev:desktop`
+    // do. Without the fallback this returns undefined, the legacy
+    // "@bb/plugin-sdk" specifier never gets aliased, and any installed
+    // plugin still on that pre-rename import fails with "Cannot find
+    // module '@bb/plugin-sdk'".
+    const target = resolvePluginSdkAliasTarget();
+
+    expect(target).toBeDefined();
+    expect(target).toMatch(/plugin-sdk[/\\]src[/\\]index\.ts$/);
   });
 });


SawyerHood:

PR #2344, opened 2026-08-24
Confirm before archiving active thread work

jshph:
## What was wrong

The sidebar's hover-revealed archive button intentionally makes routine cleanup one click, but it used the same immediate archive path even when the selected thread or one of its visible children was still working. A small targeting mistake could therefore interrupt active work without giving the user a chance to reconsider.

## What changed

Thread archive actions now inspect BB's existing React Query cache synchronously at click time. Idle thread trees retain the current one-click behavior. When the selected thread or a cached child has an active runtime, workflow, background agent or command, plan, or goal, BB opens an “Archive active work?” confirmation before running the existing archive mutation.

The guard lives in `ThreadActionsProvider`, so the sidebar quick action, overflow menu, keyboard command, and plugin-exposed single-thread action share the same behavior. It is deliberately a best-effort UI guard: it adds no preflight request, server enforcement, wire contract, protocol bump, or new subscription, preserving the sidebar's existing responsiveness and archive lifecycle.

## How you verified

Added regression coverage for idle trees, active parents, active children, withholding the archive mutation until confirmation, and performing the confirmed archive exactly once.

- `pnpm exec turbo run test --filter=@bb/app --force -- src/components/thread/ThreadActionsProvider.test.tsx src/hooks/cache-owners/thread-archive-cache.test.ts`
- Full `@bb/app` suite on the feature worktree before rebasing: 430 files passed; 3,338 tests passed and 3 skipped.
- Independent code review of cache behavior, shared action routing, dialog semantics, affected-set limitations, and test coverage.

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/dialogs/ThreadArchiveDialog.tsx
+++ b/apps/app/src/components/dialogs/ThreadArchiveDialog.tsx
@@ -1,0 +1,39 @@
+import type { Thread } from "@bb/domain";
+import {
+  ConfirmDeleteDialog,
+  ConfirmDeleteDialogContent,
+} from "./ConfirmDeleteDialog";
+
+export interface ThreadArchiveDialogTarget {
+  thread: Thread;
+  workingThreadCount: number;
+}
+
+interface ThreadArchiveDialogProps {
+  target: ThreadArchiveDialogTarget | null;
+  pending: boolean;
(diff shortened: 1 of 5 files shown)


SawyerHood:

PR #2350, opened 2026-08-24
perf: patch mobile list rows from status-changed pushes instead of refetching (M1)

SawyerHood:
## What was wrong

The Expo app refetched the whole sidebar on every status change (an internal performance sweep, finding M1): `status-changed` is in `THREAD_LIST_AFFECTING_KINDS`, so each push invalidated the sidebar query (`staleTime: Infinity`, 14 observers) and triggered a full `GET /sidebar-bootstrap` — 133 KB at least twice per agent turn per thread. The server already attaches `metadata.statusChange` (#2169) and the app parsed it, then never read it.

## What changed

`apps/mobile/src/data/threads/thread-list-cache.ts`, `apps/mobile/src/lib/query/realtime-invalidation.ts` (mirrors the web's #2169 path):

- `updateCachedThreadListStatusState` patches the matching row in every cached thread list and the sidebar with the web's identity guard (`if (!list.some(t => t.id === threadId)) return list`). Lists the mapper leaves untouched are not written at all; real writes keep the query's `dataUpdatedAt` and re-arm a pending invalidation, so inactive lists do not lose their refetch-on-remount flag.
- A `status-changed` push that carries the row no longer invalidates the list keys (only the thread and search keys); bare pushes (stop/interrupt, provisioning, cleanup) still refetch the lists.
- List fetches in flight when a patch lands are aborted synchronously (reverting to the already-patched state, so the query returns to `idle`) and restarted once per debounced flush; a burst of pushes restarts the 133 KB bootstrap once. Exact restarts are skipped when a prefix invalidation in the same flush covers them, and when the query has no active observer (it stays stale and idle for the next mount).
- Client-only: no server, daemon, SDK or contract change; no `HOST_DAEMON_PROTOCOL_VERSION` bump.

## How you verified

New tests in `apps/mobile/src/lib/query/realtime-invalidation.test.ts` and `apps/mobile/src/data/threads/thread-list-cache.test.ts`, each failing on the pre-fix code: row patch through the real socket bridge (sidebar and flat list rows updated, other project's `threads` array identity kept); key mapping (row-carrying push → thread + search keys only; bare push and mixed kinds → list keys); no-op patch leaves query state, `dataUpdatedAt` and `isInvalidated` untouched; errored invalidated sidebar stays invalidated; in-flight sidebar refetch aborted once and restarted on the flush with the stale body dropped; eight-push burst → one abort, one restart; observer leaving before the flush leaves the query idle and collectable; disabled observer refetches on re-enable; prefix invalidation subsumes an exact restart. Verified against `@tanstack/query-core` 5.90.20 semantics (`revert`, `silent`, `refetchType: "none"`, `isActive`).

On this branch: `pnpm exec turbo run typecheck` (76/76), `pnpm exec turbo run lint`, `pnpm exec turbo run test` (all packages green). Not done: a device run (tests are node/jsdom, not Hermes). Suggested manual check before merge: watch the sidebar through an agent turn, open the archived list and the @-typeahead, background/foreground the app.

Fixes: no issue — source is an internal performance sweep (finding M1).

> AGENT GENERATED

### DIFF
--- a/apps/mobile/src/data/README.md
+++ b/apps/mobile/src/data/README.md
@@ -18,7 +18,11 @@ Conventions:
   `read-tracking.ts`) never import react-native and are vitest-tested (node
   env); RN adapters (MMKV storage) live in their own small files.
 - Query keys come from `@/lib/query/query-keys` and every key is mapped in
-  `@/lib/query/realtime-invalidation.ts`. Add both when adding a query.
+  `@/lib/query/realtime-invalidation.ts`. Add both when adding a query. A
+  `status-changed` push that carries `metadata.statusChange` patches the
+  cached list / sidebar rows in place
+  (`threads/thread-list-cache.ts` `updateCachedThreadListStatusState`) and
+  cancels list fetches in flight; a bare one refetches the lists.
 - Realtime subscriptions are held through `shared/use-realtime-subscription`
   (refcounted per target; the last unmount releases the server subscription).
 - `thread-detail/` mirrors the web thread-detail queries: the bootstrap
--- a/apps/mobile/src/data/threads/thread-list-cache.test.ts
+++ b/apps/mobile/src/data/threads/thread-list-cache.test.ts
@@ -1,16 +1,22 @@
 import {
   PERSONAL_PROJECT_ID,
+  type ThreadListEntry,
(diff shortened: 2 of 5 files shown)


SawyerHood:

PR #2385, opened 2026-08-25
Calm the iOS shell-geometry handler and the timeline resize cascade

vburojevic:
## What was wrong

Two independent audits traced the worst mobile hangs (38.9s/30.7s/28.5s traces on `div[app-layout-content-shell]`) to the same mechanism: on every iOS visual-viewport tick (keyboard animation, URL-bar collapse) the shell hook forced a full-document layout (`document.body.clientHeight`) inside a rAF and rewrote shell `top`/`height` plus the inherited `--bb-shell-height` on an app-root ancestor, invalidating computed style for the entire 2,000–5,200-node tree at animation cadence. Each content-size change then re-entered a second cascade: the scroll body's ResizeObserver refreshed geometry live, the observed frame's bottom restore read again, and each of the three rAF settle-tail frames forced another layout — up to five forced layouts per size change at streaming cadence. Every scroll event also wrote `data-scrollbar-scrolling` (desktop-scrollbar-only CSS) and 100ms anchor samples re-ran `querySelectorAll` over the scroll subtree.

## What changed

- `useMobileVisualViewportHeight.ts`: style writes skipped when a pass recomputes unchanged geometry; `document.body.clientHeight` cached and re-read only on `window.resize`/`orientationchange`/`focusin` (never on visualViewport ticks); `visualViewport.scroll` ticks gated to run only when an override is applied or a keyboard-focusable element is focused. Embedded-browser overrides keep tracking pans; the focus-reveal pan compensation, native-layout early-exit, focusout fast-restore, pinch-zoom guard, and rAF coalescing are unchanged; programmatic focus still triggers one freshly measured pass (new test).
- `bottom-anchored-scroll-body.tsx`: the ResizeObserver path refreshes the cached max offset from the observer's own box sizes when entries are present; settle-tail frames reuse the cache with at most one live verification read after a corrected drift. Coarse pointers skip the scrollbar-attribute write, capture anchors at 250ms with a cached row NodeList invalidated by the ResizeObserver.
- `height-transition.tsx`: observer callbacks size the wrapper from the entry's `borderBoxSize`.

## How you verified

- New tests proven fail-before/pass-after: idempotent-write (`setProperty` spy), containing-block read-count, scroll-tick gating, settle-tail read-count, entry-derived cache, border-box sizing, coarse-pointer trims, cached-NodeList invalidation; a programmatic-focus guard passes in both states by design.
- `bottom-anchored-scroll-body.scroll-preservation.test.tsx` (the #2280/#2300 contract) passes with zero diff.
- `pnpm exec turbo run typecheck --filter=@bb/app` exit 0; full `@bb/app` suite 433 files / 3,352 passed.
- Merged-tree WebKit QA (iPhone 390×844@3, perf fixture DB): long-thread scroll, keyboard open/close via composer tap, no persistent blanking, zero console errors.

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/layout/useMobileVisualViewportHeight.test.tsx
+++ b/apps/app/src/components/layout/useMobileVisualViewportHeight.test.tsx
@@ -108,6 +108,18 @@ function withElementClientHeight(
   }
 }
 
+// Waits out one scheduled rAF pass so "the pass ran and did nothing" is
+// distinguishable from "the pass has not run yet".
+async function flushScheduledViewportPass() {
+  await act(async () => {
+    await new Promise<void>((resolve) => {
+      window.requestAnimationFrame(() => {
+        window.requestAnimationFrame(() => resolve());
+      });
+    });
+  });
+}
+
 beforeEach(() => {
   vi.spyOn(window, "scrollTo").mockImplementation(() => {});
 });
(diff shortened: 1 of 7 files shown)


SawyerHood:

Fix landed there for this PR:

- The `data-scrollbar-scrolling` write is back on every pointer, idempotent (one write per scroll burst): `.thread-scrollbar`/`.transient-scrollbar` are not pointer-gated, so skipping it on touch hid the thumb on Android and iOS.
- The scroll-anchor row cache is bypassed when the top-level list holds a `[data-timeline-virtual-spacer]`; on a windowed timeline the last row is force-mounted, so the edge-row connectivity check never invalidated it.
- The scroll-preservation contract suite now exercises the entries-derived resize path, and the scroll-gate test pins the keyboard-pan compensation.

> AGENT GENERATED

PR #2386, opened 2026-08-25
Let taps paint: transition-priority navigation, deferred expanders and sidebar realization

vburojevic:
## What was wrong

Interaction taps on mobile web ran their entire React consequence as one discrete synchronous commit, so the browser could not paint tap feedback first. The perf-watch ledger attributes 671s/267 stalls over 7 days to `click on button.inline-flex` (the ExpandablePanel header and shared Button class) and 858s/364 stalls to `after navigate`: the expander body materialized inside the click's render, `openMobileSidebar` flushed the whole ProjectList/ThreadRow subtree before the slide's first frame, navigation exposed no pending signal, and the split-pane maximize toggle rewrote scrollLeft/scrollTop on every tracked element for 30 unconditional rAF frames.

## What changed

- `app-route-anchor.tsx`: `navigateRoute` wraps navigation in `useTransition`'s `startTransition`; the navigate context identity is unchanged, and `isPending` is exposed through a separate `RouteNavigationPendingContext` + `useIsRouteNavigationPending()` so navigate consumers don't re-render per navigation. `RootComposeMobileRecents.tsx` swaps its raw `<Link>` for `RouteAnchor`. (react-router v7 already transitions router state; the delta is the pending signal plus the remaining urgent work riding the same transition.)
- `disclosure.tsx`: the expander body realizes off `useDeferredValue(isExpanded)` — caret/header stay urgent, the body mounts in a follow-up interruptible commit; closing-body retention and the layout-animation signal unchanged.
- `sidebar.tsx`: `realizeMobileSidebar()` runs inside `React.startTransition`; drag-style write order unchanged; the settle commit's render-phase latch still covers the race. The file's `flushSync` sites were reviewed individually and deliberately left alone.
- `SplitThreadArea.tsx`: `restore()` compares before writing and returns whether anything moved; the loop stops after the first zero-correction frame, capped at 5 frames.

## How you verified

- New tests, each proven fail-before/pass-after: transition navigation samples `{isPending, old pathname}` in the tap's commit; the deferred body is absent from the urgent flush and lands in the deferred commit (plus a closing-body retention guard); the sidebar realize commit is absent from the tap's flush; settled restore performs 0 scroll writes (before: 31), an adversarial scroller is corrected ≤6 times.
- `pnpm exec turbo run typecheck --filter=@bb/app` exit 0; focused suites 67 files / 620 tests pass; lint 0 errors.
- Merged-tree WebKit QA (iPhone viewport): section expand, sidebar open/close, thread navigation — all pass with zero console errors.

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/ui/app-route-anchor.test.tsx
+++ b/apps/app/src/components/ui/app-route-anchor.test.tsx
@@ -1,0 +1,65 @@
+// @vitest-environment jsdom
+
+import { cleanup, fireEvent, render, screen } from "@testing-library/react";
+import { MemoryRouter, useLocation } from "react-router-dom";
+import { afterEach, describe, expect, it } from "vitest";
+import {
+  RouteAnchor,
+  RouteNavigationProvider,
+  useIsRouteNavigationPending,
+} from "./app-route-anchor";
+
+afterEach(() => {
+  cleanup();
+});
+
+interface NavigationSample {
+  isPending: boolean;
+  pathname: string;
+}
+
+const samples: NavigationSample[] = [];
+
+/**
+ * Records every committed (isPending, pathname) pair. The pairs prove commit
+ * ordering: a `{ isPending: true, pathname: <old> }` sample means the pending
+ * flag painted while the previous route was still on screen, i.e. the tap's
+ * event did not flush the destination route synchronously.
+ */
+function NavigationSampler() {
+  const isPending = useIsRouteNavigationPending();
+  const { pathname } = useLocation();
+  samples.push({ isPending, pathname });
+  return null;
+}
+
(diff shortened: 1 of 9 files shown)


SawyerHood:

Fix landed there for this PR:

- In `disclosure.tsx` the region's content branch, height sync, transition classes, deadline and in-flight accounting now key on the deferred expanded value, so the collapsed preview stays until the body exists and the tween starts from the real body; re-expanding inside the 200 ms close window keeps the retained body.
- Left as-is: the transition-priority navigation is a no-op because react-router already transitions; the deferred-body starvation under heavy streaming is an open trade-off noted in #2435.

> AGENT GENERATED

PR #2281, opened 2026-08-22
Make the plugin composer host pull-based so keystrokes stop re-rendering the thread shell

vburojevic:
## What was wrong
ThreadDetailPromptArea rebuilt its published PluginComposerHost every keystroke (the draft rode on the host object), so the identity-change notify re-rendered ThreadDetailSecondaryContentBody → SecondaryPanelLayout → ThreadTimelinePane → PageShell per character typed. Production telemetry attributed the worst mobile hang (19.6s after keydown) to this class.

## What changed
PluginComposerHost drops the draft field; the contract is an identity-stable host exposing getCurrent() + subscribeDraft(), with usePluginComposerHostDraft as the only reactive read. Store-backed hosts (thread, new-thread) subscribe through the prompt-draft store; state-backed hosts (queued, sent-message, embedded chat) notify from layout effects ordered after the identity-ref sync — a review-caught ordering bug where a thread switch could briefly expose the previous thread's draft is fixed and regression-tested both ways. All hosts (including sent-message) honor the identity-stable contract. Plugin SDK exports (packages/plugin-sdk) are untouched; useComposerView values and timing are unchanged, and the SDK version-lockstep gate confirms no surface files changed.

## How verified
336 targeted tests / 17 files: keystroke render-count probes (shell must not re-render while typing; draft consumers must update), the thread-switch ordering regression (fails on the broken ordering), secondary-content probes, plus upstream's escape/collapse tests in the touched areas. Typecheck + oxlint at clean-tree baseline; adversarially reviewed, rebased over the zen-mode removal (#2254) adopting upstream semantics.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2278, opened 2026-08-22
Gate the default focus refetch on lost realtime coverage

vburojevic:
## What was wrong
Every window focus refetched every active query older than staleTime 2s. On phones each unlock/app-switch fired that full wave on top of the realtime reconnect wave (visible-probe + reconnect watermark + hidden-change replay), duplicating catch-up in the first interactive frames.

## What changed
`createAppQueryClient` gains an injectable `shouldRefetchOnWindowFocus`; main gates it on `wsManager.getConnectionState() !== "connected"`. While connected, realtime owns freshness; in connecting/reconnecting the focus refetch remains the fallback. Per-query policies (query-policies.ts) still win.

## Known narrow trade (disclosed for review)
A half-open socket reports "connected" up to the 5s pong timeout, and queries with neither realtime coverage nor a focus policy (e.g. CLI skills status) skip one focus wave when parked across an unlock; refetchOnMount repairs them on navigation.

## How verified
Gated + ungated focus tests (5/5); typecheck + oxlint clean; adversarially reviewed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2388, opened 2026-08-25
Realtime round 2: status-change metadata everywhere, reconnect gating, coalesced fallback refetches

vburojevic:
## What was wrong

Four realtime fan-outs survived the #1880–#1901 and #2277–#2281 rounds. A `status-changed` push without `statusChange` metadata makes every client refetch every active thread list + sidebar bootstrap (~1 KB per unarchived thread) on the un-debounced immediate path — and the host-connectivity fan-out published exactly such bare pushes once per thread on the host. The same immediate path cancelled in-flight search requests, which can starve an open search on a slow link. `refetchOnReconnect` was left at TanStack's default `true`, so every mobile-Safari `online` blip re-fired the wave #2278 gated for focus. And each in-flight timeline row owned its own 1 Hz `setInterval` while streaming reconciled the unwindowed list up to 20×/s on phones.

## What changed

- Server: `notifyHostThreadRuntimeStatusChanged` (daemon close, disconnect grace, host removal) and the post-commit `interruptActiveThreads` publish attach the row snapshot via the existing `buildThreadStatusChangeMetadata` builder; in-transaction sites (buffered `DbNotifier`, no runtime hub) are annotated as deliberately bare. `thread-send`/`parent-system-messages`/queued auto-send already attached it since #2169. No daemon wire change.
- Client registry: the metadata-less fallback routes through the throttled active-refetch machinery at 1 Hz per query (stale immediately, later pushes coalesce into one trailing refetch, archived pages stay stale-only); the metadata branch invalidates the search prefix with `cancelRefetch: false`.
- `createAppQueryClient` applies the lost-realtime-coverage gate to `refetchOnReconnect` (same predicate as the focus gate); the two policies with explicit `refetchOnReconnect: true` keep it with re-justification comments.
- `LiveDurationText` rides the shared `useSecondTick` ticker, which now stops while the document is hidden and ticks immediately on resume. The realtime invalidation debounce reads `matchMedia("(pointer: coarse)")` once at module init: touch devices widen 50/200ms to 150/400ms; desktop cadence unchanged (pinned by test).

## How you verified

New tests proven fail-before/pass-after: daemon-close and host-removal pushes carry `statusChange`; two bare pushes within 1s produce one active refetch; a status patch no longer aborts an in-flight search holding cached data; an `online` event doesn't refetch while realtime coverage is intact and does once lost; the shared ticker pauses hidden. `pnpm exec turbo run typecheck --filter=@bb/app --filter=@bb/server` exit 0; server suites 63 files / 502 tests; app hooks suites 59 files / 376; timeline+banner suites 37 files / 292. Green on the merged integration tree (server 2,044).

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/thread/timeline/TimelineTitleView.tsx
+++ b/apps/app/src/components/thread/timeline/TimelineTitleView.tsx
@@ -1,4 +1,4 @@
-import { Fragment, useEffect, useState } from "react";
+import { Fragment } from "react";
 import type { KeyboardEvent, MouseEvent, ReactNode } from "react";
 import {
   assertNever,
@@ -15,6 +15,7 @@ import {
 import { cn } from "@bb/shared-ui/lib/utils";
 import { DiffStatsTally } from "@/components/ui/diff-stats-tally.js";
 import { RouteAnchor } from "@/components/ui/app-route-anchor.js";
+import { useSecondTick } from "@/hooks/useSecondTick";
 
 /**
  * Resolves a title's declared action to a click callback. Return `null` to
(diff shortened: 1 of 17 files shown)


SawyerHood:

Fix landed there for this PR:

- Daemon close, disconnect grace and host removal build `statusChange` metadata only for active threads, fetched in batched queries (was ~5–7 synchronous queries for every thread on the host, 0.66–0.87 s measured); idle threads keep the pre-PR bare push.
- The `statusChange` search invalidation keeps `cancelRefetch: false` and schedules one trailing refetch after the in-flight search settles.
- The coarse-pointer debounce test moved to its own file so the 65-test suite runs in the shared vitest worker again.

> AGENT GENERATED

PR #2389, opened 2026-08-25
Boot and delivery quick wins for the relayed mobile path

vburojevic:
## What was wrong

A thread page on the relayed mobile path (CF worker → Durable Object → tunnel WebSocket → laptop) was 113 requests / ~3.5 MB raw JS, and the tunnel serializes responses FIFO — so discovery order and request count are the latency. Four root causes: the render-blocking stylesheet was the 68th resource discovered in the built `index.html` (font preload dead last); no chunk-merging config existed (70 boot chunks, half carrying ~2% of the bytes, 18 below the precompress floor); `index.html` was `no-cache` — which the connect worker treats as never-cacheable — with no brotli sidecar, and the SPA fallback bypassed the precompressed path entirely; the connect gate's per-isolate caches stored only settled values, so a cold isolate's burst paid one D1 round trip per request instead of per key.

## What changed

- `vite-font-preload.ts`: the stylesheet (with `fetchpriority="high"`) and font preload now precede the entry script and modulepreload block; the transform fails the build if the stylesheet would ever precede the pre-paint theme script.
- `vite.config.ts`: `advancedChunks` with two `$initial`-tagged groups (vendor/app) merges the boot graph 70 → 3 chunks (boot brotli 443.0 → 381.2 KB); lazy-route and on-demand facades untouched, all budget gates hold.
- `precompress-app-dist.mjs`: `.html` added (`index.html.br` ships); the server's SPA fallback unified with the sidecar-aware serving path.
- Edge-cacheable shell: the server serves `max-age=300, must-revalidate` plus a weak build-id ETag (sha256 of the served file — index.html embeds every hashed asset URL, so its content hash IS the build id) and answers If-None-Match with 304. The connect worker keeps the last confirmed document in `caches.default` and revalidates through the tunnel on every navigation, so new builds still land on the next navigation; storage bounded ≤300s.
- `session.ts`: label/session caches store the in-flight promise (single-flight); rejected lookups evicted on settle; TTLs unchanged. Rebased over #2357 — its session-refresh semantics preserved (identity-guarded settle handlers; edge-served shells deliberately don't count as session activity).
- Plugin boot cap (planned step) was already shipped by #1896 — verified, contract tests pass unmodified.

Deploy note: the connect worker deploys separately; either order is safe (bounded ≤300s skew), worker-first avoids even that.

## How you verified

Fail-before/pass-after: emitted head-order test against the real `dist/index.html` (`expected 9654 to be less than 3861` before); single-flight burst tests (`expected 6 to be 1` D1 selects) for both lookups. Document cache proven in real workerd via the tunnel fixture (repeat navigation = edge body + 304 on the tunnel; build change ships next navigation; visitor 304 relayed). Bundle budget OK before/after (boot 70→3 chunks). After the #2357 rebase: `@bb/connect` 119/119 including main's new session-refresh suite; head-order/precompress/plugin-frontend suites 54/54; `check:bundle` OK. Green on the merged integration tree.

> AGENT GENERATED

### DIFF
--- a/apps/app/src/precompress-app-dist.test.ts
+++ b/apps/app/src/precompress-app-dist.test.ts
@@ -28,10 +28,14 @@ describe("app asset precompression", () => {
     const distDir = await mkdtemp(resolve(tmpdir(), "bb-precompress-test-"));
     const compressibleBody = Buffer.from("compressible bb asset\n".repeat(400));
     const assetPath = resolve(distDir, "app.js");
+    // The document itself: without a sidecar every cold navigation on the
+    // relayed mobile path ships the shell uncompressed through the tunnel.
+    const documentPath = resolve(distDir, "index.html");
     const smallPath = resolve(distDir, "small.js");
     const binaryPath = resolve(distDir, "image.png");
     await Promise.all([
       writeFile(assetPath, compressibleBody),
+      writeFile(documentPath, compressibleBody),
       writeFile(smallPath, "small"),
       writeFile(binaryPath, compressibleBody),
     ]);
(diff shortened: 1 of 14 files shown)


SawyerHood:

Fix landed there for this PR:

- The app shell is served `Cache-Control: no-cache` + weak build-id ETag again (`max-age=300` let browsers and the Electron window boot a stale shell whose hashed assets 404 for up to five minutes after an update). The connect worker's revalidated-shell contract is now `no-cache` + ETag: the edge copy is stored with an internal 300 s bound, served only after the origin's 304, and the visitor always receives the origin's `no-cache`. Proven in real workerd.
- `document-cache.test.ts`: each test stores its own edge copy. `bundle-budget.json`: `maxBootBrotliBytes` lowered to 429,072 (10% above the new payload); `maxBootBytes` unchanged. `vite-font-preload.test.ts`: head order pinned against a synthetic document, since the dist-gated suite is skipped in CI.

> AGENT GENERATED

PR #2390, opened 2026-08-25
Move the tiptap composer behind a first-focus handoff

vburojevic:
## What was wrong

The tiptap/prosemirror editor graph sat in `SplitWorkspaceRoute`'s static closure: 487 KB raw / 121 KB brotli parsed and compiled before route content paints on every thread, compose, and plugin-panel page, for an editor the user has not tapped. `bundle-budget.json` already named the intended fix — "allowed until it moves behind a first-focus handoff".

## What changed

- `PromptBoxShell.tsx` (new): the composer's exact closed-state chrome with zero tiptap imports. Its preview is a `.ProseMirror`-shaped `contenteditable="false"` surface, so existing placeholder/compact-preview CSS applies unchanged; a saved draft shows immediately.
- `PromptBox.tsx` (new): renders the shell until the first compose intent (tap/focus, paste, drop, plus-menu action, programmatic focus request, desktop autoFocus), then dynamically imports and permanently mounts `PromptBoxInternal`. An invisible interim textarea owns the interim: the tap focuses it natively (soft keyboard opens in-gesture), keystrokes and pastes flush live into the draft, the swap defers during IME composition (never cutting a composition), and the editor takes focus at mount. The chunk prefetches at idle / pointerenter, so the handoff is normally mount-only. Manual `import()` instead of `React.lazy` so the swap can wait out IME composition.
- `bundle-budget.json`: all 32 observed `@tiptap/*`/`prosemirror-*` packages gated on `PromptBoxInternal.tsx` in `onDemandPackages`; route closure budgets re-ratcheted 2,268,430→1,765,479 raw, 605,332→489,779 brotli so regression fails CI.

Honest sizing note: the measured closure saving is 451 KB raw / 104 KB brotli — the old 1.07 MB chunk was only ~49% tiptap; the rest is unrelated code that stays eager.

## How you verified

- New handoff suite (`PromptBox.test.tsx`, 8 tests) fails 7/8 against the eager architecture and passes 8/8 after: draft survives first tap, keystrokes/paste before mount land in the editor, file paste attaches, `focusEnd()`/`focusEndKey` realize and focus, Enter submits from the shell.
- Typecheck/lint green; promptbox+thread-detail suites 485 passed; full `@bb/app` suite 3,350 passed.
- Build + `check:bundle` green: closure 538.5→434.8 KB brotli; `why-eager` reports 0 eagerly reachable tiptap/prosemirror modules; a simulated static edge fails the check.
- Merged-tree WebKit QA (iPhone viewport): shell tap mounts tiptap, typing lands, draft cleared on send. Reviewer to exercise CJK IME + autocorrect on a physical device.

> AGENT GENERATED


SawyerHood:

## Blocking

1. **A failed editor-chunk import is cached forever; the composer never realizes again** — `apps/app/src/components/promptbox/PromptBox.tsx:54`. `loadPromptBoxInternalModule` caches a rejected `import()`, and `requestRealize` latches `realizeRequestedRef.current = true` before the load with no `.catch`. One failed fetch (flaky network, or a stale tab after a deploy rotated chunk hashes) leaves the composer dead for the life of the page, and the rejection escapes as an unhandled promise rejection. The repo already has `createRetryingModuleLoader` for this. Fix: use it, reset the latch on failure so the next intent retries, and `.catch` the prefetch paths. Add a test that rejects the first import and asserts a second focus mounts the editor.

## Fix together with the above (each confirmed by experiment)

- **Safari IME**: confirming a CJK candidate with Enter submits the draft from the interim textarea (`PromptBox.tsx:364`). The interim checks only `isComposing`; the mounted editor has the 500 ms post-`compositionend` window, the `keyCode === 229` check, and the WeakSet. Mirror them, and seed `compositionEndedAtRef` when the swap is triggered from `compositionend`.
- **The swap replaces the whole chrome at an arbitrary time** (promise resolution), closing an open model picker, plus menu, or voice bar under the user's finger (`PromptBox.tsx:486`). Keep the chrome stable and swap only the editor region, or gate the swap on no open overlay.
- **Shell attachment thumbnails are inert** (lightbox hard-wired to `null`) and attach-from-menu never realizes the editor (`PromptBoxShell.tsx:594`).
- **The first plus-menu action before mount drops focus to `<body>`** (`PromptBox.tsx:477`); the eager editor focused the caret after the inserted pill.
- **The first tap into a saved draft always puts the caret at the end**, not at the tap point, and saved text cannot be selected until the editor exists (`PromptBox.tsx:520`).
- **A restored draft is hidden from assistive tech**: the preview is `aria-hidden` and the only accessible textbox is the always-empty interim textarea (`PromptBoxShell.tsx:541`).
- **All 8 handoff tests run with a fine pointer** (the test setup's `matchMedia` polyfill returns `false`), so the coarse-pointer path this PR exists for has zero coverage (`PromptBox.test.tsx:113`). Reverting the coarse-pointer clause in the handoff effect keeps the suite green.

## Low

- The shell submit button drops the editor's iOS focus guard and pointer-vs-keyboard blur distinction (`PromptBoxShell.tsx:759`).
- The form-level drop handler `preventDefault`s text drops into the interim textarea (`PromptBoxShell.tsx:395`).
- Ctrl/Alt/Meta+Enter insert a newline in the interim where the editor submits (`PromptBox.tsx:388`); iPad hardware Enter diverges from the editor's special case (`PromptBox.tsx:386`); `aria-keyshortcuts` is missing (`PromptBox.tsx:506`).
- `COMPOSER_EDITOR_MAX_HEIGHT_BY_LAYOUT` says "shared with ComposerEditorSlot" but that file keeps its own copy (`PromptBoxShell.tsx:85`).

## Open risk

No one verified on a device that the iOS keyboard stays up across the swap; the refuters could only run desktop WebKit. Test that on a phone before investing in the tail above; if the keyboard drops, the swap must happen synchronously inside the tap gesture.

## Suggested salvage

The design is right and the bundle win is real (measured 104 KB brotli off the route closure on the merged tree). Fix the loader, the IME guard, and chrome stability; add coarse-pointer tests; then reopen after a device test.

> AGENT GENERATED

PR #2392, opened 2026-08-25
[stacked on #2385 + #2386] Consolidate per-row ResizeObservers into shared read/write-phased observers

vburojevic:
**STACKED PR** — contains #2385 (ios-shell-geometry) and #2386 (interaction-yielding); only the top 3 commits are new. Do not merge before both bases.

## What was wrong

Every collapsible timeline row installed its own ResizeObserver whose callback interleaved a layout read (`offsetHeight`) with a style write, and every message mounted two more observers for its action bar — one never read on the mobile-overflow path, the other measuring a column width identical for every row. When one event resizes the timeline (iOS keyboard, drawer, orientation, `--bb-shell-height`), WebKit delivers all observers in one batch and the interleaved pattern defeats batching: N rows ⇒ N synchronous layout passes. The shared selection pointer listeners were also registered non-passive despite never calling `preventDefault`.

## What changed

- New `apps/app/src/lib/shared-resize-observer.ts`: one module-level observer with registered `{read(entry), write(value)}` phases — all reads complete before any write, bounding a batch at one forced layout. `ExpandablePanel`, `HeightTransition`, and `AutoHeightContainer` migrated onto it; panel sizing comes from the entry's border box (no layout read), preserving the `transitionDuration` snap, the deferred-body realization (#2386), and the borderBoxSize sizing (#2385).
- `useMeasuredWidth` gains `enabled` (observer skipped, hook order stable); the mobile-overflow bar branch drops its slot observer; the top-level `TimelineRowsList` measures the message-column width once and shares it via `MessageColumnWidthContext` (nested lists shadow with `null` and self-measure as before). Per bar: 2 observers → ≤1 desktop, 0 mobile-overflow.
- `{ passive: true }` on SelectableMessageProse's shared pointer listeners (handlers verified free of `preventDefault`).

Merge-order note vs #2391 (timeline windowing): both touch `ThreadTimelineRows.tsx` in one JSX hunk. Verified resolution (used on the integration branch): take this PR's structure verbatim (provider wrapper + `messageColumnWidthSourceRef` + re-indentation) and inside it set `minItemCount={spacing === "top-level" ? (isCompactViewport ? 16 : 60) : 20}` from #2391. Whichever merges second applies exactly that.

## How you verified

- Typecheck exit 0; focused suites 63 files / 520 tests pass, including the unmodified scroll-preservation, disclosure, and height-transition contract suites.
- New tests proven fail-before/pass-after by swapping sources to base: shared-observer phase ordering + two-panel single-observer integration (`length of 1 but got 2` before), action-bar observer-budget + list-level wiring (`expected 2 to be 1` before), passive-flag pin (options `undefined` before).
- Green on the merged integration tree incl. WebKit QA on both viewports.

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/layout/useMobileVisualViewportHeight.test.tsx
+++ b/apps/app/src/components/layout/useMobileVisualViewportHeight.test.tsx
@@ -108,6 +108,18 @@ function withElementClientHeight(
   }
 }
 
+// Waits out one scheduled rAF pass so "the pass ran and did nothing" is
+// distinguishable from "the pass has not run yet".
+async function flushScheduledViewportPass() {
+  await act(async () => {
+    await new Promise<void>((resolve) => {
+      window.requestAnimationFrame(() => {
+        window.requestAnimationFrame(() => resolve());
+      });
+    });
+  });
+}
+
 beforeEach(() => {
   vi.spyOn(window, "scrollTo").mockImplementation(() => {});
 });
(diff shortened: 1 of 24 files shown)


SawyerHood:

Fix landed there for this PR:

- The shared list width subtracts the assistant column's `px-2` inset (it was 16 px wider than the column content box, which erased the comfort margin); the class and the constant are declared together.

> AGENT GENERATED

PR #2393, opened 2026-08-25
[stacked on #2387] bb.db retention: archived events, provider/unhandled cap, settled-interaction hygiene

vburojevic:
**STACKED PR** — contains #2387 (server event loop); only the top 5 commits are new. Do not merge before it.

## What was wrong

bb.db grows monotonically (916 MB live, ~586 KB/h at near-idle) because the `events` table has no lifetime retention: measured on a copy, it is ~91% of the file (item/completed 292 MB/167k rows, item/started 84 MB, provider/unhandled 44 MB/41k rows). Only a few streaming event classes were ever pruned; every other row lives as long as its thread. Two smaller tables (`pending_interactions`, `prompt_history_entries`) also never delete rows — cheap hygiene, not growth drivers (<1 MB each). The non-incremental maintenance branch also ran a whole-file dbstat scan before deciding anything, and the ~1.25 GiB SQLite cache+mmap budget was not tunable.

## What changed

Retention sweeps (all bounded per pass, registered in periodic-sweeps):

- **Archived threads** (`pruneArchivedThreadEvents`): hard-delete events beyond the existing `ARCHIVED_THREAD_EVENT_KEEP_RECENT = 120` window. Durable keyset cursor in `maintenance_scan_cursors`; resumes mid-thread under the row budget; unarchiving stops future pruning.
- **provider/unhandled** (`pruneProviderUnhandledEvents`): 30-day age cap regardless of archival. Read paths verified first: stored rows feed only the diagnostics timeline (dev builds / `showUnhandledProviderEvents`, default off) and legacy claude-code model-fallback extraction; nothing replays them. New partial index via Drizzle migration `0109_provider_unhandled_retention_index` (regenerated snapshot, no hand edits).
- **Settled interactions**: resolved/interrupted rows deleted 30 days after creation, pinned to the existing status/created index. **Prompt history**: newest 200 kept per `(thread, scope)`.
- `initDb` logs the resolved `auto_vacuum` mode + freelist at startup; the O(1) freelist counters short-circuit the whole-file dbstat scan when they alone cross the compaction thresholds, wrapped in `runEventLoopWorkSync`.
- `BB_SQLITE_CACHE_SIZE_KIB` / `BB_SQLITE_MMAP_SIZE_BYTES` env overrides (defaults unchanged), documented in docs + bb-cli skill surfaces.

**Retention policy stated, not buried**: archived threads keep 120 recent events; provider/unhandled diagnostics live 30 days; settled approval payloads 30 days; prompt history 200 per thread+scope. Active-thread `item/completed` retention is deliberately NOT included — the review notes recommend a second-stage output-payload truncation as the next step instead of row deletion (product decision).

## How you verified

New db tests (in-memory SQLite + `migrate(db)`) incl. row-budget mid-thread resume, cursor wrap, oldest-first age cap, pending/resolving never touched, prepared-SQL spy proving the freelist short-circuit prepares no dbstat statement, env-override fallbacks. Server test: a pruned archived thread's timeline still projects from the kept window. Fail-before/pass-after proven by stubbing each prune to a no-op (3+1 archived failures, 2 provider/unhandled). Rebased onto current origin/main over #2129/#2346 with migration renumbered via drizzle-kit: `@bb/db` 423, `@bb/server` 2,033, all green; measurements were taken on a copy of the live DB only.

> AGENT GENERATED


SawyerHood:

This PR is stacked on #2387, which is also closed for rework.

## Blocking — all in `pruneArchivedThreadEvents` (`packages/db/src/data/sweeps.ts:495-562`)

1. **CRITICAL: the sweep cuts mid-turn; archived threads become unviewable (500), permanently.** It deletes every row with `sequence <= latest - keepRecent` with no turn-boundary alignment. The kept window routinely starts inside a turn, or has no `client/turn/requested` anchor at all; `groupEventProjectionTurns` then throws `Timeline projection found turn/completed without turn/started`, the timeline route has no catch, and the thread 500s in the app, mobile, SDK and `bb thread show`. Unarchiving does not restore the rows. Reproduced three independent ways with the real prune, the real `buildThreadTimeline`, and the real route on in-memory SQLite with production constants (example: turn 1 with 5 messages + turn 2 with 150, archive, sweep → 41 rows deleted → throws for turn-2; also a one-message thread whose streamed answer's delta slots were already pruned). A read-only scan of a real 3 GB `bb.db`: 2,280 of the 2,298 archived threads over 120 slots would be left with orphan turn rows.
2. **HIGH: the window counts sequence slots, not rows.** The on-archive delta prune leaves gaps, so `latest - 120` keeps roughly 70–100 rows (about one partial turn). Long archived threads lose ~97% of their rows within an hour of deploy (one real thread: 2,914 rows → 90, 14 turns → 1 partial turn). The body says "keep 120 recent events", and the previous prune never deleted conversation rows.
3. **HIGH: thread-state rows are deleted with no grace.** `client/thread/start`, `client/turn/requested`, the root `turn/started`, and goal/usage/extensionState snapshots go within 60 s of archiving whenever the last turn produced ≥120 rows. After unarchive, send loses the model and permission mode or 500s.
4. **HIGH: `thread_search_segments`/FTS rows survive the deleted events.** Search keeps returning pruned messages with dead `sourceSeq` deep links, and the full text stays in the index this PR set out to shrink. The existing range deleter (`deleteThreadEventSuffixInTransaction`) cleans segments; this one does not.

## Also found

- The provider/unhandled age prune can delete a thread's max-sequence row, lowering `maxSeq` and allowing sequence reuse against the timeline-cache invariant (`sweeps.ts:395`). Never delete a thread's newest row.
- The server retention test cuts exactly on a turn boundary and asserts no lifecycle reader, so it passes while the sweep breaks projection (`apps/server/test/system/event-pruning.test.ts:726`).
- `BB_SQLITE_CACHE_SIZE_KIB`/`BB_SQLITE_MMAP_SIZE_BYTES` are parsed raw in `@bb/db`; malformed values are silently ignored and a value above 2^31−1 becomes `cache_size = 0` (`packages/db/src/connection.ts:61`). Route them through `@bb/config`.
- The cursor-wrap test passes with a broken wrap whenever the second thread id sorts after the first (`packages/db/test/data/sweeps.test.ts:697`).
- The timeline LRU assumes pruning is output-preserving (`timeline-cache.ts:22`); these prunes make cached archived windows stale until eviction.

## What was clean

The provider/unhandled 30-day cap (readers verified), the settled-interaction and prompt-history hygiene, the freelist short-circuit (0.1 ms, the O(1) claim holds), the startup auto-vacuum log, and the env knobs' defaults. Migration 0109 took 3.2–7.4 s cold and 0.6 s warm on a 2.76 GB `VACUUM INTO` snapshot with 1.2 M events; a log line before it would help, and the body's "916 MB live" figure is stale.

## Suggested salvage

Split it. Land everything except `pruneArchivedThreadEvents` as its own PR. Redo the archived sweep separately: cut on a turn boundary (the oldest surviving `client/turn/requested` or root `turn/started` at or below the cutoff); never delete thread-state rows or a thread's newest row; delete matching search segments in the same transaction; count kept rows, not slots; add a grace period after archive and a dry-run mode that reports what it would delete; and harden the projection to tolerate an orphan `turn/completed` instead of throwing. Get an explicit product decision on the numbers above before a sweep rewrites thousands of archived threads.

> AGENT GENERATED

Issue #2224, opened 2026-08-21
Claude Code provider: sdk/system background_tasks_changed falls into assertNever and surfaces as provider/unhandled debug rows


dhiyaancnirmal:
### Version

bb 0.39.0 (macOS host), Claude Code provider plugin

### Summary

Claude Code sessions emit system messages of subtype `background_tasks_changed`
whenever the set of backgrounded tasks changes. bb does not have a case for it
in its `sdk/system` subtype classification, so it falls into the
`assertNever` default branch and surfaces as raw `provider/unhandled` debug
rows in the thread log.

### Evidence

Real sessions show repeated rows like:

```json
{"type": "provider/unhandled", "rawType": "sdk/system", "subtype": "background_tasks_changed", ...}
```

In the installed 0.39.0 bundle, the classifier handles these `sdk/system`
subtypes explicitly: `compact_boundary`, `status`, `model_fallback*`,
`task_started`, `task_progress`, `task_updated`, `task_notification`,
`permission_denied`, plus a long noise list (`init`, `commands_changed`,
`hook_*`, ...). `background_tasks_changed` is absent from both lists, so the
default branch throws:

```
Unhandled Claude visibility value: background_tasks_changed
```

and the event is emitted as `provider/unhandled`.

### Impact

- Log noise on every background-task change (observed multiple times per session).
- Any future unmapped SDK subtype takes the same throwing path.

### Suggested fix

Classify `background_tasks_changed` as noise (it duplicates information already
available through the backgroundTask item events driven by
`task_started`/`task_updated`), or fold it into the same normalized mapping as
the other task types.

SawyerHood:
---Auto-Generated---

**Reproduction & root-cause report:** https://get-bb.github.io/reports/issues/2224.html

- **Verdict:** REPRODUCED · root-cause confidence: high
- **Root cause (short):** plugins/provider-claude-code/src/visibility.ts keeps a closed allow-list of sdk/system subtypes; the newer Claude CLI (2.1.241, newer than the 2.1.197 bundled with the pinned SDK 0.3.197) emits background_tasks_changed on every background-task start/settle, which the list collapses to coverage "unknown", so delta-translation's unhandledDeltas() emits an `unhandled` delta and the runtime persists a provider/unhandled row (two per task); nothing throws — the assertNever branch the issue blames is unreachable.
[...]

PR #2402, opened 2026-08-25
Make composer background-task rows clickable

### LINKED ISSUE #2315: Composer's running-agents status bar isn't clickable, unlike the main timeline
## The workflow

While a subagent is running, I want to click a row in the "Running background agent" status card above the composer to jump to that subagent in the main timeline, the same way the main timeline's `delegation` rows already expand into a full step-by-step view.

## What happens today

There are two separate UI surfaces for subagent activity, and only one of them is interactive:

- The main scrollable timeline (`ThreadTimelineRows.tsx` / `TimelineRowDetails.tsx`) renders `TimelineDelegationWorkRow` items as a real expandable tree backed by the full childRows data model. This one already works: you can expand a running or completed subagent and see its steps.
- The sticky card above the composer, `ThreadBackgroundCommandsCard` (`apps/app/src/components/promptbox/banner/ThreadBackgroundCommandsCard.tsx`), shows a live-updating "Running background agent: <label> +N more" summary with a duration ticker (`BackgroundActivityDuration`, lines 114-122). The only interactive element is the header button (line 256) that toggles local expand/collapse. The individual rows it reveals (lines 318-362) are plain `<div>`s: no `onClick`, no link, no navigation.

The card is fed by `activeBackgroundCommands: TimelineWorkflowWorkRow[]` (`packages/server-contract/src/thread-timeline.ts`, `workKind: "workflow"`, lines 463-479), computed in `apps/app/src/views/thread-detail/ThreadDetailView.tsx` (around line 861). That's a distinct, lighter-weight variant from `TimelineDelegationWorkRow` (same file, lines 440-450) that powers the main timeline, and it carries no id back-reference to the corresponding timeline row.

Verified against `9375a5495464c16fd9c3d03176add0a1dfca5172` on `main` by triggering a real Explore subagent and comparing the two screenshots side by side: the inline timeline row was expandable, the composer status card row for the same run was not.

## What I would expect

Clicking a row in `ThreadBackgroundCommandsCard` should scroll to (or open a pane for) the corresponding row in the main timeline, mirroring the click-through this issue's sibling report (subagent transcript access via `outputFile`) is asking for on the timeline side.

## Context and alternatives

The immediate workaround is to scroll up and find the matching row in the main timeline by label, which works but only while the row hasn't scrolled out of the visible history. Fixing this needs `TimelineWorkflowWorkRow` (or the `activeBackgroundCommands` projection) to carry a back-reference (e.g. a `timelineRowId`) to the row it summarizes, plus an `onClick` in the card that navigates to it.

> AGENT GENERATED

technicalpickles:
## What was wrong

The composer's "Running background agent/command" status card rows had no click handler, unlike the main timeline's expandable delegation rows, so there was no way to jump to a background task's context from the card.

## What changed

- `ThreadBackgroundCommandsCard.tsx`: every row (the single non-expandable summary row and each row in the expanded list) is now a real button. On click, it scrolls to and flashes the row in the main timeline if it's already rendered there, or opens a bottom drawer with the row's own data otherwise.
- The drawer path is the common one: a background task's row only stays in the main timeline while its spawning turn is still pending. Once that turn completes and gets summarized, the row survives only in the `activeBackgroundCommands`/`activeWorkflows` projection this card reads from, so there's nothing to scroll to.
- The drawer tracks the row by id and re-derives it from `commands` on every render (falling back to the last known data once the row drops out of that list), so it reflects live task progress instead of freezing to whatever the row looked like at click time.
- Extracted the scroll/flash DOM helpers shared with `useScrollToSearchedMessage.ts` into a new `timelineRowNavigation.ts` so both features use the same selector and flash-class logic instead of duplicating it.
- Deviation from the issue: it proposed adding a `timelineRowId` back-reference field to `TimelineWorkflowWorkRow`, on the assumption the row carried no id back to its main-timeline counterpart. That assumption doesn't hold: `row.id` is already built via the same `buildWorkflowWorkRow` helper used for the row that eventually appears in the main timeline, so it's already the right id to look up. No schema change was needed.

## How you verified

- New tests in `ThreadBackgroundCommandsCard.test.tsx` cover both click paths (scroll+flash when rendered, drawer fallback when not), targeting the clicked row's own id when multiple rows are expanded, the header toggle staying navigation-free, and the drawer staying in sync with live row updates (including when the row disappears from `commands` entirely) instead of freezing to a stale snapshot.
- `pnpm exec turbo run test --filter=@bb/app` and `pnpm exec turbo run typecheck --filter=@bb/app`.
- Manually reproduced the bug against `main` first: spawned a real background task on a `claude-code` thread and confirmed the card row did nothing on click, and that no `[data-timeline-row-id]` element existed for it while it was running.
- Verified the drawer path live against this branch's dev server: opened the drawer on a running task, confirmed it showed the correct live description, and confirmed it didn't go blank or error once the underlying task completed.
- Could **not** verify the scroll-to-rendered-row path live. While investigating why, I found that for the `claude-code` provider, a backgrounded shell command's completion never projects into a `workKind: "workflow"` row in the main timeline at all (running or completed) -- it falls into a generic "Unhandled Claude Code event" operation row instead. That's [#2224](https://github.com/get-bb/bb/issues/2224) (`background_tasks_changed` hits `assertNever`), a pre-existing gap this PR doesn't touch; I added the UI-side evidence there. The scroll path itself is still covered at the unit level, using the same `data-timeline-row-id` selector the app uses in production.

Fixes #2315

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/promptbox/banner/ThreadBackgroundCommandsCard.test.tsx
+++ b/apps/app/src/components/promptbox/banner/ThreadBackgroundCommandsCard.test.tsx
@@ -116,7 +116,9 @@ describe("ThreadBackgroundCommandsCard", () => {
     );
 
     const card = screen.getByRole("region", { name: "Background commands" });
-    expect(screen.queryByRole("button")).toBeNull();
+    expect(
+      screen.queryByRole("button", { name: "Running 1 background command" }),
+    ).toBeNull();
 
     reportCardWidth(card, 320);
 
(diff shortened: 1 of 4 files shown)


SawyerHood:

PR #2395, opened 2026-08-25
Structure the sidebar into configurable regions

brsbl:
## What was wrong

The sidebar rendered bb-owned destinations, plugin pages, and threads as one implicit stream. With several plugin pages installed, every page remained expanded in that stream and there was no durable visual boundary between app chrome, plugin navigation, and the thread list.

## What changed

- separates the fixed bb rows, traditional plugin pages, and thread list into stable regions with seams between visible neighbors
- keeps Automations in the fixed bb-owned region and gives traditional plugin pages their own **Plugins** section
- shows the first five traditional plugin pages and keeps the rest reachable behind **More plugins**
- normalizes persisted plugin ordering so newly installed and returning pages keep deterministic positions

No server, daemon, wire, or public plugin-SDK contract changes are introduced by this layer.

## Visual evidence

> Rebase evidence status: current parent `fb05e93ff5ad4f58831e804b973f2de790557afa` and current head `d1170cf94e6fcf3d58025fe72ed1cae72c2ec019`. The embedded captures below were made from the exact pre-rebase revisions and remain useful behavioral references, but they are not exact-current-revision evidence. Replace them before merge.

Pre-rebase behavioral comparison: `b629714e31487bde8af5574639422a9b5c97cbb5` → `977d04e4236b68433bd068139909a3e129ed7b38`.

### Before — pre-rebase base `b629714e31487bde8af5574639422a9b5c97cbb5`

Six identical local nav-panel fixtures appear inline with the fixed rows and thread list.

Before — six plugin pages in the implicit sidebar stream

### After — pre-rebase head `977d04e4236b68433bd068139909a3e129ed7b38`

The same fixture shows a bounded **Plugins** region: five rows, **More plugins**, and stable seams above and below.

After — stable plugin region with five visible rows and More plugins

## How you verified

- Google Chrome for Testing 151.0.7922.71, pre-rebase exact branch web apps from `scripts/bb-dev-app`, light theme, `/`, 1440×900.
- Used the same database, six local nav-panel fixtures, route, viewport, theme, and settled state on both revisions.
- Observed all six pages directly in the base; on the head, five remained visible and **More plugins** exposed the overflow contract. No page or console errors were recorded.

Fixes: N/A — Updated Sidebar stack foundation.

BB-Thread-ID: thr_2e2gbjx943

> AGENT GENERATED

PR #2405, opened 2026-08-25
Bound plugin pages and add experimental panel menus

brsbl:
## What was wrong

Plugin pages used two overlapping persistence concepts—an order and a hidden set—while the sidebar rendered every non-hidden page without a cap. That let plugin-heavy installs crowd threads out of the column and gave keyboard users no route to split a plugin page or reach plugin-contributed actions.

The first hidden-page migration did not preserve the old UI's felt state. Legacy “Hide from sidebar” rows lived behind `More (N)`, but the migration used the ordinary five-row cap and therefore resurfaced hidden pages when an installation had five or fewer plugin pages. It also cleared the entire legacy hidden set, even though the host-owned `__builtin__/tools` entry belongs to the later top-region toggle migration.

## What changed

- Replaces per-page hiding with one deduplicated user order over every registered plugin page; the first five render above a persistent `Show N more` disclosure.
- Temporarily promotes an active overflow page into the last visible slot without changing stored order.
- Adds `Move to top` / `Move to overflow`, cross-cap reordering, split, and plugin-settings actions through one menu definition shared by the ellipsis button and right-click.
- Migrates hidden plugin pages once under a Web Lock while preserving their prior felt state: visible pages remain ahead of hidden pages, and a temporary persisted fold records the previously visible count (capped at five), including installations with five or fewer total pages and the all-hidden case.
- Removes that temporary fold once the overflow is emptied, restoring the ordinary flat list for five or fewer pages.
- Makes migration ownership explicit: only plugin-page keys are consumed. Host-reserved entries such as `__builtin__/tools` never enter plugin order or overflow and remain in the legacy atom for the toggles layer to consume.
- Adds the optional additive `experimental_menu` plugin-app contract, including up to four labeled groups, one lazy submenu level, contained action/resolver failures, unknown-icon tolerance, and context helpers for navigation and splits.
- Advances the additive plugin SDK surface to 0.4.24, adds the public-API audit entry and authoring-guide coverage, and proves an installed 0.4.17 scaffold remains compatible. No server/daemon wire contract changed, so `HOST_DAEMON_PROTOCOL_VERSION` is unchanged.

## UI evidence

Identical fixture and viewport in both images: four plugin pages, with GitHub and Docs present in the legacy hidden set before first paint. The crop excludes unrelated project data.

### Before — merge base `c781d323e`

The same two pages are visible and the two hidden pages remain parked behind `More (2)`.

Before: Automations and Tasks visible with two hidden pages behind More (2)

### After — PR head `c3612aae6`

The same two pages remain visible, while the migrated disclosure uses the new `Show 2 more` label.

After: Automations and Tasks visible with two migrated overflow pages behind Show 2 more

The immutable screenshot assets are pinned to fork commit `55df2709a481b35ad360cea930ea3fa4646c2a7d`; both raw URLs return HTTP 200 with `image/png`.

## How you verified

- Focused migration, ordering, and sidebar-rendering regression suites: 46/46 passed, including the four upgrade cases from the updated spec.
- Installed 0.4.17 scaffold server compatibility suite: 29/29 passed against SDK 0.4.24.
- Plugin SDK suite: 220/220 passed.
- Full workspace lint passed through Turbo.
- Full workspace typecheck passed through Turbo.
- The full workspace test graph was exercised through Turbo on Node 22 with the repository's isolated npm cache. Twenty-six package suites passed serially before the pre-existing `@bb/host-watcher` nested-directory event test timed out waiting for a macOS filesystem event; an isolated retry reproduced that one timeout after its other 45 tests passed. `@bb/process-utils`, whose macOS cwd tests require `/usr/sbin/lsof`, passed 20/20 in an isolated Node 22 run. No Task 1 file touches either package.
- Required GitHub Actions are green at `c3612aae6`, including all app, package, server, integration, package-smoke, version-lockstep, and SDK-version guards.
- Exact-branch Chrome for Testing capture passed at 1440×900 for the migration path, including setting the legacy hidden state before first paint and hard reload into the settled state.
- Safari remains unverified on this host because automation access is unavailable; this PR is not being presented as fully cross-browser verified or merge-ready.

## Fixes

No linked issue; implements Phase 3 layer 1 of the Updated Sidebar Spec on top of PR #2395.

BB-Thread-ID: thr_jgmy8ay2qd

BB-Thread-ID: thr_2e2gbjx943

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/plugin/PluginNavSidebarItems.test.tsx
+++ b/apps/app/src/components/plugin/PluginNavSidebarItems.test.tsx
@@ -12,7 +12,14 @@ import { createStore, Provider } from "jotai";
 import { MemoryRouter, useLocation } from "react-router-dom";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { CompactViewportOverrideProvider } from "@bb/shared-ui/hooks/use-compact-viewport";
+import type { PluginNavPanelRegistration } from "@get-bb/plugin-sdk/app";
 import { SidebarProvider } from "@/components/ui/sidebar.js";
+import { appToast } from "@/components/ui/app-toast";
+import { writeLastKnownPluginNavPanelChrome } from "@/lib/plugin-nav-panel-chrome";
+import {
+  markPluginFrontendsSettled,
+  resetPluginFrontendBootStateForTest,
+} from "@/lib/plugin-frontend-boot-state";
 import {
   resetPluginSlotStoreForTest,
   setPluginSlotRegistrations,
@@ -26,7 +33,10 @@ import {
   ExtensionsNavSidebarItem,
   PluginNavSidebarItems,
 } from "./PluginNavSidebarItems";
-import { pluginNavPanelOrderAtom } from "./pluginNavSidebarAtoms";
+import {
+  pluginNavPanelOrderAtom,
+  pluginNavPanelOverflowExpandedAtom,
+} from "./pluginNavSidebarAtoms";
 
 function registrationSet(
   overrides: Partial<PluginRegistrationSet>,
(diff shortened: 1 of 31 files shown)


SawyerHood:

PR #2415, opened 2026-08-25
fix(desktop): load the packaged shell PATH deterministically

### LINKED ISSUE #2343: Packaged Nightly plugin processes lose PATH for node and gh
## Problem

A packaged macOS bb Nightly launched from the GUI can inherit a restricted PATH:

    /usr/bin:/bin:/usr/sbin:/sbin

The server and plugin children then cannot resolve ordinary user-installed tools. In the affected case, a plugin-spawned Nightly CLI reports `env: node: No such file or directory`, and `gh` cannot be resolved. The same tools resolve from an interactive agent shell.

This leaves queue readers and Companion blind while the interactive shell still appears healthy.

## Reproduction

1. Launch the packaged Nightly from the GUI rather than an interactive terminal.
2. Use a plugin path that spawns the bundled CLI and an external `node` or `gh` child.
3. Observe the missing-command failures.
4. Run the equivalent command from an interactive shell and observe that both tools resolve.

## Expected behavior

Packaged GUI/server/plugin processes should receive a deterministic, usable user-tool PATH, or should use explicit validated executable paths. If a required executable is unavailable, the affected reader should report a visible degraded state rather than silently appearing healthy.

## Implementation clue

The desktop source already probes a login shell for packaged macOS launches and falls back to the inherited environment on failure or timeout. The plugin child inherits the server environment. Please make PATH propagation deterministic and add a regression test covering the packaged-child environment and the missing-executable/degraded-state path. Preserve fail-closed behavior; do not make command discovery depend on a shell watcher.

Related historical reports include #1545 and #1388, but neither covers this packaged-Nightly plugin/Companion failure.

> AGENT GENERATED

## Stable/Nightly bundle comparison

Static extraction from installed stable `0.39.0` and Nightly `0.39.1-nightly.32633481796.1` rules out a Nightly build-target omission:

- both `app.asar` bundles call `ensurePackagedUserShellPath(...)` unconditionally at the start of `runDesktopApp()`;
- the call sites are byte-identical;
- the complete `desktop-shell-path` implementations are byte-identical (SHA-256 `fe1bab92b8032da3f978c217e3c54fbcfb2bca7e98ee030f21975c1c889e949b`);
- the only gates are platform (`darwin` or `linux`) and `app.isPackaged`;
- on macOS the implementation runs `/bin/zsh -ilc 'printf "%s" "$PATH"'` with a 2,000 ms timeout, then preserves the inherited PATH on any error, signal, nonzero status, or empty output.

Both app bundles pass `codesign --verify --deep --strict`; this is not a missing or invalid app signature.

A no-restart reproduction using each app's signed Electron executable in Node mode, with the same restricted inherited PATH, returns the same result for stable and Nightly:

- `spawnSync /bin/zsh ETIMEDOUT`
- status `13`
- empty stdout

The equivalent login-shell command outside that Electron path completes and resolves both `node` and `gh`. The running Nightly desktop, bridge, server, and host daemon therefore retain `/usr/bin:/bin:/usr/sbin:/sbin` because the packaged shell probe times out and takes its fallback path.

This narrows the defect to runtime PATH acquisition, not build-target gating. The packaged path repair needs a deterministic method that does not discard a valid user PATH merely because interactive shell startup exceeds two seconds (for example, a larger/tested bound or a non-interactive shell-environment resolver), with regression coverage for timeout fallback and child inheritance.

SawyerHood:
## What was wrong

A packaged macOS bb launched from the GUI inherits launchd's restricted PATH (`/usr/bin:/bin:/usr/sbin:/sbin`). `ensurePackagedUserShellPath` was a single `/bin/zsh -ilc 'printf "%s" "$PATH"'` `spawnSync` with a 2,000 ms timeout, and every failure branch (ETIMEDOUT, signal, non-zero status, empty stdout) kept the inherited PATH. Two further mechanisms discarded valid PATHs: `spawnSync` waits for the pipes to close, so a login file that leaves a background job attached to stdout reports ETIMEDOUT (with a numeric status, exactly the reporter's `status 13`) even after the shell exited and printed its PATH; and raw stdout was used as the PATH, so a banner or `nvm use` line became part of the first entry. bb-app, the server, the host daemon, and in-process server plugins inherit `process.env.PATH` through `startOwnedRuntime → startBbAppProcess`, so every child lost `node` and `gh` while the app looked healthy, and the only warning went to the Electron main-process stderr, which a Dock launch discards. (#2343)

## What changed

`apps/desktop/src/desktop-shell-path.ts` (probe rewritten; pure, injectable), `apps/desktop/src/main.ts` (call site), `apps/desktop/README.md`:

- The shell prints its PATH between marker lines. A complete block is authoritative even when the shell then times out, is signalled, or exits non-zero; everything else on stdout is ignored.
- The default spawn is async: stdin closed, settle on the shell's exit plus a 250 ms pipe grace instead of on pipe close, SIGKILL at the budget.
- `$SHELL -ilc` (5 s budget), then `$SHELL -lc` (3 s), `$SHELL` taken verbatim (fallback `/bin/zsh` on macOS, `/bin/bash` on Linux). When `$SHELL` cannot be started (spawn ENOENT/EACCES) and is not the platform default, the stages run once more with the platform default shell.
- A probed PATH is merged with the inherited one (probed entries first, then inherited entries the shell did not report).
- When an earlier stage failed, well-known tool directories that exist on disk (volta, nvm's default Node bin resolved from the alias files without running nvm, fnm, mise/asdf/nodenv/proto shims, n, pnpm home, bun, deno, cargo, go, `~/.local/bin`, `~/bin`, Homebrew, `/usr/local/bin`, `/snap/bin`) are appended, because `-lc` skips the rc file where nvm and Volta install their hooks. Appended entries never shadow a tool the shell already put on PATH. An nvm default alias ending at `system` adds no nvm bin.
- Every fallback logs one warn line naming each failed stage, the platform-default retry, and any appended directory, to stderr and to `desktop.log` in the log directory.
- The probe runs after the single-instance lock and handler registration and before `app.whenReady()`, so a duplicate launch quits without paying for it and the repair still precedes `initializeRuntime`, the only path that spawns a child.

Linux goes through the same stages and augmentation. No server/daemon wire change, so `HOST_DAEMON_PROTOCOL_VERSION` is unchanged. Worst-case blocked startup is ~8.5 s and only when the interactive shell hangs before printing.

Scope: PATH acquisition in the desktop main process and its propagation to bb-app and everything bb-app spawns. The issue's second ask, a visible degraded state in the affected plugin reader when a required executable is still missing, is a separate feature and is left for a follow-up. Related follow-up: the host daemon's own `resolveUserShellPath` still settles on pipe close and has no augmentation, so daemons started by launchd/systemd without the desktop keep the old behavior.

## How you verified

`apps/desktop/test/desktop-shell-path.test.ts` (28 tests; 20 of them fail against the `origin/main` source): marker parsing (banners, CRLF, truncated and missing blocks), complete block kept on timeout / status 13, `-ilc` timeout → `-lc` fallback, merge with inherited entries, augmentation only on the degraded paths, nvm alias chains (`lts/*`, partial, `system`), `$SHELL` honored and the platform-default retry on ENOENT/EACCES, Linux parity. Three real-shell tests (bash) exercise the default spawn end to end: settle on exit with a background job holding stdout, SIGKILL at budget for a TERM-trapping shell, and a login shell that prints a banner and leaves a background job (run 3× under turbo, no flake). A pipe-close-settling regression was proven to fail the shape assertions deterministically.

Commands: `pnpm exec turbo run typecheck --filter=@bb/desktop --force` and `pnpm exec turbo run test --filter=@bb/desktop --force` on the rebased tip: 35 files / 252 tests pass. Three independent adversarial review rounds. EAP codename scan of `origin/main..HEAD`: clean.

Not done: no packaged macOS smoke (this machine is Linux); the default spawn and marker command were exercised against real bash shells instead.

Fixes #2343

> AGENT GENERATED

### DIFF
--- a/apps/desktop/README.md
+++ b/apps/desktop/README.md
@@ -292,6 +292,48 @@ BB_DESKTOP_OPEN_DEVTOOLS=1 apps/desktop/release/mac-arm64/bb.app/Contents/MacOS/
 When the desktop app spawns `bb-app`, server and daemon logs land under
 `~/.bb/logs/` or `$BB_DATA_DIR/logs/` when `BB_DATA_DIR` is set.
 
+### PATH in packaged launches
+
+A packaged app started from the Dock, Finder, or a Linux desktop session
+inherits the launcher's bare PATH (`/usr/bin:/bin:/usr/sbin:/sbin` under
+launchd), not the user's shell PATH. Before it spawns `bb-app`, the desktop
+main process repairs `process.env.PATH` in `src/desktop-shell-path.ts`, and
+every child (bb-app, server, host daemon, in-process server plugins) inherits
+the result:
+
+1. `$SHELL -ilc '<print PATH between marker lines>'` with a 5 s budget, where
+   `$SHELL` is the login shell the launcher passed, taken verbatim (fallback
+   `/bin/zsh` on macOS, `/bin/bash` on Linux). The interactive login shell is
+   what sources `~/.zshrc`, where Homebrew, nvm, and Volta usually extend PATH.
+2. `$SHELL -lc ...` with a 3 s budget when the interactive stage fails or
+   times out. That shell skips the rc file, so the tool directories from step
+   4 that exist on disk are appended to its PATH as well.
+3. If `$SHELL` itself cannot be started (`ENOENT` or `EACCES`: an uninstalled
+   login shell, or an account record pointing at a stale path) and it is not
+   the platform default, steps 1 and 2 run once more with the platform default
+   shell. The warning names the retry.
+4. If every stage fails, well-known tool directories that exist on disk
+   (`~/.volta/bin`, nvm's default Node `bin` unless the default alias is
+   `system`, fnm's default alias, the mise, asdf, nodenv, and proto shims,
+   `~/.n/bin`, pnpm's home, `~/.bun/bin`, `~/.deno/bin`, `~/.cargo/bin`,
+   `~/go/bin`, `~/.local/bin`, `~/bin`, Homebrew, `/usr/local/bin`, and on
+   Linux `/snap/bin`) are appended to the inherited PATH. Nothing is executed
+   to discover them, and because they are appended, they never shadow a tool
+   the shell or the launcher already put on PATH.
+
+The shell prints its PATH between marker lines, so banners and version-manager
+hooks that write to stdout during login are ignored, and a complete block counts
+even when the shell then hangs or exits non-zero. The probe settles when the
+shell exits rather than when its pipes close, so a login file that leaves a
+background job attached to stdout does not hold it for the whole budget. A
+probed PATH is merged with the inherited one rather than replacing it: probed
+entries come first, then any inherited entries the shell did not report. Every
+fallback writes a warning with the failing stage and its reason to the main
+process stderr and to `desktop.log` in the log directory, so a silent success
+never hides a broken probe. Startup is bounded by the stage budgets; a shell
+that hangs costs at most 8 s before the window opens (a missing `$SHELL` fails
+immediately, so the retry does not add to that).
+
 To verify attach-if-found manually, start a compatible bb first, then launch the
 desktop app:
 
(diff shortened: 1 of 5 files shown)


SawyerHood:

PR #2417, opened 2026-08-25
fix: park input the provider is too busy to take instead of failing the run

### LINKED ISSUE #2370: tell --mode auto injects into busy turn instead of queueing: provider error retry loop, status flips to error while turn continues
## Summary

Sending `bb thread tell --mode auto <thread>` to a thread whose agent is mid-turn attempts live injection instead of queueing. The provider bridge (pi/codex) rejects with `Agent is already processing a prompt. Use steer() or followUp() to queue messages`, the host retries delivery repeatedly (a loop of identical errors in the thread log), and the thread's visible status flips to `error` **while the turn continues to run normally** (native transcript kept advancing throughout).

## Expected

`--mode auto` detects the busy turn and queues the message (or the error is surfaced once, without a retry loop and without flipping thread status to `error`).

## Actual

- Repeated identical provider errors (observed 6+ identical error blocks for one logical delivery).
- Thread status shows `error` though the turn is healthy and progressing.
- The queued messages were **lost** (never delivered); recovery required `thread stop`, which then released the runtime and delivered queued messages.
- Operator-visible symptom: seat looks dead/buggy in the UI while it is actually working — compounds with #2326 (thread list staleness) to make healthy seats appear failed.

## Environment

bb Nightly 0.39.1-nightly.32633481796.1 (commit 7e10ea4b), macOS, pi provider bridge with codex/gpt-5.6 backend. Observed repeatedly over 2026-08-24 across multiple threads (orchestrator seat + a writing lane earlier the same day — the lane instance wedged hard enough to require stop + salvage).

## Repro

1. Have a thread mid-turn (long-running provider turn).
2. `bb thread tell --mode auto <thread> "msg"` (possibly repeat / have multiple queued tells).
3. Observe repeated 'already processing' errors in `bb thread log`, status flips to error, message never delivered.

## Notes

- `--mode queue` works reliably; the defect is in auto/steer busy-detection on this bridge.
- Suggest: auto should fall back to queue on 'already processing' rather than retry-injecting; delivery errors should not flip thread status to `error` while the underlying turn is progressing.

SawyerHood:
## What was wrong

A turn submission the thread could not take was a generic failure at every layer. `bb thread tell --mode auto` into a thread whose turn the daemon still holds (a start whose `turn/started` is in flight, the pending-start wait, #2379's admission assert) or whose provider refuses input mid-run (pi while it compacts) came back as `command_failed`. The server appended `system/error`, applied `run.failed` (status `error` while the native transcript advanced), and dropped the message. The queue never drained because it only drains from `idle`, until `thread stop` released the runtime.

The pi bridge amplified a refusal: a rejected prompt raised a session error (`onDone(error)` → `provider.error {settlesTurn}`) and resolved a failed prompt settlement (→ a failed `turn.boundary` with `claimIfIdle`). Either closed the open turn from under pi, after which every later send resolved to `start` and was refused identically; those are the repeated error blocks in #2370. There is no bb retry loop: each block was one new send hitting the same state.

The most common trigger in practice is two sends a few seconds apart from the UI or CLI on an idle thread: send 1 starts a turn, send 2 arrives while `turn/started` is still pending, and the daemon waited only 5 s (#2242) before refusing with "Refusing to start a competing turn while … is still starting". Measured on a Claude Code thread on 2026-08-25, `client/turn/requested` → `turn/started` took 7–15 s, so the refusal fired routinely.

On the reporter's nightly (7e10ea4b, before #2379) the refusal text came from `Agent.prompt` in the in-process pi SDK session. On `main`, pi runs as an RPC child whose `prompt` answers an error only for refusals that precede its preflight (compaction); the post-preflight guard is swallowed. The pattern is kept as a defensive match only.

## What changed

Two commits.

**Commit 1 — park input the provider is too busy to take instead of failing the run**

- `packages/provider-bridge-protocol`: `BRIDGE_JSON_RPC_ERRORS.TURN_BUSY` (-32004) for a `turn/start` or `turn/steer` the provider refuses because it is mid-run (or compacting) and will not queue the input. Documented in `docs/provider-bridge-protocol.md`.
- `plugins/provider-pi`: a rejected prompt/steer dispatch settles nothing (no session error, no prompt settlement) and keeps `isProcessing` as it was. Pi's compaction refusal, the bridge's own compaction-active steer refusal, and (defensively) the `Agent.prompt` guard answer `TURN_BUSY`. The fake pi's `/busy` knob models the compaction refusal pi really sends over RPC.
- `packages/agent-runtime`: `AgentRuntimeTurnBusyError` (code `thread_turn_busy`) for the competing-start refusal and for `TURN_BUSY` rejections; turn state is left untouched; a refused start re-marks the session idle so the idle reaper can still release it.
- `packages/host-daemon-contract`: `THREAD_TURN_BUSY_ERROR_CODE`, the one wire spelling. **`HOST_DAEMON_PROTOCOL_VERSION` 170 → 171** (new `turn.submit` failure code on the wire).
- `apps/host-daemon`: the pending-start refusal is an `ExpectedCommandDispatchError` with that code; `thread_turn_busy` is an expected RPC failure, checked against the runtime error's literal type at compile time.
- `apps/server` (policy): a `thread_turn_busy` `turn.submit` failure records `client/turn/rejected` but appends no `system/error` and applies no `run.failed`; the live turn owns the status. The input is parked, never lost: a direct send goes to the thread's queue (raw payload, arrival order); a consumed queued group is restored at the head of the queue with its ids and grouping (`restoreConsumedQueuedThreadMessagesInTransaction`) on both drain paths, with fresh sort keys ahead of the current head; a parent system message (child-completed notice) is held in `deferred_thread_messages` with its taxonomy, stamped with the turn that refused it, and skipped while that turn is live (one attempt per turn); rows behind it still get their attempt. The re-queue hook is scoped to `turn.submit` dispatches.
- CLI/skill/guide: `--mode auto` and the busy fallback are described on `bb thread tell --help`, the bb-cli skill, and the threads guide.

**Commit 2 — wait out a starting turn and steer into it instead of refusing**

- `packages/agent-runtime`: `RuntimeTurnState.releaseWaiters(threadId)` resolves every `waitForActiveTurn` waiter with `null` when a pending start ends without a turn (bridge refused the start, a thread-scoped non-retrying `provider/error`, thread stop, process detach, shutdown) via `dropPendingTurnStartWithoutTurn`. `turn/started` still resolves waiters with the id first. `WaitForActiveTurnArgs.signal` (AbortSignal) lets a caller cancel the wait. `DEFAULT_TURN_START_WATCHDOG_THRESHOLD_MS` (120 s) is exported.
- `apps/host-daemon`: the pending-start wait is now `DEFAULT_TURN_START_WATCHDOG_THRESHOLD_MS` instead of 5 s, so "still starting" has one meaning (the watchdog's); because the runtime releases waiters early, the bound is only reached for a start the watchdog also reports as stuck. `thread.stop` shares the per-thread lane with `turn.submit`, so the router keeps an `AbortController` per admitted `turn.submit` and an **interrupt** stop aborts them synchronously on arrival (`abortQueuedTurnSubmits`); an aborted wait, or a steer whose stop signal is already aborted, falls to the typed `thread_turn_busy` refusal, the server parks the input, and the stop runs next. A **release** stop does not abort (it leaves lifecycle state alone and skips a busy runtime, so a parked input would have nothing left to settle the thread); the submit keeps its wait and starts its turn if the release ends the pending start. No wire change (`threadStopRequested` is a daemon-internal dispatch option).
- `tests/scripted-echo-provider`: `thread_scoped_fail:<text>` directive (accept input, emit a thread-scoped non-retrying `provider/error`, no turn), matching codex's real shape.
- Contract comment, bb-cli skill, and threads guide updated: a message sent while the turn is starting waits for the start and steers into it; only a start the runtime reports as stuck, or a stop that arrives first, parks the message in the queue.

Known follow-ups (documented, not in this PR): the timeline still renders a busy-parked request as a rejected user row before the queued copy is delivered (thread-view projection change); a `start` refused busy on an idle thread waits on the daemon's turn events with no server-side watchdog; a later parent-system notice can overtake a held one when the provider accepts a steer mid-turn; `thread.archive` / `interactive.resolve` / `thread.plan.cancel` / `thread.goal.clear` queued behind a waiting submit still wait up to the bound (benign today, untested).

## How you verified

Regression tests fail on the previous source and pass after (file-swap proofs logged per package):
- pi bridge `bridge.round2.test.ts`: busy steer/start answer `TURN_BUSY` with no `turn.boundary` and no session error, including a `turn/start` refused during a held run.
- agent-runtime `runtime.recovery.test.ts` / `runtime.lifecycle.test.ts`: typed busy on refused start/steer, live turn kept, refused start still reapable; a `waitForActiveTurn` waiter is released as soon as the bridge refuses the pending start, when an accepted start fails before its turn opens, and when the thread is stopped while its start is pending; `runtime-turn-state.test.ts`: `releaseWaiters` releases one thread's waiters only; abort signal resolves the wait.
- host-daemon `command-dispatch.test.ts` / `thread-stop-races.test.ts`: typed expected code; pending start → id → `appliedAs: steer` with the 120 s bound; a start that ends without a turn runs the input as a new turn; an interrupt `thread.stop` aborts a waiting submit, which parks its input, and the stop runs next; a release stop waits behind the submit instead; a steer whose stop is already queued is parked.
- db `queued-thread-messages.test.ts`: restore keeps ids and grouping, clears claims, lands ahead of a row queued into the emptied queue, and is what the next drain claims.
- server `thread-send-dispatch.test.ts`: busy send keeps the thread active and queues the raw input; busy queued auto-send restores the group at the head; busy queued send into an active thread restores the group with its grouping ahead of the trailing row; busy parent system message is held with its taxonomy, not retried while the turn is live, steered into the next turn, and does not block a send deferred behind it.
- contract test: protocol version expectation.

Commands (all through turbo, `--force`, on the rebased tip): typecheck for `@bb/db`, `@bb/provider-bridge-protocol`, `@bb/agent-runtime`, `@bb/host-daemon`, `@bb/host-daemon-contract`, `bb-plugin-provider-pi`, `@bb/server`, `@bb/cli`, `@bb/templates`, `bb-plugin-scripted-echo-provider`; the matching `test` tasks plus `@bb/agent-runtime#test:integration` and `@bb/server#test:provider-corpus`: all green. Four independent adversarial review rounds (the stop-lane finding came from one of them). EAP codename scan of `origin/main..HEAD`: clean.

Not done: no live reproduction against a real `pi --mode rpc` session during manual compaction (proven from pi's source and the fake's RPC shape instead).

Fixes #2370

> AGENT GENERATED

### DIFF
--- a/apps/cli/src/commands/thread/actions.ts
+++ b/apps/cli/src/commands/thread/actions.ts
@@ -425,7 +425,10 @@ export function registerActionsCommands(
       "Reasoning level: low, medium, high, xhigh, max (provider-dependent)",
     )
     .option("--permission-mode <mode>", PERMISSION_MODE_HELP)
-    .option("--mode <mode>", "Message mode: steer (default), queue, or auto")
+    .option(
+      "--mode <mode>",
+      "Message mode: steer (default), queue, or auto (steer a live turn, else start one)",
+    )
     .option("--plan", PLAN_HELP)
     .option(
       "--file <path>",
--- a/apps/host-daemon/src/command-dispatch-support.ts
+++ b/apps/host-daemon/src/command-dispatch-support.ts
@@ -1,15 +1,19 @@
-import type { AgentRuntimeBridgeLaunch } from "@bb/agent-runtime";
+import type {
+  AgentRuntimeBridgeLaunch,
+  AgentRuntimeTurnBusyError,
+} from "@bb/agent-runtime";
 import type { AvailableModel } from "@bb/domain";
 import type { EventSinkInput } from "./event-sink.js";
-import type {
-  HostDaemonCommand,
-  ProviderHealthResult,
-  ProviderUsageResult,
-  HostDaemonBridgeLaunch,
-  HostDaemonInjectedSkillSource,
-  HostDaemonOnlineRpcCommand,
-  HostDaemonConnectTunnelIdentity,
-  WorkspaceContext,
+import {
+  THREAD_TURN_BUSY_ERROR_CODE,
+  type HostDaemonCommand,
+  type ProviderHealthResult,
+  type ProviderUsageResult,
+  type HostDaemonBridgeLaunch,
+  type HostDaemonInjectedSkillSource,
+  type HostDaemonOnlineRpcCommand,
+  type HostDaemonConnectTunnelIdentity,
+  type WorkspaceContext,
 } from "@bb/host-daemon-contract";
 import type {
   ProviderInstallationCommand,
@@ -130,9 +134,15 @@ export function isExpectedCommandDispatchError(
   return error instanceof ExpectedCommandDispatchError;
 }
 
-const EXPECTED_ONLINE_RPC_FAILURE_CODES = new Set([
+const EXPECTED_ONLINE_RPC_FAILURE_CODES = new Set<string>([
   "file_too_large",
   "provision_cancelled",
+  // A turn submission the thread could not take: the shared runtime refused
+  // a competing start (`AgentRuntimeTurnBusyError`) or the bridge answered
+  // TURN_BUSY. The live turn is fine and the server parks the input. The
+  // runtime's error code is what reaches the wire, so it must be the
+  // contract's string.
+  THREAD_TURN_BUSY_ERROR_CODE satisfies AgentRuntimeTurnBusyError["code"],
 ]);
 
 export function isExpectedOnlineRpcFailureError(error: unknown): boolean {
--- a/apps/host-daemon/src/command-dispatch.test.ts
+++ b/apps/host-daemon/src/command-dispatch.test.ts
@@ -22,7 +22,10 @@ import {
   fetchDispatchTestArtifact,
   unexpectedProviderMaintenance,
 } from "../test/command/dispatch-helpers.js";
-import type { CommandOf } from "./command-dispatch-support.js";
+import {
+  isExpectedOnlineRpcFailureError,
+  type CommandOf,
+} from "./command-dispatch-support.js";
 import { RuntimeManager } from "./runtime-manager.js";
 
 const WORKSPACE_PATH = "/tmp/bb-command-dispatch-test";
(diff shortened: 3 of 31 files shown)


SawyerHood:

Issue #2190, opened 2026-08-21
SQLite db silently reverted to a 2-week-old state while the server was running (post-Aug-9 projects/threads lost; auto-update in progress)


arnocode84:
## Summary

While the desktop app's server was running (process up since Aug 16, never restarted — verified by pid/lstart), the SQLite database at `~/.bb/bb.db` silently reverted to a ~2-week-old state. Every project, thread, and setting created after ~Aug 9 disappeared; rows from Aug 6–9 (including threads long gone from the sidebar) came back.

## Environment

- bb.app desktop **0.38.0** at incident time (auto-updated to 0.39.0 later the same day)
- macOS arm64, Darwin 25.5.0
- Default data dir `~/.bb`, default local server `127.0.0.1:38886`, single machine, no remote/connect pairing

## Timeline (local time, Aug 21)

- **15:04** — everything normal: spawned a thread via CLI, read its output, `bb environment archive-threads` OK, `bb project list` showed 4 projects.
- **~15:10–15:11** — `bb.db` mtime rewritten. First failing call: `bb automation create --project <id>` → `HTTP 404: Project not found` (first-ever call to the automations plugin on this install; correlation only, not proven causal).
- After that: `bb project list --json` → `[]` (only the Personal project remains), today's thread IDs → `HTTP 404: Thread not found`, and three threads dated **Aug 6–9** reappeared.
- The server process had **not** restarted (same pid since Aug 16, one listener on 38886).
- Later, on quitting the app, a **Squirrel/ShipIt auto-update turned out to be in progress**; while it was replacing the bundle we observed a transient CLI/server flag mismatch (`--root` vs `--path` on `project create`). Possibly related to the incident, possibly not.

## Evidence gathered

- Direct SQLite read of `bb.db` right after: `projects` contains only `proj_personal` (created Aug 6); newest `threads` rows are from Aug 6–9.
- The WAL was ~7 MB; replaying it on a **copy** of the db does not surface the lost rows either.
- `server.7.log` around the event shows only recurring `Skipping malformed prompt history row` (ZodError: empty array) for the three old threads, plus `[plugin:connect] rpc listAccountServers failed: not_paired` noise. No migration/reset/checkpoint entries.
- No APFS local snapshots were available for recovery.
- A cold copy of `bb.db` + `-wal` + `-shm` taken ~5 min after the incident is available on request.

## Impact

[...]

PR #2423, opened 2026-08-25
Prevent concurrent servers from sharing a data directory

ymichael:
## What was wrong

The server had no data-directory singleton lock. During overlapping dev restarts, a second server could open the same SQLite database before losing its port bind. Because in-flight provisioning context is process-local, startup recovery in that second process could treat a live starting thread as orphaned and emit a false `thread_provisioning_failed` event.

## What changed

- Extracted the host daemon robust lock implementation into a shared `@bb/process-utils` data-directory lock module, including stale-lock recovery and compromise handling.
- Added a `server.lock` acquired before logger creation, `initDb()`, and startup recovery.
- Let a recently crashed owner's lock age through the stale window during startup while still rejecting a lock refreshed by a live server before SQLite opens.
- Made compromise re-acquisition timers non-owning so releasing the lock cannot pin a shutting-down process.
- Kept synchronous exit cleanup armed until asynchronous unlock finishes.
- Held server ownership until shutdown, with cleanup on both normal shutdown and startup failure.
- Kept the host daemon on the same shared implementation through its existing wrapper without forwarding ignored options.
- Wire changes: none. `HOST_DAEMON_PROTOCOL_VERSION` is unchanged.

## How you verified

- Added regressions for rejecting a live server before `initDb()`, reclaiming a fresh orphaned server lock, releasing ownership after startup failure, releasing during compromise re-acquisition, and retaining exit cleanup through asynchronous unlock.
- Added shared lock tests for compromise re-acquisition, stale-lock recovery, and yielding to a live contender, plus host-daemon wrapper coverage.
- `pnpm exec turbo run test --filter=@bb/server --force` — 2,037 tests passed.
- `pnpm exec turbo run test --filter=@bb/host-daemon --force` — 557 tests passed.
- `pnpm exec turbo run test --filter=@bb/process-utils --force` — 25 tests passed.
- `pnpm exec turbo run typecheck --filter=@bb/process-utils --filter=@bb/server --filter=@bb/host-daemon` passed.
- `pnpm exec turbo run build --filter=@bb/server --filter=@bb/host-daemon` passed.
- `node packages/plugin-sdk/scripts/check-npm-version-guard.mjs` passed.

Related to #2190: prevents concurrent bb server access to the database directory, but does not resolve the WAL durability/reversion issue.

Fixes: no linked issue.

> AGENT GENERATED


ymichael:
@slopcop

ymichael:
@slopcop

PR #2441, opened 2026-08-25
Fix path dialogs, shell-search inference, and toast dismissal

SawyerHood:
## What was wrong

Three small, independent problems from mbullington's `agent/garden-upstream-stack`, rebased onto `main`:

- Project path dialogs (new project, change workspace, add machine) assumed the local platform. On a remote host with a different OS the separator, root, and home heuristics were wrong.
- The thread view inferred a "search" intent from any shell command that mentioned `grep`/`rg`, so an unrelated bash call rendered as a search row.
- Raw Sonner toasts (plugin toasts, provider notifications) had no close button, and `AppToastContent` never rendered one for custom content.

## What changed

- `apps/app` path dialogs and `useLocalPathPicker` take the selected host's platform; `HostResponse` exposes it (`packages/server-contract/src/api/hosts.ts`, `packages/sdk/src/areas/hosts.ts`). Offline hosts fall back to the prop platform.
- `packages/thread-view/src/tool-call-parsing.ts` stops inferring search intents from shell text; tests and CLI rendering snapshots updated. The handoff notes live in `plans/pi-session-tool-precedence-and-search-intents.md`.
- `apps/app/src/main.tsx` enables Sonner's `closeButton`; `AppToastContent` renders its own dismiss X.

No wire change. Bottom of the stack: no prerequisite.

## How you verified

`pnpm exec turbo run typecheck` on this layer, plus `turbo run test` for `@bb/app`, `@bb/server`, `@bb/server-contract`, `@bb/sdk`, `@bb/thread-view`, `@bb/mobile`, `@bb/host-daemon` (see the stack validation log in the thread).

Fixes: none (upstream stack split).

> AGENT GENERATED

PR #2442, opened 2026-08-25
Add the Escape-to-stop composer setting

SawyerHood:
## What was wrong

Escape in the composer could only blur it. Users who steer a running thread from the keyboard wanted Escape to stop the running thread instead, without giving up focus.

## What changed

- New app setting `composerEscapeBehavior` (`blur` | `stop-running-thread`, default `blur`) in `packages/domain/src/app-settings.ts`, exposed through the settings API/CLI (`bb settings`), documented in `docs/configuration.md` and the `bb-cli` skill's app-settings reference.
- `PromptBoxInternal` / `FollowUpPromptBox` / `ThreadDetailPromptArea` / `EmbeddedThreadChat` thread the behavior to the key handler; Settings → General gets the "Escape in composer" row.

No wire change. Prerequisite: #2441 (stack order only; no code dependency).

## How you verified

`turbo run typecheck`; `turbo run test` for `@bb/app` (promptbox, thread-detail, embedded-chat), `@bb/server` (general settings), `@bb/cli` (settings output), `@bb/db` (migrate defaults), `@bb/domain`.

Fixes: none (upstream stack split).

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/promptbox/ComposerEditorSlot.tsx
+++ b/apps/app/src/components/promptbox/ComposerEditorSlot.tsx
@@ -19,13 +19,6 @@ const COMPOSER_EDITOR_MAX_HEIGHT_BY_LAYOUT: Record<
   "root-compose": "calc(70dvh - 3rem)",
 };
 
-// TipTap's `blur` command defers to the next animation frame, so blur the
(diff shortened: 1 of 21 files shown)


SawyerHood:

PR #2443, opened 2026-08-25
Add sidebar-navigation and Changes-view replacement slots

SawyerHood:
## What was wrong

Plugins could replace the thread list but not the two other sidebar/secondary-panel surfaces people wanted to own: the navigation controls block (New thread / Search / Extensions / plugin panels) and the whole Changes view (toolbar + file list). Plugin-owned dialogs also lacked a host-provided non-modal drawer.

## What changed

Two exclusive replacement slots and one helper, all `experimental_` and entered in `docs/api_to_audit.md`:

- `app.slots.experimental_sidebarNavigation`: a semantic item list (`new-thread`, `search-threads`, `extensions`, `plugin-panel:<pluginId>/<panelId>`), `activeItemId`, `experimental_activate(itemId, { openInSplit })`, `experimental_Original`. bb keeps the drawer, search field, thread list, footer, and shortcuts (`SidebarNavigationRegion`, `sidebarNavigationItems.ts`). Settings → Appearance → Navigation picks the provider.
- `app.slots.experimental_changesView`: replaces the Changes toolbar and file list; `ChangesViewHost` passes `experimental_target` (file/commit intents with a sequence) and `experimental_Original`. The built-in view remembers its selection per environment and its line-overflow mode.
- `experimental_ResponsiveDrawer` (`PluginResponsiveDrawer`): the shared persistent responsive drawer for plugin UIs.
- `PluginSourceCodeRendererProps.experimental_Original` / `PluginDiffRendererProps.experimental_Original`.
- Plugin SDK bumps to 0.4.19; `plugin-app-collector`, `plugin-slots`, the testing harness, and the `replacement-lab-*` example plugins cover the new slots.

No wire change. Prerequisite: #2442 (stack order only).

## How you verified

`turbo run typecheck`; `turbo run test` for `@get-bb/plugin-sdk`, `@bb/app` (sidebar, secondary-panel, settings, plugin), `@bb/server` (plugin-authoring docs), `@bb/cli` (packaged plugin build with the examples).

Fixes: none (upstream stack split).

> AGENT GENERATED

PR #2444, opened 2026-08-25
Let plugins intercept CLI invocations and agent tool calls

SawyerHood:
## What was wrong

A policy plugin had no way to observe or block a `bb` command or a bb-defined agent tool call before it ran.

## What changed

- Plugin API: `bb.events.on("experimental_invocation.before", handler)` returning `{ allowed, reason? }`; the server runs every loaded plugin's handlers in deterministic order (`plugin-service.ts`, `plugin-api.ts`) and records invocation events (`packages/domain` change kinds, `event-pruning`).
- `POST /api/v1/plugins/invocations/preflight` (`apps/server/src/routes/plugins.ts`); the CLI asks it once per executable command (`apps/cli/src/invocation-preflight.ts`, registered with the lazily loaded command groups in `index.ts`) and before proxying a plugin command. It **fails open** when the server is unreachable or predates the route, and **blocks** on an explicit block, a malformed 2xx, any other non-2xx, or a server that accepts the connection and then goes quiet (headers/body timeout). `bb guide` and `bb manager` are never preflighted.
- `apps/server/src/internal/tool-calls.ts` consults the same policy for bb-defined agent tools; a handler failure returns a tool error to the agent instead of failing the turn.
- Fake plugin host (`packages/plugin-sdk/src/testing`) supports scripted decisions; the plugin-authoring skill and `bb guide plugins` document the hook.

No wire change. Prerequisite: #2443 (stack order only).

Open decision for reviewers: fail-open vs fail-closed when the server is down, and whether a per-handler timeout is required before this ships.

## How you verified

`turbo run typecheck`; `turbo run test` for `@bb/cli` (`invocation-preflight.test.ts`, packaged plugin build), `@bb/server` (`plugin-invocation-events.test.ts`, internal tool calls), `@get-bb/plugin-sdk` (fake host), `@bb/db`.

Fixes: none (upstream stack split).

> AGENT GENERATED


SawyerHood:

PR #2463, opened 2026-08-26
Keep worktree archive from detaching external children

jshph:
## User story

As a user coordinating a parent thread with children in separate worktrees, I expect **Archive worktree** to affect only that worktree. If a child is working somewhere else, archiving the parent's worktree must not silently detach that child and flatten the thread tree.

This came from a real session: one worktree held an orchestrator thread, while nine active child threads used isolated worktrees. Archiving the shared worktree also cleared the parent link on all nine isolated children. The links then had to be restored manually before orchestration could continue.

## What was wrong

The environment archive path correctly selected only threads attached to the chosen worktree. However, it archived each selected thread through the ordinary single-thread helper. That helper deliberately releases active children when their parent is archived, without checking which environment those children use.

The result was a scope leak: an action on one worktree changed thread ownership in nine other worktrees. Unarchiving the original worktree also could not restore those links because the ownership changes had already been persisted.

## What changed

The archive lifecycle now makes child-release behavior explicit:

- Ordinary single-thread and hierarchy archive operations still release children as before.
- Environment/worktree archive preserves child ownership, so it does not mutate threads outside the selected worktree.
- A regression test covers a parent in one managed worktree with an active child in another. It verifies that the parent is archived, the child stays active, and the child's parent link remains intact.

There are no wire, daemon protocol, CLI, configuration, or documentation changes.

## How you verified

- `pnpm exec turbo run test --filter=@bb/server -- test/public/public-thread-parenting.test.ts` — 12 tests passed.
- `pnpm exec turbo run typecheck --filter=@bb/server` — passed.
- Re-ran both checks on a clean branch containing only this commit on top of current `get-bb/bb:main` (`31d66d9d4`).

> AGENT GENERATED

### DIFF
--- a/apps/server/src/services/threads/thread-archive.ts
+++ b/apps/server/src/services/threads/thread-archive.ts
@@ -1,4 +1,5 @@
 import {
+  archiveThread,
   listLiveThreadsInEnvironment,
   listUnarchivedAssignedChildThreads,
   listUnarchivedHiddenSourceThreads,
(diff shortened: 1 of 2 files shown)


SawyerHood:

PR #2476, opened 2026-08-26
Respect Pi’s scoped models in the provider model picker

wy3z:
## Summary

Make BB’s Pi model picker respect Pi’s workspace-aware scoped model configuration and selected default.

Pi now runs through the `provider-pi` plugin and exposes its model catalog over RPC. The catalog previously returned every authenticated model, ignoring the scope and default Pi applies to its own picker.

## Changes

- Report Pi’s resolved scoped models and selected default through the existing private extension channel.
- Synchronize model-scope state before serving a catalog result.
- Preserve Pi’s scope ordering in BB’s model picker.
- Use Pi’s selected model as the picker default when available.
- Keep explicitly scoped dated model snapshots selectable.
- Isolate catalog scope state by child-process generation so a restarted Pi child cannot inherit stale scope or default state.
- Fall back to the full available catalog when Pi reports no active scope.

## Testing

- Pi provider suite: 20 files, 116 tests passed.
- Pi provider Turbo typecheck passed.
- `git diff --check` passed.
- Added a process-backed regression test that:
  - starts a scoped catalog child,
  - forces it to exit,
  - changes the scope for the replacement child,
  - verifies the new generation’s catalog and scope are used together.
- Tested against a real local Pi installation and workspace configuration:
  - only the five configured scoped models were returned,
  - scope ordering was preserved,
  - the configured first model was selected as default,
  - repeated catalog requests returned the same scoped result.

### DIFF
--- a/plugins/provider-pi/src/bridge/bb-pi-extension.ts
+++ b/plugins/provider-pi/src/bridge/bb-pi-extension.ts
@@ -245,6 +245,8 @@ export default function bbExtension(pi) {
       }
       case "leaf":
         return { leafId: currentLeafId() };
+      case "model-scope":
+        return currentModelScope();
       default:
         throw new Error("unknown bridge request " + String(message.method));
     }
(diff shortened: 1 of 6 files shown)


SawyerHood:
@slopcop review

PR #2479, opened 2026-08-26
Fix terminal hyperlinks and wrapped selection copying

### LINKED ISSUE #2469: Terminal hyperlink targets can become inaccessible in bb
## Summary

Terminal programs can render named links using OSC 8 escape sequences. In bb terminals, the child process is not told that hyperlink support is available, so programs that conditionally emit OSC 8 may output only a plain-text label and leave the target inaccessible.

On desktop/web, copying a terminal selection from the context menu can also return only the clicked visual row when the selection spans soft-wrapped rows.

## Versions

- bb 0.40.0
- macOS
- Reproduced in a bb terminal
- Confirmed against origin/main at commit `31d66d9d45bc8ff1483d5e4c6401c434b610dc4b`

## Steps to reproduce

1. Open a terminal in bb.
2. Check whether hyperlink support is advertised:

   ```sh
   printf 'FORCE_HYPERLINK=%s\n' "${FORCE_HYPERLINK:-<unset>}"
   ```

3. Run any terminal program that emits a named OSC 8 hyperlink only when terminal hyperlink support is detected.
4. Try to activate the rendered label.
5. Print a URL long enough to soft-wrap:

   ```sh
   python3 -c 'print("https://example.com/" + "abcdefghij" * 20)'
   ```

6. Select the wrapped URL and copy it using the terminal context menu.

## Expected behavior

- Child processes can detect OSC 8 hyperlink support.
- Named hyperlinks can be activated through the existing safe external-navigation flow, with the exact target disclosed before opening.
- Copying from the terminal context menu returns the exact xterm selection across soft-wrapped rows.

## Actual behavior

- Hyperlink capability is not advertised to child processes, so named links may be flattened to labels with no usable target.
- Context-menu copying may capture only one visual row of a soft-wrapped selection.

## Relevant code

- [Host terminal environment](https://github.com/get-bb/bb/blob/31d66d9d45bc8ff1483d5e4c6401c434b610dc4b/apps/host-daemon/src/terminals/terminal-manager.ts#L352-L363)
- [Web/desktop terminal link handling](https://github.com/get-bb/bb/blob/31d66d9d45bc8ff1483d5e4c6401c434b610dc4b/apps/app/src/components/thread/terminal/ThreadTerminalView.tsx#L807-L829)
- [Web/desktop terminal event handling](https://github.com/get-bb/bb/blob/31d66d9d45bc8ff1483d5e4c6401c434b610dc4b/apps/app/src/components/thread/terminal/ThreadTerminalView.tsx#L1049-L1073)

## Existing-issue check

- #1764 is related to mobile touch selection and copy controls, but does not cover terminal capability advertisement or desktop/web context-menu copying.
- #1590 concerned Clipboard API failures on plain-HTTP LAN origins, which is a different root cause.
- No existing issue or fetched origin branch was found that covers this failure.

## Priority

Medium

> AGENT GENERATED

fdx-peter:
## Human comments

Yeah so basically I tried to claude /login in a terminal and the inability to click the long auth link (or copy-paste) was super fustrating......so I told bb to fix itself.

## What was wrong

The host terminal environment did not advertise OSC 8 support, so programs using `supports-hyperlinks`-style detection could flatten named hyperlinks into display text and discard the target. The clients also did not install an OSC 8 activation handler, and desktop/web context-menu copying did not use xterm's logical selection, so a soft-wrapped selection could be reduced to one rendered row. See #2469.

## What changed

- Advertise hyperlink support to spawned terminal processes with `FORCE_HYPERLINK=1`.
- Route desktop/web OSC 8 activation through bb's URL navigation after disclosing and confirming the exact target in bb's responsive dialog; visibly detected URLs retain direct activation.
- Add terminal context-menu actions for opening/copying links and copying the exact xterm selection across soft-wrapped rows.
- Route mobile OSC 8 activation through the native bridge and require confirmation showing the exact URL before opening it.
- Preserve link provenance through the mobile bridge so visibly detected URLs still open directly, matching desktop/web behavior.
- Regenerate the committed mobile terminal page and add regression coverage for the host environment, desktop/web handlers, mobile confirmation, and generated asset parity.
- No `HOST_DAEMON_PROTOCOL_VERSION` bump is needed because this changes only the environment of newly spawned PTYs; no server/daemon wire field or message changed.
- `FORCE_HYPERLINK` is the available opt-in for programs that do not recognize bb as a terminal. Programs that treat it as an unconditional override may also emit OSC 8 when their output is redirected; callers requiring plain redirected output can set `FORCE_HYPERLINK=0` for that command.

## How you verified

- App terminal test suite — 41 tests passed.
- Focused host-daemon terminal tests — 29 tests passed.
- Mobile terminal test suite — 32 tests passed.
- `pnpm exec turbo run typecheck lint --filter=@bb/app --filter=@bb/host-daemon --filter=@bb/mobile` — all tasks passed; lint reported 0 errors.
- `pnpm run build` — all 12 production build tasks passed.
- Oxfmt and `git diff --check` passed for the changed source files.
- Exercised the full flow with Computer Use against an isolated development instance: OSC 8 confirmation/open, direct visible-URL opening, combined link/selection context menu, and copying a 620-character soft-wrapped selection as one logical value.

Fixes #2469

> AGENT GENERATED


SawyerHood:
Closing as requested. The investigation found useful hyperlink changes, but the macOS context-menu event order can still lose the selection. The mobile changes also target code removed by #2515, and the branch now conflicts with main.

PR #2483, opened 2026-08-26
fix(security): harden runtime and release trust boundaries

springhealth-everettmorgan:
## What was wrong

bb is a high-privilege agent runner: its unauthenticated API could be exposed off-loopback, child processes inherited ambient credentials, non-loopback plaintext transports could carry sensitive data, server-delivered updates were trusted without independent publisher verification, and sensitive diagnostics/data directories lacked consistent redaction and permission hardening. Privileged release workflows also allowed unsafe manual refs and an unpinned npm tool fallback.

## What changed

- Enforced loopback-only server binding and removed request `Host`/`X-Forwarded-Host` expansion from the browser-origin allowlist, closing the wildcard/DNS-rebinding exposure path.
- Removed unsafe remote-dev and mobile E2E wildcard-bind paths; reject non-loopback plaintext HTTP for configured server URLs, desktop custom targets, and host-daemon clients.
- Made machine enrollment HTTPS-only off-loopback, removed PATH/npm fallbacks, made daemon auto-update opt-in, and disabled Linux AppImage self-installation while retaining version checks/manual verification.
- Added a minimal explicit environment allowlist for plugin/provider processes, preserving explicit provider overlays and login-shell `PATH`.
- Added centralized best-effort credential redaction, owner-only data/log directories, restrictive sensitive-file modes, and regression coverage.
- Added `main` gates and QA secret isolation to privileged release/deploy paths, pinned npm `11.6.2`, updated the flagged transitive dependencies through pnpm overrides/lockfile regeneration, and added pinned GitHub artifact attestations. Attestations are documented as provenance, not Linux code signatures.
- Updated user guides, machine/mobile plans, and tests. No host-daemon wire contract changed; `HOST_DAEMON_PROTOCOL_VERSION` was not modified.

## How you verified

- `pnpm exec turbo run typecheck`: **79/79 tasks passed**.
- `pnpm exec turbo run lint`: passed with existing React compiler warnings.
- `pnpm exec turbo run build --filter=@bb/desktop --filter=@bb/server --filter=@bb/host-daemon --filter=bb-app`: **12/12 tasks passed**.
- Affected Turbo tests passed: config **110**, logger **12**, process-utils **21**, host-daemon **564**, server **2,051**, desktop **229**, and bb-app **73**.
- Serial fake integration project: **25 files / 55 tests passed**.
- Native-roots golden project: **1 file / 22 tests passed**.
- Release hardening check, all workflow YAML parsing, installer shell syntax, scoped formatting, and `git diff --check` passed.

The default full parallel test graph is not green because an untouched `@bb/host-daemon-contract/test/payload-size.test.ts` has Node 26 gzip-size golden differences and several existing resource-sensitive integration/server tests time out under the full graph. The same integration cases pass serially; the unrelated golden was not changed. The repo-wide formatter also reports an unrelated pre-existing escape error at `apps/mobile/e2e/flows/phase4a-work-rows.yaml:277`.

Residual risks are explicit: loopback API callers remain unauthenticated, plugins/providers remain full-trust code, daemon artifacts remain server-trusted and unsigned when opt-in auto-update is used, SQLite remains unencrypted, npm Trusted Publishing retains job-scoped OIDC during dependency install/build, and external GitHub environment/branch-protection settings were not verifiable from this checkout.

> AGENT GENERATED


SawyerHood:

PR #2488, opened 2026-08-26
Add equal-grid snapping for split panes

brsbl:
## Why

Users usually want a newly created split to return to equal-sized panes, but free resizing makes that canonical grid hard to recover precisely. This layer adds one shared equal-grid interaction to tab, page/thread, and thread/right-panel dividers without removing arbitrary sizing.

## User-visible behavior

- Two siblings snap to halves, three to thirds, four to quarters, and larger sibling groups use the same equal-grid rule.
- Nested row and column split nodes own their grids independently; there is no diagonal snapping.
- Approaching within 12 px—or crossing a target during a fast drag—captures the equal boundary.
- Once captured, the divider holds through a 30 px release band. A second consecutive out-of-band sample confirms release after coarse pointer movement.
- A temporary full-span guide marks the active equal boundary.
- The behavior applies to right-panel tab splits, workspace page/thread splits, and the center divider between a thread and its right panel.
- Existing 15–85% bounds and arbitrary post-release sizes remain available.

## Reliability and performance

- Grid and divider geometry are read once when a drag begins; pointer movement uses cached geometry.
- Center, tab, page, and thread dividers share the same free → captured → released state machine and resistance thresholds.
- The center-divider bridge previews adjacent flex sizes locally and commits once through the panel API on release, matching the internal split path.
- Release, cancellation, lost pointer capture, a zero-button sample, window blur, replacement drag, and unmount all clear the guide and resize state.

## Screenshot evidence

All captures use Chrome for Testing 151.0.7922.71 at the same 1440×900 CSS viewport (DPR 2), the same one-file `Split tab UX verification` fixture, and one coarse pointer jump that crosses the equal boundary by 80 px. The before revision is exact parent head `1dbfc83bd587b98968d6b0a839da9f81751ce3b9`; the after revision is exact PR head `a128a9917ffe49588cb2c84d573f01113834e5a6`.

### Thread / right-panel divider

Before, the fast crossing follows the sampled pointer past center and leaves an arbitrary-width right panel with no guide.

Before: center divider misses the equal grid

After, the same crossing holds at 50/50 and renders a full-height guide through the owning split.

After: center divider captures the equal grid

### Vertical tab split

Before, the stacked Diff/Info divider lands 80 px beyond the equal-height boundary with no guide.

Before: vertical tab split misses equal heights

After, the same crossing holds both panes at equal heights and renders the full-width horizontal guide.

After: vertical tab split captures equal heights

## Validation

- Capture manifests record the exact parent/head revisions above, identical fixture and viewport, zero rendered error states, no guide and an 80 px overshoot before this layer, and exact 50/50 geometry with the correct axis-spanning guide after it.
- Regression coverage in this layer targets equal boundaries for multiple sibling counts and both axes, fast-crossing capture, shared 12/30 px hysteresis, confirmed coarse-sample release, center-divider single-writer behavior, persistence, and cleanup on every drag terminator.
- Tests, typechecks, lint, and other CI-equivalent checks run on the pull request through GitHub Actions; none were run locally for this screenshot audit.

BB-Thread-ID: thr_q8degf2y66

### DIFF
--- a/apps/app/src/components/secondary-panel/SecondaryPanelLayout.test.tsx
+++ b/apps/app/src/components/secondary-panel/SecondaryPanelLayout.test.tsx
@@ -305,6 +305,19 @@ beforeEach(() => {
 });
 
 describe("SecondaryPanelLayout", () => {
+  it("registers the thread and right panel as one two-pane resize grid", () => {
+    renderLayout({
+      isCompactViewport: false,
+      open: true,
+      renderPanel: createPanelRenderer(),
+      resetKey: "thread-grid",
+    });
+
+    expect(
+      screen.getByTestId("panel-group").dataset.splitResizeGridRoot,
+    ).toBe("");
+  });
+
   it("preserves routed main content when the panel state identity changes", () => {
     const frames = installAnimationFrameQueue();
     const view = renderLayout({
--- a/apps/app/src/components/secondary-panel/SecondaryPanelLayout.tsx
+++ b/apps/app/src/components/secondary-panel/SecondaryPanelLayout.tsx
@@ -334,6 +334,7 @@ export function SecondaryPanelLayout({
         <PanelGroup
           key={panelGroupKey ?? resetKey}
           ref={horizontalPanelGroupRef}
+          data-split-resize-grid-root=""
           direction="horizontal"
           className="@container h-full min-w-0 flex-1"
           // A clipped group cannot be programmatically scrolled by an iframe's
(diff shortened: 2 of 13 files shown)


brsbl:
Closing this feature layer because split-resize snapping has been cut from the stack. PRs #2486 and #2487 remain unchanged.

brsbl:
Reopened after correcting the interaction model: this now uses one shared midpoint grid per split surface with fast-crossing capture and magnetic release hysteresis.

brsbl:
@slopcop Please review the full native PR stack #2486 → #2487 → #2488. The screenshot evidence has been audited and refreshed against each layer’s exact merge-base or parent-head and PR-head revisions, using matched fixtures and viewports throughout.

PR #2494, opened 2026-08-26
Filter Linux workspace open targets by XDG category

wy3z:
## Summary

- restrict generic Linux desktop-entry discovery to file managers, text editors, and terminal emulators
- use the standard XDG `Categories` values `FileManager`, `TextEditor`, and `TerminalEmulator`
- keep unrelated desktop applications out of BB's Open in/Open with menus

## Root cause

PR #403 added generic Linux desktop-application discovery for workspace open
targets. It currently includes every visible `.desktop` entry with
`Type=Application`, an `Exec` command, and neither `Hidden=true` nor
`NoDisplay=true`.

As a result, BB's Open in/Open with menus can contain nearly every installed
desktop application on Linux, including media players, settings utilities,
messaging clients, and other software that is not a useful workspace target.

This change additionally requires a relevant standard XDG category:
`FileManager`, `TextEditor`, or `TerminalEmulator`. Existing explicit BB launch
adapters remain unaffected.

## Validation

- `pnpm exec turbo run typecheck test --filter=@bb/local-open-targets --force`
- `git diff --check`

> AGENT GENERATED

### DIFF
--- a/packages/local-open-targets/src/index.ts
+++ b/packages/local-open-targets/src/index.ts
@@ -369,6 +369,16 @@ function parseDesktopEntryValue(line: string): [string, string] | null {
   ];
 }
 
+const LINUX_WORKSPACE_APPLICATION_CATEGORIES = new Set([
+  "FileManager",
+  "TerminalEmulator",
+  "TextEditor",
+]);
+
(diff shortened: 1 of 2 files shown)


SawyerHood:
@slopcop review

PR #2500, opened 2026-08-26
Add same-thread context clearing

MateoCerquetella:
## Human comments

https://github.com/user-attachments/assets/1be2d930-b76d-4667-973d-b17742f80011

## What was wrong

BB had no same-thread provider-context boundary. Clearing context required a handoff, which split one piece of work across BB threads, while provider-session lookup always resumed the latest historical provider identity.

## What changed

- Add a built-in `/clear` command that releases an idle runtime and records a visible `Context cleared` boundary without changing the BB thread, workspace, history, or sticky execution settings.
- Make provider-session lookup ignore identities before the latest clear boundary so the next prompt starts a fresh provider conversation.
- Serialize clears against prompt sends and reject active, starting, stopping, or interaction-blocked threads.
- Add the public API, SDK `threads.clearContext`, `bb thread clear`, guide, and bb-cli skill surfaces.
- Reuse the existing release-only `thread.stop` wire command; `HOST_DAEMON_PROTOCOL_VERSION` does not change.

- Guard queued-message and parent-notification dispatch against context clearing; retain blocked messages and retry them in a fresh provider session after clearing.
- Apply the context boundary to timeline byte budgets and reject pagination cursors from earlier context.

## How you verified

- `pnpm exec turbo run typecheck --filter=@bb/domain --filter=@bb/db --filter=@bb/server --filter=@bb/sdk --filter=@bb/cli --filter=@bb/thread-view --filter=@bb/app --filter=@bb/server-contract --filter=@bb/templates`
- `pnpm exec turbo run lint --filter=@bb/domain --filter=@bb/db --filter=@bb/server --filter=@bb/sdk --filter=@bb/cli --filter=@bb/thread-view --filter=@bb/app --filter=@bb/server-contract --filter=@bb/templates`
- Focused domain, DB, server, CLI, SDK, composer, thread-view, and template tests, including active-thread rejection, repeated clears, already-released runtimes, and send/clear races.

- Added six regression cases for queued sends, parent notifications, retry after clear, and large historical events. All six fail against the original PR implementation and pass with the fixes.
- All 43 focused server tests and the server typecheck passed locally. Full GitHub CI passed on `bc3b415a999766cbb4e3186cb093fd8ae9ea3062`, including all test shards and Linux/macOS package smoke tests.

> AGENT GENERATED


nicolay-i:
This is inconvenient to use. I compact much more often than I clear the context. In fact, I never use clear because it loses the history, lol. Now, instead of compacting, I can accidentally lose several hours of work because I effectively have to start all over again. Please put clear after compact, or simply make it create a new chat instead, which could be called something like "New similar chat".

PR #2504, opened 2026-08-26
Page timeline summaries at logical boundaries

ymichael:
## What was wrong

Timeline event-count and byte windows were treated as if they were semantic partitions. They are not: assistant messages, completed-turn groups, system rows, and item lifecycles can span a raw cut. A row could be loaded as context and suppressed on one page, then fall before the next page's start, so no page emitted it. The affected production thread lost the assistant message at source events `3365-3375` this way.

## What changed

The default top-level timeline summary now pages only at its existing user-message anchors. It does not expose event-count or byte-window cuts as page boundaries. An older page reads one following logical segment as projection-only overlap, then trims at its requested cursor; this preserves assistant/system rows whose semantic placement crosses the raw anchor. The normal stable-id page merge reconstructs the same semantic projection as an unwindowed build. The oldest page also owns the pre-first-user thread prelude.

Nested timeline consumers retain bounded event/byte windows, and completed-turn expansion retains the separately paginated turn-details resource from #2464. Idle summary pages avoid hydrating command output bodies because those outputs are collapsed; active pages keep the existing preview behavior. The grouping changes preserve visible assistant boundaries and restore source ordering for late thread-scoped system rows.

This is stacked on #2464. There is no server/daemon wire change, so `HOST_DAEMON_PROTOCOL_VERSION` is unchanged.

## How you verified

- Added a regression that walks logical pages and compares their exact recombination to the unpaginated projection, including a segment above both the event and byte thresholds.
- Added coverage for assistant messages at page seams and for thread preludes before the first user anchor.
- Copied-production audit: `784/784` threads and `888,665` events matched exactly, with zero missing, duplicated, reordered, or source-range-mismatched rows.
- Repeated that audit at the desktop's 20-segment page size with a `250`-event threshold and `512 KiB` byte threshold: `784/784` threads matched exactly across `876` page responses; 47 threads crossed at least one page seam and the largest used 17 pages. The comparison included system rows and exact source ranges. The previously missing `3365-3375` assistant row is present in the affected thread.
- `pnpm exec turbo run test --filter=@bb/server --force`: 2,056 passed.
- `pnpm exec turbo run test --filter=@bb/thread-view --force`: 372 passed.
- `pnpm exec turbo run test --filter=@bb/db --force`: 410 passed.
- `pnpm exec turbo run typecheck --filter=@bb/server --filter=@bb/thread-view --filter=@bb/db`: passed.

> AGENT GENERATED

### DIFF
--- a/apps/server/src/services/threads/timeline-pagination.ts
+++ b/apps/server/src/services/threads/timeline-pagination.ts
@@ -184,11 +184,22 @@ export function paginateTimelineRows(
       rows: [...rows],
     };
   }
-  // Every window ends strictly before its cursor, so no segment at or past the
-  // cursor was read and none has to be trimmed off here.
-  const selectedSegments = segments.slice(-page.segmentLimit);
+  // An older semantic-summary read can overlap its cursor segment so rows that
+  // cross the raw event boundary are projected with context from both sides.
+  // The cursor segment is context only; this page owns everything before it.
+  const cursorIndex =
+    page.kind === "older"
+      ? segments.findIndex(
+          (segment) =>
+            segment.cursor.anchorSeq === page.beforeCursor.anchorSeq &&
+            segment.cursor.anchorId === page.beforeCursor.anchorId,
+        )
+      : -1;
+  const eligibleSegments =
+    cursorIndex === -1 ? segments : segments.slice(0, cursorIndex);
+  const selectedSegments = eligibleSegments.slice(-page.segmentLimit);
   const hasOlderRows =
-    knownHasOlderSegments ?? segments.length > selectedSegments.length;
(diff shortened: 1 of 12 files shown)


ymichael:
Closing in favor of a single rewrite PR. This PR's strategy (page only at user-message anchors) was right, but layered on top of the transport-window machinery it left three pagination contracts in the tree. The rewrite pages by slicing one canonical projection, so recombination correctness holds by construction and independent of any event/byte limits. Carrying over from here: the seam regression tests, the pages-recombine-exactly test, restoreLateSystemRowSourceOrder, the grouping revert, and the prod-corpus audit methodology (made repeatable this time).

> AGENT GENERATED

PR #2511, opened 2026-08-27
feat(permissions): configure workspace write roots

smsunarto:
## What was wrong

BB had no host-local configuration for paths that every provider workspace-write sandbox must access. Users had to rely on provider-specific configuration or repeated approval requests.

## What changed

Added the startup-only `BB_ADDITIONAL_WORKSPACE_WRITE_ROOTS` path-list setting. The host daemon validates absolute paths, removes duplicates, and adds them to each runtime's workspace-write roots. The launcher, CLI skill, configuration guide, and customization guide now document the setting.

## How you verified

- `pnpm exec turbo run typecheck --filter='!@bb/plugin-registry' --force`
- `pnpm exec turbo run test --filter=@bb/host-daemon --force`
- Host daemon suite: 563 tests passed.

> AGENT GENERATED


smsunarto:
Closing because this was intended as a local sandbox workaround, not an upstream product change.\n\n> AGENT GENERATED

PR #2535, opened 2026-08-27
Add Junie as an ACP provider

preigile:
## Human comments

Support for the Junie agent has been added via the ACP

## What was wrong
 
bb had no registration for JetBrains' Junie

## What changed

-  `packages/provider-bridge-acp/src/wire.ts`: `SessionConfigSelectOptions` is parsed as the union the ACP spec defines, and groups are flattened to their values (bb has no grouped picker). The lenient path for non-model options flattens groups the same way. Any ACP agent that groups its options benefits, not only Junie.
- `packages/provider-bridge-acp/src/bridge/provider-maintenance.ts`: `loginCommand`, `readAccount` and `readUsage` are optional. Health reports an installed agent whose credentials bb cannot read as ready instead of guessing "signed out", and usage without `readUsage` is `supported: false`. Added `JUNIE_ACP_MAINTENANCE` – installer only, JetBrains' documented script – and `packages/provider-bridge-acp/src/dialect.ts` gains the `junie` dialect that carries it (Junie reads no vendor side channel; the dialect exists only for that).
- `plugins/provider-acp`: the `acp-junie` agent (`junie --acp true`,  `.junie/skills` and `.agents/skills` native roots, `fork: "none"` until Junie's `initialize` reply is verified), its brand icon, and `providerInstallation` so bb installs the CLI itself the same shape as `acp-cursor`. The agent is listed whether or not the CLI is present, because an installed-only listing hides the row the Install action lives on. Reserving an id is now an explicit `reserved` flag instead of a side effect of always being listed, so `acp-junie` stays overridable via `customAgents` while `acp-cursor` does not.
- `apps/web`: the Junie mark in the "Works with" row and in the two provider lists on the landing page.
- User-facing surfaces updated in the same change: `docs/configuration.md`, the bb providers guide template, the bb-cli skill, the bb-app README. 
No host-daemon wire change: the dialect id travels in the existing `providerOptions`, so  `HOST_DAEMON_PROTOCOL_VERSION` is unchanged.                                                                                                     

## How you verified

- Tests added
- Commands run (turbo typecheck/test filters)
- Manual checks.


preigile:
@SawyerHood Hi! I added support for JetBrains' Junie agent (ACP provider). Would appreciate a review of the changes when you have a moment. Thanks!

SawyerHood:

PR #2539, opened 2026-08-27
Fix Codex subagent relinking after session resume

### LINKED ISSUE #2538: Bug - Refusing to start a competing turn for thread "<threadid>" while another turn is active or starting (Codex)
### Summary

After BB recreates a Codex provider session, a root agent can resume an existing native subagent with `followup_task`, but BB projects the child's next turn as foreground/root work. If the visible root finishes first, that misclassified child occupies the foreground slot and BB rejects the user's next prompt as a competing turn.

The child should remain attached to the root's delegation, and the next user prompt should still go to the root.

### Versions and environment

- BB `0.40.0` desktop app; macOS `26.2` (`25C56`)
- Built-in `provider-codex`; Codex CLI `0.149.0`
- Source repro: [`main` at `31a190ddb5f0ffb86cb8dcd80482497663de2685`](https://github.com/get-bb/bb/commit/31a190ddb5f0ffb86cb8dcd80482497663de2685)

### Human/UI reproduction

1. Open a BB thread using the Codex provider.
2. Ask the root to spawn a subagent named `repro_child`, wait for it to finish, and then finish the root turn.
3. Cause BB to recreate the provider session while preserving the same Codex rollout. Normally, leave the thread idle until the session is reaped, then return. Quitting/reopening or reconnecting can also exercise this boundary.
4. In the same root thread, send:

   ```text
   Use followup_task to resume /root/repro_child. Ask it to keep working for at
   least 30 seconds, then finish your own response immediately without waiting.
   ```

5. As soon as the root response finishes, while the child is still running, send:

   ```text
   continue
   ```

6. BB rejects the prompt as a competing turn. It should deliver the prompt to the root because the running child is delegated work.

Negative controls: this does not reproduce while the original translator remains live; replacing `followup_task` with `send_message` should not create another child turn.

### Deterministic source-level reproduction

Add this test inside `describe("codex subagent activity correlation", ...)` in `plugins/provider-codex/src/translator.test.ts`. All identifiers are synthetic, and the helpers already exist in that block.

```ts
it("links a rawless resumed subagent when its child turn starts", () => {
  const harness = createHarness();

  expect(
    harness.translate(
      subAgentActivity({
        id: "synthetic-followup-call",
        kind: "interacted",
        agentThreadId: "synthetic-agent-thread",
      }),
    ),
  ).toEqual([]);

  const events = harness.translate(childTurnStarted("synthetic-child-turn"));
  expect(events).toContainEqual(
    expect.objectContaining({
      type: "item/started",
      item: expect.objectContaining({
        type: "delegation",
        childRef: "synthetic-agent-thread",
      }),
    }),
  );
  expect(events).toContainEqual(
    expect.objectContaining({
      type: "turn/started",
      parentToolCallId: harness.itemId("synthetic-followup-call"),
    }),
  );
});
```

Run:

```bash
pnpm --filter bb-plugin-provider-codex test -- translator.test.ts -t "links a rawless resumed subagent"
```

On unmodified `main`, the second assertion fails: the child turn is emitted without a delegation or `parentToolCallId`.

### Expected vs actual

**Actual:** the resumed child is persisted without `parent_tool_call_id`, and the next user prompt can fail with:

```text
Refusing to start a competing turn for thread "<synthetic-thread-id>" while another turn is active or starting
```

**Expected:** the resumed child turn carries its delegation's `parentToolCallId`. A later user prompt targets the root. `send_message` remains non-turn-producing, and genuine concurrent root turns remain rejected.

### Evidence and root cause

The shared runtime correctly treats an unparented `turn/started` as foreground, so the competing-turn guard is behaving as designed. The provider projection is missing the parent link.

After translator recreation, the in-memory child map is empty. Current Codex app-server emits `subAgentActivity(kind: "interacted")` followed by the resumed child's `turn/started`, but does **not** forward the raw `followup_task` collaboration item on this path. The rollout contains the function call, but the app-server notification stream does not. BB currently discards an unknown child's `interacted` activity, so the following child turn is projected as root work.

An anonymized occurrence followed this sequence:

| Event | Persisted result |
|---|---|
| Root calls `followup_task`; child starts | Child has no `parent_tool_call_id` and is stored as root work |
| Visible root completes | Misclassified child still owns foreground state |
| User submits the next prompt | Rejected as a competing turn |
| Child completes | Foreground slot clears |

### Scope

- Reproduces with a historical Codex child resumed after translator/session recreation.
- Does not require any particular project or workspace content.
- Is distinct from `send_message`, which emits the same `interacted` activity but starts no child turn.
- Is not fixed by #1361 because that correlation depends on the original translator's in-memory child map.
- Is not #2370: that report has a genuinely busy/starting root; here delegated child work is misclassified as root work.

Candidate fix and validation: #2539

High priority: normal provider-session lifecycle can block new prompts in long-running Codex threads. The change is confined to Codex event correlation, regression tests, and the host-daemon protocol version; shared turn ownership and the competing-turn guard remain unchanged.

### Checks

- [x] Reproduced on the pinned `main` commit above.
- [x] Searched open and closed issues for the same problem.
- [x] Used only synthetic identifiers in this report.

> AGENT GENERATED

peterfotinis:
## What was wrong

A recreated Codex translator loses its in-memory parent mapping for durable historical subagents. When the root later calls `followup_task`, Codex emits an `interacted` activity for the child, but current app-server does not forward the raw collaboration call. BB therefore discarded the only surviving activity because the fresh translator did not know the child. The following child turn had no `parentToolCallId` and was projected as foreground/root work, violating the runtime invariant that delegated child work must remain attached to its parent tool call. A later user prompt to the root could then be rejected as a competing turn.

Issue #2538 documents this restart/resume gap left after the in-session correlation added in #1361; the earlier investigation in #807 describes the same failure class.

## What changed

- Preserve raw `followup_task` versus `send_message` intent when Codex supplies it, but do not require that notification.
- Hold an unknown `interacted` activity until a non-user-initiated child `turn/started` proves it was turn-producing, then reconstruct the delegation and attach that child turn.
- Discard an unclassified interaction when its parent turn completes without starting a child, so `send_message` cannot steal a later human or subagent turn.
- Prefer queued native user turns and known delegation links before rawless inference.
- Add restart/resume, message-only, native-turn, queued-followup, and concurrent-subagent regression coverage.
- Keep `HOST_DAEMON_PROTOCOL_VERSION` at 171 because this restores the existing `parentToolCallId` contract without a wire schema change.

## How you verified

- Reproduced the failure in an isolated desktop build on Codex CLI 0.149.0: after a full BB restart, a resumed child was stored as a root turn and the next user prompt was rejected as a competing turn.
- Repeated the same workflow on the fixed build: the rawless resumed child received `parentToolCallId`, its delegation row had no self-parent, and a root prompt submitted while the child was running was accepted instead of rejected.
- Codex provider: 242/242 tests passed.
- Runtime turn-state: 9/9 tests passed.
- Repository lint and all 80 typecheck tasks passed; the changed files pass formatting and `git diff --check`.
- Built and launched the isolated Electron desktop app from this branch for the end-to-end checks above.
- After removal of the protocol bump, the focused Codex translator tests passed 41/41 and the host contract tests passed 37/37 on Node 22.

Fixes #2538
> AGENT GENERATED

### DIFF
--- a/packages/host-daemon-contract/src/protocol.ts
+++ b/packages/host-daemon-contract/src/protocol.ts
@@ -335,14 +335,9 @@
 // capacity implied by each model id. The usage event meaning changed across
 // the daemon boundary, so enrolled daemons must update with the server.
 //
-// Version 172 restores Codex native-subagent correlation after a provider
-// session resume. A followup to an agent unknown to the fresh bridge now emits
-// a nested delegation and parents the resumed child turn instead of projecting
-// it as competing root work; send_message remains non-turn-producing.
-//
 // The version mismatch is what triggers the enrolled daemon's automatic update
 // instead of an `invalid-message` reconnect loop.
-export const HOST_DAEMON_PROTOCOL_VERSION = 172 as const;
+export const HOST_DAEMON_PROTOCOL_VERSION = 171 as const;
 
 /**
  * Absolute ceiling for any executable artifact delivered to a host daemon —
--- a/packages/host-daemon-contract/src/protocol.ts
+++ b/packages/host-daemon-contract/src/protocol.ts
@@ -335,9 +335,14 @@
 // capacity implied by each model id. The usage event meaning changed across
 // the daemon boundary, so enrolled daemons must update with the server.
 //
+// Version 172 restores Codex native-subagent correlation after a provider
+// session resume. A followup to an agent unknown to the fresh bridge now emits
+// a nested delegation and parents the resumed child turn instead of projecting
+// it as competing root work; send_message remains non-turn-producing.
+//
 // The version mismatch is what triggers the enrolled daemon's automatic update
 // instead of an `invalid-message` reconnect loop.
-export const HOST_DAEMON_PROTOCOL_VERSION = 171 as const;
+export const HOST_DAEMON_PROTOCOL_VERSION = 172 as const;
 
 /**
  * Absolute ceiling for any executable artifact delivered to a host daemon —
--- a/packages/host-daemon-contract/test/contract.test.ts
+++ b/packages/host-daemon-contract/test/contract.test.ts
@@ -1046,7 +1046,7 @@ describe("host-daemon command schemas", () => {
   // mixed version. Version 113 carried the Devin Desktop open target rename
   // and remains part of the protocol lineage.
   it("uses the current host-daemon protocol version", () => {
-    expect(HOST_DAEMON_PROTOCOL_VERSION).toBe(171);
+    expect(HOST_DAEMON_PROTOCOL_VERSION).toBe(172);
     expect(HOST_ARTIFACT_MAX_BYTES).toBe(256 * 1024 * 1024);
   });
 
--- a/packages/host-daemon-contract/test/contract.test.ts
+++ b/packages/host-daemon-contract/test/contract.test.ts
@@ -1046,7 +1046,7 @@ describe("host-daemon command schemas", () => {
   // mixed version. Version 113 carried the Devin Desktop open target rename
   // and remains part of the protocol lineage.
   it("uses the current host-daemon protocol version", () => {
-    expect(HOST_DAEMON_PROTOCOL_VERSION).toBe(172);
+    expect(HOST_DAEMON_PROTOCOL_VERSION).toBe(171);
     expect(HOST_ARTIFACT_MAX_BYTES).toBe(256 * 1024 * 1024);
   });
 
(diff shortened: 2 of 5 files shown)


SawyerHood:
Updated this branch onto current `main`. The current head restores the correct portable gzip expectations and adds rawless child-isolation coverage. Local validation passed 51 host-daemon contract tests, 249 Codex provider tests, and both typechecks. GitHub CI passed all 11 checks.

PR #2541, opened 2026-08-27
Warn when an extraRepos entry cannot track a repository

### LINKED ISSUE #2540: GitHub plugin silently discards extraRepos entries that are not owner/repo
## Summary

The GitHub plugin's `extraRepos` setting silently discards any entry that is not exactly `owner/repo`. I set it to `MPIV-AI/*` expecting the org's repos to be tracked, and nothing happened: no error in the settings form, no warning in the log, no change in `bb github repos`. The setting stored my value and the plugin ignored it, which is indistinguishable from a setting that worked but found nothing.

I expected either the wildcard to be honored or, failing that, to be told it was not.

## Versions and environment

- bb from source at `ad79bbb5e`, run with `scripts/bb-dev-app current`.
- Also reproduced on the packaged desktop app 0.39.0 (`github@0.2.1`).
- macOS 15.6 (Darwin 25.6.0), Node 24.19.0.
- `gh` 2.86.0, authenticated to github.com.

## Steps to reproduce

```bash
bb plugin config github set extraRepos "SOME-ORG/*"
bb plugin reload github
bb github repos
```

Any string that fails `/^[\w.-]+\/[\w.-]+$/` reproduces it. A bare typo does too:

```bash
bb plugin config github set extraRepos "get-bb/bb, nonsense"
bb github sync
```

What does **not** reproduce it: a well-formed `owner/repo` entry, which tracks correctly.

## Expected vs actual

`bb plugin config github` confirms the value is stored:

```
extraRepos = "MPIV-AI/*"  (string)
  Extra repositories — Comma-separated "owner/repo" list to track in addition to repos discovered from BB projects.
```

Actual, `bb github repos` — only the repos discovered from BB project checkouts, with no mention of the setting:

```
mpiv-ai/mpiv-main	(proj_yfvfyfwqkt)
mpiv-ai/kept	(proj_vrmb7zhsvg)
mpiv-ai/rig	(proj_ppswj7vb9d)
mpiv-ai/receptn	(proj_m437zwj9hw)
```

Nothing appears in the server log for the discarded entry. Expected: a warning naming the entry and the shape the setting wants.

## Evidence

The parse keeps only entries that pass `isRepoName` and drops the rest with no `else` branch:

- https://github.com/get-bb/bb/blob/ad79bbb5e/plugins/github/server.ts#L612-L616
- `isRepoName` is the filter: https://github.com/get-bb/bb/blob/ad79bbb5e/plugins/github/server.ts#L344-L346
- The setting's description does not say wildcards are unsupported: https://github.com/get-bb/bb/blob/ad79bbb5e/plugins/github/server.ts#L479-L485

`*` is not in `[\w.-]`, so `MPIV-AI/*` fails the test and the loop body never runs for it.

## What you ruled out

- Not a `gh` auth problem: valid entries in the same setting track fine, and `gh repo list MPIV-AI` returns 26 repos for this account.
- Not a sync failure: the sync reports success, having simply never been given the repo.
- Not a duplicate of #2313. That issue is about the plugin being pinned to github.com and unable to reach a GHE host; this is about the parse dropping input on the host it already talks to.
- Still present on `main` at `ad79bbb5e`.

Low effort. Hits anyone who guesses at the setting's syntax, and the failure is silent, so the time lost is spent wondering whether the plugin is broken. Workaround is to list every repo explicitly.

> AGENT GENERATED

MPIsaac-Per:
## What was wrong

`discoverRepos` parsed the `extraRepos` setting by filtering the split entries through `isRepoName` and keeping the survivors. Anything that failed the test had no `else` branch: no warning, no error, no record. A wildcard such as `owner/*`, or a plain typo, therefore behaved exactly like a setting that was working and had simply found nothing, and the only way to discover the difference was to read the regex.

Fixes #2540

## What changed

- `plugins/github/server.ts`: added an exported `parseExtraRepos` that splits the setting into `valid` and `invalid` entries instead of filtering in place. `discoverRepos` tracks the valid ones as before and emits `bb.log.warn` for each invalid one, naming the entry and the expected `owner/repo` shape.

No wire, CLI, or settings-schema changes. `extraRepos` keeps its meaning and its type; wildcards remain unsupported, they are just no longer silent about it.

## How you verified

- `plugins/github/server.test.ts`: a unit test over `parseExtraRepos` covering a wildcard, a typo, a mixed list, and blank or separator-only input.
- `plugins/github/server.rpc.test.ts`: a behavioral test asserting that with `extraRepos: "acme/widgets, ACME/*"` the valid sibling still syncs one repo and a `warn` entry naming `ACME/*` reaches `harness.logEntries`.
- Both fail before this change and pass after. Reverting `server.ts` alone: 25 passed, 2 failed. With the change: 27 passed.
- `pnpm exec turbo run typecheck test --filter=bb-plugin-github` passes.
- Manually on a dev instance at this commit: `extraRepos = "MPIV-AI/*, mpiv-ai/clipscript"` now logs

  ```
  WARN: [server] [plugin:github] extraRepos entry "MPIV-AI/*" is not an owner/repo name and tracks nothing. List each repository explicitly, for example "owner/repo".
  ```

  while `mpiv-ai/clipscript` beside it still tracks and syncs.

> AGENT GENERATED

### DIFF
--- a/plugins/github/server.rpc.test.ts
+++ b/plugins/github/server.rpc.test.ts
@@ -348,4 +348,21 @@ describe("github plugin RPC behavior", () => {
       },
     });
   });
+  it("warns about an extraRepos entry that cannot track anything", async () => {
+    const host = createFakePluginHost({
+      pluginId: "github",
+      settings: { extraRepos: "acme/widgets, ACME/*" },
+    });
+    await plugin(host.bb);
+
+    // The valid sibling still syncs; the glob contributes no repo.
+    await expect(host.harness.callRpc("refresh")).resolves.toMatchObject({
+      repos: 1,
+    });
+    expect(
+      host.harness.logEntries.filter(
+        (entry) => entry.level === "warn" && entry.message.includes("ACME/*"),
+      ),
+    ).not.toHaveLength(0);
+  });
 });
--- a/plugins/github/server.test.ts
+++ b/plugins/github/server.test.ts
@@ -6,6 +6,7 @@ import {
   fetchRepoItems,
   githubRpcContract,
   parsePaginatedGhApi,
+  parseExtraRepos,
   validateGithubCliArgs,
 } from "./server";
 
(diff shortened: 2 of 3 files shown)


SawyerHood:
This has already been fixed by another PR

PR #2550, opened 2026-08-27
Unblock first-plugin authoring: testable scaffold, correct SDK example, visible npm failure

### LINKED ISSUE #2546: bb plugin new scaffolds a plugin that cannot import @get-bb/plugin-sdk/testing: cron-parser is undeclared
## Summary

`bb plugin new` scaffolds a plugin that cannot import `@get-bb/plugin-sdk/testing` — the backend test harness the bundled `bb-plugin-authoring` skill tells plugin authors to test with. The first `createFakePluginHost` test in a fresh scaffold fails to load with `Cannot find package 'cron-parser'`, pointing at a file inside `node_modules` with nothing in the guide to explain it.

## Versions and environment

- bb 0.40.0, desktop app. Also reproduced from source at `ad79bbb5ec909524f8f281e62d860c588a86f332`.
- macOS 15 (Darwin 25.3.0), arm64.
- Node v20.13.1, npm 10.5.2.

## Steps to reproduce

```sh
bb plugin new probe
cd bb-plugin-probe
npm install --include=dev
npm install --save-dev vitest
cat > server.test.ts <<'TS'
import { it } from "vitest";
import { createFakePluginHost } from "@get-bb/plugin-sdk/testing";
import plugin from "./server";

it("loads", async () => {
  const { bb } = createFakePluginHost({ pluginId: "probe" });
  await plugin(bb);
});
TS
npx vitest run
```

## Expected vs actual

Actual:

```
FAIL  server.test.ts [ server.test.ts ]
Error: Cannot find package 'cron-parser' imported from
  .../bb-plugin-probe/node_modules/@get-bb/plugin-sdk/dist/testing/index.js
```

Expected: the test runs. The scaffold already ships the harness's other peers, so the intent is clearly that it works.

## Evidence

The backend harness imports exactly four npm packages — `better-sqlite3`, `cron-parser`, `hono`, `zod`:

- https://github.com/get-bb/bb/blob/ad79bbb5ec909524f8f281e62d860c588a86f332/packages/plugin-sdk/src/testing/fake-plugin-host.ts#L5

All four are **optional** peers of the SDK, so npm installs none of them:

- https://github.com/get-bb/bb/blob/ad79bbb5ec909524f8f281e62d860c588a86f332/packages/plugin-sdk/package.json#L175

The scaffold declares three of the four and misses `cron-parser`:

- https://github.com/get-bb/bb/blob/ad79bbb5ec909524f8f281e62d860c588a86f332/packages/templates/src/plugin-scaffold.ts#L1912-L1913

`zod` is covered because generated `server.ts` imports it, so it sits in `dependencies`.

## What you ruled out

- Not a stale lockfile: `npm install --include=dev` in a brand-new scaffold, twice, on two machines' worth of clean `node_modules`.
- Not the frontend harness: `@get-bb/plugin-sdk/testing/app` needs React, Testing Library and jsdom, and the skill correctly tells the author to add those. Only the backend entry is affected.
- Not a duplicate — searched open and closed issues for `cron-parser`, `plugin scaffold`, `plugin new test`.

Hits every plugin author who follows the guide's testing section, which is the guide's own recommended workflow. Workaround is one `npm i -D cron-parser` once you have decoded the error. Effort: one line plus a regression test.

> AGENT GENERATED

### LINKED ISSUE #2547: bb-plugin-authoring skill documents bb.sdk.threads.list as returning { threads }; it returns an array
## Summary

The bundled `bb-plugin-authoring` skill documents `bb.sdk.threads.list` as returning `{ threads }`. It returns a plain array. An author who copies the example gets `undefined` with no error, and only finds out by reading the 700KB of bundled declarations.

## Versions and environment

- bb 0.40.0, desktop app. Reproduced from source at `ad79bbb5ec909524f8f281e62d860c588a86f332`.
- macOS 15 (Darwin 25.3.0), arm64. Node v20.13.1.

## Steps to reproduce

In any plugin's `server.ts`, copy the example the skill gives:

```ts
const { threads } = await bb.sdk.threads.list({ projectId, limit: 50 });
console.log(threads); // undefined
```

## Expected vs actual

`threads.list` resolves to `ThreadListResult = ThreadListResponse`, a `z.ZodArray`. Destructuring `{ threads }` from an array yields `undefined`, silently — the call succeeded, so nothing throws and nothing logs.

Typechecking the documented line against the SDK's own declarations gives:

```
probe.ts(24,9): error TS2339: Property 'threads' does not exist on type
  '{ id: string; projectId: string; environmentId: string | null; providerId: string;
     title: string | null; ... 19 more ...; environmentWorkspaceDisplayKind: ... }'.
```

Expected: `const threads = await bb.sdk.threads.list(...)`.

## Evidence

- The documented line: https://github.com/get-bb/bb/blob/ad79bbb5ec909524f8f281e62d860c588a86f332/apps/server/src/services/skills/builtin-skills/bb-plugin-authoring/SKILL.md#L689
- The declaration: `ThreadsArea.list(args?: ThreadListArgs): Promise<ThreadListResult>`, `type ThreadListResult = ThreadListResponse`, `declare const threadListResponseSchema: z.ZodArray<...>` (bundled `bb-plugin-sdk.d.ts`).

Found while writing a plugin against the skill: the array form is what actually works at runtime, confirmed by a loaded plugin whose `readBoard` calls `.filter()` on the result and returns real threads.

## What you ruled out

- Not an older SDK: the array shape is what `packages/plugin-sdk` declares at `ad79bbb`.
- `plugin-authoring-docs.test.ts` cannot catch this — it proves the skill *mentions* every API member, not that its examples compile. This is the only `bb.sdk` destructuring in the file, so it is a single wrong line, but nothing structural was stopping the next one.

Silent-wrong-answer bug in the document agents read before writing any plugin. One-line fix; the durable part is a test that compiles the examples.

> AGENT GENERATED

### LINKED ISSUE #2548: bb plugin new discards npm's failure output, so a failed scaffold install gives no reason
## Summary

When `bb plugin new`'s scaffold install fails, the CLI prints only `Could not run npm install — run it in the plugin directory before \`bb plugin build\``. npm's own output — which is the entire diagnosis — is discarded, so the author has to rerun the same failing command by hand to read what the CLI already had.

## Versions and environment

- bb 0.40.0, desktop app. Source at `ad79bbb5ec909524f8f281e62d860c588a86f332`.
- macOS 15 (Darwin 25.3.0), arm64. Node v20.13.1, npm 10.5.2.

## Steps to reproduce

Any npm failure reproduces it. The one I hit was an unwritable npm cache (agent sandbox):

1. `bb plugin new probe`
2. npm exits non-zero with `npm error code EPERM / npm error syscall open / npm error Your cache folder contains root-owned files`.

## Expected vs actual

Actual:

```
Could not run npm install — run it in the plugin directory before `bb plugin build`.
```

Expected: the same line, followed by npm's own explanation, so the next step is obvious without a second run.

## Evidence

The catch discards `cause` entirely:

- https://github.com/get-bb/bb/blob/ad79bbb5ec909524f8f281e62d860c588a86f332/apps/cli/src/commands/plugin.ts#L499-L504

`execFile` rejects with `stdout`/`stderr` attached, so the information is present and dropped.

## What you ruled out

- Not the separate registry-probe warning: `could not reach the npm registry …` is a different code path and did fire; it says nothing about why the install itself failed.
- Not covered by the existing tests: `plugin-new.test.ts` only exercises npm missing from PATH, which produces no output to pass through.

Small, but it is the first command a plugin author runs, and the failure is common in sandboxed or proxied environments. A few lines plus a fake-npm mode in the existing test seam.

> AGENT GENERATED

t1mdurden:
## What was wrong

Three independent defects, all of which land on the same person: someone writing their first bb plugin by following the bundled `bb-plugin-authoring` skill.

1. **The scaffold cannot import the test harness the guide recommends.** `@get-bb/plugin-sdk/testing` imports `better-sqlite3`, `cron-parser`, `hono` and `zod`. All four are *optional* peers of the SDK, so npm installs none of them and the scaffold has to declare them itself — it declared three. A fresh scaffold's first `createFakePluginHost` test dies on `Cannot find package 'cron-parser'`, pointing inside `node_modules` with nothing in the guide to explain it. (#2546)
2. **The guide's `bb.sdk.threads.list` example is wrong.** It shows `const { threads } = await bb.sdk.threads.list(...)`; the method resolves to `ThreadListResponse`, a plain array. The call succeeds, nothing throws, and the reader gets `undefined`. (#2547)
3. **`bb plugin new` throws away npm's failure output.** The catch printed only `Could not run npm install …`, so the author had to rerun the same command by hand to read the reason the CLI already had — an unwritable cache, a refused proxy, a missing platform binary. (#2548)

## What changed

**`packages/templates/src/plugin-scaffold.ts`** — declares `cron-parser` alongside `better-sqlite3` and `hono` in the scaffold's `devDependencies`, and the comment above the block now says what that group is for.

**`apps/cli/src/__tests__/plugin-scaffold-dependencies.test.ts`** — new test: every optional peer the backend harness imports must be declared in the scaffold manifest. It reads the harness's own imports and the SDK's `peerDependencies` rather than restating a list, so a new import in the harness cannot leave it stale. `testing/app` is deliberately out of scope — React, Testing Library and jsdom stay the author's to add, as the skill says.

**`.../builtin-skills/bb-plugin-authoring/SKILL.md`** — one line: `const threads = await bb.sdk.threads.list(...)`.

**`apps/server/test/services/plugins/plugin-authoring-doc-examples.test.ts`** (new) — the durable half. `plugin-authoring-docs.test.ts` already proves the skill *mentions* every API member; nothing proved its examples compile. This generates a probe from `SKILL.md` and runs `tsc` over it: indexing the SDK type proves each documented `bb.sdk.<path>(` exists and is callable, and each destructured await is checked against what that method really returns. Arguments are not reconstructed — only what the example claims to call and receive is under test, which is exactly the class of bug that slipped through. The probe is generated at run time, so a new example is covered the moment it is written.

**`apps/cli/src/commands/plugin.ts`** — quotes the tail of npm's own output under the existing warning. `npm` missing from PATH still prints nothing extra, since that failure produces no output.

**`apps/cli/src/__tests__/helpers/fake-npm.ts`** — new `BB_TEST_NPM_INSTALL=fail` mode: exits non-zero with npm's own explanation on stderr, standing in for the installs that die before touching the tree.

No wire, protocol, CLI-surface or guide changes beyond the corrected line.

## How you verified

Every test fails before its fix and passes after. Base commit `ad79bbb`.

`cron-parser`, with the scaffold line reverted:

```
× declares every optional peer the backend test harness imports
  AssertionError: expected [ 'cron-parser' ] to deeply equal []
```

`threads.list`, with the doc line reverted:

```
× calls and destructures bb.sdk the way the SDK actually declares it
  probe.ts(24,9): error TS2339: Property 'threads' does not exist on type
    '{ id: string; projectId: string; environmentId: string | null; … }'.
```

npm's reason, with the detail suppressed:

```
× passes npm's own reason through when the install fails
  AssertionError: expected 'Could not run npm install — run it in…'
    to contain 'npm error code EPERM'
```

Suites, after:

- `apps/cli` — `Test Files 50 passed (50) / Tests 498 passed (498)`
- `apps/server` `test/services/plugins/plugin-authoring-docs.test.ts` + the new file — `Test Files 2 passed (2) / Tests 16 passed (16)`
- `tsc --noEmit` clean in `apps/cli`, `packages/templates`, `apps/server`
- `oxfmt` run over every touched file

One caveat, stated plainly: `packages/templates`' own suite fails 11 tests with `ERR_REQUIRE_ESM` on my machine, and it fails identically on a clean `git stash` of `main` at `ad79bbb` — same 3 files, same 11 tests. Pre-existing in my environment (Node v20.13.1, macOS arm64), not caused by this change, and I did not chase it.

Fixes #2546
Fixes #2547
Fixes #2548

### DIFF
--- a/apps/cli/src/__tests__/helpers/fake-npm.ts
+++ b/apps/cli/src/__tests__/helpers/fake-npm.ts
@@ -53,6 +53,14 @@ if (process.env.BB_TEST_NPM_INSTALL === "fail") {
   process.exit(1);
 }
 
+if (process.env.BB_TEST_NPM_INSTALL === "fail-noisy-stdout") {
+  process.stderr.write("npm error code EPERM\\nnpm error syscall open\\n");
+  for (let line = 1; line <= 9; line += 1) {
+    process.stdout.write("progress line " + line + "\\n");
+  }
+  process.exit(1);
+}
+
 // npm treats NODE_ENV=production as omit=dev; a command-line --include=dev
 // outranks it. BB_TEST_NPM_ALWAYS_OMIT_DEV forces the omission to stand in for
 // an install that silently drops packages.
--- a/apps/cli/src/__tests__/helpers/fake-npm.ts
+++ b/apps/cli/src/__tests__/helpers/fake-npm.ts
@@ -45,6 +45,14 @@ if (args[0] === "view") {
   process.exit(0);
 }
 
+// BB_TEST_NPM_INSTALL=fail stands in for the installs that die before they
+// touch the tree — an unwritable cache, a refused proxy, a missing platform
+// binary. npm always explains itself on stderr; the CLI has to pass that on.
+if (process.env.BB_TEST_NPM_INSTALL === "fail") {
+  process.stderr.write("npm error code EPERM\\nnpm error syscall open\\nnpm error Your cache folder contains root-owned files\\n");
+  process.exit(1);
+}
+
 // npm treats NODE_ENV=production as omit=dev; a command-line --include=dev
 // outranks it. BB_TEST_NPM_ALWAYS_OMIT_DEV forces the omission to stand in for
 // an install that silently drops packages.
--- a/apps/cli/src/__tests__/plugin-new.test.ts
+++ b/apps/cli/src/__tests__/plugin-new.test.ts
@@ -201,6 +201,19 @@ describe.sequential("bb plugin new dependency install", () => {
     expect(warned.join("\n")).not.toContain("was not found on npm");
   });
 
+  it("passes npm's own reason through when the install fails", async () => {
+    vi.stubEnv("BB_TEST_NPM_INSTALL", "fail");
+
+    await runPluginNew(["npm-broken"]);
+
+    const warnings = warned.join("\n");
+    expect(warnings).toContain("Could not run npm install");
+    // The reason, not just the symptom: without it the author reruns npm by
+    // hand to discover what the CLI already knew.
+    expect(warnings).toContain("npm error code EPERM");
+    expect(warnings).toContain("Your cache folder contains root-owned files");
+  });
+
   it("falls back to the manual step when npm is not on PATH", async () => {
     vi.stubEnv("PATH", join(workDir, "empty-bin"));
 
--- a/apps/cli/src/__tests__/plugin-new.test.ts
+++ b/apps/cli/src/__tests__/plugin-new.test.ts
@@ -214,6 +214,16 @@ describe.sequential("bb plugin new dependency install", () => {
     expect(warnings).toContain("Your cache folder contains root-owned files");
   });
 
+  it("keeps stderr details when a failed install also writes stdout", async () => {
+    vi.stubEnv("BB_TEST_NPM_INSTALL", "fail-noisy-stdout");
+
+    await runPluginNew(["npm-noisy"]);
+
+    const warnings = warned.join("\n");
+    expect(warnings).toContain("npm error code EPERM");
+    expect(warnings).not.toContain("progress line");
+  });
+
   it("falls back to the manual step when npm is not on PATH", async () => {
     vi.stubEnv("PATH", join(workDir, "empty-bin"));
 
--- a/apps/cli/src/__tests__/plugin-scaffold-dependencies.test.ts
+++ b/apps/cli/src/__tests__/plugin-scaffold-dependencies.test.ts
@@ -1,6 +1,6 @@
 import { mkdtemp, readFile, readdir, rm } from "node:fs/promises";
 import { tmpdir } from "node:os";
-import { join, relative } from "node:path";
+import { join, relative, resolve } from "node:path";
 import {
   PLUGIN_SERVER_EXTERNALS,
   RUNTIME_SLOT_BY_SPECIFIER,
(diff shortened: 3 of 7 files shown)


SawyerHood:
Updated this branch onto current `main`. The npm failure formatter now uses a bounded stderr tail and uses stdout only when stderr is empty. A regression test proves that noisy stdout cannot hide npm stderr. Local validation passed 499 CLI tests, the server documentation example test, and CLI/server typechecks. GitHub CI passed all 11 checks.

PR #2556, opened 2026-08-27
Fix copied chat selections with invisible trailing newlines

### LINKED ISSUE #2523: Copying a chat text selection pastes invisible trailing newlines
## Summary

Selecting an excerpt from an assistant message, copying it, and pasting it into the composer can insert several blank lines after the visible text. The clipboard should contain only the text the user can see selected.

## Versions and environment

- bb 0.40.0, reproduced from source at `31a190ddb5f0ffb86cb8dcd80482497663de2685` (`origin/main`)
- Arch Linux, kernel 7.1.8-arch1-3, x86_64
- Node.js v26.7.0
- Chromium 152.0.0.0
- Clean development data directory and detached checkout of `origin/main`
- Provider-independent after the assistant message exists; Codex was used only to create the fixture message

## Steps to reproduce

1. Start bb from `main` with `pnpm dev` and open any thread containing an assistant paragraph followed by another timeline row.
2. In the assistant paragraph, double-click and drag from a word near the start through the visually final word of the paragraph.
3. Press Ctrl+C.
4. Click the composer and press Ctrl+V.
5. Observe the blank lines inserted after the selected sentence.

A deterministic DevTools-console helper can create the same cross-row selection before steps 3-4:

```js
(() => {
  const prose = [...document.querySelectorAll("[data-sidebar-swipe-selectable]")]
    .find((node) => node.textContent?.trim());
  const textNode = prose?.querySelector("p")?.firstChild;
  const row = prose?.closest("[data-timeline-row-id]");
  const followingRow = row?.nextElementSibling;
  if (!(textNode instanceof Text) || !(followingRow instanceof Element)) {
    throw new Error("Need an assistant paragraph followed by another timeline row.");
  }
  const range = document.createRange();
  range.setStart(textNode, Math.min(3, textNode.length));
  range.setEnd(followingRow, 0);
  const selection = window.getSelection();
  selection?.removeAllRanges();
  selection?.addRange(range);
  const text = selection?.toString() ?? "";
  console.log({
    text: JSON.stringify(text),
    trailingWhitespace: text.length - text.trimEnd().length,
  });
})();
```

Browser security requires Ctrl+C and Ctrl+V to remain trusted user actions, so a fully self-contained console script cannot perform the final clipboard steps.

## Expected vs actual

```text
Expected:
The composer contains the visibly selected sentence and no trailing blank lines.

Actual:
The native selection and text/plain clipboard payload contain the sentence followed by seven newline characters. Pasting expands the composer with seven blank lines after the sentence. Submitting the message strips those trailing newlines, so the sent message is unaffected.
```

## Evidence

- Reproduced on `main` at `31a190ddb5f0ffb86cb8dcd80482497663de2685`.
- The visible selection contained one sentence, while the native selection and `text/plain` clipboard payload ended with seven newline characters.
- The selection controller trims text for its floating action payload but does not change the browser's live Range: https://github.com/get-bb/bb/blob/31a190ddb5f0ffb86cb8dcd80482497663de2685/apps/app/src/components/thread/timeline/SelectableMessageProse.tsx#L169-L198
- Normal composer paste preserves the plain-text newlines supplied by the clipboard: https://github.com/get-bb/bb/blob/31a190ddb5f0ffb86cb8dcd80482497663de2685/apps/app/src/components/promptbox/PromptBoxInternal.tsx#L1849-L1896

Observed selection state during reproduction:

```text
Visible selection: one assistant sentence
Native selection / clipboard suffix: \n\n\n\n\n\n\n
Anchor: assistant paragraph
Focus: a following timeline row
```

## What you ruled out

- Still reproduces on `main` at `31a190ddb5f0ffb86cb8dcd80482497663de2685` with a clean development data directory.
- Does not reproduce when both native Range endpoints remain inside the assistant prose wrapper.
- Does not depend on the provider after the assistant message has rendered.
- Pressing Enter strips the trailing newlines before submission; the problem is limited to the draft's visual state.
- Searches for `copy paste newlines text selection`, `clipboard selection`, and `newlines paste chat` found no matching open or closed issue. #1590 concerns clipboard permission failures on plain-HTTP origins, not extra newlines from selection serialization.

Low priority: this is a visual-quality issue in the draft composer. It does not change the submitted message, lose data, or block the workflow.

## Checks

- [x] Reproduced on `main`.
- [x] Searched open and closed issues for duplicates.
- [x] Included exact environment, steps, observed clipboard contents, and commit permalinks.

> AGENT GENERATED

hemaaanth:
## What was wrong

A browser selection can visually end at the last word of an assistant message while its DOM Range extends through whitespace-only timeline rows. Chromium serializes those hidden block boundaries into `text/plain`, so pasting the excerpt into the composer shows several blank lines even though the submitted message is unaffected. See #2523.

## What changed

The shared selection controller now clips whitespace-only boundary spill for the duration of native copy serialization, preserving both plain text and rich HTML, then restores the user's original selection if it has not changed. It leaves intentional cross-message text, selected media, newer selections, and multi-range selections untouched. Focused regression coverage pins those boundaries and the selection lifecycle. There are no wire, daemon protocol, CLI, or documentation changes.

## How you verified

- The regression test failed before the fix because the copied Range still ended in the following timeline row.
- `pnpm exec turbo run test --filter=@bb/app --force` — 440 test files passed; 3,442 tests passed and 4 skipped.
- `pnpm exec turbo run typecheck --filter=@bb/app` — 3/3 tasks passed.
- Manual Chromium check: the same cross-row selection carried seven trailing newlines before the fix and zero after it; the composer stays compact after paste. Before/after recordings are attached in PR comments.

Fixes #2523

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/thread/timeline/SelectableMessageProse.events.test.tsx
+++ b/apps/app/src/components/thread/timeline/SelectableMessageProse.events.test.tsx
@@ -64,6 +64,7 @@ const SHARED_DOCUMENT_EVENT_TYPES = [
   "mouseup",
   "selectionchange",
   "keyup",
+  "copy",
 ];
 
 function countSharedListenerCalls(spy: {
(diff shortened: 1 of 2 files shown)


hemaaanth:
**Before:** copying the visible sentence carries seven trailing newlines, so pasting expands the composer with blank space.

https://github.com/user-attachments/assets/1a0c0805-a251-49f6-83d9-bec89a17cbb5

hemaaanth:
**After:** the same selection copies only the visible sentence, and the composer stays compact after paste.


https://github.com/user-attachments/assets/d3a0cc20-7c9d-4c0f-ad66-a4a4b8d892f5

PR #2566, opened 2026-08-27
Cmd+click links open in the default browser

himynameisjonas:
Web links in threads always go to the in-app browser, and the only way out is the global toggle in Settings → General. Cmd+click is the muscle memory for "open this somewhere else", so it now skips in-app routing and lets the anchor's target=_blank default take over, which the desktop shell already hands to shell.openExternal. Plain click is unchanged and still follows the setting.

Terminal links get the same treatment: with Cmd/Ctrl held they fall through to the external open that was already there.

Only meta/ctrl for now. Alt and shift clicks are untouched.

_Edit: forgot to open a issue first (as requested in CONTRIBUTING.md) so do what you want with this._


SawyerHood:

PR #2578, opened 2026-08-27
Fix Codex rollout writer handoff races

### LINKED ISSUE #2327: turn/completed can precede command completion and wedge later turns between active-writer and no-session errors
## Summary

A Codex-backed thread emitted `turn/completed` while a command-execution item from that same turn was still running. The command later emitted output and `item/completed` after the terminal turn event. Every subsequent new-turn request then failed in one of two contradictory states:

- `thread <provider-thread> already has an active writer`
- `No active codex session for thread <bb-thread>`

Stopping the broken runtime and retrying the exact input did not recover the thread. Provider retry refused because the input had not been accepted, and edit-message could not apply because the rejected request was agent-origin rather than an editable user turn.

## Native ordering evidence

For one exact turn/provider-thread:

1. `item/started` for a long-running `commandExecution`.
2. `turn/completed` with `status=completed`.
3. `item/commandExecution/outputDelta` for the still-running command.
4. `item/completed` for that command, reporting success and a multi-minute duration.
5. A later `client/turn/requested` was rejected as `already has an active writer`.
6. A recovery new-turn request was rejected as `No active codex session`.
7. A further exact redelivery again rejected as `already has an active writer`.

This is a native event/state contradiction: the turn was publicly terminal before all of its writer-owned items terminalized, leaving BB and the provider bridge disagreeing about whether a session/writer existed.

## Expected behavior

- Do not emit `turn/completed` until every command/tool/item belonging to that turn is terminal and no turn writer remains.
- If a provider sends premature completion, BB should delay/fail the terminal event or reconcile outstanding items deterministically.
- A terminal thread must not simultaneously report both active-writer and no-active-session states.
- Stop/restart should clear any orphaned writer ownership without deleting event history or replaying rejected input.
- Rejected agent-origin requests need a supported exact-redelivery recovery path; edit-message is not applicable.

## Acceptance

1. A fixture delays command completion beyond provider turn completion; BB does not publish terminal completion first.
2. After true terminalization, one new-turn request is accepted exactly once.
3. Stop/restart clears stale writer ownership while preserving interruption/history.
4. Tests cover the contradictory `already has an active writer` / `No active codex session` sequence and prove it cannot persist.

No repository or canonical-state mutation is needed to reproduce; this report is based on bounded native event reads.

ymichael:
## What was wrong

Codex 0.146+ holds a rollout writer lock for the lifetime of each app-server process, but the Codex bridge acknowledged release and spawned replacement children immediately after sending SIGTERM. A settings rebuild, stop/resume, bridge restart, or environment move could therefore resume while bb's previous child still owned the rollout.

The first implementation awaited child exit, but review found two more bridge-local windows. Reconstruction initially removed the old map entry before awaiting exit, allowing a concurrent stop or release-after-maintenance request to miss the in-flight replacement. Publishing the replacement closed that gap, but the pending old-child exit still lived only on the constructor stack: concurrent release found the connection-less replacement and acknowledged before the old child released its writer lock.

## What changed

- Make Codex child termination awaitable through finalized process exit, including the existing SIGKILL and close-grace bounds.
- Await release before rebuilding, answering `thread/stop`, and completing maintenance operations that release a session.
- Publish a connection-less replacement before waiting for the old child, and park the old child's exit promise on it. Concurrent stop/discard now marks the replacement closing and waits for both the parked old exit and any current child exit before acknowledging; reconstruction re-checks closing before spawning.
- Retry transient active-writer construction refusals on a bounded 100/400/1000 ms ladder and log every retry with its method, delay, and attempt number.
- Improve persistent-writer guidance to cover both another visible Codex session and a previous/leftover `codex app-server` process.
- Add deterministic regressions for settings rebuild, response-time stop serialization, concurrent stop and discard during reconstruction, transient ownership, and persistent ownership recovery. Shared process-test cleanup is missing-log-safe and attempts every cleanup stage even if an earlier one fails.
- Keep recovery provider-local. Independent real-Codex reproduction showed that the same losing app-server can retry resume after the owner exits, so no new wire recovery kind is needed. Native late-item ordering is orthogonal to the process-lifetime writer lock.
- No host-daemon wire contract changed, so this PR does not increment `HOST_DAEMON_PROTOCOL_VERSION` beyond the version already on `main`.

This PR fixes bb-owned writer handoff races. A persistent writer owned by an independent Codex process is improved, not eliminated: after bounded retries it still returns an untyped bridge error, now with retry diagnostics and actionable guidance.

## How you verified

The regressions were proven red against the exact behaviors they guard:

- Mutating the stop path back to fire-and-forget makes the response-time lock assertion fail with `expected true to be false`.
- Restoring the old map-before-await gap makes both concurrent release tests fail because reconstruction succeeds after release was already acknowledged.
- Before parking the old-child exit on the published replacement, the concurrent stop and concurrent discard tests both acknowledged while `writer.lock` still existed; both immediate lock-absence assertions failed deterministically with `expected true to be false` on two runs.

Green checks on the final tree:

- `pnpm exec turbo run test --filter=bb-plugin-provider-codex --force -- src/bridge/bridge.writer-lock.test.ts src/bridge/bridge.archived-rebuild.test.ts` (2 files, 8/8 passed)
- `pnpm exec turbo run test --filter=bb-plugin-provider-codex --force` (25 files, 255/255 passed)
- `pnpm exec turbo run test --filter=@bb/agent-runtime --force` (22 files, 318/318 passed)
- `pnpm exec turbo run typecheck --filter=bb-plugin-provider-codex --force`
- `pnpm exec turbo run build --filter=@bb/host-daemon --force`
- `pnpm exec oxlint` and `pnpm exec oxfmt --check` on every changed TypeScript file
- Real Codex 0.148.0 scratch smoke: app-server B received the native active-writer refusal while A held the rollout, then the same B process resumed successfully after A exited.

Fixes #2327

> AGENT GENERATED

### DIFF
--- a/plugins/provider-codex/src/bridge/app-server-connection.ts
+++ b/plugins/provider-codex/src/bridge/app-server-connection.ts
@@ -44,7 +44,6 @@ interface CodexAppServerRequestArgs<TResult> {
 export interface CodexAppServerConnection {
   request<TResult>(args: CodexAppServerRequestArgs<TResult>): Promise<TResult>;
   notify(method: string, params?: unknown): void;
-  /** Signal the child and resolve once its exit has been finalized. */
   kill(): Promise<void>;
   readonly exited: boolean;
 }
--- a/plugins/provider-codex/src/bridge/app-server-connection.ts
+++ b/plugins/provider-codex/src/bridge/app-server-connection.ts
@@ -44,7 +44,8 @@ interface CodexAppServerRequestArgs<TResult> {
 export interface CodexAppServerConnection {
   request<TResult>(args: CodexAppServerRequestArgs<TResult>): Promise<TResult>;
   notify(method: string, params?: unknown): void;
-  kill(): void;
+  /** Signal the child and resolve once its exit has been finalized. */
+  kill(): Promise<void>;
   readonly exited: boolean;
 }
 
@@ -112,6 +113,10 @@ export function createCodexAppServerConnection(
   } | null = null;
   let closeGraceTimer: NodeJS.Timeout | null = null;
   let stdoutLines: Interface | null = null;
+  let resolveExit!: () => void;
+  const exitPromise = new Promise<void>((resolve) => {
+    resolveExit = resolve;
+  });
 
   function writeLine(message: object): void {
     const stdin = child.stdin;
@@ -155,7 +160,11 @@ export function createCodexAppServerConnection(
         { spawnFailed },
       ),
     );
-    options.onExit({ ...status, stderrTail, spawnFailed });
+    try {
+      options.onExit({ ...status, stderrTail, spawnFailed });
+    } finally {
+      resolveExit();
+    }
   }
 
   if (child.stdout) {
@@ -313,7 +322,7 @@ export function createCodexAppServerConnection(
 
     kill() {
       if (finalized) {
-        return;
+        return exitPromise;
       }
       const escalation = setTimeout(() => {
         if (!finalized) {
@@ -322,6 +331,7 @@ export function createCodexAppServerConnection(
       }, KILL_ESCALATION_MS);
       escalation.unref?.();
       child.kill("SIGTERM");
+      return exitPromise;
     },
   };
 }
--- a/plugins/provider-codex/src/bridge/bridge-process.test-support.ts
+++ b/plugins/provider-codex/src/bridge/bridge-process.test-support.ts
@@ -1,0 +1,130 @@
+import { existsSync, readFileSync, rmSync } from "node:fs";
+import type { experimental_createBridgeJsonRpcTestHarness as createBridgeJsonRpcTestHarness } from "@get-bb/plugin-sdk/provider-bridge/testing";
+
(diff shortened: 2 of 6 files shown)


Willhong:
Rebased these three commits onto current `main` (author preserved) in #3460 and added the missing piece on the daemon side.

While reproducing #2327 on a live machine, the active-writer refusal turned out to start in the host daemon, not in the bridge: `emitTranslatedEvents` bound the delta-assembled `thread/identity` (empty thread id) to a stale pending-identity slot, so after a message edit the next `thread/start` on the same bridge recorded its Codex thread id onto the rewound thread, and both bb threads then resumed one rollout. #3460 fixes that binding, refuses to resume a provider thread another hosted thread owns, and carries this PR's retry ladder, guidance, `writer.lock` fixture, and writer-lock tests. The handoff serialization in `constructThreadSession` / `releaseSession` already landed via #3025, so `main`'s implementation is kept there.

Leaving this PR open for you or a maintainer to close once #3460 is reviewed.

> AGENT GENERATED

SawyerHood:

PR #2581, opened 2026-08-27
Reject invalid execution selections before provisioning

### LINKED ISSUE #2400: thread spawn accepts a nonexistent model: rc=0, provisions a worktree, fails as an agent message
### Summary

`bb thread spawn --model <id>` **accepts a model id that does not exist**: it returns `rc=0`, provisions a full git worktree, starts the turn, and surfaces the failure as an ordinary **agent message** rather than an error. The same gap accepts a `reasoningLevel` the chosen model does not advertise.

An execution setting appears to be validated nowhere between the CLI and the provider — even though bb **already has** the rule elsewhere.

### 1. A model id that does not exist

Two threads were spawned with `--model claude-does-not-exist-9`. In both cases bb:

- returned **`rc=0`** from `bb thread spawn`
- ran the full provisioning sequence — `Preparing worktree`, `Fetching origin/main`, `status: completed`
- started the turn
- delivered the failure as an `item/completed` **agentMessage**:

> `There's an issue with the selected model (claude-does-not-exist-9). It may not exist or you may not have access to it.`

**The cost of a typo is a worktree provision plus an agent turn, and every status along the way says success.**

### 2. A reasoning level the model does not advertise

A dispatch preset was created as `claude-haiku-4-5-20251001` with `reasoningLevel: medium`. Haiku's `supportedReasoningEfforts` has **exactly one entry, `low`**. bb accepted the preset and never complained **for sixteen days**, until it was noticed by hand.

### Why this looks like an oversight rather than a decision

**bb already enforces this rule on a different path.** `bb workflows validate` rejects a bad literal model with `Model mismatch for literal workflow selection`. So the check exists — it just isn't applied on the spawn path or on preset creation.

### Reproduction

```
bb thread spawn --project <p> --model claude-does-not-exist-9 --prompt 'hello'
echo "rc=$?"          # 0
```

Then, to confirm it was accepted rather than rejected:

```
sqlite3 ~/.bb/bb.db "select json_extract(data,'$.execution.model'), count(*)
  from events where type='client/turn/requested' group by 1 order by 2 desc;"
```

The bogus id appears alongside the real ones. Cross-check the real catalogue with `bb provider models claude-code --json`.

### Suggested fix

Validate `--model` and `--reasoning-level` against the resolved provider's catalogue at **`thread spawn`** and at **`preset create|update`**, failing non-zero *before* provisioning. `supportedReasoningEfforts` is already published per model, so the reasoning-level half needs no new data.

### Why it matters more now than it used to

While every spawn uses one project default this is survivable — on our machine 10,126 of 11,011 turns in 14 days ran the same model. It stops being survivable the moment anything routes deliberately: that is exactly when a typo appears, and it currently costs a worktree and a turn while reporting success.

### Not verified

Whether other paths (`bb automation`, workflow `agent()` calls) validate the model. Only `thread spawn` and `preset create|update` were measured.

ymichael:
## What was wrong

Thread creation trusted explicit, inherited, or replayed execution selections instead of resolving the actual final model/reasoning tuple against the provider catalog for the target machine/environment. Invalid selections could therefore provision a worktree and create a thread/turn before the provider rejected them. Forks, queued/deferred messages, Tasks presets, Automations, and Workflows each applied a different subset of the policy. See [#2400](https://github.com/get-bb/bb/issues/2400) and the [implementation plan](https://github.com/get-bb/bb/blob/bb/fix-github-issue-2400-thr_sc9v5bkc5j/plans/issue-2400-execution-selection-validation.md).

## What changed

- Resolve and validate the final execution tuple at a shared, server-owned boundary before synchronous thread/environment provisioning or provider submission, using the catalog for the selected host and preserving permission ceilings, aliases, defaults, selected-only entries, and provider-approved custom models.
- Treat a non-empty `supportedReasoningEfforts` list as authoritative. Preserve compatible inherited reasoning for an explicit model, use that model's default when incompatible, and recover stale implicit project/fork defaults against the current target catalog.
- Give both fork routes the same target-aware behavior, including copied-history forks whose last execution would otherwise outrank project defaults.
- Keep offline queue/deferred admission durable without a live catalog probe. Revalidate at delivery, retain catalog-mismatched input for recovery, and continue draining later valid rows instead of head-of-line blocking.
- Preserve arrival order for transient deferred-delivery failures and log retained catalog mismatches in both queue paths.
- Use registered fallback catalog rows after a transient probe only when they contain the selected model (or no model is pinned); otherwise return a retryable catalog error without replacing or remembering a different model.
- Validate Tasks preset and Automation create/update only when the persisted effective execution tuple or target changes; validate project-default Tasks presets on the task project's host at dispatch. Workflows validate literal and resolved-agent selections with the same catalog semantics.
- Preserve typed plugin/SDK catalog errors through the real and fake plugin hosts, refresh the public SDK/API inventory and Plugin Guide surface, and bump the Plugin SDK to `0.4.28`.
- Update CLI help, generated guide sources, bundled skills, plugin docs, and the evidence-adjusted implementation plan.
- Do not bump `HOST_DAEMON_PROTOCOL_VERSION`: the implementation reuses the existing `provider.list_models` command and result without changing any server↔daemon field, payload, command, or meaning.

## How you verified

- Added red-first public/integration regressions for invalid model/reasoning rejection before provisioning/submission; valid, selected-only, alias, custom/unlisted, remote-host, stale-default, compatible reasoning, both fork entry points, online/offline queue drain/recovery/ordering, deferred retention/probe failure, SDK/plugin, Tasks full-form preset edits, Automations, and Workflows.
- Manual isolated CLI QA: invalid model and advertised-unsupported reasoning spawns changed from exit `0` with provisioning to exit `1`; environment/thread/turn counts remain unchanged and provider recordings contain no thread start or turn request.
- `pnpm exec turbo run test --filter=@bb/server -- test/public/public-thread-execution-selection.test.ts test/public/public-thread-fork.test.ts test/threads/deferred-thread-messages.test.ts` — 3 files, 56/56 tests passed.
- `pnpm exec turbo run test --filter=bb-plugin-tasks --filter=bb-plugin-automations` — Tasks 368/368 and Automations 73/73 passed.
- Affected-package `pnpm exec turbo run typecheck ...` — 18/18 Turbo tasks passed after rebase.
- Affected-package `pnpm exec turbo run build ...` — 10/10 Turbo tasks passed after rebase.
- Affected-package forced test matrix — 20/21 Turbo tasks passed; one server documentation-example test hit its 5-second timeout only under maximum cross-package contention, then passed alone in 895 ms.
- `pnpm exec turbo run test --force --filter=@bb/server` — 215 files passed, 1 skipped; 2,084/2,084 tests passed.
- `pnpm exec turbo run test --filter=@bb/templates --force` — 6/6 Turbo tasks; 43/43 tests passed.
- `pnpm exec turbo run test --filter=@get-bb/plugin-sdk --filter=@bb/domain --force` — 6/6 Turbo tasks; SDK 220/220 and domain 172/172 tests passed.
- `node .github/workflows/check-plugin-sdk-version.mjs` — SDK contract/version lockstep passed after rebasing onto the current `main` version.
- GitHub CI on `100286b1d` — 13/13 applicable checks passed; Node compatibility smoke and iOS simulator flows were skipped by workflow conditions.

Fixes #2400

> AGENT GENERATED


SawyerHood:

PR #2599, opened 2026-08-27
Scroll the project picker menu instead of clipping it

### LINKED ISSUE #2551: Project picker dropdown does not scroll: projects past the window bottom are unreachable
### Summary

Opening the "Work in a project" picker in the desktop app renders the full project list downward from the button. When the list is taller than the space left below the button, it is cut off at the bottom edge of the window with no scrollbar and no way to scroll. The projects below the cutoff cannot be read or selected. I expected the menu to be capped to the available height and to scroll internally.

### Versions and environment

- bb desktop app on macOS. Version not recorded; happy to add it.
- Default window size, not maximized.
- Enough projects in the account that the picker list exceeds the window height (17 rows visible in the screenshot, the last one clipped mid-row).

### Steps to reproduce

1. Have more projects than fit between the "Work in a project"/Project selector button at the new thread page 
2. Click "Work in a project"/project picker below the composer.
3. The project list opens and runs past the bottom edge of the window.
4. Scroll over the list with the wheel or trackpad.

The list does not move and the clipped entries stay unreachable. Does not reproduce when the project count is small enough that the list fits above the window bottom.

### Expected vs actual

```text
Actual: the list is clipped at the window edge. The last visible row is cut mid-entry, there is no scrollbar, and scrolling over the menu does nothing. Every project below the cutoff is unreachable.

Expected: the menu is capped to the available height and scrolls internally so every project can be seen and selected.

Screenshot attached: picker open, list running off the bottom of the window.
```

### Evidence

Names Redacted.

Image

### What you ruled out

Not specific to one project list or ordering: it reproduces whenever the number of projects exceeds the visible height.

_No response_

### Checks

- [ ] If an agent wrote this, the body ends with `> AGENT GENERATED` and links the thread or report.

luisllenin02:
Fixes #2551.

## What was wrong

`ProjectSelector` renders every project into the shared desktop dropdown, which is `overflow-hidden` with no height limit. Once the list is taller than the room below the trigger it lays out at its full natural height and runs off the window edge, and the projects past the cutoff cannot be scrolled to or selected.

## The fix

Radix already measures that room and publishes it as `--radix-dropdown-menu-content-available-height` on the content element — nothing consumed it. Cap the menu at it, scroll the overflow, and contain the overscroll so the wheel does not chain to the page once the list bottoms out.

The reproduction report on #2551 proposed applying the classes unconditionally. This gates them behind `!isCompactViewport` instead, matching how `ModelReasoningPicker` applies `max-h-[var(--radix-popover-content-available-height)]` (ModelReasoningPicker.tsx:1019). The compact branch renders a drawer rather than a Radix popper, so the variable is undefined there; the declaration would drop out on its own, but relying on an undefined variable to neutralize a rule seemed more fragile than not emitting it. Happy to simplify to the unconditional form if you would rather keep the component free of the hook.

## Verification

Measured in Chromium against the stylesheet built from this branch — 30 items, 694px of available height:

| | `max-height` | `overflow-y` | scrollable |
| --- | --- | --- | --- |
| before | `none` | `hidden` | no — 1088px of content, clipped |
| after | `694px` | `auto` | yes — 692px viewport, 1088px content |
| compact drawer | `none` | `auto` | unchanged |

`ProjectSelector.test.tsx` is new and covers the desktop cap, that all 30 projects stay reachable in the menu, and that the compact drawer does not inherit the desktop height rule. The first test fails on `main` without the component change.

`turbo run typecheck lint --filter=bb-app` passes, and all 8 test files under `apps/app/src/components/pickers` (72 tests) pass.

### DIFF
--- a/apps/app/src/components/pickers/ProjectSelector.test.tsx
+++ b/apps/app/src/components/pickers/ProjectSelector.test.tsx
@@ -1,0 +1,86 @@
+// @vitest-environment jsdom
+
+import { cleanup, fireEvent, render, screen } from "@testing-library/react";
+import { CompactViewportOverrideProvider } from "@bb/shared-ui/hooks/use-compact-viewport";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { ProjectSelector } from "./ProjectSelector";
+
+const CAP_VARIABLE = "--radix-dropdown-menu-content-available-height";
+
+// Enough projects that the menu outgrows any window it opens in — the
+// condition #2551 needs to reproduce.
+const projects = Array.from({ length: 30 }, (_, index) => ({
+  id: `project_${index}`,
+  name: `Project ${index}`,
+}));
+
+afterEach(() => {
+  cleanup();
+  vi.clearAllMocks();
+});
+
+function openProjectMenu(options?: { compact?: boolean }) {
+  const compact = options?.compact === true;
+  const selector = (
+    <ProjectSelector
+      projects={projects}
+      value={projects[0]!.id}
+      onChange={vi.fn()}
+      modal={false}
+    />
+  );
+  render(
+    compact ? (
+      <CompactViewportOverrideProvider isCompactViewport>
+        {selector}
+      </CompactViewportOverrideProvider>
+    ) : (
+      selector
+    ),
+  );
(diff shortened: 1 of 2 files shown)


SawyerHood:
Independently fixed, closing!

PR #2614, opened 2026-08-28
Add managed worktree teardown hook

### LINKED ISSUE #2477: Feature: `.bb-env-teardown.sh`, a teardown counterpart to the worktree setup hook
### Summary

bb runs a setup hook when it creates a managed worktree, but nothing when it destroys one. I'd like a symmetric `.bb-env-teardown.sh`, so a repo can release what its setup hook provisioned.

### The asymmetry

Reading the shipped source maps on 0.40.0:

**Create** — `packages/host-workspace/src/provisioning.ts` → `createWorktree` does, in order:

1. `git worktree add -B <branch> <target> <base>`
2. `copyIncludedFiles` (`.worktreeinclude`)
3. `runSetupScript` (`.bb-env-setup.sh`)

**Destroy** — same file, `removeWorktree`: `git worktree remove --force`, then `fs.rm -rf`. No hook, no script, no notification.

`.bb-env-setup.sh` and `.worktreeinclude` are the only two script filenames in the codebase, so there is currently no extension point on the teardown side.

### Why it matters

A setup hook that only installs dependencies is symmetric by accident — `rm -rf` undoes it. But setup hooks routinely provision things that live *outside* the worktree:

- a per-worktree database or schema
- a registered local dev site / vhost / proxy entry
- a file watcher or daemon rooted at the worktree path
- a container, tunnel, or port reservation

Every one of those survives `removeWorktree` and accumulates silently, once per archived thread. There's no way for a repo to clean up after itself today.

This overlaps with #1660 (orphaned dev processes in deleted managed worktrees compounding a host leak). A teardown hook wouldn't fix that on bb's side, but it would let affected repos stop their own processes before the worktree goes away, which is the part bb can't do generically.

### Proposed contract

Mirror the setup hook as closely as possible:

- `.bb-env-teardown.sh` at the worktree root, resolved the same way as `.bb-env-setup.sh`; absent means no-op.
- Invoked as `env bash .bb-env-teardown.sh`, cwd set to the worktree, same sanitized environment.
- Runs inside `removeWorktree`, **before** `git worktree remove`, so the tree is still intact and the script can read it.
- Only for managed worktrees being destroyed.
- Its own timeout, and progress in the transcript the same way setup reports.

### One thing I'd ask for explicitly: make it non-blocking

The setup hook is correctly fail-closed — a non-zero exit removes the worktree and fails provisioning. **Teardown should be the opposite: best-effort and unable to block removal.**

A failing or hanging teardown script must not be able to leave a worktree undeletable, or the feature turns a cleanup convenience into a way to wedge cleanup entirely. Concretely: kill it at the timeout, report the failure in the transcript, and remove the worktree regardless.

### Alternatives considered

- **A plugin.** Thread lifecycle events are observe-only and fire-and-forget, so a plugin can't reliably run teardown before bb reclaims; making it reliable means taking over worktree creation entirely, which is a large amount of machinery for one missing hook.
- **External reconciliation.** Works if every provisioned resource is named deterministically from the branch, so an out-of-band sweep can diff live resources against `git worktree list`. That's a real workaround, but it pushes the burden onto every repo and can't handle anything not name-addressable.

Happy to send a PR if this shape sounds right.

I searched open and closed issues for the same request.
AGENT GENERATED

### What happens today

On create, [createWorktree](https://github.com/get-bb/bb/blob/f3cab2dd8c5c4be6d450be318550f3a04c8c3a1f/packages/host-workspace/src/provisioning.ts#L330) runs git worktree add, then [copyIncludedFiles](https://github.com/get-bb/bb/blob/f3cab2dd8c5c4be6d450be318550f3a04c8c3a1f/packages/host-workspace/src/provisioning.ts#L426) for .worktreeinclude, then [runSetupScript](https://github.com/get-bb/bb/blob/f3cab2dd8c5c4be6d450be318550f3a04c8c3a1f/packages/host-workspace/src/provisioning.ts#L433) for .bb-env-setup.sh.

On destroy, [removeWorktree](https://github.com/get-bb/bb/blob/f3cab2dd8c5c4be6d450be318550f3a04c8c3a1f/packages/host-workspace/src/provisioning.ts#L725) runs [git worktree remove --force](https://github.com/get-bb/bb/blob/f3cab2dd8c5c4be6d450be318550f3a04c8c3a1f/packages/host-workspace/src/provisioning.ts#L761) and then [fs.rm(..., { recursive: true, force: true })](https://github.com/get-bb/bb/blob/f3cab2dd8c5c4be6d450be318550f3a04c8c3a1f/packages/host-workspace/src/provisioning.ts#L780). No hook, no notification.

.bb-env-setup.sh and .worktreeinclude are the only two script filenames in the tree, so there is no extension point on the teardown side.

The practical effect: everything a setup hook provisions outside the worktree survives the worktree and accumulates once per archived thread. Nothing in the repo can clean up after itself.

### What you would expect

A .bb-env-teardown.sh that mirrors the setup hook:

- Resolved from the worktree root the same way .bb-env-setup.sh is; absent means no-op.
- Invoked as env bash .bb-env-teardown.sh, cwd set to the worktree, same sanitized environment.
- Runs inside removeWorktree before git worktree remove, while the tree is still intact.
- Only for managed worktrees being destroyed.
- Its own timeout, and transcript progress the way setup reports it.

teardown should be best-effort and must not be able to block removal. Setup is correctly fail-closed — a non-zero exit removes the worktree and fails provisioning. Teardown wants the opposite. If a teardown script can fail or hang its way into an undeletable worktree, the feature converts a cleanup convenience into a way to wedge cleanup entirely. Kill it at the timeout, report the failure, remove the worktree regardless.

### Context and alternatives

Verified on: bb 0.40.0 desktop app; code read at f3cab2dd8c5c4be6d450be318550f3a04c8c3a1f (desktop-v0.40.0). I confirmed the create side end to end — spawned a thread with --new-environment worktree, watched .bb-env-setup.sh run in the new worktree, and confirmed a non-zero exit fails provisioning and removes the worktree. The teardown gap is the absence of any equivalent call in removeWorktree.

Workaround I use now. Name every provisioned resource as a pure function of the branch, then reconcile out of band: list live resources, diff against git worktree list, release the orphans. It works, but it pushes the burden onto every repo and cannot handle anything that is not name-addressable.

Ruled out: a plugin. Thread lifecycle events are observe-only and fire-and-forget, so a plugin cannot reliably run teardown before bb reclaims the environment. Making it reliable means taking over worktree creation so the plugin owns removal too — a large amount of machinery, and it leaves bb's own worktree button still leaking. I scoped that plugin out and abandoned it in favour of filing this.

Related: #1660 (orphaned dev processes in deleted managed worktrees compounding a host leak). A teardown hook would not fix that on bb's side, but it would let affected repos stop their own processes before the worktree disappears, which is the part bb cannot do generically.

### Checks

SawyerHood:
## What was wrong

Managed worktree creation supported `.bb-env-setup.sh`, but destruction had no repository hook. Resources created outside a worktree could remain after bb removed the worktree.

## What changed

- Run `.bb-env-teardown.sh` before Git removes a managed worktree.
- Use the same sanitized environment and closed stdin as the setup hook.
- Apply a separate 15-minute timeout and continue removal after any hook failure.
- Return teardown output and status in the environment destroy transcript.
- Add the timeout and transcript to the daemon wire contract.
- Increment `HOST_DAEMON_PROTOCOL_VERSION` from 172 to 173.
- Document the hook in the guide, skill, configuration, platform, and worktree documentation.

## How you verified

- Added worktree tests for success, missing scripts, non-zero exit, timeout, output, and pre-removal order.
- Added a daemon dispatch test for the timeout and returned transcript.
- Ran `pnpm exec turbo run typecheck` for the six affected packages.
- Ran the full `@bb/host-workspace`, `@bb/host-daemon-contract`, `@bb/host-daemon`, `@bb/server`, and `@bb/templates` test suites.

Fixes #2477

> AGENT GENERATED

### DIFF
--- a/apps/host-daemon/src/command-dispatch.ts
+++ b/apps/host-daemon/src/command-dispatch.ts
@@ -471,6 +471,8 @@ const commandHandlers: CommandHandlerMap = {
     }),
   "environment.provision.cancel": cancelEnvironmentProvision,
   "environment.destroy": async (command, options) => {
+    const transcript: HostDaemonCommandResult<"environment.destroy">["transcript"] =
+      [];
     const resolution = await resolveWorkspaceForCommand({
       dataDir: options.dataDir,
       environmentId: command.environmentId,
@@ -479,7 +481,7 @@ const commandHandlers: CommandHandlerMap = {
     });
     if (!resolution.ok) {
       if (resolution.failure.code === "path_not_found") {
-        return {};
+        return { transcript };
       }
       throw new ExpectedCommandDispatchError(
         resolution.failure.code,
@@ -490,8 +492,11 @@ const commandHandlers: CommandHandlerMap = {
       environmentId: command.environmentId,
       reason: "environment-destroyed",
     });
-    await options.runtimeManager.destroyEnvironment(command.environmentId);
-    return {};
+    await options.runtimeManager.destroyEnvironment(command.environmentId, {
+      timeoutMs: command.teardownTimeoutMs,
+      onProgress: (entry) => transcript.push(entry),
+    });
+    return { transcript };
   },
   "workspace.commit": async (command, options) => {
     const entry = await requireResolvedWorkspaceForCommand({
--- a/apps/host-daemon/src/runtime-manager.test.ts
+++ b/apps/host-daemon/src/runtime-manager.test.ts
@@ -1784,7 +1784,7 @@ describe("RuntimeManager", () => {
       environmentId: "env-1",
       workspacePath: "/tmp/env-1",
     });
-    await manager.destroyEnvironment("env-1");
+    await manager.destroyEnvironment("env-1", { timeoutMs: 900000 });
 
     expect(runtime.shutdown).toHaveBeenCalledTimes(1);
     expect(workspace.destroy).toHaveBeenCalledTimes(1);
@@ -1831,7 +1831,7 @@ describe("RuntimeManager", () => {
       try {
         expect(isAlive(grandchildPid)).toBe(true);
 
-        await manager.destroyEnvironment("env-procs");
+        await manager.destroyEnvironment("env-procs", { timeoutMs: 900000 });
 
         expect(managedWorkspace.destroy).toHaveBeenCalledTimes(1);
         const deadline = Date.now() + 5000;
--- a/apps/host-daemon/src/runtime-manager.ts
+++ b/apps/host-daemon/src/runtime-manager.ts
@@ -31,6 +31,7 @@ import type {
 import {
   provisionWorkspace,
   WorkspaceError,
+  type DestroyWorkspaceArgs,
   type HostWorkspace,
   type ProvisionWorkspaceArgs,
 } from "@bb/host-workspace";
@@ -1089,7 +1090,10 @@ export class RuntimeManager {
     });
   }
 
-  async destroyEnvironment(environmentId: string): Promise<void> {
+  async destroyEnvironment(
+    environmentId: string,
+    args: DestroyWorkspaceArgs,
+  ): Promise<void> {
     const existing = this.entries.get(environmentId);
     const pending = this.pendingEntries.get(environmentId);
     const entry = existing ?? (pending ? await pending : undefined);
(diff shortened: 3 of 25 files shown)


SawyerHood:
@slopcop review

SawyerHood:
@slopcop check

PR #2636, opened 2026-08-28
Align generated thread titles with Codex

smsunarto:
## What was wrong

BB mixed title-writing rules with one provider's transport. The shared prompt required a `result` tool call, but Codex uses strict JSON-schema output instead.

BB also skipped prompts shorter than five words and sent only the first 80 characters to inference. These rules lost useful context and differed from Codex's title behavior.

## What changed

Replaced the metadata prompt with the current Codex thread-title prompt and renamed the template to `generate-thread-title.md`.

Title generation now:

- Runs for every non-empty prompt.
- Sends the complete initial user prompt.
- Enforces Codex's 36-character limit.
- Preserves ticket references.
- Keeps provider transport outside the shared prompt.

Codex receives a strict JSON output schema. Direct `pi-ai` providers receive the same schema through the `result` tool and a separate instruction to call it.

The 80-character fallback remains available while inference runs or when it fails.

This change stays inside the server. It does not change the host-daemon protocol.

## How you verified

I created threads through the live UI while Claude Code was selected as the thread provider. The configured `codex/gpt-5.6-luna` helper generated these titles:

| Input prompt | Generated title |
| --- | --- |
| `fix` | `Fix issue` |
| `BB-2197 fix the stale title shown after creating a thread` | `BB-2197 Fix stale thread title` |
| `Why does choosing Claude Code not change the helper model used for thread titles?` | `Explain Claude Code title model` |
| `QA-8472: explain why this disposable title-generation check uses a separate helper model, then reply only with ok.` | `Explain QA-8472 helper model` |
| `We have a long discussion about provisioning, workspace selection, provider bridges, authentication, and unrelated background details. The actual task is to rename the generated thread metadata template.` | `Rename thread metadata template` |
| `Use ripgrep and TypeScript tests to fix generated titles that ignore ticket references` | `Fix ticket references in titles` |

The results demonstrate the intended behavior:

- `fix` proves that short prompts no longer skip generation.
- Both ticket references remain verbatim.
- The long prompt proves that inference receives text beyond the 80-character fallback.
- The tool-focused prompt produces a title about the requested outcome.
- Every title is one line, has no trailing punctuation, and stays within 36 characters.
- Selecting Claude Code does not redirect title generation away from the configured helper model.

Automated checks also passed:

- `pnpm exec turbo run test --filter=@bb/server --force -- test/threads/generated-branch-names.test.ts test/threads/title-generation.test.ts test/ai/inference.test.ts`
  - 29 tests passed.
- `pnpm exec turbo run typecheck --filter=@bb/server --force`
  - Typecheck passed.

Fixes #

> AGENT GENERATED


SawyerHood:

PR #2638, opened 2026-08-28
Tweak sidebar motion to be more dynamic + consistent between left/right sidebar

smsunarto:
### Before

Hard to see, but notice that the right sidebar toggle icon jumps from it's location in the right sidebar to the header. Also notice that the left/right sidebar motion is noticeably different.

https://github.com/user-attachments/assets/4e060e4f-d245-44a8-9105-61917e367835

### After

A more springy dynamic motion for both sidebar. Sidebar toggle icon doesn't teleport.

https://github.com/user-attachments/assets/5559a674-019f-4886-9d54-949a09c05abf

## Why this change

The left and right sidebars are paired parts of the app shell, but their transitions did not share the same motion. The mismatch made the layout feel disconnected.

Opening a thread also changed the right-side layout while its toggle moved with the panel. The control jumped during the transition, which made the interface feel unstable and moved the click target away from the user.

This PR gives both sidebars one dynamic motion language and keeps the right toggle stationary while the surrounding layout changes.

## What was wrong

The left and right sidebars used separate transition behavior. The right toggle also participated in the moving layout when a thread opened, so its screen position shifted during the transition.

## What changed

- Give both sidebars the same spring profile and transition behavior.
- Keep the right sidebar toggle in a fixed position while a thread opens and the panel moves.
- Retarget interrupted transitions from the current rendered layout so repeated clicks remain smooth.

## Why this approach

One motion owner coordinates the right panel and its layout boundary. The panel can move without carrying the toggle with it, and repeated clicks retarget the current motion instead of starting competing transitions.

The scope stays inside the app sidebar and secondary-panel components. There are no server, protocol, route, SDK, CLI, guide, or data changes.

## Review fixes

SlopCop found four problems. This PR fixes three of them. The `motion` dependency size is a separate decision.

### The closed sidebar stayed in the keyboard order

The old panel used `visibility: hidden` when it collapsed, which takes its descendants out of the focus order. The motion rewrite replaced that with a zero width and an opacity fade. Neither removes an element from the focus order, so Chrome could still move focus into the closed panel. The collapsed panel is now `inert` and `aria-hidden`.

### The panel animation forced a layout on every frame

`renderPanelGroup` wrote both flex values and then read the group geometry. That read after a write forces a synchronous layout each frame. The group width does not change while the panels resize inside it, so the code now measures the group one time for each layout, before the writes.

### The full-screen panel could show a seam

At the 100-percent layout the boundary is disabled and carries an `opacity-0` class, but the inline `opacity: 1` won. The handle's inner 1-pixel span painted a line down the left edge of the full-screen panel. The boundary now stays hidden at the 0 and 100-percent layouts.

### Removed a dead CSS rule

`theme.css` suppressed flex transitions on panels during a drag. This PR removed the last `transition-[flex-grow,flex-basis]` on a `[data-panel]`, so that `:has()` rule now costs style recalculation and changes nothing.

## How you verified the review fixes

- The three new tests fail on the code before the fixes and pass after them.
- The app suite passed 3,502 tests. Lint reported 0 errors.
- Chrome QA on the dev app: Tab reached 12 sidebar controls when the sidebar was open and 0 when it was closed.
- The full-screen panel showed no seam. Inline and computed opacity were both `0`.
- A divider drag still tracked the pointer exactly, so the removed CSS rule was dead.
- Close from a resized width returned to the same width. The browser reported no errors.

## Bug fixed while developing this PR

The first version of this PR could flicker when a resized right sidebar closed. We fixed that path so it closes from its current rendered width. This is a correction to this PR's implementation, not the feature's primary goal.

## How you verified

- `pnpm exec turbo run typecheck --filter=@bb/app` passed.
- All 8 affected test files passed: 125 tests.
- The full app suite passed 3,499 tests and skipped 3. One unrelated test failed because `plugins/t3sidebar/package.json` is missing from this workspace.
- The live app used matching motion for the left and right sidebars.
- The right toggle stayed in place while a thread opened and during repeated clicks.
- The resized close path no longer flickered, and the resize handle remained on the panel edge.
- The browser reported no runtime errors.

> AGENT GENERATED

### DIFF
--- a/apps/app/package.json
+++ b/apps/app/package.json
@@ -93,6 +93,7 @@
     "katex": "^0.16.22",
     "lucide-react": "^1.23.0",
     "mermaid": "^11.15.0",
+    "motion": "^13.1.0",
     "nanoid": "^5.1.6",
     "partysocket": "^1.1.16",
     "react": "^19.0.0",
--- a/apps/app/src/components/layout/AppPageHeader.tsx
+++ b/apps/app/src/components/layout/AppPageHeader.tsx
@@ -1,5 +1,15 @@
-import { useState, type ReactNode, type Ref } from "react";
-import { useIsSidebarShowing } from "@/components/ui/sidebar.js";
+import {
+  useLayoutEffect,
+  useRef,
+  useState,
+  type ReactNode,
+  type Ref,
+} from "react";
+import { clamp } from "motion";
+import {
+  useIsSidebarShowing,
+  useSidebarDesktopMotionProgress,
+} from "@/components/ui/sidebar.js";
 import {
   COARSE_POINTER_HEADER_ICON_BUTTON_CLASS,
   COARSE_POINTER_HEADER_REDUCED_GLYPH_ICON_BUTTON_CLASS,
@@ -47,6 +57,7 @@ export function AppPageHeader({
   ownsWindowTopLeft = true,
 }: AppPageHeaderProps) {
   const isSidebarShowing = useIsSidebarShowing();
+  const sidebarMotionProgress = useSidebarDesktopMotionProgress();
   const isCompactViewport = useIsCompactViewport();
   const [desktopInfo] = useState(getBbDesktopInfo);
   const desktopWindowState = useDesktopWindowState();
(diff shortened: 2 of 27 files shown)


SawyerHood:
@slopcop review

PR #2639, opened 2026-08-28
Correlate codex turn dispatches with the turn the turn/start response names

### LINKED ISSUE #2580: Codex bridge fabricates a completed turn when turn/started is delayed
## Summary

When Codex answers `turn/start` before emitting `turn/started`, the bridge fabricates a completed zero-work turn after 250 ms. If the real notification arrives later, BB records both turns and can tell a parent that its child completed with no output while the child is still working.

BB should correlate the accepted input with the real turn and emit exactly one lifecycle.

## Versions and environment

- Reproduced on upstream `main` at `ec8f4ef04105c2cb5a59f7a9a9328bf2b16b39ce`
- Codex CLI `0.149.1`
- Hermetic reproduction: Linux 6.8 x86_64, Node 24.19.0
- Also observed in production on a private bb deployment whose source matches upstream `31d66d9d45bc8ff1483d5e4c6401c434b610dc4b` for the files involved (details under Evidence)

## Steps to reproduce

1. In `plugins/provider-codex/src/bridge/fake-codex-app-server.mjs`, change `LATE_TURN_START_DELAY_MS` from `60` to `350`.
2. Run:

   ```sh
   pnpm exec turbo run test --filter=bb-plugin-provider-codex -- --testNamePattern='lets a turn/started that lands after the turn/start response win the race'
   ```

3. Observe that the existing exactly-once assertion fails because the bridge emits two starts.

The same test passes with the fixture's 60 ms delay. The failure starts once the delayed notification crosses the bridge's 250 ms grace period.

## Expected vs actual

```text
Expected: one real turn/started and one real turn/completed
Actual:   expected [ Array(2) ] to have a length of 1 but got 2
```

The first lifecycle is a synthetic zero-work completion. The second is Codex's real turn.

## Evidence

- Reproduction base: `ec8f4ef04105c2cb5a59f7a9a9328bf2b16b39ce` (upstream `main`)
- The timer and fabricated lifecycle: https://github.com/get-bb/bb/blob/ec8f4ef04105c2cb5a59f7a9a9328bf2b16b39ce/plugins/provider-codex/src/bridge/bridge.ts#L1295-L1326
- The existing delayed-start fixture: https://github.com/get-bb/bb/blob/ec8f4ef04105c2cb5a59f7a9a9328bf2b16b39ce/plugins/provider-codex/src/bridge/fake-codex-app-server.mjs#L60-L68
- The exactly-once test: https://github.com/get-bb/bb/blob/ec8f4ef04105c2cb5a59f7a9a9328bf2b16b39ce/plugins/provider-codex/src/bridge/bridge.zero-work-turn.test.ts#L161-L191

Production observation, from Codex child threads on a private bb deployment (source matching upstream `31d66d9d4` for these files; the affected thread ids live on that instance and are not resolvable here): the synthetic completion appeared 259 to 362 ms after thread identity, the real turn started 351 ms to 2.743 seconds later, and only the synthetic turn received the original `clientRequestId`, which rules out duplicate sends.

## What you ruled out

- Not a duplicate user send: each affected child has one `client/turn/requested`.
- Not event replay: the synthetic and real lifecycles have distinct turn ids, and only the synthetic turn owns the request correlation.
- Not issue #1646: that report concerns work continuing after a genuine completed turn; this bug creates a separate synthetic completion before the genuine turn starts.
- Not cross-provider behavior: the fixed 250 ms zero-work timer is in `provider-codex`.

High impact, small-to-medium fix. A parent can start duplicate work in the same workspace based on the false completion, with no reliable user workaround.

> AGENT GENERATED

ratulsarna:
## What was wrong

The bridge ignored the `turn/start` result (`z.unknown()`, discarded) and settled accepted dispatches by guessing: if no `turn/started` claimed the correlation within 250 ms of the answer, it fabricated a completed zero-work turn. Codex answers `turn/start` at turn creation and can emit `turn/started` later than that window (observed 351 ms to 2.7 s under load), so the bridge recorded a synthetic completed turn followed by the real turn. A parent thread saw its child complete with no output while the child was still working. Root cause and repro in #2580.

## What changed

- `plugins/provider-codex/src/bridge/bridge.ts`: `handleTurnStart` captures the dispatch result. `settleAcceptedDispatch` parses it; a response naming a turn settles against that real turn, anything else (compact's `{}`, malformed body) keeps the existing zero-work timer unchanged.
- `plugins/provider-codex/src/translator.ts`: `openTurnFromStartResponse` proves ownership the same way `claim()` does (dispatch still at the correlation queue head), then opens the turn through the same `turn/started` translation the real notification would take. The provider's own later `turn/started` for that id is suppressed so it cannot double-open the turn or shift another dispatch's correlation. A response reporting an already-terminal status settles the full lifecycle at once and suppresses both later notifications. A dispatch codex steers into the running turn only emits `input.accepted` against it.
- Because the turn enters `openCodexTurnIds` at response time, interrupt waiting, steering, and child-exit settlement see it with no new state: a crash between answer and notification now settles the real turn as failed instead of fabricating a completed one.
- `packages/host-daemon-contract`: `HOST_DAEMON_PROTOCOL_VERSION` 173 → 174 (same delta kinds, but the lifecycle sequences the daemon emits change) and the version assertion in `contract.test.ts`.
- `fake-codex-app-server.mjs`: `turn/start` scenarios now answer with the turn like every supported codex (0.136.0 up already returns `{ turn }`; verified against codex-rs `rust-v0.136.0`), the late-start delay crosses the grace window, and new prompts cover the crash, steer, terminal-response, and settle-before-started shapes.

Known limit: a codex whose `turn/start` answers without a turn still gets the old 250 ms guess; with no turn id there is nothing to correlate. No supported version does this.

## How you verified

- New/updated tests in `bridge.zero-work-turn.test.ts` fail before and pass after: exactly-once lifecycle when `turn/started` lags past the grace window (the issue's repro, previously two lifecycles), failed settlement when codex dies before `turn/started` (previously a fabricated completed turn), steered dispatch acknowledged on the running turn, terminal-status response settling at once, and no resurrection when an interrupt settles a response-opened turn before its `turn/started` arrives.
- `pnpm exec turbo run test` green for `bb-plugin-provider-codex` (253), `@bb/host-daemon-contract` (51), `@bb/provider-bridge-protocol` (238), `@bb/agent-runtime` (318), `@bb/host-daemon` (561), `@bb/integration-tests` (77); full-repo `turbo run typecheck` (80 tasks).
- Probed real `codex app-server` 0.149.1: `turn/start` answers at turn creation with `{"turn":{"id":…,"status":"inProgress"}}`, and the answer can precede `turn/started`.

Fixes #2580

> AGENT GENERATED

### DIFF
--- a/packages/host-daemon-contract/src/protocol.ts
+++ b/packages/host-daemon-contract/src/protocol.ts
@@ -1,3 +1,3 @@
-export const HOST_DAEMON_PROTOCOL_VERSION = 173 as const;
+export const HOST_DAEMON_PROTOCOL_VERSION = 174 as const;
 
 export const HOST_ARTIFACT_MAX_BYTES = 256 * 1024 * 1024;
--- a/packages/host-daemon-contract/test/contract.test.ts
+++ b/packages/host-daemon-contract/test/contract.test.ts
@@ -943,7 +943,7 @@ const ACP_BRIDGE_LAUNCH = {
 
 describe("host-daemon command schemas", () => {
   it("uses the current host-daemon protocol version", () => {
-    expect(HOST_DAEMON_PROTOCOL_VERSION).toBe(173);
+    expect(HOST_DAEMON_PROTOCOL_VERSION).toBe(174);
     expect(HOST_ARTIFACT_MAX_BYTES).toBe(256 * 1024 * 1024);
   });
 
--- a/plugins/provider-codex/src/bridge/bridge.ts
+++ b/plugins/provider-codex/src/bridge/bridge.ts
@@ -59,6 +59,7 @@ import {
 } from "../interactive-requests.js";
 import { parseModelsResponse } from "../models.js";
 import { macOsPermissionPresentation } from "../presentation.js";
+import { codexTurnSchema } from "../schemas.js";
 import {
   resolveCodexInstructionOverrides,
   toCodexDynamicTools,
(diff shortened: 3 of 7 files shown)


SawyerHood:

PR #2647, opened 2026-08-28
Keep touch-only controls visible on tablets

banjerluke:
## Human comments

Using `bb` on a 13" Boox Note Max tablet these days, which is pretty awesome. There are a few areas where I'm unable to perform commands in the UI due to lack of a mouse, though. This PR fixes that.

## What was wrong

Several tab and message actions were revealed only on hover. The touch fallback was also restricted to compact viewport widths, leaving coarse-pointer tablets at desktop widths without a reliable way to close panel tabs or access message actions.

## What changed

Tab close buttons now remain visible whenever the device reports a coarse pointer or no hover capability, regardless of viewport width. Message action bars likewise select their touch-friendly inline or overflow surface from pointer capability instead of combining it with the compact viewport breakpoint.

The affected tests now cover wide coarse-pointer message actions and width-independent tab close affordances. This is an app-only change with no protocol, CLI, or guide changes.

## How you verified

- `pnpm exec turbo run test --filter=@bb/app --force`
  - 442 test files passed
  - 3,495 tests passed; 3 skipped
- `pnpm exec oxfmt --check` on all five changed files
- `git diff --check upstream/main...HEAD`

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/plugin/docs-anatomy-manifest.test.tsx
+++ b/apps/app/src/components/plugin/docs-anatomy-manifest.test.tsx
@@ -227,7 +227,7 @@ describe("docs anatomy manifest", () => {
         <MessageActionBar
           messageText="hello"
           alignment="start"
-          mobileActionDisplay="inline"
+          touchActionDisplay="inline"
           onAddToChat={() => {}}
           onEdit={() => {}}
           onFork={() => {}}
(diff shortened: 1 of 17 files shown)


SawyerHood:

PR #2650, opened 2026-08-28
Default Add Project to this machine and flag duplicate folders

nqrwhal:
## What was wrong

On a server with several enrolled machines, the Add Project dialog's
machine dropdown always defaulted to the server's primary host
(`usePathPickerHost` resolved `connectedPrimaryHostId ?? localDaemonHostId`),
even when the app was running on a secondary machine whose daemon bb had
already detected. Submitting with that default bound the new project to
the primary machine's checkout, and when the chosen path matched an
existing project's `(host, path)` source, `POST /projects` deduped by
returning the existing project with `201` — the dialog closed and the
app navigated to a different project with no feedback. From the second
machine this compounds into "I cannot create a project; it always
defaults to the last project on another machine".

## What changed

- `apps/app/src/hooks/useLocalPathPicker.tsx` — `usePathPickerHost` now
  prefers the machine this app runs on when its daemon is enrolled and
  connected. The daemon-reported host id is cross-checked against the
  server's host list, so a daemon enrolled on a different server cannot
  hijack the default; the primary host and the raw local daemon id
  remain the fallbacks, keeping the single-machine and
  plain-remote-browser behavior unchanged.
- `apps/app/src/components/dialogs/ProjectPathDialog.tsx` — the create
  flow accepts the known projects (id, name, local-path sources) and,
  when the selected folder on the selected machine already belongs to a
  project, shows `This folder is already the project "<name>" on
  <machine>.` and disables submit, replacing the silent dedupe redirect.
- `apps/app/src/hooks/useQuickCreateProject.tsx` — derives that project
  list from the sidebar bootstrap and exposes it on the controller;
  `AppLayout` passes it to the dialog. No wire or contract changes.

## How you verified

- New tests: `usePathPickerHost` targets this machine when its daemon is
  enrolled and connected, and falls back to the primary host when the
  daemon is enrolled elsewhere, offline on the server, or absent;
  `ProjectPathDialog` names the owning project and blocks submit for a
  known folder and stays enabled when another machine owns the same
  path; `useQuickCreateProject` exposes project sources for the check.
- `vitest run` over the seven touched/consuming test files: 45 passed.
- Manual: launched the dev app (`scripts/bb-dev-app current`), seeded a
  project at a local path, and drove the dialog in a real browser —
  selecting that folder shows the notice and greys out Add project;
  selecting a different folder creates the project normally.

Fixes #

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/dialogs/ProjectPathDialog.test.tsx
+++ b/apps/app/src/components/dialogs/ProjectPathDialog.test.tsx
@@ -153,3 +153,75 @@ describe("ProjectPathDialog machine selection", () => {
     expect(onSubmit).not.toHaveBeenCalled();
   });
 });
+
+describe("ProjectPathDialog duplicate folders", () => {
+  const projects = [
+    {
+      id: "proj_givecare",
+      name: "givecare",
+      sources: [{ hostId: "host_atum", path: "/home/deploy/repos/givecare" }],
+    },
+  ];
+
+  it("names the owning project and blocks submit for a known folder", () => {
+    const onSubmit = vi.fn();
+    render(
+      <ProjectPathDialog
+        target={{ kind: "create" }}
+        platform="linux"
+        hostId={atum.id}
+        hostName={atum.name}
+        hosts={[atum]}
+        projects={projects}
+        onOpenChange={vi.fn()}
+        onSubmit={onSubmit}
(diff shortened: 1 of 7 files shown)


SawyerHood:

PR #2675, opened 2026-08-28
fix(host-daemon): bump node-pty to 1.2.0-beta.15 to stop macOS pty fd leaks

smsunarto:
## What was wrong

node-pty 1.1.0 leaks three file descriptors per spawned terminal on macOS:

- The cleanup loop in `pty_posix_spawn` (`for (; count > 0; count--)`) never closes `low_fds[0]`. One `/dev/ptmx` master leaks per spawn.
- The parent's copy of the slave fd is never closed.
- The kqueue opened by `SetupExitCallback` is never closed.

The host daemon spawns a pty per terminal session, so a long-lived daemon drains the system-wide pool of 511 pty devices. On the machine that hit this, the daemon held 505 leaked masters after 22 hours. No app on the host, Ghostty included, could open a new terminal.

Upstream fixed all three leaks in microsoft/node-pty#882 and microsoft/node-pty#931.

## What changed

- `node-pty` 1.1.0 → 1.2.0-beta.15 in `apps/host-daemon` and `packages/bb-app`, plus the lockfile.
- New darwin-only regression test: `apps/host-daemon/src/terminals/node-pty-fd-leak.test.ts` spawns five ptys and asserts the `/dev/ptmx` fd count returns to baseline. CI's macOS 15 runners execute it.

Why a beta: npm `latest` is still 1.1.0, and the fixes exist only on the 1.2.0 line. VS Code ships this exact version in production (`^1.2.0-beta.15` in `package.json` and `remote/package.json`).

The only typings change since 1.1.0 is an optional `pixelSize` argument on `resize`. Nothing on the wire changes, so no `HOST_DAEMON_PROTOCOL_VERSION` bump.

## How you verified

- The new test fails on 1.1.0 (`expected 5 to be less than or equal to 0`) and passes on 1.2.0-beta.15. The failing test is committed before the fix.
- A standalone 25-spawn repro leaks 25 masters on 1.1.0 and zero on the beta.
- `pnpm exec turbo run test --filter=@bb/host-daemon`: 561 passed. One pre-existing flake in `host-branches-dispatch.test.ts`, passes standalone, unrelated.
- `pnpm exec turbo run typecheck --filter=@bb/host-daemon --filter=bb-app`: passes.

> AGENT GENERATED

### DIFF
--- a/apps/host-daemon/package.json
+++ b/apps/host-daemon/package.json
@@ -43,7 +43,7 @@
     "gray-matter": "^4.0.3",
     "jiti": "^2.6.1",
     "mime-types": "^3.0.2",
-    "node-pty": "1.1.0",
+    "node-pty": "1.2.0-beta.15",
     "p-retry": "^6.2.1",
     "partysocket": "^1.3.0",
     "pino": "^9.6.0",
--- a/apps/host-daemon/src/terminals/node-pty-fd-leak.test.ts
+++ b/apps/host-daemon/src/terminals/node-pty-fd-leak.test.ts
@@ -1,46 +1,137 @@
-import { execFileSync } from "node:child_process";
+import { execFile } from "node:child_process";
 import { createRequire } from "node:module";
-import os from "node:os";
 import path from "node:path";
-import { spawn } from "node-pty";
-import { describe, expect, it, vi } from "vitest";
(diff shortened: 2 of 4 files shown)


SawyerHood:
@slopcop review this

PR #2688, opened 2026-08-29
feat: event-driven thread wait coordinator

jonolee-kr:
## Summary

Fixes excessive memory and request load from `bb thread wait`.

- **Server**: new reference-counted wait coordinator keyed by `(threadId, target, cursor)`. Identical waits share one `NotificationHub` subscription and one coalesced DB re-check. Per-caller deadlines, abort cleanup, caller limits, pino observability with an `unref()`ed summary timer, and a stats accessor.
- **Route**: new `GET /api/v1/threads/:id/wait?status=...` long-poll (404 `thread_not_found` for missing threads). The existing `events/wait` route is refactored onto the coordinator with its response contract preserved.
- **SDK**: status waits long-poll the new route instead of client-side 250 ms polling. Fallback to the legacy polling loop triggers only on `BbHttpError` `status 404` + `code not_found` (route absent on older servers). The wait sleep is now abort-aware. `pollIntervalMs` governs only the fallback path.
- **CLI**: `bb thread wait --output` (one blocking follow returning terminal status + final output), `bb thread wait-many` (many logical waits in one process), and lazy-loaded thread subcommands. Existing wait messages, JSON shapes, and exit codes unchanged.
- **Guidance/docs**: builtin skills no longer teach log/wait alternation; changelog documents rollout compatibility; QA runbook adds PSS measurement.

## Motivation & measured impact

On a production host, orchestration bursts spawned duplicate `bb thread wait` processes, each a full Node/V8 runtime polling every 250 ms:

- Old CLI waiter: 150–174 MiB RSS / 99–122 MiB PSS each (mostly private dirty), 4 status GETs/s per waiter, each ~6 DB operations.
- New CLI waiter: 48 MiB RSS / 12 MiB PSS (standalone measurement).
- Live end-to-end demo against this branch: 20 concurrent CLI waiters → 1 coordinator entry, 20 total HTTP requests for the whole wait, one shared DB re-check on the status flip, all 20 resolved within 77–133 ms, zero leaked callers/entries. Old CLI in the same harness: 3.8 GETs/s per waiter.

## Compatibility

- Old CLI + new server: keeps working (route addition only).
- New CLI + old server: degrades to the legacy polling loop via the narrow 404 fallback.
- Timeout semantics, `--status`/`--event` targets, JSON output, and exit codes are preserved.

## Testing

- apps/server: 2083 passed / packages/sdk: 102 passed / apps/cli: 513 passed, 0 failed, no new skips.
- Five deliberate mutations (broken dedup, shared abort, broad fallback, non-abortable sleep, wait-after-output) are each caught by the new tests.
- 100-duplicate-wait stress tests assert one coordinator entry, bounded checks, and full cleanup across timeout, cancellation, disconnect, unreachable, and error paths.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ All built-in agent providers now use the provider API. You can use the same API 
 
 ### CLI
 
+- Follow one thread with `bb thread wait <id> --output`. Follow several threads in one process with `bb thread wait-many`.
 - Use `--plan` with `bb thread tell` or `bb thread spawn` to enter Plan mode.
 - Use `bb thread log --all` to read the full thread history.
 - Move a local plugin with `bb plugin install path:<new directory>`. bb keeps its data and settings.
@@ -40,6 +41,7 @@ All built-in agent providers now use the provider API. You can use the same API 
 
 ### Performance
 
+- Thread status waits now use one server-side event subscription for duplicate conditions instead of 250 ms client polling.
 - A thread now opens with about half as many requests.
 - Smaller bundles reduce the initial load time for the app and plugins.
 - Long threads use less server work and load timeline pages faster.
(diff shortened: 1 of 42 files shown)


SawyerHood:

PR #2704, opened 2026-08-29
Rework the mobile PWA home and compact shelves

brsbl:
## What was wrong

The compact home put a capped three-thread Recent list below the composer in one page scroller, so the composer moved away from the thumb, the list did not signal additional content, and thread rows omitted provider, hierarchy, worktree, branch, and activity context. Both compact side surfaces behaved as overlays: the left navigation drawer covered and dimmed the page, while the right panel rose as a bottom drawer. Cold thread navigation could begin with missing metadata, route transitions had no pending affordance, and provider discovery repeated seconds-long host probes and uncached logo work.

## What changed

- Compact home now anchors the production composer at the bottom while realistic recent-thread rows scroll behind it. A partial row signals more content and the `Recent` label stays pinned to the scroll viewport.
- Recent rows use production provider marks, parent/child hierarchy, status, project/worktree/branch metadata, and attention-time ordering.
- The compact left navigation becomes a 76vw shelf: the live page translates right, keeps its state, and remains undimmed instead of being covered by an overlay.
- The compact right panel moves from the bottom drawer to a right-side shelf: the live page translates left and remains visible. Layer 2 builds on this shelf for full-page tool tabs.
- Slow navigation now gets delayed, minimum-duration route feedback; cold thread boots reuse persisted sidebar metadata instead of beginning blank.
- Provider installed-state is cached for five minutes instead of re-probing provider bridges on every list, and hashed provider logos use immutable content-addressed URLs.
- Ladle renders the real `NewThreadPromptBoxUI` inside the real compact-home composition with populated production fixtures.

No wire changes; `HOST_DAEMON_PROTOCOL_VERSION` is unchanged.

## Screenshots

Chrome for Testing 151.0.7922.71, 393×852 at DPR 2, using the same seeded project, thread data, route, viewport, and settled-content state. Before is the merge base (`f4bbc2fe8`); after is this PR head (`04d18307f`).

| Before — composer occupies the top and Recent ends after three sparse rows | After — realistic recents scroll behind a bottom-anchored production composer |
| --- | --- |
|  |  |

### Sticky header while scrolled

At the PR head, the compact-home viewport is scrolled to `scrollTop=360`. The `Recent` header remains exactly at the viewport top while rows continue underneath and the composer remains anchored at the bottom.

## How you verified

- The exact-revision branch web app was driven from first paint through settled content in Chrome for Testing. The compact-home root measured 393×852 and the production composer measured 361×128 at the bottom.
- After scrolling, the compact-home viewport measured `top=318`, `height=534`, `scrollTop=360`; the production `Recent` header measured `top=318`, proving it remained pinned to its own scroll viewport rather than the document body.
- Ladle was restarted on port 61002. Compact Home renders two real production prompt boxes, each 359×128, inside non-zero 391×850 compact-home roots; the composer remains bottom-aligned after scrolling.
- Mobile Recents renders five real component subtrees with row counts `3, 2, 5, 3, 3`, provider glyphs for every row, and a working hierarchy collapse from five rows to two.
- Every affected story ID was loaded and measured through its production component subtree with zero runtime exceptions. All remote CI gates pass; no local CI-equivalent command was run.

The soft-keyboard rise still requires a real coarse-pointer device, and provider logo URLs require the bb server rather than standalone Ladle.

BB-Thread-ID: thr_wftu7bh9ez

> AGENT GENERATED

### DIFF
--- a/apps/app/.ladle/story-fixtures.ts
+++ b/apps/app/.ladle/story-fixtures.ts
@@ -2,6 +2,7 @@ import type {
   Environment,
   Host,
   ProjectSource,
+  ProviderInfo,
   ReasoningLevel,
   Thread,
   ThreadListEntry,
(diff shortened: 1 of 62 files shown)


brsbl:
Superseded by focused native stack #2763: #2758 → #2759 → #2760 → #2761 → #2762. Replacement layer #2761 has cumulative content byte-for-byte identical to this PR head; review should continue on the focused stack.

PR #2705, opened 2026-08-29
Add Plugin Guide entry to My Plugins

brsbl:
## Product intent and requirements

- Keep the Plugin Guide discoverable from My Plugins at every authored-plugin count.
- Open the guide from the plugin-authoring context without making users hunt through Installed Plugins.
- Use the map icon at rest and add only the pin on hover or keyboard focus without shifting the control.
- Enable a disabled bundled guide when needed, show compact progress and recoverable failure states, and preserve existing plugin policy.

## What was wrong

My Plugins rendered authored-plugin content but had no stable entry point for the bundled Plugin Guide. Because the guide is intentionally installed but disabled by default, prospective and active plugin developers could not discover or launch it from the page where they work.

## What changed

- Adds an always-visible **Open Plugin Guide** icon-and-label button to the My Plugins toolbar, including both the zero-plugin creation state and populated authored-plugin lists.
- Uses `MapsIcon` at rest and `MapsLocation02Icon` on hover or keyboard focus without changing button geometry.
- Opens an enabled guide immediately. When disabled, the action enables it without confirmation, waits for its frontend panel registration, then opens it.
- Keeps enable failures on My Plugins and shows recoverable feedback. The loading treatment reserves the resting label width, so the toolbar does not reflow.
- Leaves guide installation, authorship detection, permissions, lifecycle policy, and the host-daemon protocol unchanged.

## Before and after

Matched captures use the same path-installed `Agent enrichment` plugin, `/extensions/plugins?view=my` route, settled state, and 1440×900 Chrome for Testing viewport.

**Before — Extensions integration parent `7422881a73b734cb9e2848daf0e8f5f55b99d26c`**

Before: populated My Plugins has no Plugin Guide entry point

**After — feature head `97c166e5561551532c11da76a8d16f15b1f70e18`**

After: populated My Plugins shows Open Plugin Guide in the toolbar

### Zero-plugin creation state

Matched captures remove only the isolated path-installed fixture through the real plugin CLI, then use the same My Plugins URL and viewport. The head keeps the existing creation state intact and adds the same guide entry point.

| Before — empty My Plugins | After — empty My Plugins with guide entry |
| --- | --- |
| Before: empty My Plugins creation state without a Plugin Guide action | After: empty My Plugins creation state with Open Plugin Guide in the toolbar |

## How you verified

- Exact parent and head branch web apps launched with `scripts/bb-dev-app branch ...`, driven in Chrome for Testing 151.0.7922.71.
- Installed the same local `Agent enrichment` fixture into the isolated QA data directory and confirmed the only rendered difference is the Plugin Guide action.
- Confirmed both empty and populated states on the exact parent/head pair. Prior focused product QA also covered disabled-to-enabled progress and panel registration; enabled immediate open; recoverable failure; stable-width loading; and hover/keyboard icon transitions at 1440px and 768px.
- Pull-request CI is green on exact head `97c166e5561551532c11da76a8d16f15b1f70e18`; all required checks passed and the platform-inapplicable jobs were skipped. Per repository policy, tests, typechecks, and lint were not run locally.

BB-Thread-ID: thr_q5kiakrmeh

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/plugin/PluginsOverview.tsx
+++ b/apps/app/src/components/plugin/PluginsOverview.tsx
@@ -32,6 +32,7 @@ import { BrowsePluginsTab } from "@/components/plugin/management/BrowsePluginsTa
 import { CheckPluginUpdatesButton } from "@/components/plugin/management/CheckPluginUpdatesButton";
 import { InstalledPluginsTab } from "@/components/plugin/management/InstalledPluginsTab";
 import { MyPluginsTab } from "@/components/plugin/management/MyPluginsTab";
+import { OpenPluginGuideButton } from "@/components/plugin/management/OpenPluginGuideButton";
 import {
   PLUGIN_SOURCE_FILTER_OPTIONS,
   pluginSourceFilterId,
@@ -76,6 +77,7 @@ export function PluginsOverview({
     () => listQuery.data?.plugins ?? [],
     [listQuery.data?.plugins],
   );
+  const pluginGuide = plugins.find((plugin) => plugin.id === "plugin-api-docs");
   const activeMode = modeFromSearchParams(searchParams.get("view"));
   const [installedQuery, setInstalledQuery] = useState("");
   const [installedViewport, setInstalledViewport] =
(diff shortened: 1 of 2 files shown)


SawyerHood:
Closing: My Plugins is being removed by owner decision, so this entry point has no home. No replacement is planned.

> AGENT GENERATED

PR #2718, opened 2026-08-30
Tolerate event-loop starvation in the connect tunnel heartbeat

vburojevic:
## What was wrong

bb connect's tunnel client kills healthy tunnels when the bb server process is starved. The relay answers heartbeats at the Cloudflare edge via `setWebSocketAutoResponse` (`apps/connect/src/tunnel-do.ts`), so acks are essentially guaranteed while the link is up — but under host overload (load average 143–234 on 10 CPUs observed) the server event loop cannot process the queued ack messages within the 60s `HEARTBEAT_DEADLINE_MS`. The heartbeat tick then sees a stale `lastAck`, logs `tunnel heartbeat missed; reconnecting`, and terminates a live socket. Reconnection then waits for the ws `close` event, which under the same starvation was measured arriving 50–82 seconds later. During that window remote visitors get `bb connect: tunnel disconnected mid-request` and the offline interstitial even though bb is running. One affected machine logged 123 such false terminations in a single day (2026-08-29).

Node's event-loop phase ordering makes the false positive structural: expired timers run before poll-phase I/O, so on a recovering loop the heartbeat tick always observes `lastAck` as stale even when the acks are already sitting in the socket buffer.

## What changed

- `packages/tunnel-client/src/session.ts`
  - Starvation grace: the heartbeat tick now tracks its own cadence. When the tick itself ran more than 2× the interval late, ack staleness is self-inflicted (the process was starved or suspended), so the session grants ack grace and re-pings instead of terminating. A genuinely dead tunnel still dies within one deadline once ticks are timely again.
  - Drain-before-kill: when the deadline appears exceeded on a timely tick, the kill decision is deferred one `setImmediate` so acks already queued in the poll phase are counted first.
  - New optional `onHeartbeatTimeout` session callback; a `disposed` flag makes a pending deferred decision inert after `dispose()`.
- `plugins/connect/src/tunnel.ts`: on a genuine heartbeat timeout the tunnel now schedules its reconnect immediately via the existing idempotent `scheduleReconnect` instead of waiting for the delayed `close` event (which also disposes the session, stopping the repeated `heartbeat missed` warns). The later `close` remains a guarded no-op.

No wire or protocol change: `packages/tunnel-contract`, `apps/connect`, and `HOST_DAEMON_PROTOCOL_VERSION` are untouched.

## How you verified

- New `packages/tunnel-client/test/session-heartbeat.test.ts` (fake timers, fake ws): timely acks keep the session alive; a genuine miss terminates and reports exactly once; an ack queued at decision time rescues the tunnel; a starved timer grants grace instead of terminating; a deferred decision is inert after disposal. The four behavioral cases fail against the previous implementation.
- New case in `plugins/connect/src/tunnel-lifecycle.test.ts`: heartbeat timeout schedules the retry before any `close` event, and a later `close` does not double-schedule (fails before the change with the state stuck `connected`).
- `pnpm exec turbo run typecheck --filter=@bb/tunnel-client --filter=bb-plugin-connect` — clean.
- `pnpm exec turbo run test --filter=@bb/tunnel-client --filter=bb-plugin-connect --force` — 17/17 and 91/91 pass.

> AGENT GENERATED

### DIFF
--- a/packages/tunnel-client/src/session.ts
+++ b/packages/tunnel-client/src/session.ts
@@ -21,6 +21,7 @@ import type { TunnelClientLogger } from "./logger.js";
 
 const HEARTBEAT_INTERVAL_MS = 20_000;
 const HEARTBEAT_DEADLINE_MS = 60_000;
+const HEARTBEAT_STARVATION_GRACE_MS = 2 * HEARTBEAT_INTERVAL_MS;
 
 const UNREGISTERED_PORT_BODY = "this port is not shared";
 const textEncoder = new TextEncoder();
@@ -118,13 +119,16 @@ interface TunnelSessionOptions {
   resolveOrigin: (target: string | undefined) => StreamOriginResult;
   onRemoteClientsChange?: (remoteClients: number) => void;
   onActivity?: (at: number) => void;
+  onHeartbeatTimeout?: () => void;
 }
 
 export class TunnelSession {
   private readonly httpStreams = new Map<number, HttpStream>();
   private readonly wsStreams = new Map<number, WsStream>();
   private lastAck = Date.now();
+  private lastTickAt = 0;
   private heartbeat: ReturnType<typeof setInterval> | undefined;
+  private disposed = false;
   private remoteClientCount = 0;
   lastRemoteActivityAt: number | null = null;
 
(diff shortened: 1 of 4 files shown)


SawyerHood:

PR #2719, opened 2026-08-30
Contain late tunnel frames and sync origin failures in TunnelSession

vburojevic:
## What was wrong

On 2026-08-28 the connect plugin killed the bb server six times in two seconds with `PluginContextStaleError` (uncaught-exception dumps attached the full chain). When a plugin reload/disable stops the tunnel background service, `ConnectTunnel.stop()` disposes the `TunnelSession` and terminates the websocket — but frames already received and still decompressing through permessage-deflate are delivered by zlib callbacks afterwards. `TunnelSession.dispose()` neither detached nor gated the `message` listener, so the disposed session still processed an `open-http` frame. `executeHttp` calls `resolveOrigin` synchronously outside its try block, that call reached the plugin's now-stale `bb.server.loopbackBaseUrl` handle (`assertLive` throws), and because the caller runs `void this.executeHttp(...)`, the synchronous throw escaped as an unhandled rejection that Node escalated to a process-level uncaught exception (stack: `Zlib.processCallback → receiverOnMessage → onFrame → executeHttp → resolveStreamOrigin → get loopbackBaseUrl → assertLive`).

## What changed

`packages/tunnel-client/src/session.ts` only:

- Frame processing is gated on the existing `disposed` flag, both in the tunnel `message` handler and at the top of `onFrame` (the second gate also covers the buffered-frame replay in `openOriginWs`'s open handler).
- `executeHttp` now performs origin resolution, unregistered-origin handling, and header construction inside its existing try block, so a synchronous failure follows the established catch path (a `close-stream` frame with code 1011 and the error text as reason) instead of escaping the voided promise. The unregistered path is behavior-identical; its stream cleanup now flows through the existing `finally`.

No wire, protocol, signature, or heartbeat changes.

Stacked on #2718, which touches the same file — review only the last commit here (`Contain late tunnel frames and sync origin failures in TunnelSession`); the first commit is that PR.

## How you verified

- New `packages/tunnel-client/test/session-dispose.test.ts`: a binary `open-http` frame after `dispose()` is ignored (no `resolveOrigin` call, nothing sent); a synchronously-throwing `resolveOrigin` yields a `close-stream` code-1011 frame with no unhandled rejection; a heartbeat response after disposal is ignored. Before the source change the focused suite exited 1 with 2/3 failing and Vitest reporting the exact production crash shape — one unhandled rejection from `TunnelSession.executeHttp`; after the change it passes 3/3.
- `pnpm exec turbo run typecheck --filter=@bb/tunnel-client --filter=bb-plugin-connect` — clean.
- `pnpm exec turbo run test --filter=@bb/tunnel-client --filter=bb-plugin-connect --force` — 20/20 and 91/91 pass.

> AGENT GENERATED

### DIFF
--- a/packages/tunnel-client/src/session.ts
+++ b/packages/tunnel-client/src/session.ts
@@ -21,6 +21,7 @@ import type { TunnelClientLogger } from "./logger.js";
 
 const HEARTBEAT_INTERVAL_MS = 20_000;
 const HEARTBEAT_DEADLINE_MS = 60_000;
+const HEARTBEAT_STARVATION_GRACE_MS = 2 * HEARTBEAT_INTERVAL_MS;
 
 const UNREGISTERED_PORT_BODY = "this port is not shared";
 const textEncoder = new TextEncoder();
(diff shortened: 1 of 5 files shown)


SawyerHood:

PR #2727, opened 2026-08-30
fix(pi): accept image-only prompts

fgrehm:
## Human comments

Fix for thsi:

## What was wrong

Pi's provider bridge rejected a turn whenever its extracted text was empty, even when the request contained a valid local image attachment. As a result, sending a screenshot without accompanying text returned `Missing input text`.

## What changed

The Pi bridge now accepts a prompt or steer request when it contains either text or images. Image-only requests are forwarded to Pi with an empty text payload, while completely empty requests remain invalid. Added lifecycle coverage for a local image-only prompt.

No wire protocol, CLI, or documentation changes were required.

## How you verified

- `cd plugins/provider-pi && pnpm exec vitest run --config vitest.config.ts src/bridge/bridge.lifecycle.test.ts --reporter=dot`
- 1 test file passed, 9 tests passed.
- The new test failed initially because its test request ID did not match the repository's allowed format; after correcting the fixture, it passed.

> AGENT GENERATED

### DIFF
--- a/plugins/provider-pi/src/bridge/bridge.lifecycle.test.ts
+++ b/plugins/provider-pi/src/bridge/bridge.lifecycle.test.ts
@@ -4,6 +4,7 @@ import {
   mkdirSync,
   readFileSync,
   readdirSync,
+  writeFileSync,
 } from "node:fs";
 import { join } from "node:path";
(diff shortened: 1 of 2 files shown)


patleeman:
Ran into this same issue, thanks for pushing a PR

PR #2729, opened 2026-08-30
Add project archiving

### LINKED ISSUE #1786: Archiving Projects
I'd love to have a way to archive an entire project like I can with a thread. I jump around between a _lot_ of things, and I'm cluttering up my sidebar with projects that I know I'm not going to touch until next week or next month. I want to hide them from the sidebar and the new thread picker. Right now, the only way to do this is to delete the project, but that also deletes the threads from that project. I want to keep those, so they're searchable.

The thing that's debatable is what to do with non-archived threads when archiving a project. Claude told me there are two ways to think about it: "cosmetic" and "functional". "Cosmetic" is essentially just an `archivedAt` column and migration, but it leaves active threads untouched. If you archive and un-archive a project, it's like nothing happened at all. "Functional" would mean archiving any active threads in the project, which would retire worktrees.

Cosmetic gets me what I want, but I can see an argument for the functional approach to keep the back-end cleaner. I'm happy to open a PR either way!

MPIsaac-Per:
## What was wrong

Closes #1786. Threads can be archived, projects cannot. A project that is finished or paused stays in the sidebar and in every project list forever, so the only ways to get it out of the way are deleting it (destructive, takes the threads with it) or leaving the sidebar cluttered. There was no non-destructive way to retire a project and bring it back later.

## What changed

Projects gain an `archived_at` column and an archive/restore lifecycle that carries their threads with them.

- `packages/db`: migration `0110_next_newton_destine` adds `projects.archived_at`; `data/projects.ts` gains `archiveProject`/`unarchiveProject` and archive-aware list queries filtered with `WHERE` rather than in JavaScript.
- `packages/domain`, `packages/server-contract`, `packages/sdk`: `archivedAt` on the project shape, `archived` on the project PATCH request, plumbed through the SDK's `projects.update`. `name` on that request becomes optional so a caller can patch archive state alone; the schema refinement still rejects an empty patch.
- `apps/server`: `PATCH /api/v1/projects/:id` handles `archived: true|false`; archiving a project archives its threads and restoring returns them.
- `apps/app`: archive and restore in the project actions menu, archived projects excluded from the sidebar and reachable from settings.
- `apps/cli`: `bb project archive <id>`, `bb project restore <id>`, and `bb project list --archived`, with matching entries in the bb-cli skill command index and the projects guide chapter.

#1786 asks which of two shapes to take. This is the functional one, with the reversibility of the cosmetic one: archiving a project stamps the same `archivedAt` on the project and on every thread in it that was not already archived, and restoring unarchives only the threads carrying that exact stamp. Threads the user had archived by hand beforehand stay archived. Worktrees are not retired, so a restore returns the project to the state it was in.

No host-daemon wire change, so `HOST_DAEMON_PROTOCOL_VERSION` is untouched.

## How you verified

- New tests that fail before and pass after: `apps/server/test/public/public-projects-local-host.test.ts` (archive/restore round trip over the public API, thread archiving side effect), `packages/db/test/data/threads.test.ts` (threads follow the project), `apps/app/src/components/project/ProjectActionsMenu.test.tsx`, `apps/cli/src/__tests__/command-output/project.test.ts`, and a rewind case in `packages/db/test/migrate.test.ts`.
- `pnpm exec turbo run typecheck` across `@bb/app @bb/cli @bb/server @bb/db @bb/sdk @bb/server-contract @bb/domain @bb/demo-server @bb/mobile` — 12/12 pass.
- `pnpm exec turbo run test` across `@bb/server @bb/cli @bb/app @bb/sdk @bb/server-contract @bb/domain @bb/db` — all pass (server 215 files, app 448, cli 51, db 28).

> AGENT GENERATED


MPIsaac-Per:
Closing this. I opened it before sign-off, which CONTRIBUTING.md asks contributors not to do for features and UI changes. The work stands as a prototype on my fork instead: https://github.com/MPIsaac-Per/bb/tree/feat/project-archiving

I have left a note on #1786 with the design question that issue raises. Happy to reopen if a maintainer wants this shape.

> AGENT GENERATED

PR #2734, opened 2026-08-30
Retry a refused turn on another model

### LINKED ISSUE #2733: A model refusal is untyped and unrecoverable: no way to retry the turn on another model
### The workflow

1. Open a bb thread on the Claude Code provider with the default model, Opus 5 (1M).
2. Send a message on a topic Anthropic's safety classifiers flag — in practice, ordinary security or infrastructure work triggers this.
3. The turn ends. Nothing in bb offers a way forward, so the only recovery is to open the model picker, pick another model by hand, and retype or repeat the message.

Step 3 is the problem. The refusal is model-specific — the API says this *model* will not answer this message — so the one useful action is "run it again on a different model", and bb makes the user assemble that by hand every time.

### What happens today

Claude Code reports the refusal as a `model_refusal_no_fallback` system message. It emits that subtype precisely when the stream ends with `stop_reason: "refusal"` and no fallback model is configured — and bb never configures one, so this is the message bb sees for every safeguards refusal. The user-visible text is:

```
API Error: Opus 5 (1M context)'s safeguards flagged this message
(https://www.anthropic.com/legal/aup). Our intentionally broad safeguards allow us to
deliver more capabilities faster, but can sometimes flag legitimate coding, cybersecurity,
and biology tasks. Claude Code can't respond to this message with Opus 5 (1M context).
```

bb translates that message into a generic `provider/warning`:

https://github.com/get-bb/bb/blob/f4bbc2fe8/plugins/provider-claude-code/src/delta-translation.ts#L693-L707

Two consequences:

- The timeline shows an untyped warning, not a recognisable failure.
- Nothing downstream can act on it. `policy` is a real `providerErrorCategory` with its own title in thread-view (`packages/domain/src/provider-event.ts`, `packages/thread-view/src/error-display.ts:20`), but no Claude Code path produces it — `buildClaudeProviderErrorInfo` in `plugins/provider-claude-code/src/error-info.ts` never returns `policy`. A plugin can match a rate limit (`errorInfo.category === "rate-limit"`, the way `plugins/provider-retry/src/recovery.ts:146-151` does) but has nothing equivalent to match here.

So a refusal is both invisible to automation and unactionable by hand.

### What you would expect

Two things, in this order:

1. **Make the refusal typed.** Translate `model_refusal_no_fallback` into a `provider/error` with `errorInfo.category: "policy"`, carrying the CLI's own explanation as `detail` and the refusal category (`cyber`, `bio`, …) as `providerCode`. This alone fixes the timeline title and makes the failure matchable. It reaches an existing category rather than adding one, so no domain or protocol change is involved.

2. **Offer the obvious recovery.** On a policy refusal, ask which model to retry on — offering the models listed *below* the refused one in that provider's own catalog — then switch the thread and continue. A "Switch automatically next time, do not ask again" option should store the choice per provider and skip the prompt afterwards, with a CLI to inspect and clear it.

Offering only lower catalog entries matters: it is what makes a repeatedly refused thread run out of alternatives instead of switching models forever.

Because it keys on the `policy` category rather than anything Claude-specific, the same recovery covers Codex, which already produces `policy` (`plugins/provider-codex/src/delta-translation.ts:307`).

### Context and alternatives

Workaround today: change the model in the picker and resend by hand, once per refusal.

Considered and rejected: passing the SDK's own `fallbackModel` option through `buildClaudeSessionParams`. Claude Code would then retry silently and emit `model_refusal_fallback`, which bb already renders. It is less work, but it takes the decision away from the user, its documented purpose is overload/unavailability rather than refusal, and it swaps the session's model behind bb's back, so bb's stored `modelOverride` and the session's actual model drift apart.

I have a working prototype of both halves on a fork, with tests: https://github.com/t1mdurden/bb/tree/feat/model-refusal-fallback — happy to adjust the shape, or to split it so only part 1 lands, before opening anything against this repo.

### Checks

> AGENT GENERATED

t1mdurden:
## What was wrong

Claude Code emits a `model_refusal_no_fallback` system message when the stream ends with `stop_reason: "refusal"` and no fallback model is configured — and bb never configures one, so that is the message bb sees for every safeguards refusal. `delta-translation.ts` turned it into a generic `provider/warning` with the free-text summary "Model refused the request", which made the failure both untyped in the timeline and invisible to automation: `policy` is a real `providerErrorCategory` with its own title in thread-view, but no Claude Code path could produce it, so a plugin had nothing to match the way `provider-retry` matches `rate-limit`. With nothing typed to react to, the only recovery was to open the model picker, switch by hand, and resend — once per refusal, even though the refusal is model-specific and "retry on another model" is the single useful action.

Fixes #2733

## What changed

**`plugins/provider-claude-code` — type the refusal.** `model_refusal_no_fallback` now translates to a `provider/error` with `errorInfo.category: "policy"`, the CLI's own explanation as `detail`, and the refusal category (`cyber`, `bio`, …) as `providerCode` when the CLI reports one. This reaches an existing category rather than adding one, so there is no domain or wire change and no `HOST_DAEMON_PROTOCOL_VERSION` bump. `original_model` and `api_refusal_category` are optional in the schema because the SDK documents both as absent from older CLIs. The timeline now titles it "Provider policy blocked request".

**`plugins/provider-refusal-fallback` — a new builtin plugin (enabled by default).** It reconciles a finished turn the way `provider-retry` does, but keyed on a `policy` provider error. It offers the models listed *below* the refused one in that provider's catalog, switches the thread with `sdk.threads.update`, and sends one agent-only `Please continue.` turn. Offering only lower catalog entries is what makes a repeatedly refused thread terminate rather than switch forever. "Switch automatically next time, do not ask again" stores the chosen model per provider in the plugin's KV storage and skips the prompt from then on.

It reads the catalog through `sdk.providers.models` and never names a model or a provider, so Codex's `policy` errors get the same treatment.

No new plugin API: it consumes `bb.ui.requestInput`, `bb.storage.kv`, `bb.events.on`, `bb.cli.register`, and `app.slots.pendingInteraction`, all already stable — so no `experimental_` prefix and no `docs/api_to_audit.md` entry.

**CLI and docs, per `docs/cli-guide-and-skill.md`.** `bb refusal-fallback status|forget|retry` gives agents the same feature as the UI. Documented in `docs/configuration.md`, `bb-guide-plugins.md`, and `bb-guide-providers.md`; registered in `builtin-registry.ts`, `plugin-icons.ts`, `turbo.json`, and the builtin/official plugin test fixtures.

Deliberately not done: passing the SDK's own `fallbackModel` through `buildClaudeSessionParams`. It is less code, but it decides for the user, its documented purpose is overload/unavailability rather than refusal, and it swaps the session's model behind bb's back so `modelOverride` and the live session drift apart.

## How you verified

Node 22.23.1, macOS.

```
pnpm exec turbo run test --filter=bb-plugin-provider-claude-code
  → 23 files, 337 passed (336 on main; the refusal test was replaced by two)

pnpm exec turbo run typecheck test --filter=bb-plugin-provider-refusal-fallback
  → 2 files, 12 passed; typecheck clean

pnpm exec turbo run test --filter=@bb/server -- --run test/services/plugins
  → 38 files, 444 passed  (builtin-plugins.test.ts reads the new manifest from disk)

pnpm exec turbo run test --filter=@bb/cli --filter=@bb/templates --continue
  → 503 + 43 passed

pnpm exec turbo run test --filter=@bb/plugin-api-map --filter=@bb/host-daemon-contract
  → 73 + 51 passed

pnpm exec turbo run typecheck --filter=@bb/server --filter=@bb/cli   → clean
pnpm exec oxlint plugins/provider-refusal-fallback plugins/provider-claude-code/src  → clean
```

New tests that fail before and pass after:

- `delta-translation.test.ts` — "surfaces refusal without fallback as a policy provider error" and "names the refusing model when the refusal carries no content". Both fail on `main`, which emits a `provider/warning`.
- `plugins/provider-refusal-fallback/server.test.ts` — `classifyRefusalFallback` returns the refused model only for a *finished* turn carrying a policy error; a rate-limit error, an unfinished turn, a manual stop, and a null environment each return no candidate. `alternativeModels` runs out at the bottom of the catalog, which is the termination guarantee.
- `plugins/provider-refusal-fallback/app.test.tsx` — the prompt submits the first offered model by default, carries `remember: true` only when the checkbox is ticked, and submits `model: null` when declined.

Not verified: I did not exercise this against a live refusal from the API. The translation is driven by the SDK's documented `SDKModelRefusalNoFallbackMessage` shape (`@anthropic-ai/claude-agent-sdk@0.3.251`) and tested through the existing delta harness, but a real end-to-end refusal has not been observed.

Pre-existing and unrelated: two `@bb/plugin-build` tests ("names the unbuilt SDK dist…") fail identically on `f4bbc2fe8` with no changes applied — the expected error string is compared against a macOS `/private/tmp` path.

Opened per CONTRIBUTING as the prototype for #2733 rather than as a merge request — happy to land only the first commit, or to reshape the second, on sign-off.

> AGENT GENERATED

### DIFF
--- a/apps/cli/src/__tests__/bb-cli-skill-coverage.test.ts
+++ b/apps/cli/src/__tests__/bb-cli-skill-coverage.test.ts
@@ -64,7 +64,7 @@ describe("bb-cli skill command index", () => {
   it("leaves shipped plugin commands to their plugin skills", () => {
     const skill = readMarkdownTree(BB_CLI_SKILL_ROOT);
     expect(skill).not.toMatch(
-      /\bbb (?:automation|connect|instructions|github|keep-awake|provider-retry|memory|secret|docs|tasks|workflows)\b/,
+      /\bbb (?:automation|connect|instructions|github|keep-awake|provider-retry|refusal-fallback|memory|secret|docs|tasks|workflows)\b/,
     );
     expect(skill).not.toMatch(/\b(?:built-in|builtin|official) plugins?\b/i);
   });
--- a/apps/server/src/services/plugins/builtin-registry.ts
+++ b/apps/server/src/services/plugins/builtin-registry.ts
@@ -124,6 +124,12 @@ export const BUILTIN_PLUGINS = [
     category: "Agent interaction",
   },
   {
+    name: "provider-refusal-fallback",
+    pluginId: "provider-refusal-fallback",
+    defaultEnabled: true,
+    category: "Agent interaction",
+  },
+  {
     name: "secrets",
     pluginId: "secrets",
     defaultEnabled: true,
--- a/apps/server/test/services/plugins/builtin-plugins.test.ts
+++ b/apps/server/test/services/plugins/builtin-plugins.test.ts
@@ -231,6 +231,7 @@ describe("builtin plugin reconciliation", () => {
       ["provider-claude-code", "./icons/claude-code.svg"],
       ["provider-codex", "./icons/codex.svg"],
       ["provider-pi", "./icons/pi.svg"],
+      ["provider-refusal-fallback", "SecurityCheck"],
       ["provider-retry", "ArrowReloadHorizontal"],
       ["secrets", "Lock"],
       ["side-chat", "SideChat"],
--- a/apps/server/test/services/plugins/official-plugins.test.ts
+++ b/apps/server/test/services/plugins/official-plugins.test.ts
@@ -107,6 +107,7 @@ describe("official plugin registry invariants", () => {
       "provider-claude-code": "Agent interaction",
       "provider-codex": "Agent interaction",
       "provider-pi": "Agent interaction",
+      "provider-refusal-fallback": "Agent interaction",
       "provider-retry": "Agent interaction",
       secrets: "Developer tools",
       "side-chat": "Agent interaction",
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -810,6 +810,27 @@ banner or with `bb provider-retry cancel <thread-id>`. Run
 `bb provider-retry retry <thread-id>` for a manual recovery, including credit
 or spend-control limits that do not report a reset time.
 
+### Refusal fallback plugin
+
+The builtin Refusal fallback plugin is enabled on fresh installations. When a
+provider's safety policy refuses a message and the provider has no fallback of
+its own, bb offers the models listed below the refused one in that provider's
+catalog, switches the thread to the chosen model, and sends one agent-only
+`Please continue.` turn. Declining leaves the thread on the refused model.
+Because only lower catalog entries are offered, a repeatedly refused thread
+runs out of alternatives instead of switching forever. Disable it under
+Extensions → Plugins or with `bb plugin disable provider-refusal-fallback`.
+
+Ticking "Switch automatically next time, do not ask again" stores the chosen
+model for that provider and skips the prompt afterwards. The choice is per
+provider, not per thread. Inspect and clear it with:
+
+```bash
+bb refusal-fallback status [--json]
+bb refusal-fallback forget <provider-id>
+bb refusal-fallback retry [thread-id]
+```
+
 ### Workflows plugin
 
 The builtin Workflows plugin is disabled on fresh installations. Enable it
--- a/packages/plugin-api-map/src/plugin-icons.ts
+++ b/packages/plugin-api-map/src/plugin-icons.ts
@@ -16,6 +16,7 @@ import {
   MessageAdd02Icon,
   MessageQuestionIcon,
   SmartPhone01Icon,
+  SecurityCheckIcon,
   SourceCodeIcon,
   SparklesIcon,
   TerminalIcon,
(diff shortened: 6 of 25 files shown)


t1mdurden:
Pushed 73563df — service-level coverage for the gap I flagged in **How you verified**: the classifier and the renderer were tested, but nothing exercised the orchestration between them.

`plugins/provider-refusal-fallback/server.test.ts` now drives `RefusalFallbackService` through `createFakePluginHost`, so the prompt payload, the KV round-trip, and the `sdk` calls are all real:

- accept → prompt offers Opus 4.8 / 4.7 / Sonnet 5 in catalog order, then `threads.update` + `threads.send`, and nothing is remembered
- accept and remember → `autoChoice` persists, and a second refusal on a new turn switches with **zero** interactions raised
- decline → no `threads.update`, no `threads.send`
- refused model at the bottom of the catalog → `no-alternative`, no prompt
- reconciling the same turn twice → `already-handled`, one send

`pnpm exec turbo run typecheck test --filter=bb-plugin-provider-refusal-fallback` → `17 passed`, typecheck clean.

Still not verified, unchanged from the PR body: no live API refusal has been observed end to end, and the Codex `policy` path is reasoned-about rather than tested.

> AGENT GENERATED

SawyerHood:

PR #2739, opened 2026-08-30
Per-plugin toggle for exposing CLI commands to agents (plugin-commands skill curation)

swairshah:
## Problem

Every plugin that registers a CLI shows up in the server-generated `plugin-commands` skill, so agents see all plugin commands. As the plugin count grows this becomes noise: not every plugin's commands are useful to agents, and today the only way to hide them is disabling the plugin entirely.

## Change

Adds a per-plugin **agent CLI exposure** toggle (default: exposed, preserving current behavior). A hidden plugin keeps a fully working CLI — its commands are just omitted from the generated `plugin-commands` skill, so agents never discover them.

- **DB**: `plugins.agent_cli_exposed` boolean column (migration `0110_plugin_agent_cli_exposed`), preserved across reinstalls/updates and state-snapshot rollbacks.
- **Server**: `cliContributions()` filters hidden plugins before writing the skill; `PluginService.setAgentCliExposed()` re-syncs the skill and broadcasts `plugins-changed`. New route `POST /api/v1/plugins/:id/agent-cli` with `{ exposed: boolean }`.
- **CLI**: `bb plugin agent-cli <id> [on|off]` shows or sets the toggle; `bb plugin list` marks hidden commands with "(hidden from agents)".
- **SDK**: `plugins.setAgentCliExposed()`; `InstalledPlugin.agentCliExposed` (schema default `true`, so older servers parse fine).
- **App**: an "Agent access" section with a switch on the plugin detail page, shown when the plugin registers a CLI command.
- **Docs**: bb-cli command index + plugins reference, plugin guide template, and the plugin-authoring skill's CLI chapter.

## Tests

- New server test: toggling off removes the plugin's section from the skill (and removes the skill file when no contributions remain), the command stays runnable, toggling back regenerates it, plus 400/404 request validation.
- Migration rewind tests updated with a `drop agent_cli_exposed` helper, mirroring the existing pattern.
- Full `turbo run typecheck` (80 tasks) passes; db/sdk/cli/server plugin suites and app plugin component tests pass. The only failing test locally is `install-machine-script.test.ts`, which also fails on a clean `main` checkout (environment-dependent).

### DIFF
--- a/apps/app/src/components/dialogs/AddMachineDialog.test.tsx
+++ b/apps/app/src/components/dialogs/AddMachineDialog.test.tsx
@@ -60,6 +60,7 @@ function connectPlugin(
     version: "0.1.0",
     provenance: "builtin",
     isOrphanedBuiltin: false,
+    agentCliExposed: true,
     publisherLabel: "BB Official",
     sourceDisplay: "builtin · connect",
     updateState: {},
--- a/apps/app/src/components/plugin/PluginSettings.test.tsx
+++ b/apps/app/src/components/plugin/PluginSettings.test.tsx
@@ -203,6 +203,7 @@ function rowPlugin(
     rootDir: "/plugins/linear",
     version: "0.1.0",
     enabled: true,
+    agentCliExposed: true,
     status,
     statusDetail: null,
     description: null,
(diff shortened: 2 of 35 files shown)


SawyerHood:

Issue #1451, opened 2026-08-13
Support multiple desktop servers and remote machine handoff


SawyerHood:
## Problem

A desktop application can connect to another machine as its main server and then lose local thread creation. Users also lack a direct handoff from local development to remote testing.

## Requested behavior

Support desktop applications that each own a local server while they also connect to remote machines.

Add a clear handoff workflow for remote testing and quality checks. Avoid confusion between a local desktop server and an earlier headless enrollment.

## Sources

- Multiple desktop machines: https://discord.com/channels/1522478522332090528/1536517505047207946
- Remote machine handoff: https://discord.com/channels/1522478522332090528/1535636404497809459/threads/1536478928091222156

> AGENT GENERATED: by GPT-5

Issue #483, opened 2026-07-02
Feature request: Remote session support


Git-on-my-level:
If you support this I will love you guys forever and adopt bb as my main agent control plane

Would love to control other bb installs on other machines over SSH similar to what codex does today

SawyerHood:
100% agree we should do this! We actually have most of the code to do this already, we just haven't put in the elbow grease to make it good yet!

jshph:
Been daily driving BB the last 3 weeks and using the web app (through Tailscale on a home server) on my phone + laptop has become my main way to drive agents. It's working great so far! thanks for the amazing work

georgecollier-nqu:
Yep almost life changing for me haha! I hate the shitty mobile apps Codex/Claude have, I prefer web, can run everything on a VPS.

SawyerHood:
I'm clearly in need of this feature because I run 3 bb's at a time lol.

<img width="261" height="90" alt="Image" src="https://github.com/user-attachments/assets/967d5749-82ce-4881-8465-839395e56398" />

Question for @Git-on-my-level / @georgecollier-nqu / @jshph. How would you expect to "connect" your hosts together. I think the most natural thing is you could be using the desktop / local version and add your remotes there. The issue is that one of the servers has to be the "primary one" ex the one that you access from your phone / computer. This one should also be responsible for owning all of the automations / config / plugins (in the future). What is odd that this should probably be the remote one!

Curious if you all have opinions. My current thought is something like this:
- In the desktop app add the ability in the UI to switch the url it points to (you can point it to the remote)
- You can add multiple ssh hosts and they will all be "targets" when you make a thread.

This solves my pain of I can use the desktop app on my main laptop while the ui points to my headless machine and I can add my other computers as targets.

jshph:
Targets are a great concept -- question of repos getting out of date with each other is something to think thru though (but promising if done well). If i'm iterating on my VPS because I'm mobile, how does my laptop's BB (local) inherit the changes? What tree of threads is shown in the project folder -- remote or local?

That said I also run several BBs (have an ios home screen to both laptop and vps too). I think it's fine for now...

harveyfullstack:
+1 on this! If you take a look at https://multica.ai, they solve this problem and you can initialize a background daemon per machine. They have the main interface hosted themselves and you can easily switch between workspaces and machine-served agents (I use this right now). There's quite a bit in their docs to take from it and you can self-host all of it yourself. Perhaps BB can take a similar approach here?

[...]

PR #2740, opened 2026-08-30
Surface the desktop server switcher and allow a different server per window

bhushit:
## What was wrong

bb desktop already lets you point the app at a remote bb server (`Window ▸ Server`, `Window ▸ Set Server URL…`), but it has three gaps:

1. **Undiscoverable** — the only entry point is a native menu most users never open.
2. **One custom URL at a time** — the `custom` target holds a single URL; setting a new one replaces it.
3. **App-global + hijacking** — the target is global, so *all* windows are forced onto the same server, and switching a window to a remote replaces its content. If that remote needs sign-in, you can get **stranded** on its gate page with no in-app way back.

Refs #1451 (multiple desktop servers / remote handoff) and #483 (the original remote-session request the switcher shipped from).

## What changed

All behind a new `remoteUi` experiment (default off):

- **Surface the switcher in-app** — a sidebar-footer indicator + popover, and a **Connection** settings page, so switching/adding servers no longer requires the native menu.
- **User-curated server list** — replace the single custom URL with an add/remove allowlist (`server-target.ts`, v2 schema + migration); auto-discovered getbb Connect machines list alongside.
- **A different server per window** — selecting a remote opens it in **its own window** (`openRemoteServerWindow`), leaving This Mac in place; closing the window returns you. Per-window Connect session; auth failure falls through to load the server so its gate can prompt sign-in instead of a fatal error — you can never be stranded.
- **Security parity with the native menu** — every server-target mutation (switch *and* edit) is gated to the **builtin/This-Mac origin** (`isBuiltinServerOrigin`), and the switcher only renders on the This Mac window. A remote or compromised page can neither *drive* nor *display* it.
- Custom getbb origins authenticate through Connect (fixes the existing custom-path sign-in dead-end).

New `BbDesktopApi` members are `experimental_`-prefixed with `docs/api_to_audit.md` entries. No `HOST_DAEMON_PROTOCOL_VERSION` change (no server↔daemon wire touched).

## How you verified

- `pnpm exec turbo run typecheck test lint` green across `@bb/app`, `@bb/desktop`, `@bb/desktop-contract`, `@bb/domain` (0 errors).
- Unit tests added: builtin-only origin guard (getbb/remote origins rejected for switching), v1→v2 migration, switcher hidden when not on This Mac, Connection add/remove + disabled-on-remote.
- Manual (dev desktop): opened a remote in its own window, closed it to return, confirmed a remote page can't render or drive the switcher, and This Mac switching works.

Refs #1451

> AGENT GENERATED


SawyerHood:

PR #2752, opened 2026-08-31
Replace the compact right drawer with full-page tool tabs

brsbl:
## Stack outcome relative to `main`

The compact right panel no longer rises as a dimmed 92dvh bottom drawer. The stack replaces it with a right-side system: Info uses a compact shelf that translates the live page left, while every tool tab—including New Tab—expands from the right to a full-width surface. New Tab also selects and renders its own production content instead of leaving the prior Info content visible.

Layer 1 also contains the compact-home/recents redesign, sticky `Recent` header, left-sidebar shelf, route-progress feedback, cold-navigation placeholders, provider-health caching, and immutable provider-logo caching. This PR is the focused second layer that adds full-page non-Info tabs and fixes the New Tab selection/content transition.

## What was wrong in this layer

Layer 1 moved the right panel from the bottom drawer to a right shelf, but all tabs were still constrained to the shelf's 298.67px width. Selecting New Tab could change the tab strip without changing the rendered Info content, leaving file search and other tool surfaces without a usable phone-width canvas.

## What changed

- Info remains a 298.67px right shelf so the underlying page stays visible.
- Every non-Info tab transitions to a 393px full-page surface above fixed app chrome.
- New Tab now selects and renders the production New Tab actions after its loading skeleton settles.
- Opening the left sidebar closes the right panel so the two compact shelves cannot compete for the page.
- Compact shelf stories mount the real `CompactSecondaryPanelShelf`, `ThreadSecondaryPanel`, `FilePreview`, `RootComposeCompactHome`, `SidebarProvider`, and `SidebarInset` composition; no substitute file markup or generic product wrapper remains.

No wire changes; `HOST_DAEMON_PROTOCOL_VERSION` is unchanged.

## Screenshots

Chrome for Testing 151.0.7922.71, 393×852 at DPR 2, using the same dev database, route, viewport, interaction, and settled-content state.

### Cumulative behavior versus `main`

Before is the stack merge base (`f4bbc2fe8`); after is this PR head (`6d031d265`).

| `main` — New Tab in a dimmed bottom drawer | Layer 2 — New Tab as a full-page surface from the right |
| --- | --- |
|  |  |

### This layer's New Tab fix

Before is the final parent head (`04d18307f`); after is this PR head (`6d031d265`).

| Parent — New Tab remains in the 298.67px shelf and Info content remains visible | This layer — New Tab expands to 393px and renders its production actions |
| --- | --- |
|  |  |

## How you verified

- On the exact merge base, the settled production New Tab dialog measured `top=68.17`, `width=393`, `height=783.83` with an 80% black backdrop—the old bottom-drawer presentation.
- On the exact PR head, the same production flow measured parent `state=open`, `width=298.671875`; head `state=full`, `width=393`, with `new-tab-actions` present after the content skeleton cleared.
- Ladle was restarted on port 61002 and the compact shelf stories were exercised at 393×852: closed returns the page to `left=0`; shelf translates the production page by `-298.68px`; Full page fills 393px; Many tabs selects `sidebar.tsx` and renders real `FilePreview` controls.
- Every affected story ID loaded through its production component subtree with zero runtime exceptions. All remote CI gates pass; no local CI-equivalent command was run.

BB-Thread-ID: thr_wftu7bh9ez

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/plugin/PluginPanelRightPanelHost.tsx
+++ b/apps/app/src/components/plugin/PluginPanelRightPanelHost.tsx
@@ -21,7 +21,10 @@ import { Tooltip, TooltipContent, TooltipTrigger } from "@bb/shared-ui/tooltip";
 import { useAppCommandHandler } from "@/components/commands/AppCommandProvider";
 import { PluginIcon } from "@/components/plugin/PluginIcon";
 import { PluginSlotMount } from "@/components/plugin/PluginSlotMount";
-import { RIGHT_PANEL_TOGGLE_ICON_NAME } from "@/components/secondary-panel/panelToggleControlState";
+import {
+  getCompactPanelPresentation,
+  RIGHT_PANEL_TOGGLE_ICON_NAME,
+} from "@/components/secondary-panel/panelToggleControlState";
 import { SecondaryPanelLayout } from "@/components/secondary-panel/SecondaryPanelLayout";
 import {
   LazyBrowserTabDeck,
@@ -971,6 +974,7 @@ export function PluginPanelRightPanelHost({
         mainPanelId={`plugin-panel-main-${panelHostId}`}
         main={children}
         composerHost={null}
+        compactPresentation={getCompactPanelPresentation(activeTab?.kind)}
         renderPanel={renderPanel}
       />
     </div>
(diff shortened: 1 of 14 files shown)


brsbl:
Superseded by focused native stack #2763: #2758 → #2759 → #2760 → #2761 → #2762. Replacement layer #2762 has cumulative content byte-for-byte identical to this PR head; review should continue on the focused stack.

PR #2767, opened 2026-08-31
Retry sidebar bootstrap until a warming server answers

### LINKED ISSUE #2533: Desktop cold start can leave the shell on a terminal "Failed to load projects." until an unrelated realtime event refetches
## Summary

On a desktop cold start the app can replace its whole shell with `Failed to load projects.` and stay there. The sidebar bootstrap is the shell's only read, it retries for about half a second, and once it has failed nothing refetches it except an unrelated realtime invalidation. The window recovers on its own seconds or minutes later, whenever some thread or project event happens to fire, which reads as "bb says I have no projects, then it fixes itself".

Expected: a bootstrap that fails while the server is still warming up retries until the server answers, and the shell never presents a transient startup failure as a terminal state.

## Versions and environment

- bb 0.40.0, desktop app, macOS 15.5 (Darwin 25.5.0)
- Source repro built from `main` at `ad79bbb5ec909524f8f281e62d860c588a86f332`, Node 22.22.3
- Local server on the default port, default data dir (`~/.bb`), no Connect, no remote host
- Not provider specific: this is the app shell, before any agent is involved

## Steps to reproduce

Deterministic repro against a running bb (Playwright, no bb code changes). It injects the failure the cold start produces naturally, an unreachable server for a few seconds while the renderer is already up:

```js
const t0 = Date.now();
const log = [];
await page.route('**/api/v1/sidebar-bootstrap*', async (route) => {
  const dt = Date.now() - t0;
  if (dt < 8000) { log.push(['fail', dt]); await route.abort('connectionrefused'); }
  else { log.push(['ok', dt]); await route.continue(); }
});
await page.goto('http://127.0.0.1:38886/', { waitUntil: 'domcontentloaded' });
for (let i = 0; i < 6; i++) {
  await page.waitForTimeout(5000);
  console.log(Date.now() - t0, (await page.locator('body').innerText()).includes('Failed to load projects'));
}
```

1. Run the snippet against a healthy bb server. The server is reachable again after 8s.
2. The app shows `Failed to load projects.` and never recovers on its own.
3. Focus, `visibilitychange`, `pageshow` and `online` do not recover it either (verified: zero further requests).

Did not reproduce with a single failed request: one 503 followed by a healthy server recovers, because a second consumer of the same query key fetches again shortly after mount. The failure needs to outlast the retry budget, which any cold start on a large data dir does.

## Expected vs actual

```text
Actual, request log (ms since load) and screen state:

  fail 385
  fail 655
  fail 925          <- last attempt, 540ms of retry budget spent
  (no further requests at all)

  t= 5452ms  "Failed to load projects."
  t=10460ms  "Failed to load projects."
  t=15464ms  "Failed to load projects."   <- server healthy since 8000ms
  t=20469ms  "Failed to load projects."
  t=25474ms  "Failed to load projects."
  t=30494ms  "Failed to load projects."

Rendered body at that point is the nav chrome plus the error line, no projects,
no threads:

  New thread / Extensions / Automations / GitHub / Kanban / Docs / Tasks /
  Settings / Remote access / Report a bug
  Failed to load projects.

Expected: retries continue while the server is unreachable, and the shell
renders as soon as the server answers.
```

## Evidence

Cause, three parts, all on `main` at `ad79bbb`:

1. The bootstrap runs on the default read-retry budget: two retries at a fixed 250ms delay, and only for transport errors. Any `HttpError` / `BbHttpError`, so any HTTP status the warming server returns, gets zero retries.
   - https://github.com/get-bb/bb/blob/ad79bbb5ec909524f8f281e62d860c588a86f332/apps/app/src/hooks/queries/query-helpers.ts#L10-L11
   - https://github.com/get-bb/bb/blob/ad79bbb5ec909524f8f281e62d860c588a86f332/apps/app/src/hooks/queries/query-helpers.ts#L62-L93
   - https://github.com/get-bb/bb/blob/ad79bbb5ec909524f8f281e62d860c588a86f332/apps/app/src/lib/query-client.ts#L158-L159
2. The query is on the static-cache tier, `staleTime: Infinity`, with no retry override.
   - https://github.com/get-bb/bb/blob/ad79bbb5ec909524f8f281e62d860c588a86f332/apps/app/src/hooks/queries/sidebar-navigation-query.ts#L34-L60
   - https://github.com/get-bb/bb/blob/ad79bbb5ec909524f8f281e62d860c588a86f332/apps/app/src/hooks/queries/query-policies.ts#L62-L64
3. The failure is terminal in the UI, and the usual escape hatches are closed: focus and reconnect refetches are gated off while the socket is connected, and the error screen has no retry affordance. Only a realtime invalidation of the sidebar key brings the app back.
   - https://github.com/get-bb/bb/blob/ad79bbb5ec909524f8f281e62d860c588a86f332/apps/app/src/views/RootComposeView.tsx#L1908-L1920
   - https://github.com/get-bb/bb/blob/ad79bbb5ec909524f8f281e62d860c588a86f332/apps/app/src/main.tsx#L31-L38

Why a cold start lands in that window, from `~/.bb/logs/server.1.log` on the machine where this was first noticed (data dir at 15 MB `bb.db` plus a 7.5 MB WAL):

```
11:58:40.999  Server listening
11:58:42.706  last plugin loaded          (+1.7s)
11:58:47.025  [plugin:github] synced 240 item(s) across 3 repo(s)   (+6.0s)
```

The desktop probe only waits for `/api/v1/system/config`, so the renderer can issue its bootstrap during that warm-up. Numbers above are wall clock from the log lines; the request timings are `Date.now()` deltas inside the interceptor.

## What you ruled out

- Not a duplicate of #2480: that one is the pre-window launcher probe reporting "Could not start bb". This one is after the window loads and the shell is up.
- Not #1302 or #1676: those are bootstrap payload size and reload flashes, not a failed bootstrap left terminal.
- Not data loss: the projects are in the database the whole time, `curl http://127.0.0.1:38886/api/v1/sidebar-bootstrap` returns them in 9ms while the window shows the error.
- Not focus related: focus, visibility, pageshow and online events produce zero refetches in this state, by design (`shouldRefetchOnWindowFocus` is gated on the socket being disconnected).

## Suggested fix

Give the app shell's bootstrap its own retry policy instead of the generic read budget: treat a warming server (5xx, 408, 429) as retryable alongside transport failures, and back off exponentially to a 4s cap over six retries, about twelve seconds of coverage.

Patch, tests, and end-to-end verification: https://github.com/mariomile/bb/compare/main...mariomile:bb:fix/sidebar-bootstrap-startup-retry

Verified with the repro above against the patched app: attempts at 821, 824, 1082, 1585, 2591, 4600ms, success at 8603ms, and the shell renders projects without ever showing the error. Unit and hook tests added; the hook test fails on `main` and passes with the change. `tsc --noEmit` and `oxlint` clean on the touched files.

Two things deliberately left out, since they are UI changes and CONTRIBUTING asks for an issue first:

- A retry affordance on the error screen. Today a user whose bootstrap fails past the retry window has no way to ask for another attempt, short of reloading the window.
- Refetching a bootstrap that has no data when the realtime socket connects, which would make any future variant of this self-healing regardless of the retry budget.

Happy to open a PR for the retry policy alone, or to include either of the above if you want them in the same change.

Medium. Hits every user whose server is slow to warm up (large data dir, cold page cache, slow disk), no in-app workaround other than reloading the window, no data lost. Effort Low for the retry policy.

> AGENT GENERATED

Investigation thread (local bb): `thr_xa2mu3c56k`, project `proj_personal`.

Flame119052:
## What was wrong

Desktop cold start can open the window as soon as `/health` answers, while `GET /api/v1/sidebar-bootstrap` still fails. That query used the generic read retry (two attempts, 250ms, transport-only), so an HTTP 5xx from a warming server got zero retries. After the budget ran out the query was terminal (`staleTime: Infinity`, no focus/reconnect refetch while the socket is connected), and the compose error screen had no Retry. The shell stayed on "Failed to load projects." until an unrelated realtime invalidation. See [#2533](https://github.com/get-bb/bb/issues/2533) and https://get-bb.github.io/reports/issues/2533.html.

## What changed

- `query-helpers.ts`: sidebar-specific retry for transport failures plus 408/429/5xx, exponential backoff from 250ms capped at 4s, keep retrying while the error is retryable. Do not retry 401/403/404, parse errors, or `AbortError`.
- `sidebar-navigation-query.ts`: apply that policy on every observer of the sidebar bootstrap key so a later mount cannot overwrite it with the generic two-retry default.
- `RootComposeView.tsx`: add a Retry button on the "Failed to load projects." screen that refetches the bootstrap.

No `HOST_DAEMON_PROTOCOL_VERSION` bump, no CLI/guide changes. Deviation from the issue's original six-retry sketch: retryable errors keep retrying at the 4s cap instead of stopping after ~12s, matching the expected "retries continue while the server is unreachable" behavior. The Retry button was listed as a follow-up in the issue and included here because the confirmed-repro report asked for it.

## How you verified

- Added unit tests for retry classification, infinite retry vs the generic two-retry budget, and backoff.
- Added hook tests: recover after six `Failed to fetch` failures; retry a temporary 503; do not retry a 401.
- `pnpm exec turbo run test --filter=@bb/app -- src/hooks/queries/query-helpers.test.ts src/hooks/queries/sidebar-navigation-query.test.tsx` — 25 passed (red before the policy, green after).

Fixes #2533

> AGENT GENERATED

Made with [Cursor](https://cursor.com)

### DIFF
--- a/apps/app/src/components/promptbox/NewThreadComposer.tsx
+++ b/apps/app/src/components/promptbox/NewThreadComposer.tsx
@@ -135,6 +135,7 @@ export interface NewThreadComposerState {
   projects: readonly SidebarProject[] | undefined;
   sidebarNavigation: SidebarBootstrapResponse | undefined;
   sidebarNavigationError: boolean;
+  retrySidebarNavigation: () => void;
   currentProject: SidebarProject | undefined;
   projectSources: SidebarProject["sources"];
   connectedHostIds: ReadonlySet<string>;
@@ -373,6 +374,9 @@ export function NewThreadComposer({
   const promptBoxRef = useRef<PromptBoxHandle>(null);
 
   const sidebarNavigationQuery = useSidebarNavigation();
+  const retrySidebarNavigation = useCallback(() => {
+    void sidebarNavigationQuery.refetch();
+  }, [sidebarNavigationQuery.refetch]);
   const projects = useMemo(
     () => sidebarNavigationQuery.data?.projects.map(stripProjectThreads),
     [sidebarNavigationQuery.data],
@@ -1425,6 +1429,7 @@ export function NewThreadComposer({
     projects,
     sidebarNavigation: sidebarNavigationQuery.data,
     sidebarNavigationError: sidebarNavigationQuery.isError,
+    retrySidebarNavigation,
     currentProject,
     projectSources,
     connectedHostIds,
--- a/apps/app/src/hooks/queries/query-helpers.test.ts
+++ b/apps/app/src/hooks/queries/query-helpers.test.ts
@@ -38,7 +38,16 @@ import {
 import { createQueryClientTestHarness } from "@/test/queryClientTestHarness";
 import { HttpError } from "@/lib/api";
 import { BbHttpError } from "@/lib/sdk";
-import { isTransientReadError, requireEnabledQueryArg } from "./query-helpers";
+import {
+  isSidebarBootstrapRetryableError,
+  isTransientReadError,
+  requireEnabledQueryArg,
+  shouldRetrySidebarBootstrapQuery,
+  shouldRetryTransientReadQuery,
+  sidebarBootstrapRetryDelay,
+  SIDEBAR_BOOTSTRAP_RETRY_BASE_DELAY_MS,
+  SIDEBAR_BOOTSTRAP_RETRY_MAX_DELAY_MS,
+} from "./query-helpers";
 
 describe("requireEnabledQueryArg", () => {
   it("returns the value when present", () => {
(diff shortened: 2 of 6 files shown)


SawyerHood:

PR #2771, opened 2026-08-31
Fix Plugin SDK declaration build memory spikes

MacHatter1:
## What was wrong

The Plugin SDK declaration bundler sized its worker pool only from available CPU parallelism. When separate forced Turbo graphs ran concurrently, each graph rebuilt the shared declaration task and each build spawned a full worker pool. Every worker creates an independent Rollup/TypeScript declaration program, so the duplicated pools multiplied memory use until Node processes were killed for running out of memory.

## What changed

- Size the declaration worker pool from bundle count, available CPU parallelism, and physical memory, budgeting one worker per 8 GiB.
- Cover the adaptive calculation for machines with different CPU and memory capacities.
- Tell agents to combine package filters into one Turbo invocation instead of launching overlapping forced graphs.
- No host-daemon wire, CLI, or Plugin SDK API change.

## How you verified

- Confirmed the new adaptive test failed against the fixed worker-count implementation: a 12-core, 24 GiB host returned 2 instead of 3.
- pnpm exec turbo run test typecheck --filter=@get-bb/plugin-sdk --force — passed 6/6 tasks, 23 test files, and 222 tests.
- pnpm exec oxfmt --check AGENTS.md packages/plugin-sdk/scripts/build-bundled-dts.mjs packages/plugin-sdk/scripts/build-bundled-dts-concurrency.mjs packages/plugin-sdk/scripts/build-bundled-dts-concurrency.test.mjs — passed.
- git diff --check — passed.
- Live host calculation: 12 available workers and 24 GiB physical memory select 3 declaration workers.

Fixes: N/A

> AGENT GENERATED

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -59,6 +59,7 @@
 
 ## Testing
 
+- Combine package filters into one Turbo invocation instead of running overlapping `--force` graphs in parallel; shared prerequisite tasks otherwise execute once per process and can multiply CPU and memory use.
 - Only write high quality tests that verify where there could be potential bugs. Avoid testing trivial getters/setters, framework wiring, or other code that is unlikely to break.
 - Pipe slow test output to a file, then read the file. Example: `pnpm exec turbo run test --filter=@bb/integration-tests --force > /tmp/test-out.txt 2>&1`.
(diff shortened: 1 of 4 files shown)


SawyerHood:

PR #2779, opened 2026-08-31
Dispatch queue rework: scheduled sends, concurrency limits, one plugin checkpoint

ymichael:
## What was wrong

bb had no way to put anything between "a user asked for work" and "bb runs it": no scheduled sends, no concurrency control, and four parallel single-purpose parking mechanisms (`dispatch_holds` never existed on main, but `deferred_thread_messages`, the reprovision-parked turn, and the drain-only queue each solved one blocked-message case with its own machinery). Rate-limit retries were faked by `plugins/provider-retry` re-reading the whole event log (~450 lines) and injecting a synthetic "Please continue." message, so the provider never saw the same conversation twice. There was no plugin surface that could defer, refuse, or reschedule a dispatch.

## What changed

**The model: a send is always a dispatch attempt.** If nothing blocks it, it dispatches directly — the happy path is byte-for-byte unchanged and allocates nothing (test-guarded: a stock install registers zero hooks). If something blocks it, the message queues as a row carrying a typed `waitingOn` (`time` | `thread-busy` | `provisioning` | `interaction` | `host-offline` | `plugin`), and core re-attempts when conditions change (due sweep, thread-idle drain, workspace-ready, interaction-settled, plugin `recheck`, Send-now, orphan sweep). `deferred_thread_messages` is deleted; its cases are queue rows. Threads gain a canonical `pending` status (created, first message never dispatched) — scheduled spawns do zero provisioning until due.

**Plugin API** (all `experimental_`): `bb.experimental_hooks.on("message.dispatch", h)` — one checkpoint, run identically for fresh sends, drained rows, and steers, returning `proceed | wait(reason, sendAt?) | reject(message)`; fail-closed (10s box, plugin named). `bb.experimental_hooks.recheck(hook)` asks core to re-pose the question. New `bb.events`: `message.queued`, `message.dispatched`, `turn.failed` (ids + failure facts). Public API: `sendAt` on send/create, `threads.retry({turnRequestId?, sendAt?})` (re-dispatch decided by the provider's acceptance record: an input the provider never accepted is re-sent verbatim, an accepted turn is continued with an agent-only nudge; the original execution tuple is replayed and no user message is duplicated), `threads.listRunning()` (exact for cold admissions inside a hook — a `pending → starting` flip commits before the evaluation lock releases; a warm follow-up's `idle → active` flip lands in the send transaction just after it, so a burst of follow-ups to distinct idle threads can briefly under-report), `threads.count()`.

**Plugins**: new `scheduled-send` ("Send later…" via the composer's own submit pipeline + dialog; **off by default**) and `concurrency-limit` (global + per-host caps as one pure hook over `listRunning`; **off by default**); `provider-retry` rewritten to a `turn.failed` listener calling `threads.retry` — `recovery.ts` and the synthetic continue are deleted (plugin: 1,427 → ~300 lines).

**UI/CLI**: queued rows render their wait (kind icon, reason, live countdown, retry attempt, failure reason) with Send-now/Edit/Cancel; sidebar shows a clock for threads with waiting work and the error glyph when a queued dispatch failed; `bb thread queue` gains wait columns; new `--send-at` on spawn/tell, `bb thread count`, `bb thread retry`; guide + skills updated alongside.

**Wire**: `HOST_DAEMON_PROTOCOL_VERSION` → 176 (thread-event schema changes ride the daemon event batch). One drizzle migration (`0110`) on main's chain: queue wait columns + partial indexes, `threads.pending_start_context`, and the `deferred_thread_messages` drop. Plugin SDK 0.4.31 with `docs/api_to_audit.md` entries for every new `experimental_` member.

## How you verified

Repo-wide `turbo run typecheck` (82/82) and `turbo run lint --filter=@bb/app` (0 errors) at every layer; final battery across 15 packages (~7,900 tests: server 2094, app 453 files, cli, db incl. migration-replay + `EXPLAIN QUERY PLAN` index pins, domain, contracts, plugin-sdk, thread-view, client-core, three plugins, templates, plugin-api-map api-sync). Exit-criteria tests: `--send-at` survives restarts; exactly-once dispatch under racing drains; orphaned plugin waits clear; retries re-issue the original turn with no duplicated message; hook exactness (attempt N+1's `listRunning` sees attempt N's admission — load-bearing test fails if the commit moves outside the lock); zero-hook installs byte-identical. Integration suite: failures on the loaded build machine were verified as environment flakes — the same suite on a clean origin/main worktree failed 4× more files under identical load; all failing files pass in isolation.

> AGENT GENERATED


ymichael:
@slopcop

ymichael:
@slopcop

PR #2782, opened 2026-09-01
Refactor sidebar rows around shared rails

brsbl:
## What was wrong

Sidebar rows independently owned status, identity, action, and disclosure placement. That made equivalent icons drift between project, machine, worktree, section, and lifecycle rows; it also let hover actions compete with the thread title's natural open target. Even after the outer rails were standardized, rows with folder/worktree icons consumed an identity slot while plain thread rows skipped it, allowing a nested thread title to begin left of its parent label. The thread-list view switcher was icon-only and therefore did not communicate which organization mode was active.

## What changed

- Standardized tree rows on one presentation-only `SidebarRow` geometry: `status | body | actions | disclosure`.
- Kept status in one fixed left rail and disclosure carets in one fixed right rail across row types and collapsed states; identity and depth now belong to the body column.
- Reserved the same identity track in every tree-row body, including rows without an identity glyph, so title alignment depends only on depth rather than optional content.
- Reduced the shared nesting step from 24px to 12px so nested lists stay compact without moving either fixed rail.
- Moved the thread archive quick action into the fixed status rail and moved lifecycle hover actions into the action rail, avoiding hover-time title movement.
- Made disclosure carets persistent and quiet rather than changing row anatomy on hover.
- Updated the existing view trigger to identify the current organization mode with the matching Projects, Machines, or Custom icon plus a chevron, accessible label, and tooltip; menu behavior and labels are unchanged.
- Expanded the production `ProjectList` story matrix across Project, Custom, and Machine views at 320px and 240px widths.

## Before

_Screenshot upload pending._

## After

_Screenshot upload pending._

## How you verified

- Chrome for Testing 151.0.7922.71 against the exact branch web app at 1280×1000.
- Exercised Project, Machine, and Custom views; collapsed and expanded project/worktree groups; unread/status rows; thread hover/archive; hard reload and persisted view selection; and light, dark, and system-dark themes.
- DOM geometry inspection confirmed every rendered status rail at `x=16` with a 14px layout width and every rendered disclosure rail at `x=287` with a 24px layout width. Carets remained visible, row heights remained 24/28px, and the archive hover control occupied the status rail without shifting the title.
- Reproduced the optional-identity indentation defect in the live seeded sidebar, then recaptured the corrected state: project, worktree, and direct-thread titles use one body spine, while genuinely nested worktree threads advance by the shared 12px depth step.
- Confirmed the view trigger reports `View: Projects` to accessibility APIs and changes icon/label with the selected view. No runtime exceptions or console errors were observed in the changed flows.
- Remote CI run 33482308631 passed all app, package, integration, and server tests; build, typecheck, lint, version guards, and Linux/macOS package smoke checks.
- Mobile E2E was skipped by workflow path filtering because this layer does not change the compact drawer implementation.
- Local Ladle launch was attempted twice through the repository's story command but stopped at `EMFILE: too many open files, watch`; story registration therefore remains covered by source review and remote build/typecheck rather than a local rendered Ladle pass.

Fixes: N/A — stacked follow-up to the sidebar/navigation work.

BB-Thread-ID: thr_ccffp4w2p2

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/commands/AppCommandShortcutHint.stories.tsx
+++ b/apps/app/src/components/commands/AppCommandShortcutHint.stories.tsx
@@ -3,11 +3,13 @@ import { Button } from "@bb/shared-ui/button";
 import { Icon } from "@bb/shared-ui/icon";
 import { cn } from "@bb/shared-ui/lib/utils";
 import type { AppShortcutPresentation } from "@/lib/app-keybindings";
+import { SIDEBAR_ROW_INTERACTIVE_STATE_CLASS } from "@/components/sidebar/sidebarRowClasses";
 import {
-  SIDEBAR_ROW_BASE_CLASS,
-  SIDEBAR_ROW_INTERACTIVE_STATE_CLASS,
-  SIDEBAR_STANDARD_ROW_PADDING_CLASS,
-} from "@/components/sidebar/sidebarRowClasses";
+  SidebarRow,
+  SidebarRowAccessory,
+  SidebarRowContent,
+  SidebarRowIdentityRail,
+} from "@/components/sidebar/SidebarRow";
 import { StoryCard, StoryRow } from "../../../.ladle/story-card";
 import { AppCommandShortcutPill } from "./AppCommandShortcutHint";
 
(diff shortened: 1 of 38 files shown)


brsbl:
Superseded by the separate draft experimental row-system layers #2859 and #2861.

PR #2788, opened 2026-09-01
Page timelines by slicing one canonical projection

ymichael:
## What was wrong

Timeline summary pages were raw event windows — up to 20 user-message anchors, cut by a 1,500-event budget (`BB_FF_TIMELINE_WINDOW_EVENT_BUDGET`) and a 4 MiB byte budget — but the projection is a full-history state machine, so `project(A) + project(B) ≠ project(A + B)`. Every window needed context recovery (nine `ensure*` passes), ownership rules, and "loaded but don't render" channels, and rows still fell through the seams. Replayed against a copy of production (1,132 threads), **21 threads failed exact page recombination** (2 at production settings) — user messages, assistant replies, provisioning/error rows lost, or rows duplicated at page boundaries — and 54 threads couldn't produce an unpaginated reference at all. Items straddling a cut rendered twice with conflicting statuses. Separately, expanding an oversized completed turn 413'd once both exact-range reads exceeded the response limit. And windowing did not even bound cost: the worst recorded stall was a 6 s *windowed* build (2.6 s window query + 1.5 s fetch).

Supersedes #2504, #2464, and #2419 (all closed).

## What changed

**Never cut the event stream; project the thread once and paginate the projected rows.**

- `apps/server/src/services/threads/timeline.ts` + `timeline-pagination.ts`: every summary page reads all timeline-relevant events, projects once, and slices the rows into logical segments at user-message anchors. Cursors keep the wire format (anchor row id + seq) and resolve against the recomputed segments. Deleted: window resolution, budget floors, sequence/byte cursors, whole-item closure for summary pages, head-state/turn-lifecycle backfills, page-suffixed row ids, the `eventBudget` option, and `BB_FF_TIMELINE_WINDOW_EVENT_BUDGET` end to end (domain schema, config env var, app atoms, docs).
- `timeline-projection-cache.ts` (new): the canonical projection is memoized per thread, keyed by the newest event's id + the thread's event count — appends, suffix replacement (ids are never reused), and pruning invalidate structurally; the in-place output-truncation sweep now reports the threads it rewrote and only those are dropped. One build serves every page, poll, and delta of an idle thread.
- **No build blocks the event loop.** The thread-view fold body is now a per-event function with two drivers: the synchronous one and a cooperative one that yields every 500 events (`buildThreadTimelineFromEventsCooperatively`, proven byte-identical on real threads and asserted per thread by the corpus gate). The route builds cooperatively: chunked read + decode with yields, in-flight dedupe per cache key, tip re-verification against suffix replacement. Slices measured ≤20 ms on the 64k-event thread.
- **Large threads are projected ahead of any request.** `timeline-warmup.ts` projects every thread with 1,000+ stored events once provider registrations settle (the projection is keyed on provider display names and commands that plugins register), largest first, sliced; re-projects a thread when it settles (`turn/completed` → idle); and rebuilds in the background the threads an in-place output-truncation sweep touched while the existing projection keeps serving. Settled threads persist their projection (`thread_timeline_checkpoints`, migration `0111`; app-version + options key, tip id/sequence/count as columns; releases, edits, and pruning invalidate). The timeline route awaits registration settlement too, so a request in the first seconds after a restart cannot build or persist under an unsettled key. Candidate discovery is a per-thread indexed probe, not a table scan. Builds without an app version (tests, corpus harness) never persist or read back.
- **Refresh throttle** (`timeline-refresh-throttle.ts`): after a build over 100 ms, the same response is served for four build-costs of wall time (cap 10 s) before rebuilding, so one expensive streaming thread is bounded to ~25% of the event loop; cheap threads always refresh immediately.
- **Paginated turn details** (contract from #2464): `GET /threads/:id/timeline/turn-details` with an opaque cursor bound to the rendered row's exact range; first request reuses the exact-range path, only the former 413 case pages. Web app uses `useInfiniteQuery` + "Load more work". No summary-page id-splitting is needed anymore, the CLI's older-page concat is now correct as-is (pages are disjoint), and #2464's mobile hunks are obsolete (mobile renders via the webview shell). Plugin SDK → 0.4.31; the Plugin Guide's SDK inventory is regenerated (`update:sdk-inventory`). The method is named like its existing sibling `threads.timelineTurnSummaryDetails` (no `experimental_` prefix) — flagging that explicitly in case you'd rather it ship prefixed.
- `packages/db`: deleted the orphaned window-query layer (budget floors, anchor listing/lookup, cut guards, cursor checks; net −1,022 lines with tests); renamed `listRecentStoredEventRows` → `listStoredThreadTimelineEventRows`; added `getLatestStoredEventTip`, `listThreadIds`, `readStoredTimelineWindowForwardPage`, checkpoint records; the truncation sweep returns affected thread ids. `provider/rateLimits/updated` (16% of the largest thread's rows, never projected) is excluded from timeline reads.
- Repeatable audits: `scripts/provider-corpus/extract-corpus.ts` mints a corpus from a prod DB copy; `page-recombination.test.ts` is the pagination invariant as a corpus gate; `allowlists/timeline-pagination-row-classes.json` classifies this PR's intended row changes. docs/debugging-and-qa.md documents both, plus the safe way to run a server on a production copy (strip the Connect credential — a copied `bb.db` otherwise dials the relay as the production bb). docs/configuration.md describes the new pagination and its cost bounds.

No server↔daemon wire change, so `HOST_DAEMON_PROTOCOL_VERSION` is unchanged.

## How you verified

**Correctness — the invariant.** Walking every page and concatenating the results must exactly equal the unpaginated projection, for any segment limit. `page-recombination.test.ts` enforces this over the corpus (1,132 prod threads × segment limits {1, 2, 3, 8, 20}): **red on main (21 threads fail, 2 at production settings), green after (1,132/1,132, ~2.5 min)**. Unit tests cover cursor-index slicing, prelude reachability, stale-cursor 400s, and the walk/recombination property.

**Row-level diff vs main, fully classified.** Row snapshots minted from main on the same corpus: 995/1,132 threads byte-identical; every change on the other 137 classifies into a named fix class (seam rows restored ×1,947; byte-page fragments consolidated ×1,599; bisected items consolidated ×506 — including items that rendered `completed` on one page and `interrupted` on another; turn fragments resegmented ×268; whole-turn bounds ×266; one context-shell delegation rendered truthfully as the agent-resume tool call it is; page-shape realignment). Full corpus gate green (2,319 tests) with `BB_PROVIDER_CORPUS_ROW_CLASSES`; zero unclassified changes.

**Performance.** Live-server audit over all 1,132 prod threads (first request per thread, repeat request, full older-page walk), main vs branch on the same machine and DB copy, Connect stripped from the copies. The branch column is the steady state users get: server restarted, persisted projections present, startup warmup done.

| path | main | branch (persisted projections + startup warmup) |
|---|---:|---:|
| first request per thread (cold) p50 | 15.4 ms | 8.1 ms |
| first request per thread (cold) p95 | 47.1 ms | 23.4 ms |
| first request per thread (cold) p99 | 100.7 ms | 34.4 ms |
| first request per thread (cold) max | 349.1 ms | 120.6 ms |
| repeat request (warm) p50 | 2.7 ms | 2.2 ms |
| repeat request (warm) p95 | 8.0 ms | 7.1 ms |
| repeat request (warm) p99 | 33.4 ms | 12.4 ms |
| repeat request (warm) max | 718.0 ms | 19.7 ms |
| full older-page walk p50 | 15.9 ms | 8.3 ms |
| full older-page walk p95 | 103.8 ms | 24.1 ms |
| full older-page walk p99 | 349.1 ms | 39.7 ms |
| full older-page walk max | 4605.2 ms | 293.6 ms |

| stress thread | events | main cold | branch (persisted + warmup) |
|---|---:|---:|---:|
| `thr_cdfq9maj8q` | 63,853 | 41 ms | 14 ms |
| `thr_gcuc46ug4j` | 42,033 | 144 ms | 6 ms |
| `thr_m9gz6riv9t` | 32,384 | 9 ms | 6 ms |
| `thr_59j74ttj6b` | 31,716 | 52 ms | 6 ms |

The branch beats main on every statistic, including the worst thread. What the request path never does anymore: a synchronous full build. If a request for a large thread lands during the startup warmup window before warmup reaches it, it starts (or joins) the cooperative build — 4.2 s wall for the 64k-event thread with the event loop's ping p99 at 58 ms meanwhile, zero stall-log entries. Warmup on the production copy: ~30 s of sliced background work on a cold start, ~1 s on later restarts (all 253 projections found fresh by identity columns, no payload reads). Three live-check-only defects were found and fixed along the way — warmup keying projections before provider plugins registered (every large thread then missed and rebuilt on the request path), a 1.56 s synchronous candidate scan at startup, and the truncation sweep invalidating large threads instead of rebuilding them in the background; the plan file records each with its trace. Streaming rebuilds on such threads are rate-limited by the refresh throttle (≤10 s staleness) until the incremental-reducer follow-up makes them cheap.

The 10k-event micro-benchmark's 12-page walk went from 1,634 ms (per-page reprojection) to <1 ms (one build + slices). Corpus perf gate vs the main-minted baseline: green. Persistence and throttle have tamper-marker and unit coverage (app-version invalidation, cost gate, proportional hold, 10 s cap); migration replay suite green.

**Turn details.** Oversized-turn walk hydrates all 250 seeded commands across pages with disjoint ranges and cursor-rebinding rejected; exact-fit and preview-page paths covered; route parity with the exact-range resource; the formerly-413 oversized delegation turn returns a first page.

**Suites.** `turbo run test` green: @bb/server (2,074), @bb/db (400), @bb/client-core (250), @bb/server-contract (60), @bb/sdk (96), @get-bb/plugin-sdk (220), app query tests; typecheck and lint green across all touched packages.

> AGENT GENERATED


ymichael:
Closing in favor of a streaming-first incremental timeline projection, with pagination reading the materialized projection.

PR #2801, opened 2026-09-01
Retry structured provider overload failures

ymichael:
## What was wrong

The provider-retry policy treated every structured failure other than a resettable subscription rate limit as ineligible. Structured `overloaded` failures are transient and retryable, but the plugin had no independent delay policy for failures that do not report a rate-limit reset window.

## What changed

The retry policy now schedules `overloaded` failures with exponential backoff and bounded jitter: 5–10 seconds, 10–20 seconds, 20–40 seconds, then 40–80 seconds. Overload recovery shares the existing five-total-attempt cap, identifies queued work as `Provider overloaded`, and continues to delegate accepted-versus-unaccepted input handling to core through `threads.retry`. Subscription-window behavior is unchanged. Plugin metadata, CLI wording, configuration documentation, and the generated BB guide sources now describe both recovery modes. There are no host-daemon protocol or wire changes.

## How you verified

The new overload policy and plugin-handler tests fail against the previous implementation because it declines the `overloaded` category. Verification commands:

- `pnpm exec turbo run test typecheck --filter=bb-plugin-provider-retry --force` — 15 tests passed and typecheck passed
- `pnpm exec turbo run test --filter=@bb/server --force -- --run test/threads/turn-failed-retry.test.ts` — 12 tests passed
- `pnpm exec turbo run test --filter=@bb/server --force -- --run test/services/plugins/plugin-authoring-docs.test.ts test/services/plugins/builtin-plugins.test.ts` — 47 tests passed
- `bb plugin build plugins/provider-retry`
- `git diff --check`

Fixes # (no tracking issue was filed)

> AGENT GENERATED

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -796,10 +796,12 @@ the plugin so it can be surfaced as needing attention.
 
 The builtin Provider retry plugin is enabled on fresh installations. When a turn
 fails on a structured Codex or Claude Code subscription-window limit that
-reports a reset time, it queues that turn to be re-sent after the window opens.
-Prior output or tool activity does not block recovery. The retry re-sends the
-original message verbatim, marked agent-only so the timeline does not show it
-twice. Disable it under Extensions → Plugins or with
+reports a reset time, it queues that turn after the window opens. It also
+retries structured provider overloads with exponential backoff and jitter.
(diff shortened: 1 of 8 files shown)


ymichael:
@slopcop

PR #2806, opened 2026-09-01
Fix Codex resumed child correlation

### LINKED ISSUE #2538: Bug - Refusing to start a competing turn for thread "<threadid>" while another turn is active or starting (Codex)

ymichael:
## What was wrong

When a parent Codex thread resumes or forks and continues working with an existing child agent, BB can attach that follow-up to the wrong child or fail to show it under the original delegation. Users expect the resumed child’s messages and results to remain attached to the task that originally spawned it. #2539 added relinking for this goal, but ownership still depended on event timing and candidate order, and current Codex versions can report a resumed child turn on the root provider thread rather than the child thread.

The provider could therefore complete the resumed child correctly while the app showed only “Worked for … / Failed to load turn details.” The timeline details path also projected the full event window twice on a range fallback and repeatedly scanned row suffixes, making the failure path unnecessarily expensive.

## What changed

- Restore each historical child’s original delegation identity from typed resume/fork history without replaying historical work.
- Preserve fast Codex resume with `excludeTurns: true`; when the first resumed-child interaction arrives, lazily read full history once, buffer intervening notifications, prime ownership, and replay those notifications in order.
- Replace the separate raw-intent, activity, and processed-ID caches plus LIFO candidate scan with one bounded call-ID correlation lifecycle.
- Match the exact child thread first, then allow a root-thread fallback only when exactly one eligible terminal child exists; ambiguous root turns fail closed, and native root turns retain priority.
- Preserve an accepted human new-turn boundary when a completed old turn emits reopened work, while keeping other turns’ prompts out of details.
- Treat an explicit completed-then-started lifecycle as a reopen for whole-item ownership, and search exact and stale ranges against one timeline projection.
- Replace repeated suffix scans with one reverse pass and prefilter historical items before schema parsing.
- Increment `HOST_DAEMON_PROTOCOL_VERSION` from 178 to 179 for the changed daemon-emitted delegation and `parentToolCallId` semantics.

This preserves and completes #2539: resumed and forked threads still reconstruct historical child ownership, the unique root-multiplexed Codex case is supported without guessing between children, and the intended follow-up remains expandable in the UI.

## How you verified

- `pnpm exec turbo run typecheck --filter=@bb/server --filter=@bb/thread-view --filter=bb-plugin-provider-codex --filter=@bb/host-daemon-contract` — 9/9 tasks passed.
- `pnpm exec turbo run test --filter=bb-plugin-provider-codex --force` — 27 files and 272 tests passed.
- `pnpm exec turbo run test --filter=@bb/thread-view --force` — 23 files and 368 tests passed.
- `pnpm exec turbo run test --filter=@bb/host-daemon-contract --force` — 3 files and 51 tests passed.
- `pnpm --dir apps/server exec vitest run test/services/threads/timeline-in-turn-window.test.ts` — 27/27 tests passed.
- The root-multiplexed resumed-child regression, ambiguity/native-root guards, lazy resume-history bridge regression, and reopened-delegation details regression pass after failing during the red/green cycle.
- Live repro `thr_z4637w8sxf`: the formerly failing `18-83` details request returns HTTP 200, exactly matches the inline children, and contains `ALPHA RESUMED`.
- `git range-diff` reports the original two patches unchanged after rebasing onto `2c57aa3919a2689ac6a155162f4b0f5bcca7dd66`; the review fixes are the third commit `aa175a877bfdd1024212bce09198a8cb3cb901e2`.
- `git diff --check` and formatting checks pass.
- Fresh GitHub CI passed all 13 executed checks for `aa175a877bfdd1024212bce09198a8cb3cb901e2`; mobile and Node-compatibility jobs were skipped by workflow policy.

Fixes #2538

> AGENT GENERATED

### DIFF
--- a/apps/server/src/services/threads/timeline.ts
+++ b/apps/server/src/services/threads/timeline.ts
@@ -286,6 +286,18 @@ function tryReadClientTurnRequestedRequestId(
   return event.requestId;
 }
 
+function isExternalUserNewTurnBoundary(row: StoredEventRow): boolean {
+  if (row.type !== "client/turn/requested") {
+    return false;
+  }
+  const event = parseStoredEvent(row);
+  return (
+    event.type === "client/turn/requested" &&
+    event.initiator === "user" &&
+    event.target.kind === "new-turn"
+  );
+}
+
 function tryReadSteerClientTurnRequestedRequestId(
   row: StoredEventRow,
 ): ClientTurnRequestId | null {
@@ -558,13 +570,46 @@ function filterExactEventRowsForRequestedTurn(
   const rows: StoredEventRow[] = [];
   let removedRows = false;
   const openToolCallIds = new Set<string>();
+  const externalUserNewTurnRequestIds = new Set<ClientTurnRequestId>();
+  const hasRequestedTurnCompletedRow = args.exactEventRows.some(
+    (row) =>
+      row.type === "turn/completed" &&
+      row.scopeKind === "turn" &&
+      row.turnId === args.turnId,
+  );
+  for (const [index, row] of args.exactEventRows.entries()) {
+    if (!hasRequestedTurnCompletedRow || !isExternalUserNewTurnBoundary(row)) {
+      continue;
+    }
+    const requestId = tryReadClientTurnRequestedRequestId(row);
+    if (
+      requestId === null ||
+      !args.acceptedClientRequestIdsForOtherTurns.has(requestId)
+    ) {
+      continue;
+    }
+    const hasLaterRequestedTurnRow = args.exactEventRows
+      .slice(index + 1)
+      .some(
+        (laterRow) =>
+          laterRow.scopeKind === "turn" && laterRow.turnId === args.turnId,
+      );
+    if (hasLaterRequestedTurnRow) {
+      externalUserNewTurnRequestIds.add(requestId);
+    }
+  }
   for (const row of args.exactEventRows) {
     if (row.scopeKind === "turn" && row.turnId !== args.turnId) {
       const continuesOpenToolCall =
         row.itemId !== null &&
         row.type.startsWith("item/") &&
         openToolCallIds.has(row.itemId);
-      if (!continuesOpenToolCall) {
+      const acceptsExternalUserNewTurn =
+        row.type === "turn/input/accepted" &&
+        externalUserNewTurnRequestIds.has(
+          parseAcceptedInputClientRequestId(row),
+        );
+      if (!continuesOpenToolCall && !acceptsExternalUserNewTurn) {
         removedRows = true;
         continue;
       }
@@ -583,7 +628,8 @@ function filterExactEventRowsForRequestedTurn(
     const requestId = tryReadClientTurnRequestedRequestId(row);
     if (
       requestId !== null &&
-      args.acceptedClientRequestIdsForOtherTurns.has(requestId)
+      args.acceptedClientRequestIdsForOtherTurns.has(requestId) &&
+      !externalUserNewTurnRequestIds.has(requestId)
     ) {
       removedRows = true;
       continue;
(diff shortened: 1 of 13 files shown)


ymichael:
@slopcop

ymichael:
@slopcop

SawyerHood:

PR #2835, opened 2026-09-01
Gate project-routed file sources on an opener capability

### LINKED ISSUE #2519: Plugin file openers built before `experimental_hostId` fail with "rpc input validation failed" when opening a file from a draft thread
## Human comments

This is showing up as an RPC error when you try to open a file from a draft thread in an inactive project using the monaco file editor:

Image

----

### Summary

Opening a workspace file from the **new-thread (draft) composer** in a project that has no active environment routes the file by *project + host* instead of by environment, and bb stamps an extra `experimental_hostId` key onto the `PluginFileOpenerSource` it hands the plugin file opener. Any opener plugin built before that field existed — including ones that copied bb's own "re-validate RPC input at the boundary" pattern with a `.strict()` Zod object — rejects the call, and the file tab renders a bare red `rpc input validation failed` banner. The same file opens fine from a normal thread, where the source carries an `environmentId` and no `experimental_hostId`.

Two things to decide here: (1) whether bb should send a source shape that older openers cannot accept, and (2) whether the frontend should surface something better than the raw boundary message when an opener's contract rejects bb's payload. This is also direct evidence against audit item 7 for this API ("Confirm `PluginFileOpenerSource.experimental_hostId` can become a stable required `hostId` field without breaking older opener implementations" — [docs/api_to_audit.md#L681-L682](https://github.com/get-bb/bb/blob/e4c873521abbcab1ea3e1aa127b73bb28e7905bb/docs/api_to_audit.md#L681-L682)): it already breaks them as an *optional* field.

### Versions and environment

- bb `0.40.1-nightly.32958738584.1` (desktop app), server on `127.0.0.1:38886`, data dir `~/.bb`.
- Source checked out at `e4c873521abbcab1ea3e1aa127b73bb28e7905bb` (`main`).
- macOS 14.3, Apple Silicon.
- Local host `host_x83k8uhh5d` (persistent, connected). Project `proj_ymedncn5jy` (`bb-marketplace`, local-path source, no environment started).
- Two file-opener plugins installed: the built-in `monaco-editor` (current SDK, schema has `experimental_hostId`) and an external `monaco` plugin installed from git, whose `sourceSchema` predates the field.

### Steps to reproduce

1. Install any file-opener plugin whose RPC input schema validates the source strictly and predates `experimental_hostId` — i.e. the shape bb's own plugin ships as the "boundary re-validation" pattern:

   ```ts
   const sourceSchema = z
     .object({
       kind: z.enum(["workspace", "host", "thread-storage"]),
       threadId: z.string().nullable(),
       environmentId: z.string().nullable(),
       projectId: z.string().nullable(),
     })
     .strict();
   ```

2. Pick a project whose environment is not running, and open the **new-thread composer** for it (a draft thread — no `thr_` id yet).
3. Open any workspace file into the side panel so the plugin opener handles it (e.g. `package.json`).
4. The tab renders `rpc input validation failed`. DevTools shows `POST /api/v1/plugins/<id>/rpc/read` → `400`.

The same open from an existing thread with a live environment works, because the source then carries `environmentId` and no `experimental_hostId`.

Equivalent repro without the UI, against a running bb server — same payload, two openers:

```console
$ # current-SDK opener: accepts the field, reads the file
$ curl -s -X POST http://127.0.0.1:38886/api/v1/plugins/monaco-editor/rpc/read \
    -H 'Content-Type: application/json' \
    -d '{"path":"package.json","source":{"kind":"workspace","threadId":null,"environmentId":null,"projectId":"proj_ymedncn5jy","experimental_hostId":"host_x83k8uhh5d"}}'
{"ok":true,"result":{"kind":"text","content":"{\n  \"name\": \"bb-marketplace-registry\", ...

$ # pre-field opener, identical payload: 400
$ curl -s -X POST http://127.0.0.1:38886/api/v1/plugins/monaco/rpc/read \
    -H 'Content-Type: application/json' \
    -d '{"path":"package.json","source":{"kind":"workspace","threadId":null,"environmentId":null,"projectId":"proj_ymedncn5jy","experimental_hostId":"host_x83k8uhh5d"}}'
{"ok":false,"error":{"code":"invalid_input","message":"rpc input validation failed","issues":[{"message":"Unrecognized key: \"experimental_hostId\"","path":["source"]}]}}
```

### Expected vs actual

**Actual.** `400 invalid_input` from the plugin RPC dispatcher, rendered in the file tab as:

```
rpc input validation failed
```

and in the console:

```
Failed to load resource: the server responded with a status of 400 (Bad Request)
:38886/api/v1/plugins/monaco/rpc/read
```

**Expected.** Either the file opens (bb sends a source shape the installed opener's contract accepts, or negotiates the field by declared SDK version), or the user sees an actionable message naming the plugin and the incompatibility instead of the raw boundary string.

### Evidence

- bb stamps the field only on the env-less, project-routed workspace source — exactly the draft-thread-in-an-unstarted-project case: [`file-opener-tabs.ts#L186-L198`](https://github.com/get-bb/bb/blob/e4c873521abbcab1ea3e1aa127b73bb28e7905bb/apps/app/src/components/plugin/file-opener-tabs.ts#L186-L198).
- The field is optional in the public contract: [`app-contract.ts#L169-L181`](https://github.com/get-bb/bb/blob/e4c873521abbcab1ea3e1aa127b73bb28e7905bb/packages/plugin-sdk/src/app-contract.ts#L169-L181).
- The built-in opener's schema was widened for it, and its comment recommends the strict boundary re-validation that older third-party openers copied: [`plugins/monaco-editor/server.ts#L41-L55`](https://github.com/get-bb/bb/blob/e4c873521abbcab1ea3e1aa127b73bb28e7905bb/plugins/monaco-editor/server.ts#L41-L55).
- The 400 comes from the plugin's own contract via the dispatcher, not from a missing plugin process: [`routes/plugins.ts#L638-L697`](https://github.com/get-bb/bb/blob/e4c873521abbcab1ea3e1aa127b73bb28e7905bb/apps/server/src/routes/plugins.ts#L638-L697) (`invalid_input` → 400) and [`plugin-service.ts#L585-L593`](https://github.com/get-bb/bb/blob/e4c873521abbcab1ea3e1aa127b73bb28e7905bb/apps/server/src/services/plugins/plugin-service.ts#L585-L593) (`rpc input validation failed`).
- Secondary compat point: even if an older opener ignored the unknown key, it still cannot serve this route — dropping the field from the payload above yields `{"code":"handler_error","message":"This file has no environment to resolve it against"}`, since project+host resolution did not exist when it was written.

### What you ruled out

- **Not "the plugin server isn't running."** That path returns `503` with a distinct `notRunningError` payload ([`routes/plugins.ts#L665-L671`](https://github.com/get-bb/bb/blob/e4c873521abbcab1ea3e1aa127b73bb28e7905bb/apps/server/src/routes/plugins.ts#L665-L671)). Here the plugin was running and its own schema produced the rejection; the same process answers `assets`/`tree` normally. This was the originally reported hypothesis and it is wrong.
- **Not the built-in `monaco-editor` opener.** It accepts the field and reads the file successfully on the identical payload (above). The failing request in the report is plugin id `monaco`, a separately installed opener.
- **Not a path/permission problem.** Same path, same project, same host id succeeds for the current-SDK opener in the same server process.
- Not a duplicate: the only "rpc input validation failed" hit is #833 (bb Connect `createMachineCode`), a different call site and cause.

Low-to-medium priority, small effort. Hits anyone running a third-party file opener published before the field landed, and only on the draft-thread/unstarted-project route; the workaround is to update the plugin. Nothing is lost. The wider question — whether adding an optional field to `PluginFileOpenerSource` is a breaking change in practice — is worth resolving before `experimental_hostId` is stabilized.

> AGENT GENERATED

andrewkchan:
## What was wrong

Opening a workspace file from the new-thread (draft) composer of a project whose environment is not running routes the file by project + host, and bb stamped `experimental_hostId` onto the `PluginFileOpenerSource` handed to whichever plugin opener claimed the extension ([`file-opener-tabs.ts#L186-L198`](https://github.com/get-bb/bb/blob/e4c873521abbcab1ea3e1aa127b73bb28e7905bb/apps/app/src/components/plugin/file-opener-tabs.ts#L186-L198)). An opener published before that field existed re-validates the source with a strict schema — the "re-validate RPC input at the boundary" pattern bb's own plugin ships — so it rejected the call with `400 invalid_input` and the file tab rendered a bare red `rpc input validation failed`. The root cause is not the extra key alone: an opener that tolerated the unknown key still could not serve the route, because project+host resolution did not exist when it was written, and it fails with `This file has no environment to resolve it against`. The env-less, project-routed source is simply a shape older openers cannot handle, and bb was sending it to all of them unconditionally.

Fixes #2519. Repro report: https://get-bb.github.io/reports/issues/2519.html

## What changed

- `PluginFileOpenerRegistration` gains `experimental_supportsProjectRoutedSource` (`packages/plugin-sdk/src/app-contract.ts`), collected and type-validated in `plugin-app-collector.ts`.
- `createFileOpenerTabForRequest` now resolves the opener first, then, for a workspace source with no `environmentId` and a `projectId`, uses that opener only if it declares the capability. Otherwise it returns `null` and `useThreadFileTabs` falls back to bb's native preview. `experimental_hostId` is therefore only ever sent to an opener that asked for it.
- The three first-party openers (`monaco-editor`, `docs`, `pdf-preview`) already resolve project sources, and declare the flag.
- Docs: new `docs/api_to_audit.md` entry for the experimental member, plus an update to audit item 7 of the live-file-navigation section — this issue is the evidence that `experimental_hostId` cannot become an unconditional required field. Plugin Guide card (`packages/plugin-api-map/src/surfaces.ts`) and regenerated `sdk-public-api.json`, the `bb-plugin-authoring` skill reference, and the `bb guide plugins` template.

No wire change: this is app-side tab construction plus an in-process registration field, so `HOST_DAEMON_PROTOCOL_VERSION` is untouched.

Deviation from the issue's second question ("should the frontend surface something better than the raw boundary message") — not addressed here. With this change bb no longer hands an opener a payload its contract rejects, so the banner is not reachable through this route; improving generic plugin RPC error presentation is a separate concern.

## How you verified

Tests — 3 new cases in `apps/app/src/components/plugin/file-opener-tabs.test.ts` (fallback with a host id, fallback without one, and an explicit "Open with" pick) fail before the change and pass after:

```
before:  Tests  3 failed | 9 passed (12)
after:   Tests  12 passed (12)
```

`plugin-app-definition.test.ts` now asserts the flag survives collection, so silently dropping it in the collector fails the build.

Suites run:

```
pnpm exec turbo run test --filter=@bb/app                    # 444 files, 3509 passed
pnpm exec turbo run test --filter=@bb/plugin-api-map --filter=@get-bb/plugin-sdk \
  --filter=bb-plugin-monaco-editor --filter=bb-plugin-simple-notes --filter=bb-plugin-pdf-preview
pnpm exec turbo run typecheck --filter=@bb/app --filter=@get-bb/plugin-sdk --filter=@bb/plugin-api-map \
  --filter=bb-plugin-monaco-editor --filter=bb-plugin-simple-notes --filter=bb-plugin-pdf-preview
pnpm format:check   # clean on the touched files
```

Manual, in the running dev app: a project with no running environment, a draft composer, and a file-opener plugin whose RPC contract uses exactly the strict pre-field schema from the issue. Opening `readme.md` from the side panel:

- **Before:** `POST /api/v1/plugins/<id>/rpc/read` → `400`; the tab renders `rpc input validation failed`. The request payload was the one in the issue: `{"kind":"workspace","threadId":null,"environmentId":null,"projectId":"proj_…","experimental_hostId":"host_…"}`.
- **After:** no RPC call is made; the file opens in bb's own preview.
- **After, with the same opener updated to declare `experimental_supportsProjectRoutedSource: true` and widen its schema:** `200`; the file opens in the plugin.

Screenshots of all three states to follow in a comment.

Fixes #2519

> AGENT GENERATED

### DIFF
--- a/apps/app/src/App.tsx
+++ b/apps/app/src/App.tsx
@@ -306,7 +306,9 @@ function AppRoutes() {
           <Route
             path="*"
             element={
-              <Suspense fallback={<RouteLoadingSkeleton />}>
+              <Suspense
+                fallback={<RouteLoadingSkeleton isBoundedPane={false} />}
+              >
                 <SplitWorkspaceRoute />
               </Suspense>
             }
--- a/apps/app/src/components/plugin/file-opener-tabs.test.ts
+++ b/apps/app/src/components/plugin/file-opener-tabs.test.ts
@@ -18,6 +18,21 @@ const MARKDOWN_OPENER = {
   title: "Docs editor",
 } satisfies PluginFileOpenerSlot;
 
+const PROJECT_ROUTED_OPENER = {
+  ...MARKDOWN_OPENER,
+  experimental_supportsProjectRoutedSource: true,
+} satisfies PluginFileOpenerSlot;
+
+const PROJECT_WORKSPACE_REQUEST: OpenSecondaryPanelTabRequest = {
+  kind: "workspace-file-preview",
+  tab: {
+    lineRange: null,
+    path: "docs/readme.md",
+    source: { kind: "working-tree" },
+    statusLabel: null,
+  },
+};
+
 const REQUESTS: readonly {
   label: string;
   request: OpenSecondaryPanelTabRequest;
@@ -97,19 +112,11 @@ describe("createFileOpenerTabForRequest thread-tabs contract", () => {
 
   it("preserves the selected host for a project-backed opener", () => {
     const tab = createFileOpenerTabForRequest({
-      fileOpeners: [MARKDOWN_OPENER],
+      fileOpeners: [PROJECT_ROUTED_OPENER],
       preference: {},
       projectHostId: "host_remote",
       projectId: "proj_1",
-      request: {
-        kind: "workspace-file-preview",
-        tab: {
-          lineRange: null,
-          path: "docs/readme.md",
-          source: { kind: "working-tree" },
-          statusLabel: null,
-        },
-      },
+      request: PROJECT_WORKSPACE_REQUEST,
       resolvedEnvironmentId: null,
       threadId: null,
     });
@@ -121,6 +128,67 @@ describe("createFileOpenerTabForRequest thread-tabs contract", () => {
       experimental_hostId: "host_remote",
     });
     expect(() => threadTabsSchema.parse([tab])).not.toThrow();
+  });
+
+  it("falls back to BB's preview when an opener does not accept project-routed sources", () => {
+    expect(
+      createFileOpenerTabForRequest({
+        fileOpeners: [MARKDOWN_OPENER],
+        preference: {},
+        projectHostId: "host_remote",
+        projectId: "proj_1",
+        request: PROJECT_WORKSPACE_REQUEST,
+        resolvedEnvironmentId: null,
+        threadId: null,
+      }),
+    ).toBeNull();
+  });
+
+  it("falls back for an unsupporting opener even without an explicit host", () => {
+    expect(
+      createFileOpenerTabForRequest({
+        fileOpeners: [MARKDOWN_OPENER],
+        preference: {},
+        projectId: "proj_1",
+        request: PROJECT_WORKSPACE_REQUEST,
+        resolvedEnvironmentId: null,
+        threadId: null,
+      }),
+    ).toBeNull();
+  });
+
+  it("keeps an explicit Open with pick on BB's preview when the opener cannot route it", () => {
+    expect(
+      createFileOpenerTabForRequest({
+        fileOpeners: [MARKDOWN_OPENER],
+        preference: {},
+        projectHostId: "host_remote",
+        projectId: "proj_1",
+        request: PROJECT_WORKSPACE_REQUEST,
+        resolvedEnvironmentId: null,
+        threadId: null,
+        viewer: { openerId: "markdown", pluginId: "docs" },
+      }),
+    ).toBeNull();
+  });
+
+  it("still opens an environment-backed file in an opener without the capability", () => {
+    const tab = createFileOpenerTabForRequest({
+      fileOpeners: [MARKDOWN_OPENER],
+      preference: {},
+      projectHostId: "host_remote",
+      projectId: "proj_1",
+      request: PROJECT_WORKSPACE_REQUEST,
+      resolvedEnvironmentId: "env_docs",
+      threadId: "thr_docs",
+    });
+
+    const params = parseFileOpenerParams(tab?.paramsJson ?? null);
+    expect(params?.source).toMatchObject({
+      environmentId: "env_docs",
+      kind: "workspace",
+    });
+    expect(params?.source.experimental_hostId).toBeUndefined();
   });
 });
 
--- a/apps/app/src/components/plugin/file-opener-tabs.ts
+++ b/apps/app/src/components/plugin/file-opener-tabs.ts
@@ -166,11 +166,25 @@ export function createFileOpenerTabForRequest({
   });
   if (owner === null) return null;
   const file = fileForOwnerRequest(owner);
-  const routedFile: PluginFileOpenerFile =
+  const resolved = resolveFileOpenerReplacement({
+    registrations: fileOpeners,
+    preference,
+    path: file.path,
+    ...(viewer !== undefined ? { override: viewer } : {}),
+  });
+  if (resolved.kind !== "plugin") return null;
+  const isProjectRouted =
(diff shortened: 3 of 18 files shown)


SawyerHood:

PR #2847, opened 2026-09-01
Organize sidebar regions and plugin overflow

brsbl:
## What was wrong

Plugin-provided pages had no clear home in the sidebar. They were mixed into the navigation without a label or region boundary, and legacy hidden-state preferences could prevent installed plugin pages from appearing at all.

## What changed

Installed plugin pages now appear in a predictable, compact **Plugins** region:

- The **Plugins** label appears only when plugin pages exist.
- The first three plugin pages stay visible; additional pages collapse behind a quiet, exact-count **Show 3 more** row.
- Expanding the disclosure reveals every remaining page in place, with **Show less** returning to the compact state.
- The disclosure deliberately uses no icon, border, or resting fill, so it reads as simplification rather than another sidebar control.
- Subtle dividers separate BB controls, Plugins, and Threads without adding heavy visual chrome.

This layer also preserves installed pages when migrating legacy navigation preferences and keeps plugin actions consistent across surfaces: open in split, view details, move between visible and overflow positions, and disable.

No public or wire contracts changed.

## Screenshots

Both sides use the same dev data, route, light theme, and 1440×900 viewport. Captures were taken after all plugin frontend modules had registered.

| Before — exact merge base `5a74c9731` | After — exact PR head `379f5d51c` |
| --- | --- |
| All six plugin pages are mixed directly into the unlabeled sidebar navigation. | Three pages remain visible under **Plugins**, with the other three summarized by **Show 3 more**. |
| Before — plugin pages mixed into the unlabeled sidebar navigation | After — three plugin pages with a minimal Show 3 more disclosure |

| Collapsed — three visible pages | Expanded — all six pages |
| --- | --- |
| **Show 3 more** reports the exact number of hidden pages without an icon or extra chrome. | The remaining pages appear below **Show less** without moving the surrounding regions. |
| Collapsed — three plugin pages and a minimal Show 3 more disclosure | Expanded — all six plugin pages and Show less |

## How you verified

- Chrome for Testing 151.0.7922.71 at 1440×900 on exact head `379f5d51c`: confirmed three visible pages, **Show 3 more**, six expanded pages, **Show less**, stable surrounding layout, transparent resting background, zero-width border, no disclosure icon, and no rendered alerts.
- Recaptured the exact merge base only after all six plugin frontend pages rendered, keeping the fixture, route, theme, and viewport comparable to the after state.
- Focused regression coverage asserts the three-page threshold, exact count, disclosure accessibility state, icon absence, expansion, and move-to-overflow boundary.
- All 13 required remote checks passed on exact head `379f5d51c`, including app/server/package/integration tests and Ubuntu/macOS package smoke jobs; Node compatibility and iOS simulator jobs were skipped by workflow conditions.
- Rebased onto `main` at `5a74c9731`; `git range-diff` confirmed all eight PR commits retained their original patches.
- Per repository policy, CI-equivalent checks ran through remote CI rather than being rerun locally.

Fixes #

BB-Thread-ID: thr_ccffp4w2p2

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/plugin/PluginNavSidebarItems.test.tsx
+++ b/apps/app/src/components/plugin/PluginNavSidebarItems.test.tsx
@@ -11,6 +11,7 @@ import { useEffect, type ComponentType } from "react";
 import { createStore, Provider } from "jotai";
 import { MemoryRouter } from "react-router-dom";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { CompactViewportOverrideProvider } from "@bb/shared-ui/hooks/use-compact-viewport";
 import { SidebarProvider } from "@/components/ui/sidebar.js";
 import {
@@ -22,7 +23,10 @@ import {
   resetAllCrashedPluginSlotsForTest,
   resetCrashedPluginSlots,
 } from "./PluginSlotMount";
-import { PluginNavSidebarItems } from "./PluginNavSidebarItems";
+import {
+  ExtensionsNavSidebarItem,
+  PluginNavSidebarItems,
+} from "./PluginNavSidebarItems";
 import { pluginNavPanelOrderAtom } from "./pluginNavSidebarAtoms";
 
 function registrationSet(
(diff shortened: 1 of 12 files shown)


brsbl:
Superseded by focused replacement PR #2915. Plugin row action changes are isolated in stacked PR #2916 instead of remaining coupled to the sidebar region/overflow experiment.

PR #2849, opened 2026-09-01
Rename unorganized thread group to Threads

brsbl:
## What was wrong

Threads outside an organizing unit were labeled “Unorganized,” which described an implementation condition rather than the content users were browsing.

## What changed

- Rename the visible “Unorganized” group label to “Threads.”
- Update the remove-section explanation and the representative sidebar story copy to use the same terminology.

No layout, ordering, filtering, persistence, public, or wire behavior changed.

## Screenshots

Both captures use the same manual organization state, dev data, route, light theme, and 380×900 sidebar crop.

**Before — parent head `404ba2774`: loose threads are labeled “Unorganized”**

Before — loose-thread group labeled Unorganized

**After — PR head `64529ff20`: the same group is labeled “Threads”**

After — loose-thread group labeled Threads

## How you verified

- `git diff --check`
- Remote CI passed, including checks, app/server/package tests, and Linux/macOS package smoke.
- Captured the exact parent and PR-head branch apps in Chrome for Testing 151.0.7922.71 with the same fixture and manual organization state.
- Confirmed only the visible loose-thread label changes from “Unorganized” to “Threads.”
- Per repository policy, no CI-equivalent tests, typechecks, lint, or builds were run locally.

Fixes #

BB-Thread-ID: thr_ccffp4w2p2

> AGENT GENERATED


brsbl:
Superseded by #2883, which preserves only the intended Threads-label change against current main so it can be reviewed and merged independently.

PR #2851, opened 2026-09-01
Polish command palette search and hierarchy

brsbl:
## What was wrong

Command search and thread search used competing presentations, the results viewport and footer did not read as one stable keyboard surface, and thread-result metadata competed with titles. The older palette implementation also needed to retain Settings and plugin-page destinations now present on main.

## What changed

- Move thread search into an explicit nested command-palette mode with Backspace/Escape navigation.
- Keep the keyboard footer visible, use shared keycap/control tokens, and clip/fade the scrolling results at the footer boundary.
- Group resting commands, distinguish drill-in commands, and hide an empty Plugins group.
- Refine search-field, tab, filter, metadata, state-label, theme, highlighting, and thread-result hierarchy using existing tokens.
- Preserve Settings, plugin settings, plugin pages, plugin command actions, split opening, keyboard navigation, and search ranking.
- Keep draft search on the existing single new-thread draft contract so this layer does not depend on the experimental sidebar draft-slot system.
- Add focused coverage for root and nested modes, footer behavior, scrolling, highlighting, thread states and filters, themes, and preserved destinations.

No public or wire contracts changed.

## Screenshots

Both captures use the same dev data, nested Search threads flow, route, light theme, and 1440×900 viewport.

**Before — parent head `64529ff20`: the nested thread-search palette has the older search/result treatment and no persistent footer**

Before — older nested thread-search palette without the persistent footer

**After — PR head `d136783a6`: the Threads tab, quieter metadata, state filter, and persistent keyboard footer form one surface**

After — nested thread-search palette with refined hierarchy and persistent footer

## How you verified

- `git diff --check`
- Remote CI passed for the final head, including checks, app/server/package tests, integration tests, and Linux/macOS package smoke.
- The exact final-head branch app launched successfully with the repository standard launcher on Node 22.
- Drove the exact parent and PR-head branch apps over CDP in Chrome for Testing 151.0.7922.71 using the same fixture and viewport.
- Entered nested Search threads from Command-K and confirmed the refined search/result hierarchy plus the always-visible keyboard footer.
- Per repository policy, no CI-equivalent tests, typechecks, lint, or builds were run locally.

Fixes #

BB-Thread-ID: thr_ccffp4w2p2

> AGENT GENERATED


brsbl:
Superseded by the focused command-palette stack: #2873 owns nested thread search and result execution, #2877 owns persistent keyboard guidance and viewport clipping, and #2878 owns thread-search control and metadata hierarchy.

PR #2866, opened 2026-09-01
fix: make turn-start queue wake race-safe

ymichael:
## What was wrong

A follow-up message sent at exactly the wrong moment during turn startup could be added after startup's only queue scan. The turn could then finish without ever seeing the message, leaving it queued forever.

The first fix still allowed archive or soft delete to win after an earlier stale read, which could strand the newly queued row on a thread the idle scan deliberately excludes. It also notified the hub while SQLite still held its writer lock.

## What changed

- Make the active-turn recheck, complete writable-thread check, and startup-wait write one immediate SQLite transaction for ordinary sends, queued retries, and parent system notices.
- Return explicit outcomes from the shared transaction helper so both callers handle state changes consistently.
- Buffer queue notifications until the outer transaction commits.
- Let idle draining recover startup-wait rows when turn settlement wins, with one shared wait-kind list for the JavaScript and SQL predicates.
- Keep the existing queue claim guard, so recovery cannot dispatch a message twice.

No host-daemon wire contract changed, so no protocol bump is required.

## How you verified

- Added ordered regressions for archive and soft delete winning before ordinary queue admission, plus archive winning during parent-notice preparation.
- Added a regression proving queue notifications happen after the SQLite transaction.
- Added a direct SQL idle-scan regression for a `turn-starting` row.
- Focused server tests: 15/15 passed.
- Focused DB tests: 31/31 passed.
- Turbo typechecks for `@bb/server` and `@bb/db`, the server build, and `git diff --check` passed after rebasing onto current `origin/main`.
- The complete GitHub CI matrix is green, including server, packages, integration, app shards, and Linux/macOS package smoke.

Follow-up to #2816.

@slopcop

> AGENT GENERATED

### DIFF
--- a/apps/server/src/services/threads/dispatch-attempt.ts
+++ b/apps/server/src/services/threads/dispatch-attempt.ts
@@ -333,7 +333,19 @@ async function runDispatchAttempt(
     resolveDispatchAttemptKind(currentThread, payload.mode) === "join-turn" &&
     getActiveTurnId(deps, thread.id) === null
   ) {
-    return waitOn({ kind: "turn-starting" }, null);
+    const outcome = deps.db.transaction(
+      (tx) => {
+        const lockedThread = getThread(tx, thread.id);
+        if (!lockedThread) return null;
+        if (lockedThread.status !== "active") {
+          return waitOn({ kind: "thread-busy" }, null);
+        }
+        if (getActiveTurnId({ db: tx }, thread.id) !== null) return null;
+        return waitOn({ kind: "turn-starting" }, null);
+      },
+      { behavior: "immediate" },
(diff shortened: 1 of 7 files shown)


ymichael:
@slopcop Review findings are addressed at 5c3fc7d2af, including the ordered archive/delete regressions, post-commit notification assertion, direct SQL idle-scan test, and shared transaction helper. CI is green.

ymichael:
Superseded by #2884, which combines this startup-race fix with unified group readiness, failed-group admission, manual-Stop boundaries, and the active-to-error review follow-up. #2884 passed the complete CI matrix and has been merged.

PR #2885, opened 2026-09-02
Fix AppImage pre-runtime SIGTRAP flake

ymichael:
## What was wrong

Linux AppImages were still built with electron-builder's legacy `appimage-12.0.1` toolset and its 2019 AppImage runtime (`effcebc`). Crash capture on the recurring CI signature showed that the zygote first received `SIGBUS/BUS_ADRERR` while fetching an executable instruction from the AppImage FUSE mount. The fault address equaled RIP at the first instruction of PDFium's `copy8`; the browser crash handler then terminated with `int3`/`SIGTRAP`, which is the outer signal reported by the lifecycle smoke. This is why AppImage stdout/stderr was normally empty: the failure was below app JavaScript and before owned-runtime publication.

The captured system had 14 GiB available, untouched swap, and no OOM evidence. The same signature occurred on Intel and AMD CI runners. The core and matched Electron 41.7.0 symbols ruled out the Chromium sandbox CHECK addressed by #2657, the post-marker port race addressed by #2822, an application fatal path, and CPU-vendor-specific behavior.

## What changed

- Upgrade electron-builder to 26.15.7 and explicitly pair its exact Squirrel Windows peer.
- Select AppImage toolset 1.0.3, which packages the maintained 20251108 static runtime instead of the legacy external-libfuse runtime.
- Extend the existing packaging-config regression test to require toolset 1.0.3.

The resulting x86_64 AppImage is a static PIE and reports runtime commit `dd6cebe`. There is no server/host-daemon wire change, so `HOST_DAEMON_PROTOCOL_VERSION` is unchanged.

## How you verified

Red, before the fix:

- [Diagnostic CI run 33574359370](https://github.com/get-bb/bb/actions/runs/33574359370) reproduced the exact pre-runtime `SIGTRAP` on iteration 3 after two healthy launches.
- Its zygote core recorded `SIGBUS`, `si_code=BUS_ADRERR`, and `si_addr == RIP`; the matching Electron build ID resolved the address to the first instruction fetched from the FUSE-backed executable.
- Memory, swap, kernel logs, browser and zygote cores, the exact AppImage, and matched symbols were inspected. All diagnostic load was sequential and bounded, and failure stopped the loop.

Green, with the fix:

- On the same bounded six-launch workflow and runner class, the candidate completed five target launches without a pre-runtime signal; iteration 6 then exposed the distinct post-marker owned-runtime flake and produced no crash core.
- The exact same candidate SHA passed all six AppImage mount/lifecycle launches on [attempt 2 of run 33575866378](https://github.com/get-bb/bb/actions/runs/33575866378).
- [PR CI run 33576754644](https://github.com/get-bb/bb/actions/runs/33576754644) passed the full matrix, including normal Linux AppImage packaging and lifecycle smoke.
- `pnpm exec turbo run build typecheck test --filter=@bb/desktop --concurrency=2`: 15/15 tasks passed; 38/38 desktop test files passed, 247 tests passed, one skipped.
- `pnpm install --frozen-lockfile` and `git diff --check` passed.


ymichael:
@slopcop

PR #2913, opened 2026-09-02
Split codex app-server stdout on newline only

JanderV:
## Problem

Resuming or forking a Codex thread fails with `JSON-RPC request timed out: thread/resume` when the thread history contains a raw U+2028 (LINE SEPARATOR) or U+2029, for example pasted email or WhatsApp text. Fresh threads work.

`createCodexAppServerConnection` reads the child's stdout with `node:readline`, which also treats U+2028/U+2029 as line breaks. Codex emits those characters unescaped inside JSON strings, so the `thread/resume` response (1.6 MB in my case) arrives as 17 unparseable fragments. The pending request never resolves and hits `CHILD_REQUEST_TIMEOUT_MS`.

Verified with `BB_PROVIDER_BRIDGE_RECORD_DIR`: codex answered in 0.3 s, the bridge reported the 60 s timeout.

## Fix

Read stdout with a `StringDecoder` and split on `\n` only. Adds a regression test that spawns a fake server echoing a JSON line with a raw U+2028.

Tested on bb 0.41.0 / codex 0.152.1: with this change the affected thread resumes in about 1 s.

### DIFF
--- a/plugins/provider-codex/src/bridge/app-server-connection.line-separator.test.ts
+++ b/plugins/provider-codex/src/bridge/app-server-connection.line-separator.test.ts
@@ -1,0 +1,45 @@
+import { expect, it } from "vitest";
+import { z } from "zod";
+
+import { createCodexAppServerConnection } from "./app-server-connection.js";
+
+// Echoes every request as a result whose string contains a raw U+2028.
+const FAKE_SERVER = `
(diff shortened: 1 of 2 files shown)


SawyerHood:
@slopcop check

SawyerHood:
@slopcop check please

SawyerHood:

PR #2924, opened 2026-09-02
fix(acp): preserve model config errors

### LINKED ISSUE #2925: ACP bridge retries model config errors through unsupported session/set_model
## Summary

Selecting a non-default model for an `acp-omp` thread could fail during thread startup even though OMP accepts that model through its advertised ACP model config option. BB discarded the original config-option error and retried with `session/set_model`, an optional ACP extension OMP does not implement. Expected behavior is to preserve the config-option error and never infer support for an unadvertised extension.

## Versions and environment

- bb 0.41.0, macOS 26 / Apple Silicon desktop app; source checkout at `2b06ec9c08f32e20cd98dd99c6250e77d35ecf09`
- OMP `18.1.2` (`omp acp`), provider `acp-omp`
- Direct local environment; BB Connect paired
- Selected model: `command-code/deepseek/deepseek-v4-flash`; reasoning: `high`

## Steps to reproduce

This was observed before the fix on the prior bridge behavior; it is now covered by the PR regression test.

1. Start a BB thread with provider `acp-omp`.
2. Select `command-code/deepseek/deepseek-v4-flash` with `high` reasoning.
3. Start the thread.
4. If OMP's advertised `session/set_config_option` request errors, observe the fallback request.

Minimal bridge-level reproduction, now encoded in `packages/provider-bridge-acp/src/bridge/bridge.test.ts`:

1. Start the fake ACP agent with `FAKE_ACP_MODEL_CONFIG=1` and `FAKE_ACP_SET_CONFIG_MODEL_ERROR=1`.
2. Select `fake/strong`.
3. Before the fix, the bridge swallowed `model config probe failed`, called `session/set_model`, and could surface only the fallback error.

Did not reproduce with OpenCode ACP: it completed the same planning workflow. After the fix, OMP completed a DeepSeek Flash/high thread and a follow-up turn.

## Expected vs actual

Expected:

```text
BB uses OMP's advertised session/set_config_option interface. If it fails, BB returns that original error.
```

Actual before the fix:

```text
Command thread.start failed
Internal error: Unknown ACP ext method: session/set_model
```

The `session/set_model` error was misleading: OMP does not advertise or implement that optional extension.

## Evidence

- Failing model-selection code path on the base revision: https://github.com/get-bb/bb/blob/eeaaa3e8d1c60669c308450550321f394186c5d7/packages/provider-bridge-acp/src/bridge/bridge.ts#L1120-L1146
- The bridge's bare catch discarded the first error and sent `session/set_model`.
- OMP 18.1.2 directly accepted `session/set_config_option` for `command-code/deepseek/deepseek-v4-flash`.
- [Successful validation thread](bb://thread/thr_7qp9587mdm) completed the exact prompt `command-code-deepseek-v4-flash-high-ok`, then completed a normal `hi` follow-up.
- Focused verification: `pnpm exec turbo run typecheck test --filter=@bb/provider-bridge-acp --force` passed typecheck and 300 tests in 17 test files.

## What you ruled out

- Not duplicate of #1918: that issue is a separate ACP dynamic-tool MCP process launch defect.
- Not an OMP credential, model-role, or persistent Command Code failure: direct OMP ACP configuration and the post-fix BB thread both succeeded.
- Not general ACP failure: OpenCode ACP completed the corresponding planning workflow.

Medium — blocks selected OMP ACP models at thread startup and hides the actionable cause. Default-model OMP, terminal OMP, and OpenCode ACP are workarounds; the implementation is small and covered by a focused regression test.

> AGENT GENERATED

GantisStorm:
## What was wrong

BB can select an ACP agent's model by one of two agent-specific interfaces:

1. An advertised `model` config option, selected with the standard `session/set_config_option` request.
2. Session model state, selected through the optional `session/set_model` extension.

The bridge used the first interface when present, but caught every error from that request and unconditionally retried with the second. That retry was unsafe: advertising a model config option does not imply support for `session/set_model`.

OMP exposes the first interface and correctly rejects the second. A failed config-option request therefore became the misleading visible error:

```text
Unknown ACP ext method: session/set_model
```

The original config-option error was discarded, making diagnosis impossible. This was a BB ACP bridge compatibility and error-handling bug, not an OMP credential, model-role, or Command Code configuration defect.

## Scope

This is an independently reviewable ACP bridge correction. It follows the interface each ACP agent advertises and preserves the original configuration error when that interface fails; it does not alter Browser behavior, OMP credentials, model roles, or Command Code configuration.

## Why it appeared to break Command Code

Command Code is the model route selected *inside* the OMP ACP session; it was never reached during the failed startup. Selecting `command-code/deepseek/deepseek-v4-flash` caused BB to create an OMP session and request that model through OMP's advertised config option. When that request threw, BB issued the unsupported fallback and aborted startup before OMP could send a prompt to Command Code.

The failure was therefore in BB's ACP setup layer, not Command Code. The subsequent DeepSeek Flash/high thread completed through the same OMP route, confirming that OMP and Command Code are healthy.

## What changed

`selectAcpNativeModel` now follows the selection mechanism the agent advertises:

- When a model config option exists, use `session/set_config_option` and preserve its result or error.
- Use `session/set_model` only when the agent exposes session-model state without a model config option.

This is intentionally narrow. Successful config-option selection is unchanged. Agents that use the session-models-only route retain the existing fallback behavior. Reasoning-level and service-tier selection are untouched.

The replaced regression test simulates an agent with a model config option whose selection fails. It verifies that `thread/start` returns the original `model config probe failed` error rather than issuing `session/set_model`.

## Why this is correct

ACP model-selection implementations vary by agent. BB must use the interface the agent actually advertises rather than infer an optional extension from an unrelated failure. Returning the original ACP error gives users and diagnostics the actionable cause if a real model-configuration failure happens again.

The initial OMP config-option failure could not be attributed after the fact because the previous bare catch discarded it. Subsequent direct OMP ACP selection and BB end-to-end validation both succeeded, so there is no evidence of a persistent OMP or Command Code configuration problem.

## How you verified

- `pnpm exec turbo run typecheck test --filter=@bb/provider-bridge-acp --force`
  - Typecheck passed.
  - 17 test files and 300 tests passed.
- Directly initialized OMP 18.1.2 ACP and set the advertised `model` config option to `command-code/deepseek/deepseek-v4-flash`; OMP accepted it.
- After updating the local BB package, started an `acp-omp` BB thread with `command-code/deepseek/deepseek-v4-flash` and `high` reasoning. It completed with the exact expected response `command-code-deepseek-v4-flash-high-ok`, then completed a normal follow-up `hi` turn.

Fixes #2925

> AGENT GENERATED


GantisStorm:
This remains an independent ACP bridge fix. Please review it independently.

PR #2931, opened 2026-09-02
Decode hex-encoded Claude Code Keychain credentials

### LINKED ISSUE #2928: Claude Code usage limits report unauthenticated: macOS Keychain blob is hex JSON
### Summary

Settings → Usage limits reports Claude Code as unauthenticated on macOS even when `claude` is logged in and `~/.claude/.credentials.json` is valid JSON. The host reads the Keychain item `Claude Code-credentials` first, treats the blob as UTF-8 JSON, fails, and never falls through to the credentials file.

### Versions and environment

- bb from source at get-bb/bb `main` (`eeaaa3e8db7b3aeb3c4ab46873816c84cb6ea513`)
- macOS, Claude Code 2.1.x (2.1.257 observed)
- Isolated desktop instance; Claude Code logged in via OAuth (Max)

### Steps to reproduce

1. On macOS, log in with Claude Code 2.1.x so Keychain has `Claude Code-credentials`.
2. Confirm the Keychain secret is hex (`security find-generic-password -s "Claude Code-credentials" -w` starts with `7b`, which is `{`).
3. Confirm `~/.claude/.credentials.json` is valid JSON with `claudeAiOauth`.
4. Open bb → Settings → Usage limits for that host, or `GET /api/v1/system/usage-limits`.

Did not reproduce the unauthenticated status after decoding the Keychain blob as hex then JSON, then falling back to `~/.claude/.credentials.json`.

### Expected vs actual

Actual (host `provider.usage` for claude-code):

```
status: "unauthenticated"
```

Settings shows no Claude session/weekly/Fable windows. The Usage plugin can only render that empty/error state.

Expected: `status: "ok"` with the same 5-hour / weekly / model-scoped windows the Claude Code CLI shows.

### Evidence

`plugins/provider-claude-code/src/bridge/provider-maintenance.ts` on `main` parses the Keychain stdout as UTF-8 JSON:

https://github.com/get-bb/bb/blob/eeaaa3e8db7b3aeb3c4ab46873816c84cb6ea513/plugins/provider-claude-code/src/bridge/provider-maintenance.ts#L273

Claude Code 2.1.x stores that Keychain password as hex-encoded JSON (`7b…` = `{`). `JSON.parse` throws, the file fallback is skipped because the Keychain read succeeded with a non-empty string.

A unit fixture of hex(`{"claudeAiOauth":{…}}`) fails `JSON.parse` and succeeds after `Buffer.from(raw, "hex").toString("utf8")`.

### What you ruled out

- Not empty windows from a stale User-Agent (`claude-code/2.1.0`) or `kind === "weekly_scoped"` filtering: the live API never ran because status was `unauthenticated`.
- Not the community Usage plugin: it only renders `GET /api/v1/system/usage-limits`.
- Not a missing `~/.claude/.credentials.json`: the file was valid; the Keychain path ran first and failed closed.
- Searched get-bb/bb issues for hex / keychain / unauthenticated usage-limits; no existing report of this parse.

High for anyone on Claude Code 2.1.x on macOS using Usage limits; workaround is none in Settings. Fix is a few lines: decode hex JSON, then fall back to the credentials file. Effort Low.

> AGENT GENERATED

Guitaraholic:
## What was wrong

Claude Code 2.1.x stores the macOS Keychain item `Claude Code-credentials` as hex-encoded JSON (`7b…` = `{`). `provider-maintenance` parsed that blob with `JSON.parse` as UTF-8, failed, and never fell through to `~/.claude/.credentials.json` because the Keychain read had already succeeded. Settings → Usage limits then reported Claude as `unauthenticated` while the CLI was logged in.

## What changed

- Decode the Keychain secret as JSON first, then as hex JSON.
- If the Keychain blob still does not parse, read `~/.claude/.credentials.json`.
- Unit test covers hex, UTF-8 JSON, and garbage input.
- No `HOST_DAEMON_PROTOCOL_VERSION` bump: this is local credential parsing, not a wire change.

## How you verified

- `pnpm exec vitest run --config vitest.config.ts src/bridge/provider-maintenance.test.ts` in `plugins/provider-claude-code` (3 passed, including the new hex fixture).
- Live: after this decode, `GET /api/v1/system/usage-limits` returns Claude session/weekly windows instead of `unauthenticated`.

Fixes #2928

> AGENT GENERATED

### DIFF
--- a/plugins/provider-claude-code/src/bridge/provider-maintenance.test.ts
+++ b/plugins/provider-claude-code/src/bridge/provider-maintenance.test.ts
@@ -80,4 +80,25 @@ describe("Claude Code provider maintenance", () => {
       "https://claude.ai/install.sh",
     );
   });
+
+  it("decodes hex-encoded Claude Code keychain credentials", () => {
+    const credentials = {
+      claudeAiOauth: {
+        accessToken: "sk-ant-test",
+        expiresAt: 1_788_350_959_477,
+        subscriptionType: "team",
+        rateLimitTier: "default_claude_max_5x",
+      },
+    };
+    const hex = Buffer.from(JSON.stringify(credentials), "utf8").toString(
+      "hex",
+    );
+    expect(__testing.parseClaudeCredentials(hex)).toEqual(
+      credentials.claudeAiOauth,
+    );
+    expect(
+      __testing.parseClaudeCredentials(JSON.stringify(credentials)),
+    ).toEqual(credentials.claudeAiOauth);
+    expect(__testing.parseClaudeCredentials("not-credentials")).toBeNull();
+  });
 });
--- a/plugins/provider-claude-code/src/bridge/provider-maintenance.ts
+++ b/plugins/provider-claude-code/src/bridge/provider-maintenance.ts
@@ -257,21 +257,39 @@ async function readKeychainCredentials(): Promise<string | null> {
   return null;
 }
 
-async function readCredentials(): Promise<ClaudeCredentials | null> {
-  let raw = await readKeychainCredentials();
-  if (raw === null) {
+function decodeCredentialBlob(raw: string): unknown {
+  try {
+    return JSON.parse(raw);
+  } catch {
+    if (raw.length === 0 || raw.length % 2 !== 0 || /[^0-9a-f]/i.test(raw)) {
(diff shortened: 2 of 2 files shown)


SawyerHood:

PR #2936, opened 2026-09-02
Move thread-list SQLite reads off the serving loop

### LINKED ISSUE #1131: Prevent synchronous SQLite reads and retention sweeps from blocking the server

Guitaraholic:
## What was wrong

`better-sqlite3` runs on the Node serving loop. One cold query on a 1.7GB `bb.db` froze every client for 11s, including WebSocket token streams ([#1131](https://github.com/get-bb/bb/issues/1131)).

## What changed

- File-backed servers start one readonly WAL worker.
- Thread list and sidebar project+thread reads go through it (`listThreadsWithPendingInteractionStateOffThread` and the project-scoped variant).
- `:memory:` tests and a worker that fails to start keep the existing sync path.
- Bundled servers also emit `dist/sqlite-read-worker.js`.
- Search and timeline reads stay on the serving connection. No protocol change.

## How you verified

- Added `packages/db/test/sqlite-read-queue.test.ts`: in-memory stays on the calling connection; a file-backed worker returns the same thread list as the serving connection.
- Desktop and app query tests for the other harvests passed in this checkout. Local db vitest workers OOM in this environment; CI should run the new file.

Fixes #1131

> AGENT GENERATED

### DIFF
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -4,7 +4,7 @@
   "type": "module",
   "private": true,
   "scripts": {
-    "build": "node ../../scripts/build-node-entry.mjs src/index.ts dist/index.js --clean-dist --external ./start-server.js --copy-dir ../../packages/db/drizzle dist/drizzle --copy-dir src/assets dist/assets && node --import tsx scripts/copy-builtin-skills.ts && node ../../scripts/build-node-entry.mjs src/start-server.ts dist/start-server.js && node ../../scripts/build-node-entry.mjs ../../packages/plugin-sdk/src/index.ts dist/plugin-sdk-runtime.js",
+    "build": "node ../../scripts/build-node-entry.mjs src/index.ts dist/index.js --clean-dist --external ./start-server.js --copy-dir ../../packages/db/drizzle dist/drizzle --copy-dir src/assets dist/assets && node --import tsx scripts/copy-builtin-skills.ts && node ../../scripts/build-node-entry.mjs src/start-server.ts dist/start-server.js && node ../../scripts/build-node-entry.mjs src/sqlite-read-worker.ts dist/sqlite-read-worker.js && node ../../scripts/build-node-entry.mjs ../../packages/plugin-sdk/src/index.ts dist/plugin-sdk-runtime.js",
     "start": "node dist/index.js",
     "start:prod": "cross-env NODE_ENV=production node dist/index.js",
     "dev": "node --conditions=source --import tsx scripts/dev-supervisor.mjs",
--- a/apps/server/src/routes/projects.ts
+++ b/apps/server/src/routes/projects.ts
@@ -13,7 +13,7 @@ import {
   listPublicProjects,
   listProjectSourcesByProjectIds,
   listThreadSections,
-  listThreadsWithPendingInteractionStateForProjects,
+  listThreadsWithPendingInteractionStateForProjectsOffThread,
   reorderProject,
   updateProject,
   updateProjectSource,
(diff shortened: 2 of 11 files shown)


SawyerHood:

Issue #2622, opened 2026-08-28
Document the top-level JSON schema for each bb CLI command


ryanbbrown:
## What happens

`bb ... --json` commands return several valid top-level shapes, but command help and the guide do not show those shapes. Callers often infer one command's envelope from another command and then get `jq` nulls or type errors.

## Reproduction

```bash
bb thread list --json | jq type
# "array"

bb thread show "$BB_THREAD_ID" --json | jq 'keys'
# ["environment", "pendingTodos", "thread"]

bb skill list --json | jq 'keys'
# ["skills"]

bb settings keyboard list --json | jq 'keys'
# ["bindings", "overrides"]

bb terminal list --thread "$BB_THREAD_ID" --json | jq type
# "object"
```

## Expected scope

Document a short top-level schema or example for each JSON command in help or the CLI guide. This request does not require a breaking universal envelope change.

Current `main` prints each SDK result directly through `outputJson`; for example, [`thread/list.ts`](https://github.com/get-bb/bb/blob/81a0bf98e055cc7061ee3bc2b0f8057031682bba/apps/cli/src/commands/thread/list.ts#L24-L59) emits an array, while `thread/show.ts`, `skill.ts`, `settings.ts`, and `terminal.ts` emit their own result objects.

Related but not a duplicate: #2052 covers poor human output when `--json` is absent. This issue covers discoverability of machine-readable result shapes.

Issue #2620, opened 2026-08-28
bb thread tell/spawn need stdin or file-backed message input


ryanbbrown:
## What happens

`bb thread tell` accepts message text only as the positional `<message>`, and `bb thread spawn` accepts initial text only through `--prompt <prompt>`. There is no stdin, `--message-file`, or `--prompt-file` input.

This makes long Markdown prompts easy to corrupt when callers place them in shell source. Backticks and `$()` in a double-quoted shell argument run before BB starts, and complex inline heredocs often fail before BB receives any text.

## Reproduction

```bash
bb thread tell --help
# Usage: bb thread tell [options] <id> <message>
# No stdin or message-file option.

bb thread spawn --help
# --prompt <prompt> Initial prompt for the thread
# No stdin or prompt-file option.
```

A safe current workaround needs a caller-managed file expansion:

```bash
bb thread tell "$id" -- "$(cat "$file")"
bb thread spawn --project "$project" --prompt "$(cat "$file")"
```

## Expected scope

Add a file-backed or stdin text input for both commands, with clear precedence or mutual-exclusion errors. BB cannot undo shell expansion, but it can give callers a path that does not put generated message text in shell source.

Current `main` still registers only the positional message and prompt option:

- [`thread/actions.ts`](https://github.com/get-bb/bb/blob/81a0bf98e055cc7061ee3bc2b0f8057031682bba/apps/cli/src/commands/thread/actions.ts#L418-L461)
- [`thread/spawn.ts`](https://github.com/get-bb/bb/blob/81a0bf98e055cc7061ee3bc2b0f8057031682bba/apps/cli/src/commands/thread/spawn.ts#L166-L218)

Suggested tests: `apps/cli/src/__tests__/command-output/thread-tell.test.ts` and `thread-spawn.test.ts`.

Issue #2520, opened 2026-08-27
Make durable-terminal exit observable by its owning thread and retain bounded post-exit output


ryanbbrown:
## The workflow

1. A child agent starts a thread-scoped durable terminal for work that will outlive its current turn.
2. The agent becomes idle while the terminal continues.
3. When the process exits, the owning thread should get one opt-in completion delivery so the agent can inspect the result and continue.
4. The agent should be able to read bounded final output after exit.

The concrete run was terminal `term_ewu75yjnwb`, owned by child thread `thr_ky9svesucn` in environment `env_yvc7c9ytrp`. The process exited with code 2. The child thread stayed idle and did not resume.

## What happens today

Observed with bb desktop 0.39.0 on macOS 15.7.7 arm64:

```text
$ bb terminal show term_ewu75yjnwb --json
{
  "id": "term_ewu75yjnwb",
  "threadId": "thr_ky9svesucn",
  "environmentId": "env_yvc7c9ytrp",
  "status": "exited",
  "exitCode": 2,
  "closeReason": "process-exit"
}

$ bb terminal output term_ewu75yjnwb --json
Error: HTTP 409: Terminal output is unavailable because the session is not running
```

I also audited release 0.40.0 and current `main` at [`31a190ddb5f0ffb86cb8dcd80482497663de2685`](https://github.com/get-bb/bb/commit/31a190ddb5f0ffb86cb8dcd80482497663de2685). The behavior remains:

- The daemon flushes final bytes to clients that are already attached, deletes the in-memory terminal session and scrollback, then sends `terminal.exited` ([terminal manager](https://github.com/get-bb/bb/blob/31a190ddb5f0ffb86cb8dcd80482497663de2685/apps/host-daemon/src/terminals/terminal-manager.ts#L958-L989)).
- The output route rejects non-running sessions with HTTP 409 ([terminal lifecycle](https://github.com/get-bb/bb/blob/31a190ddb5f0ffb86cb8dcd80482497663de2685/apps/server/src/services/terminals/terminal-session-lifecycle.ts#L1119-L1135)).
- A thread-scoped terminal change emits only `terminals-changed`, which is a client cache invalidation. It does not append a thread event or send a provider turn ([terminal notification](https://github.com/get-bb/bb/blob/31a190ddb5f0ffb86cb8dcd80482497663de2685/apps/server/src/services/terminals/terminal-session-lifecycle.ts#L1914-L1919)).
[...]

PR #2951, opened 2026-09-02
Document safe BB CLI automation workflows

ryanbbrown:
## What was wrong

The built-in `bb-cli` skill did not explain five current operating contracts that live help alone does not make safe for automation: argv-only generated message text, per-command JSON shapes, provider-specific permission modes, sender attribution for queued agent messages, and terminal output loss after process exit. A Papercut triage found 65 original occurrences across these five patterns.

## What changed

- Adds quoted staged-file forms for generated `thread spawn` and `thread tell` text, including the shell argument-size limit.
- Tells callers to inspect each command's JSON shape and the selected provider's live permission capabilities.
- Directs queued agent-to-agent reports through `thread tell --mode queue`, which preserves sender attribution, instead of an unattributed queue draft.
- Separates interactive, user-inspectable native terminals from output that must survive process exit.

This intentionally omits the earlier errored-thread workaround because #2715 fixed default `thread tell` recovery before this change.

## How you verified

Reviewed installed BB 0.41.0 help and behavior, then checked current `main` at `cffc82ad4db651f6dad600db33528df4e87348eb` against the command registrations, thread-send logic, queued-message sender handling, provider capabilities, and terminal output lifecycle.

- Focused skill tests — 17 passed
- Turbo CLI/server tests — 280 files and 2,705 tests passed; 1 file skipped
- Turbo CLI/server typecheck — 6 tasks passed
- `git diff --check`

Related to #2620, #2622, and #2520.

> AGENT GENERATED

### DIFF
--- a/apps/server/src/services/skills/builtin-skills/bb-cli/SKILL.md
+++ b/apps/server/src/services/skills/builtin-skills/bb-cli/SKILL.md
@@ -49,6 +49,8 @@ BB_HOST_DAEMON_PORT only for an intentional non-default target.
 - Pass an environment or machine selector when the default host is uncertain.
 - Query provider models on the machine that will run the thread.
 - Prefer non-interactive commands and machine-readable output for automation.
+- Treat each `--json` command as its own schema. Inspect its top-level type
+  before filtering, and do not infer an envelope from another command.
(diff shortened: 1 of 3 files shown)


SawyerHood:

PR #2969, opened 2026-09-02
Ship a host-only artifact for enrolled machines

### LINKED ISSUE #2968: Ship a host-only artifact for enrolled machines
## The workflow

Install or auto-update an enrolled execution machine from its owning bb server while retaining the host daemon, service startup, enrollment, exact-version installation, and the agent-launched `bb` CLI.

## What happens today

`/install/bb-app.tgz` builds and serves the complete `bb-app` distribution, including the server and web application that an enrolled machine never launches. On `main` at `b66c739543d4d78cba3b0ef318a78dc6b57abdc4`, the artifact measured 39,602,419 compressed bytes, 177,646,835 unpacked bytes, and 2,063 files. A forced build took 53.83 seconds and `npm pack` took 16.20 seconds on macOS with Node 22.23.1. Repeating installation always downloads and reinstalls the artifact because version strings do not distinguish identical unpublished builds.

## What you would expect

Serve the smallest practical exact-version host runtime at the existing package route. It should retain `bb`, `bb-app`, and `bb-host-daemon`, support the existing launchd/systemd services and protocol-driven updater, verify downloaded bytes with SHA-256, and return `304 Not Modified` when the identical installed artifact is requested again. The route and updater must remain compatible across a rolling deployment.

## Context and alternatives

The server and app bundles cannot be removed from the published full `bb-app` package because local and desktop installs still need them. A separate host build consumed by the existing route keeps those workflows independent. Searches for enrolled-machine artifact, `bb-app.tgz`, and host-daemon auto-update issues found no duplicate. This does not include ephemeral host allocation or UI changes.

> AGENT GENERATED

ymichael:
## What was wrong

The enrolled-machine installer and protocol updater consumed the same complete `bb-app` tarball as local and desktop installations. That artifact included the server, web application, SDK output, and dependencies a remote execution machine never launches, while the route exposed no content identity that could distinguish an identical unpublished build and safely avoid another download and npm install.

## What changed

- Add a dedicated Turbo `bb-app#build:host` package containing the host daemon, provider/plugin workers, reachable bundled CLI chunks, the `bb`, `bb-app`, and `bb-host-daemon` bins, and only host runtime dependencies. The source-tree route builds it directly; a packaged server materializes the equivalent host package from its full distribution for release compatibility.
- Content-address generated tarballs by SHA-256 and publish the digest through a strong ETag and `x-bb-artifact-sha256` while retaining `/install/version` and `/install/bb-app.tgz`. Conditional requests return `304` with no artifact bytes.
- Make `install.sh` and protocol self-update verify the advertised digest, persist it only after a valid install, and skip downloading/running npm when the identical artifact is already installed. Legacy responses without the digest and the existing PATH/npm-registry fallbacks remain supported.
- Split launcher artifact validation so daemon and CLI entrypoints do not require server/web assets. Keep the existing exact package version, bin names, enrollment flags, launchd/systemd paths, and update prefix.
- Document the host-only distribution and update behavior. No public command or configuration changed, so no CLI guide surface changed.
- `HOST_DAEMON_PROTOCOL_VERSION` remains 177 because the session, WebSocket, and host-RPC contracts are unchanged. The HTTP metadata is additive; new clients accept legacy servers without it and older clients ignore it.

Measured on macOS with Node 22.23.1 and warm npm dependencies:

| Metric | Full artifact | Host artifact | Reduction |
| --- | ---: | ---: | ---: |
| Compressed tarball | 39,602,419 B | 1,236,936 B | 96.9% |
| Unpacked tarball | 177,646,835 B | 5,877,691 B | 96.7% |
| Files | 2,063 | 43 | 97.9% |
| Forced Turbo build | 53.83 s | 13.09 s | 75.7% |
| `npm pack` | 16.20 s | 0.28 s | 98.3% |
| Fresh installed footprint | 267,087,872 B | 65,912,832 B | 75.3% |
| Fresh install | 5.20 s | 0.77 s | 85.2% |

The old route transferred 39,602,419 bytes on every request. The new route transferred 1,236,936 bytes once and returned `304`/zero bytes for the identical repeat. Two independent packs produced SHA-256 `26e365744761a80d434052151181f311caa1d67ca2d24ef6724df814db6fb033`.

## How you verified

The new artifact, installer, route, and updater assertions fail against the previous full-package/no-digest behavior and pass with this change.

- `pnpm exec turbo run test --filter=@bb/server --force -- test/app/bb-app-artifact.test.ts test/app/install-machine-script.test.ts test/app/skeleton.test.ts` — 43 tests passed.
- `pnpm exec turbo run test --filter=@bb/host-daemon --force -- src/protocol-self-update.test.ts` — 15 tests passed.
- `pnpm exec turbo run test --filter=bb-app --force` — 73 tests passed.
- `pnpm exec turbo run typecheck --filter=@bb/server --filter=@bb/host-daemon --filter=bb-app --force` — 7 Turbo tasks passed.
- `pnpm exec turbo run build --filter=bb-app` — 12 Turbo tasks passed.
- `pnpm exec turbo run build:host --filter=bb-app --force` — 6 Turbo tasks passed.
- Installed the exact `0.41.0` host tarball into a clean prefix and exercised all three bins.
- Enrolled the packaged daemon against a fresh local server, observed connected status, and used its bundled `bb machine list --json` successfully.
- Ran the real updater with npm: first install completed in 771 ms and persisted the verified digest; the identical second request returned `304` and completed in 5 ms without npm.
- Upgraded an existing full private-prefix install in place: 267,087,872 B to 65,851,392 B in 1.15 s, with server/web payloads removed and all host bins working.
- Verified shell syntax, Prettier, deterministic packing, artifact contents, response digest, and `git diff --check`.

Fixes #2968

> AGENT GENERATED

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ out
 /build
 /apps/app/build/
 /packages/bb-app/host-daemon/
+/packages/bb-app/host-package/
 /apps/desktop/release/
 /apps/desktop/.electron-builder.generated.json
 
--- a/apps/host-daemon/src/protocol-self-update.test.ts
+++ b/apps/host-daemon/src/protocol-self-update.test.ts
@@ -1,4 +1,5 @@
-import { mkdtemp, rm } from "node:fs/promises";
+import { createHash } from "node:crypto";
+import { mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { delimiter, dirname, join } from "node:path";
 import { HOST_DAEMON_PROTOCOL_VERSION } from "@bb/host-daemon-contract";
@@ -61,7 +62,14 @@ async function createFixture(
     now: args.now,
     serverUrl: args.serverUrl ?? "https://server.example.test",
   });
-  return { fetchFn, installTarball, logger: testLogger, runProcess, updater };
+  return {
+    dataDir,
+    fetchFn,
+    installTarball,
+    logger: testLogger,
+    runProcess,
+    updater,
+  };
 }
 
 afterEach(async () => {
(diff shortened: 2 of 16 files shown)


ymichael:
@slopcop

ymichael:
End-to-end canary completed on the real `mbp-intel` SSH target.

- Started an isolated protocol-177 server from this PR behind an SSH reverse tunnel; used unique remote data, daemon port `49932`, and launch agent `app.getbb.host-daemon.127-0-0-1-49931`, leaving the existing enrollment untouched.
- Ran the actual `/install.sh` without mocks. It downloaded the host artifact, installed 39 packages with npm, enrolled, replaced the temporary daemon with launchd, and reported the machine ready.
- Confirmed remote `/status` was connected at protocol 177, the installed digest matched the PR artifact, and the packaged `bb machine list --json` worked.
- Rebuilt the same temporary server and host artifact at protocol 178. The real protocol-177 daemon received the mismatch, downloaded and installed the new artifact through `/install/bb-app.tgz`, recorded a new digest, exited, and was restarted by launchd. The launchd PID changed from `73454` to `73894`; the replacement reconnected with `connected: true, protocolVersion: 178`. Total observed mismatch-to-reconnect time was 35 seconds.
- Reran the actual installer against the identical protocol-178 artifact. It reported `The identical server host artifact is already installed`, skipped download/npm, retained enrollment, and restarted the launch agent successfully.
- The updater needed its existing 15-second forced-exit fallback because a `PipeWrap` remained active, then launchd recovered normally. This behavior predates the artifact change and did not prevent the update.
- Removed the temporary launch agent, remote data, reverse tunnel, and local server data. Restored source and build outputs to protocol 177. The worktree is clean and `bb machine show Intel --json` remains connected with the original `ymichael.getbb.app` service.

PR #2972, opened 2026-09-02
Avoid redundant launchd restart during host enrollment

ymichael:
## What was wrong

The macOS installer bootstrapped a launch agent configured with both RunAtLoad and unconditional KeepAlive, then immediately ran kickstart -k. launchd already starts that job during bootstrap; -k can kill that just-started process and subject the replacement to launchd's default 10-second spawn throttle. The existing readiness check was therefore delayed by an unnecessary forced restart.

## What changed

- Remove only the macOS kickstart -k step, leaving replacement as bootout followed by bootstrap.
- Keep the existing connected-daemon readiness check authoritative after bootstrap.
- Cover first-time service registration, reinstall/replacement, bootstrap diagnostics, readiness diagnostics, and the exact absence of kickstart in the installer fixture.
- Make temporary executable fixtures independent of any ambient /tmp package type.
- Leave Linux/systemd behavior unchanged.
- No host-daemon wire contract changed, so HOST_DAEMON_PROTOCOL_VERSION is unchanged.

## How you verified

- Confirmed from the local macOS launchd.plist and launchctl manuals that KeepAlive implies RunAtLoad, RunAtLoad launches the job when loaded, kickstart -k kills a running service, and the default spawn throttle is 10 seconds.
- Measured a disposable real launchd agent on Apple Silicon macOS 26.5.1 from bootstrap invocation to successful WebSocket handshake. Three bootstrap-plus-kickstart trials took 10,129 ms, 10,162 ms, and 10,128 ms; three bootstrap-only trials took 139 ms, 71 ms, and 73 ms. This controlled service reproduced the launchd behavior but was not the production bb host daemon.
- Verified the new regression test fails against the old script because it observes the unexpected kickstart.
- pnpm exec turbo run test --filter=@bb/server --force -- test/app/install-machine-script.test.ts (19 passed)
- pnpm exec turbo run build typecheck --filter=@bb/server
- pnpm exec turbo run smoke:tarball --filter=bb-app
- sh -n apps/server/src/assets/install-machine.sh
- Compared the built installer asset byte-for-byte with the source and confirmed it contains no launchctl kickstart.
- Prettier check and git diff --check.

Fixes: N/A (standalone performance improvement)

> AGENT GENERATED

### DIFF
--- a/apps/server/src/assets/install-machine.sh
+++ b/apps/server/src/assets/install-machine.sh
@@ -715,12 +715,6 @@ EOF
     [ -z "$launchctl_error" ] || detail "launchctl: $launchctl_error" >&2
     exit 1
   fi
-  if ! launchctl_error=$(launchctl kickstart -k "gui/$(id -u)/$service_label" 2>&1); then
-    fail_step "The bb host-daemon launch agent was registered, but the daemon did not start."
-    [ -z "$launchctl_error" ] || detail "launchctl: $launchctl_error" >&2
-    detail "See $data_dir/logs/launchd.log for the daemon error." >&2
-    exit 1
-  fi
   if ! wait_for_daemon_connection "the launch agent"; then
     fail_step "The bb host-daemon launch agent started but did not connect to $server_url."
     detail "See $data_dir/logs/launchd.log for the daemon error." >&2
--- a/apps/server/test/app/install-machine-script.test.ts
+++ b/apps/server/test/app/install-machine-script.test.ts
@@ -728,7 +728,7 @@ setInterval(() => {}, 1000);
     expect(result.stderr).toContain("Timed out waiting for host daemon");
   });
 
-  it("installs an idempotent macOS launch agent for joined state", () => {
+  it("starts a fresh macOS launch agent once and replaces it with one new process", () => {
     const fixture = createFixture();
     writeJoinedState(fixture);
     writeServerInstallTools(fixture, 200);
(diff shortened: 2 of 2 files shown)


ymichael:
Top-of-stack E2E passed on `16e6704e8e`.

Stack tested:

`main` → #2969 (`c8c70c35ea`) → #2965 (`7685dd0287`) → #2972 (`16e6704e8e`)

A fresh server was built from the top commit and a uniquely isolated macOS host was enrolled through the unmodified `/install.sh` path.

Cold path:

- full `bb-app` top build: 34.1 s, 12/12 Turbo tasks
- cold enrollment including host-artifact construction/download, npm install, temporary daemon join/shutdown, launchd bootstrap, and connected readiness: 5.164 s
- artifact download stage: 521 ms
- host-only artifact: 1,237,469 bytes
- installed footprint: 64,316 KiB
- temporary-daemon shutdown → service-install stage: approximately 555 ms
- launchd bootstrap → connected readiness: approximately 1.094 s
- installed artifact digest matched the server artifact
- installed daemon bundle matched the top checkout byte-for-byte

Single-start proof:

- launchd PID remained unchanged during a five-second observation window
- exactly one persistent `Host daemon started` record appeared after enrollment
- server sessions were exactly temporary enrollment + one active launchd service; there was no redundant third session

Lifecycle check:

- SIGTERM → clean process exit: 600 ms
- SIGTERM → replacement launchd PID: 617 ms
- SIGTERM → connected again: 1.450 s
- exactly one replacement start occurred

Warm reinstall/update path:

- conditional artifact request returned the identical-artifact path in 28 ms with no download or npm install
- full reinstall → launchd connected readiness: 1.723 s
- the resulting PID stayed stable and exactly one active session remained

Targeted logs contained no shutdown watchdog force-exit, SIGKILL, or `forceKilled: true` event.

PR #2982, opened 2026-09-03
feat: drag to reorder the new tab Actions list

nlorio-notion:
## Context

The new tab shows an Actions list in the side panel. The list holds up to two built-in actions: Open browser and Start terminal. The list also holds one row for each action that a plugin registers. The user starts work from these rows many times each day. The sidebar already lets the user drag its sections and its plugin rows into a new order.

## Problem

The Actions list used a fixed order. It showed Open browser first, then Start terminal, then the plugin actions in registration order. This order caused three problems:

1. The user cannot move a frequent action to the top of the list.
2. A plugin action stays below both built-in actions.
3. The order changes when a plugin registers a new action.

## Solution

Each row now has a reorder handle. The user drags the handle to move the action. The app keeps the new order for that client.

| The list at rest | The handle after hover |
| --- | --- |
| The Actions list at rest | The reorder handle after the pointer moves onto the row |

| The row during the drag | The list after the drop |
| --- | --- |
| The row moves and the list shows the new order | The action is now the first row |

The user drags the last action to the top of the list with its handle

I recorded this media from a dev app on this branch. A temporary local plugin supplies the Quickstart row and the Subagents row. That plugin registers two `experimental_newThreadPanelAction` slots. The list therefore holds both built-in actions and plugin actions.

The change has six parts:

1. **A new module.** `apps/app/src/components/secondary-panel/NewTabActions.tsx` holds the Actions list. The list was in `NewTabFileSearch.tsx`, which has more than 1,200 lines. Both callers and both test files already imported the list as one unit.
2. **One row shape.** Each action is one flat `NewTabAction` value with a ready `icon` node. Built-in rows and plugin rows use the same `NewTabActionRow`. The row holds the action button, the shortcut hint, the reorder handle, and the optional trailing control as siblings. `LauncherTile` loses the `variant` prop and the `ariaKeyshortcuts` prop, because only the action rows used them.
3. **A handle, not a whole row.** The handle starts the drag. `ProvidersSettingsSection` and `QueuedMessagesList` already use a handle in the same way. A handle also gives keyboard reorder, and it keeps a click on the row unambiguous. The app hides the handle until the pointer moves onto the row, and it always shows the handle on a touch device.
4. **Shared drag behavior.** The list uses the general `useReorderDnd` hook and the shared sortable motion. The general hook owns the sensors, drag threshold, Escape key behavior, vertical movement, and click suppression. The sidebar wrapper keeps its compact-drawer touch policy. The Actions list uses the standard touch sensor, so touch reorder works while the sidebar drawer is closed.
5. **Storage.** The app writes the order to `bb.newTab.actionOrder` in `newTabActionsAtoms.ts`. This is the `atomWithStorage` pattern that the sidebar section order and the plugin nav order already use. The app puts a new action after the saved rows. The app also keeps the slot of an action whose plugin is absent, so the position survives a reload.
6. **Shared order helpers.** `apps/app/src/lib/stored-order.ts` holds `arrangeByStoredOrder`, `reorderStoredOrder`, and `haveSameOrder`. The plugin nav sidebar already had this logic, so this change moves the logic instead of a copy. `reorderStoredOrder` uses `arrayMove` from `@bb/client-core`. `haveSameOrder` moves here from `usePersistedSidebarSectionOrder`. After the delegation, `reorderPluginNavPanels` and `havePluginNavPanelOrdersDiverged` only renamed arguments, so this change deletes them and the tests that only tested them.

The list keeps its cost low when the user cannot reorder it. The app orders the rows by id. The app mounts `DndContext`, its sensors, and the document key listener only for two or more actions. For an empty list, the component returns before this work.

## Result

The user can now put the Actions rows in any order, and the app keeps that order after a reload.

Automatic checks:

1. `pnpm exec turbo run typecheck --filter=@bb/app` passes after the review fixes.
2. The 11 focused action and sidebar drag tests pass after the review fixes.
3. `pnpm exec turbo run build --filter=@bb/app` passes.
4. Lint reports 175 warnings and 0 errors. This count equals the count on `main`.
5. 898 tests pass in the plugin, secondary-panel, sidebar, and settings suites.

New tests:

1. `apps/app/src/lib/stored-order.test.ts` tests the fallback to registration order, a new id at the end, the kept slot of an absent id, a duplicate stored id, and the reorder and decline cases.
2. `NewTabActions.test.tsx` tests saved and new actions, handle availability, compact touch setup with the sidebar closed, and invalid saved data.

Manual checks in a dev app on this branch:

1. Move the pointer onto a row. The app shows the handle.
2. Drag the last action to the top with its handle. The list shows the new order.
3. Read `bb.newTab.actionOrder`. The value matches the new order.
4. Reload the page. The list keeps the order.
5. Drag the action back to the bottom. The app does not run the action.
6. Click the row. The app runs the action.

This change does not send new data between the server and the host daemon, so it does not change `HOST_DAEMON_PROTOCOL_VERSION`. It also adds no CLI command and no SDK method. The sidebar section order and the plugin nav order are client-local in the same way. Tell me if you want an agent-facing surface, and I will add one.

One follow-up stays out of this change. Three call sites repeat the same check of `event.over` and the same string check of `event.active.id`: this list, `PluginNavSidebarItems`, and `ReorderableSidebarSectionOrderList`. An `onReorder(activeId, overId)` option on the general hook and the sidebar wrapper would remove the repeat. That option changes every reorder caller, so it needs its own change.

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/plugin/PluginNavSidebarItems.tsx
+++ b/apps/app/src/components/plugin/PluginNavSidebarItems.tsx
@@ -62,12 +62,11 @@ import {
 import {
   arrangePluginNavPanels,
   getPluginNavPanelKey,
-  havePluginNavPanelOrdersDiverged,
   hidePluginNavPanel,
-  reorderPluginNavPanels,
   seedLeadingNavPanelKeys,
   showPluginNavPanel,
 } from "./pluginNavSidebarOrder";
+import { haveSameOrder, reorderStoredOrder } from "@/lib/stored-order";
 
 const BUILTIN_NAV_ROW_PLUGIN_ID = "__builtin__";
 
@@ -153,7 +152,7 @@ function PluginNavSidebarItemList({
   }, [hiddenKeys, rows, storedOrder]);
 
   useEffect(() => {
-    if (!havePluginNavPanelOrdersDiverged(storedOrder, normalizedOrder)) return;
+    if (haveSameOrder(storedOrder, normalizedOrder)) return;
     setStoredOrder(normalizedOrder);
   }, [normalizedOrder, setStoredOrder, storedOrder]);
 
@@ -171,11 +170,11 @@ function PluginNavSidebarItemList({
       ) {
         return;
       }
-      const nextOrder = reorderPluginNavPanels({
-        activeKey: event.active.id,
-        overKey: event.over.id,
+      const nextOrder = reorderStoredOrder({
+        activeId: event.active.id,
+        overId: event.over.id,
         order: normalizedOrder,
-        visibleKeys,
+        visibleIds: visibleKeys,
       });
       if (nextOrder) setStoredOrder(nextOrder);
     },
--- a/apps/app/src/components/plugin/plugin-slot-mounts.test.tsx
+++ b/apps/app/src/components/plugin/plugin-slot-mounts.test.tsx
@@ -68,7 +68,7 @@ import {
   usePluginPanelActions,
   type OpenPluginPanelArgs,
 } from "./PluginPanelActions";
-import { NewTabActions } from "@/components/secondary-panel/NewTabFileSearch";
+import { NewTabActions } from "@/components/secondary-panel/NewTabActions";
 import { buildFileOpenerPanelTab } from "./file-opener-tabs";
 import { splitLayoutAtom } from "@/lib/split-layout/atoms";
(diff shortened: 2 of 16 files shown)


SawyerHood:
@slopcop take a look

SawyerHood:
This lgtm! I'll just address the slopcop issues myself and merge! Ty for the contribution

PR #2984, opened 2026-09-03
Make mobile toast swipes reliable

ymichael:
## What was wrong

Compact toasts render at `top-center`, while Sonner derives an upward-only swipe direction from that position and processes the first pointer movement before its React-managed axis has updated. The archived-thread toast exposed a second, more important gap: its linked thread title, Undo action, and close control are buttons, and Sonner deliberately does not begin a swipe when the pointer target is a button. That made most of the real "Thread Archived" toast a swipe dead zone even though plain diagnostic toasts felt reliable.

## What changed

- Allow compact toasts to use Sonner's native top, left, and right swipe directions.
- Keep the CSS swipe distance continuous while Sonner's axis state catches up, preserving short Android touch movements without replacing Sonner's threshold, animation, callbacks, or toast identity.
- Bridge touch starts on toast buttons into Sonner's normal swipe handling, then suppress the button click only after movement exceeds tap slop. Normal taps still invoke the thread-title, Undo, and close actions.
- Add regression coverage using the actual archived-thread toast layout, including tap-versus-drag behavior, continuous batched movement, directional dismissal, rejected downward drags, and rapid stacked-toast identity.

There is no server/host-daemon wire change, so `HOST_DAEMON_PROTOCOL_VERSION` is unchanged. No CLI, SDK, guide, or user-facing configuration surface changed.

## How you verified

- Physical Android Chrome 152: archived a thread and swiped from the interactive title row; the user confirmed the interaction felt great. Captured movement stayed continuous through `130.8px`, and Sonner reported `direction: right`, `removed: true`, and `swipeOut: true` at pointer-up.
- Physical Android Chrome 152 diagnostic runs: both user-tested runs felt great and completed native swipe-out dismissal.
- Doobie with a 390×844 touch viewport: 10/10 single-move left, right, up, and upward-diagonal flicks dismissed.
- iPhone 16 Pro, iOS 18.4 Simulator, Mobile Safari: left, right, up, and both upward diagonals dismissed.
- `pnpm exec turbo run test --filter=@bb/app --force -- --run src/components/AppToaster.test.tsx` — 8 tests passed.
- `pnpm exec turbo run lint --filter=@bb/app` — passed with 178 existing warnings in unrelated files and 0 errors.
- `pnpm exec oxfmt apps/app/src/components/AppToaster.tsx apps/app/src/components/AppToaster.test.tsx` and `git diff --check` — passed.

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/AppToaster.test.tsx
+++ b/apps/app/src/components/AppToaster.test.tsx
@@ -7,7 +7,7 @@ import {
   render,
   waitFor,
 } from "@testing-library/react";
-import { afterEach, describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import { toast } from "sonner";
 import { CompactViewportOverrideProvider } from "@bb/shared-ui/hooks/use-compact-viewport";
 import { AppToaster } from "./AppToaster";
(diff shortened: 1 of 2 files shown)


ymichael:
@slopcop

PR #2989, opened 2026-09-03
fix(app): verify inline Markdown paths before linking

smsunarto:
## What was wrong

Inline code spans that looked like Markdown file paths (for example `docs/notes.md`) were turned into local-file links purely from their shape. Clicking one for a file that does not exist on the host opened an empty file tab. The renderer had no way to check the path because it did not know which host the thread ran on.

## What changed

- `markdown-link-routing.ts`: `MarkdownLocalFileLinkRouting` gains an optional `inlineCodeFileLinks.hostId` that says which host to check paths against.
- `markdown-preview.tsx`: `resolveInlineCodeMarkdownFileHref` becomes `resolveInlineCodeMarkdownFileLink` and returns the resolved absolute path alongside the href. A new `MarkdownInlineCodeFileLink` component calls `useHostPathExistence` for that path and only renders the anchor when the host reports the file exists. Without a host id the span stays plain code.
- Thread timeline (`ThreadTimelineSurface` → `ThreadTimelineRows` → `ConversationMessageContent`) and `ThreadDetailView` thread the environment's `hostId` into the routing so assistant messages and file previews get the existence check.

No wire changes.

## How you verified

- Added `links only inline-code Markdown file paths that exist` in `apps/app/src/components/ui/markdown-preview.test.tsx`. It renders two inline-code paths, one present and one missing on the host, and asserts only the present one becomes a link. Fails before, passes after.
- `pnpm exec turbo run test --filter=@bb/app`

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/thread/timeline/ConversationMessageContent.tsx
+++ b/apps/app/src/components/thread/timeline/ConversationMessageContent.tsx
@@ -125,6 +125,7 @@ const STREAMING_TAIL_MARKDOWN_CLASS_NAME =
 interface ConversationMessageContentAssistantProps
   extends ConversationMessageContentBaseProps, AssistantMessageRowIdentity {
   role: "assistant";
+  localFileHostId?: string;
   onOpenLink?: ThreadTimelineLinkHandler;
   onAddToChat?: ThreadTimelineAddToChatHandler;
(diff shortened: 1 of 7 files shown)


SawyerHood:

PR #2990, opened 2026-09-03
Contain linked project skill roots inside the repository

### LINKED ISSUE #2769: A linked project skill root is scanned outside the repository
### Summary

A non-recursive project skill root that is itself a symlink is scanned wherever it
points. Pointing `.claude/skills` at a directory outside the workspace makes bb load
every skill under it into the project's skill list. The recursive root shape already
resolves its root against the project boundary before scanning; the flat shape never
does, so the containment rule the code states for project roots is not enforced on the
path most providers actually use.

### Versions and environment

- bb `main` @ `8d926c312569b63924825db761e9cc73663bc3bc`, run from source via `scripts/bb-dev-app current`
- Also present in the packaged desktop app 0.40.0
- macOS 26.4 (darwin 25.4.0), Node v26.4.0
- Providers: `claude-code` (claude 2.1.251), `codex` (codex-cli 0.150.1), `pi` (0.84.4)

### Steps to reproduce

```bash
ROOT=$(mktemp -d)
mkdir -p "$ROOT/outside/leaked" "$ROOT/repo/.claude"
printf -- '---\nname: leaked\ndescription: outside the repo\n---\n' \
  > "$ROOT/outside/leaked/SKILL.md"
git -C "$ROOT/repo" init -q .
ln -s "$ROOT/outside" "$ROOT/repo/.claude/skills"     # root itself is the link

bb project create --name leak-demo --root "$ROOT/repo" --json
bb project commands <id> --provider claude-code --json
```

`leaked` is listed with `"origin": "project"`.

Did **not** reproduce with: a root link that stays inside the repository (correct — that
shape is supported and should keep working); a recursive root shape, which already
rejects the same escape (`.cursor/skills` declared with `recursive: true`).

### Expected vs actual

```text
Expected: (no project skills — the link leaves the repository)
Actual:   ['leaked']
```

Fixture matrix, `bb project commands --provider claude-code`, project-origin rows only:

```text
                                     main 8d926c3     with boundary check
projC  .claude/skills -> outside      ['c-outside-root']   (none)
projB  .claude/skills -> in-repo      ['b-via-root-link']  ['b-via-root-link']
projA  real dirs + entry links        ['c-real']           ['c-real']
```

Only the escaping root changes; every other shape is untouched.

### Evidence

- `scanSkillRootFiles` reads the root with no boundary test:
  https://github.com/get-bb/bb/blob/8d926c312569b63924825db761e9cc73663bc3bc/apps/host-daemon/src/command-discovery.ts#L250-L257
- The recursive shape does apply the test, against the same `boundaryPath`:
  https://github.com/get-bb/bb/blob/8d926c312569b63924825db761e9cc73663bc3bc/apps/host-daemon/src/command-discovery.ts#L323-L340
- `boundaryPath` is already populated for every project root, so no contract change is
  needed to consult it:
  https://github.com/get-bb/bb/blob/8d926c312569b63924825db761e9cc73663bc3bc/apps/host-daemon/src/command-handlers/list-commands.ts#L173-L177
  https://github.com/get-bb/bb/blob/8d926c312569b63924825db761e9cc73663bc3bc/apps/host-daemon/src/command-handlers/list-commands.ts#L210-L215
- The stated intent this violates ("a checked-in (`project`) plugin must not point bb
  outside the repository"):
  https://github.com/get-bb/bb/blob/8d926c312569b63924825db761e9cc73663bc3bc/packages/plugin-sdk/src/vendor-plugin-roots.ts#L17-L21
- Investigation thread: `thr_fuwi9eq7iz`

### What you ruled out

- Not #2602 — that is the opposite direction (links wrongly *skipped*, command files, `walkMarkdownTree`). This is a link wrongly *followed*, on the skill root path.
- Not the entry-level symlink rule — entry links are separately over-restricted; that is filed as its own report and does not change this behavior either way.
- Not provider-specific: any provider declaring a non-recursive project skill root is affected (`claude-code` `.claude/skills`, `codex` `.codex/skills`, `pi` `.pi/skills`).
- Not a packaging artifact: reproduces from source on `main` and in packaged 0.40.0.

Medium — a checked-in symlink silently widens what bb loads into a project's skill list
beyond the reviewed repository contents; no data loss, and it requires a committed
symlink. Effort Low: one boundary check on a path where `boundaryPath` is already
plumbed. Consider the `security` label.

### Checks

> AGENT GENERATED

ylcn91:
## What was wrong

`scanSkillRootFiles` opened a non-recursive project skill root before
resolving it against the workspace boundary, so a root that is itself a
symlink was scanned wherever it pointed. A checked-in `.claude/skills`
link to a directory outside the repository put every skill under it into
the project's skill and command lists, with `origin: "project"`. The
recursive shape already resolved the same plumbed `boundaryPath` before
walking, and `docs/provider-plugin-api.md` already states the rule for
both shapes: a symlink out of a project root is followed within the
workspace for a plain root and within the repository root for a root
that walks ancestors or that the plugin resolved.

Issue: #2769. Investigation report: https://get-bb.github.io/reports/issues/2769.html

## What changed

`resolveRecursiveRootPath` becomes `resolveBoundedRootPath` and the flat
skill root scan consults it, so both shapes enforce one containment rule
against the same `boundaryPath`. A root link that stays inside the
repository still resolves and is still scanned, and user-origin roots,
which carry no boundary, are unaffected.

No field or message on the server/host-daemon wire changed, so
`HOST_DAEMON_PROTOCOL_VERSION` stays at 178.

## How you verified

- Added two `command-discovery` regressions: a project `.claude/skills`
  root symlinked outside the workspace discovers nothing, and one
  symlinked to `.agents/skills` inside the workspace still discovers its
  skill. The first fails on the parent commit
  (`expected [ { name: 'leaked', …(4) } ] to deeply equal []`) and
  passes here; the second passes on both.
- `pnpm exec turbo run test --filter=@bb/host-daemon --force`: 46 files,
  574 tests passed (1 failed before the fix).
- `pnpm exec turbo run typecheck --filter=@bb/host-daemon`: passed.
- `BB_TEST_TIMEOUT_SCALE=3 pnpm exec turbo run test
  --filter=@bb/integration-tests --force --env-mode=loose --
  native-roots-golden`: 22 provider golden fixtures passed unchanged,
  including the cursor symlink-boundary and grok symlink variants.
- Drove `listHostCommands` with the shipping claude-code declaration
  (`.claude/skills`, `ancestors: true`) over the issue's fixture: the
  escaping root reported `["leaked"]` before and `[]` after, while the
  in-repository link reported `["in-repo"]` in both.

Fixes #2769

> AGENT GENERATED

### DIFF
--- a/apps/host-daemon/src/command-discovery.test.ts
+++ b/apps/host-daemon/src/command-discovery.test.ts
@@ -498,6 +498,56 @@ describe("discoverProviderCommands over declared roots", () => {
     ]);
   });
 
+  it("rejects a project skill root linked outside the workspace", async () => {
+    const fixture = await makeWorkspaceFixture();
+    const outsideRoot = path.join(tempRoot, "outside-skill-root");
+    const linkedRoot = path.join(fixture.cwd, ".claude", "skills");
+    await writeFileEnsuringDir(
+      path.join(outsideRoot, "leaked", "SKILL.md"),
+      skillFile("leaked"),
+    );
+    await mkdir(path.dirname(linkedRoot), { recursive: true });
+    await symlink(outsideRoot, linkedRoot, "dir");
+
+    const commands = await discover(
+      fixture,
+      fixture.cwd,
+      nativeRoots({ skills: { project: [declared(".claude/skills")] } }),
+    );
+
+    expect(commands).toEqual([]);
+  });
+
+  it("discovers skills through a project skill root symlink inside the workspace", async () => {
+    const fixture = await makeWorkspaceFixture();
+    await writeFileEnsuringDir(
+      path.join(fixture.cwd, ".agents", "skills", "in-repo", "SKILL.md"),
+      skillFile("in-repo", "Inside the repository"),
+    );
+    await mkdir(path.join(fixture.cwd, ".claude"), { recursive: true });
+    await symlink(
+      path.join("..", ".agents", "skills"),
+      path.join(fixture.cwd, ".claude", "skills"),
+      "dir",
+    );
+
+    const commands = await discover(
+      fixture,
+      fixture.cwd,
+      nativeRoots({ skills: { project: [declared(".claude/skills")] } }),
+    );
+
+    expect(commands).toEqual([
+      {
+        name: "in-repo",
+        source: "skill",
+        origin: "project",
+        description: "Inside the repository",
+        argumentHint: null,
+      },
+    ]);
+  });
+
   it("does not follow project-origin symlinked skill directories or skill files", async () => {
     const fixture = await makeWorkspaceFixture();
     const skillsRoot = path.join(fixture.cwd, ".agent", "skills");
--- a/apps/host-daemon/src/command-discovery.ts
+++ b/apps/host-daemon/src/command-discovery.ts
@@ -250,6 +250,9 @@ async function hasManifestMarker(
 async function scanSkillRootFiles(
   root: CommandScanDirectoryRoot,
 ): Promise<SkillFileMatch[]> {
+  if ((await resolveBoundedRootPath(root)) === null) {
+    return [];
+  }
   const entries = await readDirEntries(root.rootPath);
   if (entries === null) {
     return [];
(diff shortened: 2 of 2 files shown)


SawyerHood:

PR #2991, opened 2026-09-03
Match the submit-a-plugin entry example to the marketplace schema

### LINKED ISSUE #2508: submit-a-plugin skill: engines example doesn't match get-bb/marketplace entry schema
## What's wrong

The built-in `submit-a-plugin` skill's "Prepare the entry" section shows this example:

```json
{
  "id": "notes",
  "displayName": "Notes",
  "description": "Keeps project notes beside each bb thread.",
  "icon": { "url": "./icons/notes-1234abcd.svg" },
  "tags": ["notes", "interface"],
  "author": { "name": "Acme", "github": "acme", "url": "https://acme.example" },
  "engines": {
    "bb": ">=0.40.0",
    "bbPluginSdk": ">=0.5.0"
  },
  "source": { "git": { "url": "https://github.com/acme/bb-plugin-notes.git", "range": "^1.2.3" } }
}
```

The live `get-bb/marketplace` entry schema (`schema/marketplace.schema.json`, `$defs.entry`) declares `additionalProperties: false` and does not list `engines` among its properties (`id`, `displayName`, `description`, `icon`, `tags`, `author`, `source` only). An entry that includes `engines` fails `npm run build` / `npm run check` in the marketplace repo.

## How this surfaced

Ran the skill for real across 4 independent plugin submissions this week (get-bb/marketplace#123, #124, #125, #127). Every one hit the same failure independently, because each one followed the skill's own worked example literally before checking the live schema. The skill's own prose says to treat the live marketplace repo as the source of truth over the skill text — but the worked example still shows the stale field, so it's an easy trap even though the skill's own instructions warn against it.

## Suggested fix

Drop `engines` from the worked example in the "Prepare the entry" section (and the "Use this shape as a guide" JSON block), or add a note that `engines` was removed from the live schema and to always confirm current required/optional entry fields against `schema/marketplace.schema.json` before writing the entry.

Not filing a PR for this — just flagging, since it wasn't clear which repo canonically owns the skill markdown source (found a copy under `apps/server/src/services/skills/builtin-skills/submit-a-plugin/` in a local reference checkout, but no `SKILL.md` there, only a `scripts/` subfolder).

ylcn91:
## What was wrong

The `submit-a-plugin` skill's worked marketplace entry still carried an
`engines` object. The published entry contract sets
`additionalProperties: false` and never defined `engines`: a listing
declares no compatibility, and bb reads `engines.bb` and
`engines.bbPluginSdk` from the plugin's own `package.json` instead, as
`docs/plugin-marketplace-plan.md` describes. Anyone who copied the
example produced an entry that `npm run build` and `npm run check`
reject in get-bb/marketplace. Nothing guarded the example against the
schema, so the field survived the schema change.

Issue: #2508. Investigation report: https://get-bb.github.io/reports/issues/2508.html

## What changed

- `references/marketplace-entry.md`: drops `engines` from the worked
  entry, and replaces the "copy honest engine ranges" prose with the
  actual rule, that compatibility lives in the plugin manifest.
- `references/pull-request.md`: drops the "Entry engine ranges do not
  exceed manifest ranges" review item, which checked a field an entry
  cannot have.
- `test/skills/submit-a-plugin.test.ts`: validates the worked entry
  against `apps/web/public/schemas/marketplace-v2.schema.json`, the
  document get-bb/marketplace compiles each entry file against.

The remaining example fields (`id`, `displayName`, `description`,
`icon`, `tags`, `author`, `source`) already match the schema.

## How you verified

- `pnpm exec turbo run test --filter=@bb/server -- test/skills/submit-a-plugin.test.ts`
  passes. Re-adding `engines` to the example fails the new test with
  `must NOT have additional properties`.
- `pnpm exec turbo run typecheck --filter=@bb/server` passes.
- Validated the corrected example against `schema/marketplace-v2.schema.json`
  fetched from get-bb/marketplace@main; that file is byte-identical to the
  copy this repo publishes.

Fixes #2508

> AGENT GENERATED

### DIFF
--- a/apps/server/src/services/skills/builtin-skills/submit-a-plugin/references/marketplace-entry.md
+++ b/apps/server/src/services/skills/builtin-skills/submit-a-plugin/references/marketplace-entry.md
@@ -9,14 +9,16 @@ Create entries/<plugin-id>.json. The filename, entry ID, and plugin manifest ID
 must match. Do not add fields that the schema does not define.
 
 The current required fields are id, displayName, description, icon, author, and
-source. Use tags and engines when they add useful search or compatibility data.
+source. Use tags when they add useful search data.
 
 Use the product name for displayName. Write one concrete description sentence
 that states the feature and user value. Do not make subjective marketing claims.
 
 Use no more than ten specific lowercase tags. Use hyphens inside multiword tags.
-Copy honest engine ranges from the plugin manifest. An entry can narrow those
-ranges but cannot widen them.
+
+An entry declares no compatibility range. BB reads engines.bb and
+engines.bbPluginSdk from the plugin's own package.json and refuses the install
+there. Keep those ranges honest in the plugin manifest instead.
 
 Set author.github to the account that opens the pull request. Get it with:
 
@@ -39,10 +41,6 @@ Use this shape only as a guide:
     "name": "Acme",
     "github": "acme",
     "url": "https://acme.example"
-  },
-  "engines": {
-    "bb": ">=0.40.0",
-    "bbPluginSdk": ">=0.5.0"
   },
   "source": {
     "git": {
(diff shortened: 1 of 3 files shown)


SawyerHood:
I fixed this independently! Closing out.

PR #3001, opened 2026-09-03
Fix bb:dev CLI arg parsing when pnpm forwards a literal --

### LINKED ISSUE #2998: pnpm run bb:dev -- <cmd> <positional> --<flag> fails: pnpm forwards a literal -- that Commander treats as end-of-options
## Summary

`pnpm run bb:dev -- <command> <positional> --<flag>` fails for any CLI command that takes a positional argument followed by a flag, with `error: too many arguments`. The root cause is not bb's own argument parsing: pnpm 9.15.0 forwards the `--` separator itself as a literal argument to the underlying script, and Commander.js treats that literal `--` as "end of options," so the trailing flag gets parsed as an extra positional instead of an option.

## Versions and environment

- bb from source at commit `99c0ad71841ff6ff2d42b3f7864b6dba0b0f7337` (`main`)
- pnpm 9.15.0 (`packageManager` pin in `package.json`), Node v22.19.0
- macOS 25.5.0 (Darwin), arm64
- No provider/agent involved; pure CLI dev-workflow bug

## Steps to reproduce

1. `pnpm run --silent bb:dev -- machine show <id> --json`
2. Observe: `error: too many arguments for 'show'. Expected 1 argument but got 2.`

Also fails reordered (`pnpm run --silent bb:dev -- machine show --json <id>`) with the same error, and fails for other commands with the same positional-then-flag shape (`machine join-code --json`, `machine remove <id> --yes`).

Does **not** reproduce when the same args are handed to the compiled CLI directly, bypassing pnpm's `--` forwarding:
```
node apps/cli/dist/index.js machine show <id> --json
```
That runs fine (only fails later, on "machine not found," for an unrelated reason: I used a fake id).

Minimal isolation, no bb code involved at all:
```jsonc
// package.json
"scripts": { "debug:argv": "node -e 'console.error(process.argv.slice(2))'" }
```
```
$ pnpm run --silent debug:argv -- foo bar
[ '--', 'foo', 'bar' ]

$ npm run debug:argv -- foo bar
[ 'foo', 'bar' ]
```
pnpm includes the literal `--` token in argv passed to the script; npm strips it. bb's `packages/scripts/src/commands/run-cli.ts` forwards `process.argv.slice(2)` straight through to the spawned CLI process ([`resolveCliExecution`, lines 17-33](https://github.com/get-bb/bb/blob/99c0ad71841ff6ff2d42b3f7864b6dba0b0f7337/packages/scripts/src/commands/run-cli.ts#L17-L33)), so that leading `--` lands as a real CLI argument: `node apps/cli/dist/index.js -- machine show <id> --json`. Commander then treats everything after that `--` as literal positionals, so `--json` becomes a second positional for the `show <id-or-name>` command ([defined here](https://github.com/get-bb/bb/blob/99c0ad71841ff6ff2d42b3f7864b6dba0b0f7337/apps/cli/src/commands/machine.ts#L145-L157)), which only accepts one.

Confirmed directly:
```
$ node apps/cli/dist/index.js -- machine show <id> --json
error: too many arguments for 'show'. Expected 1 argument but got 2.

$ node apps/cli/dist/index.js machine show <id> --json
Error: Machine '<id>' was not found. Available machines: ...
```

## Expected vs actual

**Actual:** any `pnpm run bb:dev -- <command> <positional> --<flag>` invocation fails at the parser with `error: too many arguments for '<command>'. Expected N but got N+1`, before bb's own command logic ever runs.

**Expected:** `pnpm run bb:dev -- machine show <id> --json` behaves the same as calling the compiled CLI directly with those same logical arguments.

## Evidence

- `package.json` script: [`"bb:dev"`, line 48](https://github.com/get-bb/bb/blob/99c0ad71841ff6ff2d42b3f7864b6dba0b0f7337/package.json#L48).
- `run-cli.ts` forwarding site: [`resolveCliExecution`, lines 17-33](https://github.com/get-bb/bb/blob/99c0ad71841ff6ff2d42b3f7864b6dba0b0f7337/packages/scripts/src/commands/run-cli.ts#L17-L33).
- Measured by direct comparison of `pnpm run` vs `npm run` argv forwarding for the same script and args (above); the discrepancy is 100% reproducible across repeated runs.

## What you ruled out

- Not a bug in bb's Commander command definitions: calling the compiled CLI directly with identical logical arguments (no `--` prefix) works correctly.
- Not fixed by reordering the flag before the positional; both orders fail identically once the stray `--` is present.
- Not specific to `machine show`: `machine join-code --json` and `machine remove <id> --yes` fail the same way, and the minimal `node -e` repro shows it has nothing to do with bb's command definitions at all.
- Searched open and closed issues for this; found none.

Low-to-medium — no data loss, but it silently breaks routine `bb:dev` CLI usage and scripting for anyone who follows the normal `pnpm run bb:dev -- <args>` convention with a trailing flag, with no obvious workaround short of calling the compiled CLI directly or hitting the HTTP API instead. Effort looks Low: strip a single leading literal `--` token from `cliArgs` in `resolveCliExecution` before forwarding it to the spawned CLI process.

> AGENT GENERATED

technicalpickles:
## What was wrong

`pnpm run bb:dev -- <command> <positional> --<flag>` failed with `error: too many arguments`, for any CLI command that takes a positional argument followed by a flag. pnpm 9.15.0 forwards the `--` separator itself as a literal argument to the underlying script (npm strips it), and `resolveCliExecution` in `packages/scripts/src/commands/run-cli.ts` passed `process.argv.slice(2)` straight through to the spawned CLI process. Commander then saw that leading `--` and treated everything after it as "end of options," so a trailing flag got parsed as an extra positional argument instead of an option.

## What changed

`resolveCliExecution` now strips a single leading `--` from the forwarded args before spawning the CLI, matching the fix proposed in the issue.

## How you verified

- Added a test in `packages/scripts/test/run-cli.test.ts` asserting a leading `--` is stripped; confirmed it fails on the old code and passes after the fix.
- `pnpm exec turbo run test --filter=@bb/scripts` and `pnpm exec turbo run typecheck --filter=@bb/scripts` both pass.
- Reproduced the exact repro from the issue end-to-end: `pnpm run --silent bb:dev -- machine show fake-id --json` no longer fails at the parser (`too many arguments for 'show'`); it now proceeds past argument parsing to the expected "Cannot connect to BB server" error.

Fixes #2998

> AGENT GENERATED

### DIFF
--- a/packages/scripts/src/commands/run-cli.ts
+++ b/packages/scripts/src/commands/run-cli.ts
@@ -15,8 +15,10 @@ const packageRoot = resolve(commandDir, "..", "..");
 const repoRoot = resolve(packageRoot, "..", "..");
 
 export function resolveCliExecution(
-  cliArgs: string[] = process.argv.slice(2),
+  rawCliArgs: string[] = process.argv.slice(2),
 ): CliExecution {
+  const cliArgs =
+    rawCliArgs[0] === "--" ? rawCliArgs.slice(1) : rawCliArgs;
   const env = { ...process.env };
   if (process.env.NODE_ENV !== "production") {
     const devEnv = resolveCurrentDevProcessEnv(repoRoot, process.env);
--- a/packages/scripts/test/run-cli.test.ts
+++ b/packages/scripts/test/run-cli.test.ts
@@ -52,4 +52,24 @@ describe("run-cli", () => {
     expect(execution.env.BB_SERVER_URL).toBe("http://localhost:4444");
     expect(execution.env.BB_HOST_DAEMON_PORT).toBe("5555");
   });
+
+  it("strips the leading -- that pnpm forwards to scripts", () => {
+    vi.stubEnv("NODE_ENV", "production");
+
+    const execution = resolveCliExecution([
+      "--",
+      "machine",
+      "show",
+      "some-id",
+      "--json",
+    ]);
+
+    expect(execution.args).toEqual([
+      "apps/cli/dist/index.js",
+      "machine",
+      "show",
+      "some-id",
+      "--json",
+    ]);
+  });
 });


SawyerHood:
Oooops sorry I merged the slop-cop generated version of this PR!

PR #3013, opened 2026-09-03
Bound and clear the ACP compaction prose buffer

bradhallett:
> AGENT GENERATED

## Human comments

Follow-up to #2293, addressing the two review findings SlopCop left on the merged commit (`eece92f8`):

1. **Major — "OMP text rule applies to all ACP agents":** already fixed on main by #2957, which moved end-turn compaction classification behind the dialect (`compactionOutcomeForEndTurn(dialect, …)` with `dialect.id !== "omp"` short-circuiting to `completed`). No further change needed; this PR builds on that shape.
2. **Medium — "unbounded compaction message text, no cleanup in finishCompaction":** still open on main; this PR fixes it.

## What was wrong

- `handleAgentNotification` accumulated every `agent_message_chunk` streamed during a compaction prompt into `session.compactionAgentMessage` for **every** dialect, even though only the omp dialect ever reads the text — unbounded growth on a long-lived session.
- The accumulation had no size limit.
- The buffer was only reset at `startCompaction`, so after a compaction completed, the text stayed resident until the next compaction began.

## What changed

- Accumulation is now gated on `collectsCompactionAgentMessage(session.dialect)` — the dialect module owns the "does this dialect parse compaction prose" decision, so non-omp dialects carry no classifier state at all.
- Accumulation clamps at 64 KiB (`COMPACTION_AGENT_MESSAGE_CAP`), keeping the head of the message: omp's failure/no-op prose ("Compaction failed: …", "Nothing to compact") leads its compaction report.
- The buffer is cleared in `finishCompaction` — the single settlement funnel that success, failure, and interrupt all route through. The now-redundant reset at `startCompaction` is removed; session-creation initialization covers the first compaction.
- `fake-acp-agent.mjs` gains `FAKE_ACP_COMPACT_AGENT_MESSAGES` (JSON string array; one message per `/compact` request, last entry repeats) so one session can exercise sequential compactions with different prose.
- SDK bump 0.4.40 → 0.4.41 (bridge dist output changes).

## How you verified

- Three new regression tests, each mutation-checked (revert the guard, confirm the test fails, restore):
  - failure prose within the cap still classifies the turn failed — kills a tail-keeping clamp;
  - failure prose past the cap is ignored and the turn completes with `thread/compacted` — kills cap removal;
  - a second compaction in the same session classifies independently after a failed first one — kills removal of the `finishCompaction` clear.
- The accumulation dialect gate is behaviorally invisible by construction (no non-omp code reads the buffer), so it is covered by #2957's existing generic/OpenCode `completed`-outcome regressions rather than a test that could not fail.
- Package gates: `turbo run test typecheck --filter=@bb/provider-bridge-acp --force` — 18/18 test files, typecheck clean; `oxfmt --check` clean; `oxlint` zero new findings (the pre-existing `bb(no-comments)` hits in the `.mjs` harness are identical to main: 15 before, 15 after); `check-npm-version-guard` PASS at 0.4.41.

### DIFF
--- a/packages/domain/src/plugin-sdk-version.ts
+++ b/packages/domain/src/plugin-sdk-version.ts
@@ -1,3 +1,3 @@
-export const PLUGIN_SDK_VERSION = "0.4.40";
+export const PLUGIN_SDK_VERSION = "0.4.41";
 
 export const PLUGIN_SDK_MAJOR = Number(PLUGIN_SDK_VERSION.split(".", 1)[0]);
--- a/packages/plugin-sdk/package.json
+++ b/packages/plugin-sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@get-bb/plugin-sdk",
-  "version": "0.4.40",
+  "version": "0.4.41",
   "homepage": "https://github.com/get-bb/bb#readme",
   "bugs": {
     "url": "https://github.com/get-bb/bb/issues"
--- a/packages/provider-bridge-acp/src/bridge/bridge.test.ts
+++ b/packages/provider-bridge-acp/src/bridge/bridge.test.ts
@@ -2531,6 +2531,78 @@ describe("acp bridge", () => {
     expect(threadEventsOfType("thread/compacted")).toEqual([]);
   });
 
+  it("classifies failure prose that arrives within the compaction buffer cap", async () => {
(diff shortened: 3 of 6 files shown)


SawyerHood:

PR #3028, opened 2026-09-03
Preserve trusted bots in PR approval gate

ymichael:
## What was wrong

The approval gate added in #3022 seeded only human contributors, so trusted pull requests from `bb-slop-cop[bot]` and `dependabot[bot]` were rejected. The maintainer escape hatch could not repair this because `/approve @name[bot]` was parsed only through the opening bracket and accepted otherwise-invalid trailing content.

## What changed

- Add the two active, repository-evidenced automation authors to the exact allow list without exempting arbitrary bots.
- Require `/approve` to contain one complete GitHub-style human or `[bot]` login, rejecting malformed identities and trailing content.
- Exercise the workflow shell directly for trusted and untrusted bot gating plus valid and invalid approval commands, and include the workflow inputs in the Turbo test cache key.

No permissions, approval-file writes, human allow-list behavior, fork safety, wire contracts, CLI surfaces, or mobile behavior changed.

## How you verified

- Reproduced both trusted bots being rejected and `[bot]` approval names being truncated before the fix.
- `pnpm exec turbo run test --filter=@bb/scripts --force` — 22 files and 129 tests passed.
- `pnpm exec turbo run typecheck build --filter=@bb/scripts` — both tasks passed.
- Parsed both workflows as YAML, checked their shell blocks with `bash -n`, and ran `git diff --check`.

Fixes #3022

> AGENT GENERATED

### DIFF
--- a/.github/APPROVED_CONTRIBUTORS
+++ b/.github/APPROVED_CONTRIBUTORS
@@ -2,6 +2,7 @@ amadad
 andrewkchan
 AndrewSB
 barelyhuman
+bb-slop-cop[bot]
 ben-vargas
 bradhallett
 brsbl
@@ -9,6 +10,7 @@ builtui
 charpeni
 Danielalnajjar
 davidondrej
+dependabot[bot]
 DevVig
 Diffuzmetall
 dillonzq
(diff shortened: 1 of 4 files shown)


ymichael:
@slopcop please review.

PR #3036, opened 2026-09-04
Add Account Pool plugin: multi-account Claude hub

SawyerHood:
Stack layer 2 of 3. Prerequisite: the provider env contribution hook PR (layer 1) below it in this stack; this layer does not use the hook yet, layer 3 wires the plugin to it.

## What was wrong

BB had no server-side account pool for Claude traffic, so one Claude Code process could use only its local credential and could not rotate away from an exhausted subscription. BB-88 establishes the first stack layer: a manually configured Anthropic-compatible hub with durable account, secret, and quota state.

## What changed

- Added the disabled-by-default `account-pool` built-in plugin with provider-keyed account metadata in plugin KV, per-account 0600 token files, a generated 0600 hub bearer key, and SQLite quota/hold/error state.
- Added authenticated Anthropic-compatible routes for Messages and token counting plus an unauthenticated 200 response for Claude Code's `HEAD /api/hello`, with request-header rewriting, query forwarding, byte-streaming responses, quota tracking, threshold-aware selection, quota-429 rotation, per-minute pacing, OAuth refresh serialization/writeback, and graceful drain. The hub accepts requests before account setup, reports a 503 while no account is enabled, and uses newly added or enabled accounts without reload.
- Added `bb pool account add|list|remove|enable|disable` and `bb pool status`, mirrored through plugin RPC, plus built-in catalog registration and CLI guide, skill, and configuration documentation.
- No Plugin SDK or host-daemon wire contract changed. The requested wildcard 404 handler is omitted because the current `bb.http.route` SDK supports exact paths only and this stack layer explicitly forbids an SDK change; unknown plugin paths retain BB's standard 404.

## How you verified

- Added fake-upstream and Plugin SDK harness tests covering immediate use after first-account CLI add, hub authentication, unauthenticated hello probes, header/query/body forwarding, first-chunk latency and byte-identical SSE, quota bucket parsing, threshold skip, quota-429 resend, per-minute same-account pacing, serialized OAuth refresh with 0600 writeback, refresh/auth failures, CLI/RPC operations, content-type-aware stream failures, and stop/drain behavior. These tests have no implementation to load before this change and pass with the new plugin.
- `pnpm exec turbo run typecheck test --filter=bb-plugin-account-pool --continue` — 9 tests passed; typecheck passed.
- `pnpm exec turbo run typecheck test --filter=@bb/server --filter=@bb/plugin-api-map --filter=@bb/templates --continue` — 13 tasks passed; server 227 files/2202 tests, template and plugin-map suites passed.
- `pnpm exec oxlint plugins/account-pool/src` and targeted Prettier checks passed.
- In the running dev app, `HEAD /api/hello` returned 200; the hub streamed fake upstream SSE chunks every 20 seconds through the real mounted plugin route for at least 200 seconds; `bb pool status` showed the service accepting traffic and recorded 42% 5-hour / 64% 7-day quota utilization.
- Imported the server host's Claude OAuth credential, restored the upstream to `https://api.anthropic.com`, and ran `ANTHROPIC_BASE_URL=<dev hub> ANTHROPIC_AUTH_TOKEN=<hub key> claude -p "Reply with exactly: POOL OK"`. It exited 0 with `POOL OK`; account quota changed from unknown to 18% five-hour and 32% seven-day utilization.

Task: BB-88

> AGENT GENERATED


SawyerHood:
CI for this head ran via manual dispatch after the restack (the pull_request trigger did not fire): https://github.com/get-bb/bb/actions/runs/33832576981 — all jobs passed.

> AGENT GENERATED

PR #3040, opened 2026-09-04
Declare plan support for ACP agents and translate it via session modes

### LINKED ISSUE #3019: ACP providers can't use plan mode: supportsPlan is never declared or translated from session modes
### The workflow

I run orchestrator-grade agents through ACP providers (`acp-omp` today; the same shape applies to any ACP agent that supports planning). The workflow I want:

1. `bb thread spawn --provider acp-omp` and start an ambitious task.
2. Ask for a plan first — `bb thread tell <id> --plan` (plan-then-approve), or `/plan` in the composer.
3. Review the proposed plan, approve or redirect, then let the agent execute.

### What happens today

bb's plan lifecycle is wired for the first-party providers: `supportsPlan` is declared through the native capability chain and plan-mode enter/exit reaches the plan approval card. For ACP-hosted providers nothing declares plan support, and the channel ACP agents actually use to signal mode changes — session-mode updates on the side channel — is never translated into bb's plan lifecycle. So `/plan` and `--plan` stay gated off for providers whose agents can plan, and an ACP agent that enters plan mode on its own initiative does not surface a plan approval flow in bb.

Observed on `main` at f6868ad0c: `acp-omp` threads expose no plan path, while the same agent run through a first-party provider does.

### What you would expect

- ACP providers participate in the same capability chain: plan support is declared (handshake `supportsPlan`) rather than assumed absent.
- Session-mode updates are translated into bb's plan lifecycle, so plan-then-approve and `/plan` work uniformly across first-party and ACP providers; `acp-omp` un-gates.
- Capability-faithful: providers that cannot plan keep the gate closed.

### Context and alternatives

I prototyped this on my fork — branch [`feat/acp-capability-parity` @ `d49cca076`](https://github.com/bradhallett/bb/commit/d49cca076) (based on `main` f6868ad0c): `supportsPlan` declarations through the ACP capability layer plus the session-modes → plan-lifecycle bridge translation, with the full surface-parity set (CLI/SDK/docs per the repo convention) and the turbo gate green. Happy to open the PR — looking for sign-off first, per CONTRIBUTING.

Ruled out:

- Not a duplicate of #2019 (closed, Claude-specific: plan ignored on already-loaded sessions; CLI-parity path) — that is first-party behavior; this is the ACP capability gap.
- Not a duplicate of #1712 (closed, Claude Plan mode proposal flow) — same distinction.

Tested against `main` at f6868ad0c; investigation thread `thr_wr92wddupy`.

> AGENT GENERATED

bradhallett:
## Human comments

I hit this running omp as an ACP provider day-to-day: plan-then-approve works on first-party providers and silently doesn't exist for ACP threads, which made the orchestration workflows I use bb for (scouts → review → implement) awkward to gate. The slop-cop repro matches what I saw before filing.

Two things I'd flag for reviewers: the capability only un-gates what the agent actually advertises (no-modes agents get byte-identical passthrough — that was a hard requirement for me, not a nice-to-have), and rewind is deliberately left undeclarable rather than approximated. Built agent-assisted; I reviewed the wire-shape decisions myself.

## What was wrong

The plan action and edit-past-message gates were already capability-driven (`composerActions: ["plan"]`, `capabilities.fork`), but ACP agents could not declare either: the ACP provider hardcoded `composerActions` to `[]` and typed `AcpAgentDefinition.fork` as `none|tip`, so every ACP agent was excluded by declaration shape rather than by capability. The bridge also dropped `promptMode` and discarded current-mode updates as noise. Reproduced independently by the slop-cop triage at base `43e3fc519` ([report](https://get-bb.github.io/reports/issues/3019.html)); fixes #3019.

## What changed

- `AcpAgentDefinition` gains `supportsPlan` (mirroring `supportsManualCompaction`), threaded through the `customAgents` setting and mapped to `composerActions: ["plan"]` in the declaration; `acp-omp` declares it.
- The ACP bridge now reads the session modes agents advertise (`session/new`, `session/load`, `session/fork` results and `current_mode_update` notifications) and, when the agent advertises a plan mode, drives it with `session/set_mode` on plan turns, strips the `/plan` command mention from the prompt, and returns the session to its initial mode afterward; agents that report no modes keep byte-identical text passthrough.
- Session rewind stays undeclarable for ACP agents: ACP v1 `session/fork` is tip-only with no fork-point selector and reports no per-turn checkpoints, so no shipped or custom ACP agent can honestly declare checkpoint rewind today.

## How you verified

`pnpm exec turbo run typecheck test lint --continue` plus `pnpm format:check`. provider-bridge-acp typecheck clean, all 317 tests pass, including the new ones: declaration mapping (`composerActions: ["plan"]` only for agents declaring it), wire modes/`current_mode_update` parsing, and bridge plan-turn translation (`set_mode "plan"` precedes `session/prompt` with the mention stripped; a no-modes agent gets no `set_mode` and untouched text; the mode resets to the initial mode after the turn). Server suites that consume provider declarations pass solo (first-party-provider-plugins, plugin-install, plugin-update). Remaining failures are pre-existing on main `f6868ad0c`, each verified by stash-rerun on a clean tree: secret-storage concurrent-creators, server install-machine-script (macOS `Abort trap: 6` on the spawned host-daemon), host-workspace workspace-diff/provisioning, and 183 files flagged by `format:check` from the pinned oxfmt 0.64.0 at head (formatter default drift); the new code matches the dominant committed style of the files it touches.

Fixes #3019

> AGENT GENERATED

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -408,8 +408,13 @@ entry is a path or `{"path": ..., "recursive": true, "ancestors": true}` for
 an agent that nests skills or reads them from every ancestor directory),
 `permissionCli` (permission-mode launch flags), `supportsManualCompaction`
 (only if the agent accepts an explicit compaction request — bb hides
-`/compact` otherwise), and `dialect` (the vendor side channels bb reads for
-the agent: `cursor`, `opencode`, `omp`, or `grok`).
+`/compact` otherwise), `supportsPlan` (only if the agent exposes a `plan`
+session mode over ACP — bb offers the composer's `/plan` action and switches
+the session into that mode for plan turns), and `dialect` (the vendor side
+channels bb reads for the agent: `cursor`, `opencode`, `omp`, or `grok`).
+Editing a past message (session rewind) stays unavailable for ACP agents:
+ACP `session/fork` can only fork a session's current end, never an earlier
+point, so no `fork` declaration is offered for these entries.
 
 The change applies immediately: the plugin re-registers its providers when the
 setting changes, with no restart and no `config refresh`.
--- a/packages/provider-bridge-acp/src/bridge/bridge.test.ts
+++ b/packages/provider-bridge-acp/src/bridge/bridge.test.ts
@@ -11,7 +11,10 @@ import { tmpdir } from "node:os";
 import { dirname, join, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
-import { createStandaloneBuiltinCompactCommandInput } from "@bb/domain";
+import {
+  createBuiltinPlanCommandTextInput,
+  createStandaloneBuiltinCompactCommandInput,
+} from "@bb/domain";
 import type { DynamicTool, ReasoningLevel } from "@bb/domain";
 import {
   PROVIDER_BRIDGE_PROTOCOL_VERSION,
@@ -145,6 +148,7 @@ function executionOptions(args: {
   envVars?: Record<string, string>;
   instructions?: string;
   model?: string;
+  promptMode?: "plan";
   reasoningLevel?: ReasoningLevel;
   serviceTier?: "default" | "fast";
   providerOptions?: Record<string, unknown>;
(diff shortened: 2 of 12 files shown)


willadamskeane:
I tested this against the shipped OpenCode ACP provider and found two follow-up gaps: `acp-opencode` is not opted into `supportsPlan`, and the host-daemon protocol is not advanced even though an older daemon will silently ignore the now-exposed `promptMode`. I opened a minimal four-file patch directly against this PR branch: https://github.com/bradhallett/bb/pull/1. Focused provider/contract tests and typechecks pass (133 tests total).

SawyerHood:

Issue #1552, opened 2026-08-13
Pool multiple Claude accounts: limit-aware load balancing across provider connections


babebort:
## Problem

Heavy users run several Claude accounts in parallel (e.g. two Max subscriptions) because one account's usage limit runs out mid-day. Today each bb thread is effectively pinned to whichever account the underlying provider session is logged into: when that account hits its limit, the thread stalls even though another connected account still has plenty of headroom. Spreading the work is fully manual — notice the limit error, switch profiles by hand, restart or re-route the thread.

## Requested behavior

Building on the named provider connections proposed in #1447, add an opt-in **connection pool** for same-provider connections — a routing wrapper that balances threads across all of the user's Claude accounts:

- Group several explicitly-enabled Claude connections into one pool that a thread can target instead of a single account.
- **Limit-aware routing**: start new threads on the connection with the most remaining usage (bb already knows how to read account usage — use it as the balancing signal), with round-robin as the fallback when usage is unknown.
- **Automatic failover**: when a turn fails with a rate/usage-limit error, retry it on the next connection in the pool instead of surfacing the error and stalling.
- **Session affinity by default**: keep a thread on one connection while it works (provider-side prompt caching and context live there); rebalance at thread start or on limit errors, never mid-turn.
- Surface pool state in the UI: remaining usage per connection and which connection each thread is currently on.

The explicitness rules from #1447 still apply: pooling only ever spans connections the user explicitly enabled, no silent account discovery, authentication stays provider-owned.

## Why this belongs in bb

An external wrapper/proxy can't do this well: good balancing needs bb's own view of thread lifecycle (turn boundaries, safe retry points) and its per-connection usage data, while provider auth must stay provider-owned rather than funneled through a third-party proxy. bb already orchestrates the sessions — it is the natural place for the scheduling decision.

> AGENT GENERATED: filed by Claude Code on behalf of a bb user

tonydzi:
hi, mycroft here - anton's synthetic cofounder (an agent leaving field notes on an agent IDE seemed only fair).

we run this exact setup manually today: three claude subscriptions plus codex/gemini/grok rails across 6 machines, so some field numbers for why limit-aware beats round-robin.

[...]

PR #3042, opened 2026-09-04
Route Claude Code through Account Pool with per-machine tokens

SawyerHood:
Stack layer 3 of 3 (top). Prerequisites: #3035 (provider env contribution hook) and #3036 (Account Pool hub) below it in this stack.

## What was wrong

The Account Pool hub existed, but Claude Code sessions were not routed to it
automatically: the plugin still exposed one global bearer key, had no
per-thread bypass, and could not replace an unauthenticated machine's local
Claude health with the readiness supplied by an enabled pool account. This is
the BB-90 stack layer on `bb/account-pool-hub` and
`bb/provider-env-contribution`.

## What changed

- Registered the Account Pool's Claude Code environment contribution with the
  exact server-relative hub route and a secret bearer token minted lazily per
  machine. The resolver contributes nothing when no account is enabled or the
  thread is bypassed.
- Replaced the global hub key with per-machine 0600 token files loaded once
  into a single-process cache, constant-time bearer matching across every
  current and grace token, serialized mint/use/rotation updates, a ten-minute
  rotation grace window, and removal of the obsolete global key file. Last-use
  persistence is throttled to once per machine per minute, expired grace tokens
  are pruned with the next write, and status never returns token values.
- Added `bb pool token rotate --machine <id-or-name>` and
  `bb pool bypass <thread-id> [--off]`, with matching plugin RPC operations and
  discoverability updates in the CLI skill, generated guide source, and
  configuration docs.
- Tracked the last 24 hours of routed threads and made unsafe disablement
  visible in `bb pool status`, including proxied hosts that are relying on the
  pool, plus a plugin warning log when a disable leaves those threads on hosts
  without usable local Claude credentials. The dispose-time inspection is
  guarded against failures and bounded to two seconds.
- Added the single experimental SDK member
  `bb.providers.experimental_contributeEnvHealth`, with fake-host support,
  Plugin Guide documentation, API audit notes, API map inventory, and SDK
  version 0.4.45. The server presents an `unauthenticated` or `expired` Claude
  provider as ready/`Proxied` only when the same live plugin also has an env
  resolver and the Account Pool has an enabled account.
- Reused the lower stack layers' env command payload and hub drain. This PR
  does not change the server-to-daemon wire, so it does not bump
  `HOST_DAEMON_PROTOCOL_VERSION`.

## Not in this PR

- BB's plugin SDK has no system-notice append surface or fitting backend thread
  action. `bb.sdk.threads.send` would create a real provider turn, so this PR
  does not fake the requested disable-time thread notice or bypass action. The
  safety gap is instead explicit in `bb pool status` and the disable log.
- Disabling a plugin reloads the Claude bridge, which rebuilds the session but
  discards the previous attachment before it can emit the same cosmetic
  rebuild warning used by an in-place bypass change. In e2e thread
  `thr_stxrb938ec`, the fresh `provider.env-resolved` event without pool
  entries and the successful `LOCAL AGAIN` response demonstrate the actual
  disable behavior. The bypass path in that thread does retain the attachment
  and emits the expected rebuild warning.
- A packaged remote daemon cannot be enrolled into an isolated dev app. The
  multi-machine path is verified by the daemon-side runtime test that expands
  `serverPath` against its loopback machine-auth URL, plus a live 200 response
  from the dev hub's `HEAD /api/hello`; this does not claim a live remote
  thread.

## How you verified

- Added Account Pool regressions for empty resolver/health output, exact
  per-host contributions and secret masking, distinct machine tokens, bypass
  through CLI and RPC, token rotation and grace expiry, the single-process
  cache and last-use write throttle, secret-free status, proxied and
  unauthenticated routed-host warnings, bounded/failure-safe disable logging,
  and the existing authenticated streaming/drain paths. These behaviors had no
  implementation before this change; all 16 tests pass.
- Added server and fake-plugin-host coverage proving health is ignored without
  an adjacent env resolver and marks only credential-blocked provider health
  ready when a contribution is active.
- `pnpm exec turbo run typecheck --filter=@get-bb/plugin-sdk --filter=@bb/server --filter=@bb/plugin-api-map --filter=bb-plugin-account-pool --continue` — 9 tasks passed.
- `pnpm exec turbo run test --filter=@get-bb/plugin-sdk --filter=@bb/server --filter=@bb/plugin-api-map --filter=bb-plugin-account-pool --continue --force` — 11 tasks passed; Account Pool 13 tests, SDK 22 files, API map 11 files, and server 227 files / 2205 tests.
- Review follow-up: `pnpm exec turbo run typecheck test --filter=bb-plugin-account-pool --filter=@bb/server --filter=@get-bb/plugin-sdk --continue --force` — 13 tasks passed; Account Pool 16 tests, SDK 22 files, and server 227 files / 2205 tests.
- `pnpm exec turbo run typecheck test --filter=@bb/agent-runtime --continue --force` — typecheck passed and 22 files / 319 tests passed, including loopback `serverPath` expansion and secret event masking.
- `node .github/workflows/check-plugin-sdk-version.mjs` — passed with the SDK surface change and version bump.
- `git diff --check origin/bb/account-pool-hub...HEAD` — passed.
- In the isolated dev app, Claude Code thread `thr_stxrb938ec` returned
  `POOL OK` through the hub with an expanded loopback URL and masked token in
  `provider.env-resolved`; quota utilization moved from unknown to 28%/34%.
  Bypass returned `DIRECT OK` with no pool env and the rebuild warning. Plugin
  disable returned `LOCAL AGAIN` using local credentials with no pool env, and
  the plugin was re-enabled. Per-machine token rotation also completed, and a
  live hub hello request returned HTTP 200. BB-90 has the reduced JSON evidence
  and both rendered UI screenshots attached.

Part of #1552

Task: BB-90

> AGENT GENERATED


SawyerHood:
CI for this head ran via manual dispatch after the restack (the pull_request trigger did not fire): https://github.com/get-bb/bb/actions/runs/33832575481 — all jobs passed.

> AGENT GENERATED

SawyerHood:
Added `ENABLE_TOOL_SEARCH=true` to the contributed env: Claude Code disables tool search behind a non-first-party `ANTHROPIC_BASE_URL` by default, and the hub forwards `tool_reference` blocks and the beta header unchanged. Docs: https://code.claude.com/docs/en/agent-sdk/tool-search

> AGENT GENERATED

PR #3069, opened 2026-09-04
Toggle focused composer with its keyboard shortcut

swairshah:
## What was wrong

The `composer.focus` command always sent focus to the active composer. The follow-up composer already had a collapse path, but the shortcut command could not use it when the editor was focused.

## What changed

The composer command now tries to collapse the active follow-up composer before it falls back to focusing it. A focused composer captures its height, releases editor focus, and uses the existing compact state. The next shortcut press focuses the editor and expands the composer again. There are no wire, CLI, or plugin API changes.

## How you verified

Added controller and follow-up composer tests that fail before the change and cover both shortcut directions. Ran `pnpm exec turbo run test --filter=@bb/app -- src/components/plugin/ComposerExtensionHost.test.tsx src/components/promptbox/FollowUpPromptBox.test.tsx` and `pnpm exec turbo run typecheck --filter=@bb/app` after rebasing onto `get-bb/bb:main`. All 37 focused tests and the typecheck passed. The shortcut was also confirmed in the browser dev build.

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/plugin/ComposerExtensionHost.test.tsx
+++ b/apps/app/src/components/plugin/ComposerExtensionHost.test.tsx
@@ -16,6 +16,7 @@ import {
 } from "./ComposerExtensionHost";
 
 const mocks = vi.hoisted(() => ({
+  collapseIfFocused: vi.fn(() => false),
   focusDefault: vi.fn(() => true),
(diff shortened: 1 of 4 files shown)


SawyerHood:
@slopcop review

PR #3070, opened 2026-09-04
Offer Show more when a toast title is truncated

### LINKED ISSUE #3068: Toast error titles truncate with no way to read the rest
### Summary

Creating a thread that checks out a branch in an unmanaged workspace already held by a live thread makes the server return HTTP 409. The app shows an error toast reading `Cannot checkout branch while another thr…` — clamped to one line and elided mid-sentence, with no way to read the rest. The toast auto-dismisses after 4s and the elided text cannot be selected or copied, so the user never learns what the problem was. I expected the full message to be readable, either by wrapping or via the existing "Show more" link into the Notification Center.

### Versions and environment

- bb 0.41.0, installed via `npx bb-app` (binary resolves to `~/.npm/_npx/…/node_modules/bb-app/host-daemon/dist/bb`)
- Ubuntu 26.04 LTS, Linux 7.0.0-22-generic
- Node v22.22.1
- App surface: bb web/desktop app UI, dark theme, wide viewport (~2047px). **I could not determine which browser engine rendered the attached screenshot** — the bug is layout/DOM-level (a `truncate` class) and is not engine-specific, but I have not confirmed it in Safari or on mobile widths.
- No provider/agent involvement — this fires in the new-thread composer before any agent starts.
- Code analysis performed against `main` at `ea6ce5b0c528b95a419d1513333c169ca8468936`. All cited lines verified at that commit.

### Steps to reproduce

Any 409 whose message exceeds roughly 40 characters reproduces this. The path I hit:

1. Open a project pointing at an **unmanaged** workspace path (a plain checkout, not a bb-managed worktree).
2. Start a thread in that workspace and leave it live.
3. From the new-thread composer, select the same workspace path and set the environment to check out a branch ("New branch from: …").
4. Submit the thread.
5. Server returns `409` with `{"code":"invalid_request","message":"Cannot checkout branch while another thread is using this workspace"}`.
6. Watch the toast in the bottom-right corner.

Minimal restatement: **any** mutation error whose message is a bare sentence longer than ~40 characters and that is not one of the eight lifecycle error codes reproduces this. The specific 409 above is just a convenient trigger.

Did **not** reproduce with: errors carrying a structured lifecycle code (e.g. `environment_not_ready`) — those get a `{title, body}` pair, so the long text lands in the description slot, which *does* get a "Show more" link. Also does not reproduce for `GENERIC_REQUEST_FAILED_MESSAGE`, which is hardcoded to a short title plus a description.

### Expected vs actual

```
Actual (toast, verbatim as rendered):

  (!)  Cannot checkout branch while another thr…            [x]

Actual (server response body):

  HTTP/1.1 409 Conflict
  {"code":"invalid_request","message":"Cannot checkout branch while another thread is using this workspace"}

Expected:

  The full sentence readable, e.g.

  (!)  Cannot checkout branch while another thread is
       using this workspace                    Show more  [x]

  or at minimum a "Show more" link that opens the Notification
  Center, where the same text already wraps and has a copy button.
```

Screenshot of the toast follows in a comment on this issue. Look at the bottom-right toast: the title ends in an ellipsis after `another thr`, and there is no "Show more" affordance next to it — only the close button.

### Evidence

Root cause chain, all permalinks at `ea6ce5b0c528b95a419d1513333c169ca8468936`:

1. The server builds a bare, unstructured string message:
   [`apps/server/src/services/threads/workspace-path-claims.ts#L51-L55`](https://github.com/get-bb/bb/blob/ea6ce5b0c528b95a419d1513333c169ca8468936/apps/server/src/services/threads/workspace-path-claims.ts#L51-L55)

2. It is thrown as a flat `ApiError` with code `invalid_request`:
   [`apps/server/src/services/threads/thread-create.ts#L362-L364`](https://github.com/get-bb/bb/blob/ea6ce5b0c528b95a419d1513333c169ca8468936/apps/server/src/services/threads/thread-create.ts#L362-L364)

3. `invalid_request` is not one of the eight members of `lifecycleApiErrorSchema`:
   [`packages/server-contract/src/errors.ts#L170-L179`](https://github.com/get-bb/bb/blob/ea6ce5b0c528b95a419d1513333c169ca8468936/packages/server-contract/src/errors.ts#L170-L179)
   so `describeLifecycleError` returns `null` and no structured `{title, body}` is derived:
   [`apps/app/src/lib/lifecycle-errors.ts#L460-L467`](https://github.com/get-bb/bb/blob/ea6ce5b0c528b95a419d1513333c169ca8468936/apps/app/src/lib/lifecycle-errors.ts#L460-L467)

4. `getMutationErrorMessage` therefore falls through to `getHttpErrorMessage` and returns the raw server sentence:
   [`apps/app/src/lib/mutation-errors.ts#L147-L153`](https://github.com/get-bb/bb/blob/ea6ce5b0c528b95a419d1513333c169ca8468936/apps/app/src/lib/mutation-errors.ts#L147-L153)

5. `showMutationErrorToast` passes it to `appToast.error` as a **single** argument, so the whole sentence lands in the toast's *title* slot with no description:
   [`apps/app/src/lib/mutation-errors.ts#L197`](https://github.com/get-bb/bb/blob/ea6ce5b0c528b95a419d1513333c169ca8468936/apps/app/src/lib/mutation-errors.ts#L197)

6. The title is clamped to one line with `truncate`:
   [`apps/app/src/components/ui/app-toast.tsx#L206-L211`](https://github.com/get-bb/bb/blob/ea6ce5b0c528b95a419d1513333c169ca8468936/apps/app/src/components/ui/app-toast.tsx#L206-L211)
   and the "Show more" affordance is implemented **only** inside `AppToastDescription`, with the overflow measurement (`scrollWidth - clientWidth > 1`) taken against the description element alone:
   [`apps/app/src/components/ui/app-toast.tsx#L137-L167`](https://github.com/get-bb/bb/blob/ea6ce5b0c528b95a419d1513333c169ca8468936/apps/app/src/components/ui/app-toast.tsx#L137-L167)
   That whole block is rendered only when `description || hasActions`:
   [`apps/app/src/components/ui/app-toast.tsx#L212-L216`](https://github.com/get-bb/bb/blob/ea6ce5b0c528b95a419d1513333c169ca8468936/apps/app/src/components/ui/app-toast.tsx#L212-L216)
   **A toast with a truncated title and no description therefore has no recovery path at all.**

7. The Notification Center — the destination "Show more" would have reached — already renders the same text correctly: it wraps (`break-words`) and offers a copy button.
   [`apps/app/src/components/notifications/NotificationCenter.tsx#L111`](https://github.com/get-bb/bb/blob/ea6ce5b0c528b95a419d1513333c169ca8468936/apps/app/src/components/notifications/NotificationCenter.tsx#L111) ·
   [`NotificationCenter.tsx#L49-L81`](https://github.com/get-bb/bb/blob/ea6ce5b0c528b95a419d1513333c169ca8468936/apps/app/src/components/notifications/NotificationCenter.tsx#L49-L81)
   The data is present; only the affordance to reach it is missing.

**Why it truncates so early.** sonner 1.7.4 (`apps/app/package.json` pins `^1.7.4`) hardcodes a default toast width of 356px — in the shipped bundle, `node_modules/…/sonner/dist/index.mjs` minifies it to `he=356`, applied as the `--width` custom property. bb mirrors that on the card as `w-[var(--width,356px)]` ([`app-toast.tsx#L196`](https://github.com/get-bb/bb/blob/ea6ce5b0c528b95a419d1513333c169ca8468936/apps/app/src/components/ui/app-toast.tsx#L196)). After the 16px icon, gaps, padding and the close button, the title line holds roughly 40 characters. The message is 66 characters. (Character estimate is from the rendered screenshot, not measured programmatically.)

**Related second-order impact (not separately fixed here):** the truncated text also cannot be recovered by selection. The DOM text is genuinely elided by `text-overflow`, the toast auto-dismisses after `DEFAULT_TOAST_DURATION = 4000` ([`app-toast.tsx#L82`](https://github.com/get-bb/bb/blob/ea6ce5b0c528b95a419d1513333c169ca8468936/apps/app/src/components/ui/app-toast.tsx#L82); sonner's own default is the same 4000, `Wt=4e3`), and sonner's pointer handling turns a mouse drag across the toast into a swipe-to-dismiss. So there is no select-and-copy fallback either.

### What you ruled out

- **Not a duplicate.** Searched open and closed issues on `get-bb/bb` for `toast truncat`, `truncated error`, `Show more toast`, `error message cut`, `toast`, `toast title`, `unreadable`, `notification center`, `ellipsis`, `409 workspace`, `Cannot checkout branch`. Nothing covers toast text truncation or the missing "Show more" on titles.
- **Not a duplicate of #2068** ("Archived idle threads block branch checkout after the workspace is released", closed). That issue is about the 409 being raised when it should not be. This one is about the message being unreadable once raised, and is independent of whether the 409 is correct.
- **Not a missing lifecycle description.** `invalid_request` is genuinely outside `lifecycleApiErrorSchema`, so the existing structured title/body machinery correctly does not apply. Adding a code to that union would fix this one message but not the general class.
- **Not a sonner `user-select: none` problem.** sonner only applies that while `data-swiped='true'`; the text is unselectable because it is elided, not because selection is disabled.
- **Not a Notification Center rendering bug.** It renders the full text with wrapping and a copy button already; the data reaches it fine.
- **Not specific to this one 409.** Any non-lifecycle error message over ~40 characters hits the same path.
- **Still present on `main`** at `ea6ce5b0c528b95a419d1513333c169ca8468936` — all cited code verified at that commit.

Medium — hits any user who trips a non-lifecycle server error longer than one toast line, no workaround at all (cannot read, select, copy, or navigate to the text), no data lost; effort probably Low.

### Checks

> AGENT GENERATED

IlyaM:
## Human comments

Minimal fix — it just restores a way to read the message. Two things I think
still need doing:

1. The toast is too narrow. 356px is only Sonner's default, not a decision
anyone here made, and there's plenty of screen going spare.

2. These should really be a short title plus a longer body. Part of that
already exists and gets thrown away — lifecycle errors do emit
`{title, body}`, then `formatLifecycleErrorDescription` glues them back into
one string. That part's a one-function fix. The flat `ApiError(…, message)`
throws never had a split at all, and sorting those out is a much bigger job.

## What was wrong

The toast card clamps its title to a single line with `truncate`, but the
"Show more" affordance — the only route to the Notification Center, where
toast text wraps and can be copied — lived inside `AppToastDescription` and
was measured against the description alone. A toast with no description
therefore had no recovery path at all.

`showMutationErrorToast` produces exactly that shape: it calls
`appToast.error(message)` with a single argument
(`apps/app/src/lib/mutation-errors.ts:197`), so the whole server message
lands in the title slot. Flat errors take this path because they carry no
structure to split on — `thread-create.ts:363` raises
`ApiError(409, "invalid_request", …)` for a busy workspace, and
`invalid_request` is not one of the eight `lifecycleApiErrorSchema` codes,
so `describeLifecycleError` returns null and `getMutationErrorMessage`
falls through to the raw server sentence (`mutation-errors.ts:152`).

At Sonner's default 356px width the title holds roughly 40 characters, so
"Cannot checkout branch while another thread is using this workspace"
rendered as "Cannot checkout branch while another thr…" with no way to read
the rest. Line references are against `ea6ce5b0c`.

## What changed

**`apps/app/src/components/ui/app-toast.tsx`** — extracted the overflow
measurement out of `AppToastDescription` into a `useOverflowTruncation`
hook and applied it to the title as well as the description.
`AppToastContent` now owns a single "Show more" button, rendered when
either region overflows, and emits the secondary row for a title-only toast
that overflowed. `AppToastDescription` collapsed into `AppToastContent` (it
had no consumers outside this file), and an unused `data-testid` was
dropped. Net -10 lines.

Descriptions behave exactly as before, and `mutation-errors.ts` is
untouched, so the inline consumers of `getMutationErrorMessage` render
identical text. App-only change: no host daemon wire impact (no
`HOST_DAEMON_PROTOCOL_VERSION` bump needed), and no CLI, guide, or
documentation surface is affected.

Deliberately out of scope: the deeper fix of giving flat server errors a
real title/body split, and promoting `invalid_request` refusals to
structured lifecycle codes. Both would narrow the class of errors that can
overflow, but neither removes the need for a recovery path on the title.

## How you verified

- Four tests in `apps/app/src/lib/notifications/notifications.test.tsx`
  covering title-only overflow, title-overflows-while-description-fits,
  description-only overflow, and the both-fit negative case. The first two
  fail before this change and pass after.
- Replaced the prototype-wide width stub with a per-element one, so a title
  and a description can disagree about fitting; the previous stub forced
  them to agree and could not exercise the new branch.
- Mutation-tested the assertions: reverting `canShowMore` to
  description-only fails exactly the two title tests and leaves the
  description test green; hardcoding `titleTruncated = true` fails exactly
  the both-fit test.
- `pnpm exec turbo run test --filter=@bb/app` — 18/18 across
  `app-toast.test`, `notifications.test`, `AppToaster.test`.
- `pnpm exec turbo run typecheck --filter=@bb/app` clean;
  `lint --filter=@bb/app` reports no warnings in either touched file.

Fixes #3068

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/ui/app-toast.tsx
+++ b/apps/app/src/components/ui/app-toast.tsx
@@ -4,6 +4,7 @@ import {
   useState,
   type MouseEvent,
   type ReactNode,
+  type RefObject,
 } from "react";
 import { toast as sonnerToast, type Action, type ExternalToast } from "sonner";
 import { Button } from "@bb/shared-ui/button";
@@ -54,12 +55,6 @@ interface AppToastContentProps {
   onDismiss?: () => void;
   title: ReactNode;
   tone: AppToastTone;
-}
-
-interface AppToastDescriptionProps {
-  description: ReactNode;
-  notificationId: string | null;
-  onShowMore: () => void;
 }
 
 interface ShowAppToastParams {
@@ -129,44 +124,22 @@ function AppToastActionButton({
   );
 }
 
-export function AppToastDescription({
-  description,
-  notificationId,
-  onShowMore,
-}: AppToastDescriptionProps) {
-  const bodyRef = useRef<HTMLDivElement | null>(null);
+function useOverflowTruncation(
+  content: ReactNode,
+): [RefObject<HTMLDivElement | null>, boolean] {
+  const ref = useRef<HTMLDivElement | null>(null);
   const [truncated, setTruncated] = useState(false);
 
   useLayoutEffect(() => {
-    const body = bodyRef.current;
-    if (body === null) {
+    const node = ref.current;
+    if (node === null) {
+      setTruncated(false);
       return;
     }
-    setTruncated(body.scrollWidth - body.clientWidth > 1);
-  }, [description]);
+    setTruncated(node.scrollWidth - node.clientWidth > 1);
+  }, [content]);
 
-  return (
-    <>
-      <div
-        ref={bodyRef}
-        data-testid="app-toast-description"
-        className="min-w-0 flex-1 truncate"
-      >
-        {description}
-      </div>
-      {truncated && notificationId !== null ? (
-        <Button
-          type="button"
-          variant="link"
-          size="sm"
-          className="h-auto shrink-0 px-0 py-0 text-xs text-muted-foreground underline underline-offset-4"
-          onClick={onShowMore}
-        >
-          Show more
-        </Button>
-      ) : null}
-    </>
-  );
+  return [ref, truncated];
 }
 
 export function AppToastContent({
@@ -180,6 +153,11 @@ export function AppToastContent({
   title,
   tone,
 }: AppToastContentProps) {
+  const [titleRef, titleTruncated] = useOverflowTruncation(title);
+  const [descriptionRef, descriptionTruncated] =
+    useOverflowTruncation(description);
+  const canShowMore =
+    notificationId !== null && (titleTruncated || descriptionTruncated);
   const hasActions = action !== undefined || cancel !== undefined;
   const actions = hasActions ? (
     <>
@@ -205,21 +183,33 @@ export function AppToastContent({
         </div>
         <div className="min-w-0 flex-1">
           <div className="flex min-w-0 items-center gap-2">
-            <div className="min-w-0 flex-1 truncate text-sm font-medium leading-5">
+            <div
+              ref={titleRef}
+              className="min-w-0 flex-1 truncate text-sm font-medium leading-5"
+            >
               {title}
             </div>
           </div>
-          {description || hasActions ? (
+          {description || hasActions || canShowMore ? (
             <div className="mt-0.5 flex min-w-0 flex-nowrap items-center gap-2 text-xs leading-5 text-muted-foreground">
               {description ? (
-                <AppToastDescription
-                  description={description}
-                  notificationId={notificationId}
-                  onShowMore={() => {
+                <div ref={descriptionRef} className="min-w-0 flex-1 truncate">
+                  {description}
+                </div>
+              ) : null}
+              {canShowMore ? (
+                <Button
+                  type="button"
+                  variant="link"
+                  size="sm"
+                  className="h-auto shrink-0 px-0 py-0 text-xs text-muted-foreground underline underline-offset-4"
+                  onClick={() => {
                     dismissToast(id);
                     openNotificationCenter(notificationId);
                   }}
-                />
+                >
+                  Show more
+                </Button>
               ) : null}
               {actions}
             </div>
--- a/apps/app/src/lib/notifications/notifications.test.tsx
+++ b/apps/app/src/lib/notifications/notifications.test.tsx
@@ -51,47 +51,88 @@ describe("toast history", () => {
   });
 });
 
-describe("truncated toast descriptions", () => {
-  function renderToast() {
+describe("truncated toast text", () => {
+  const FITTING_WIDTH = 300;
+  const OVERFLOWING_WIDTH = 600;
+  const LONG_TITLE =
+    "Cannot checkout branch while another thread is using this workspace";
+  const SHORT_TITLE = "Installing the plugin failed";
+  const DESCRIPTION = "a very long esbuild error";
+
+  function mockOverflowingText(overflowingText: string | null) {
+    vi.spyOn(HTMLElement.prototype, "scrollWidth", "get").mockImplementation(
+      function (this: HTMLElement) {
+        return this.textContent === overflowingText
+          ? OVERFLOWING_WIDTH
+          : FITTING_WIDTH;
+      },
+    );
+    vi.spyOn(HTMLElement.prototype, "clientWidth", "get").mockReturnValue(
+      FITTING_WIDTH,
+    );
+  }
+
+  function showMoreButton() {
+    return screen.queryByRole("button", { name: "Show more" });
+  }
+
+  it("offers Show more when only the title does not fit", () => {
+    mockOverflowingText(LONG_TITLE);
     render(
       <AppToastContent
-        title="Installing the plugin failed"
-        description="a very long esbuild error"
+        title={LONG_TITLE}
+        description={DESCRIPTION}
         tone="error"
         notificationId="notification-7"
       />,
     );
-  }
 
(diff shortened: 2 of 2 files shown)


SawyerHood:

PR #3066, opened 2026-09-04
Show completed thinking in thread history

IlyaM:
## What was wrong

[#1248](https://github.com/get-bb/bb/issues/1248) requests expandable
`Thought for <duration>` sections while retaining the existing streaming
thinking presentation.

bb already displayed provider-visible reasoning through the active
`Thinking…` disclosure, but discarded that text from the timeline projection
when the reasoning lifecycle completed. Completed threads therefore retained
the surrounding work and assistant response, but not the reasoning shown while
the turn was running.

## What changed

Completed, non-empty reasoning lifecycles now produce expandable
`Thought for <duration>` entries in thread history through the existing generic
system-operation timeline machinery.

- Titles use persisted lifecycle timestamps and the existing compact duration
  formatter.
- Expandable details contain only reasoning text exposed by the provider.
- Entries retain their event range, turn scope, effective parent scope, and
  completed or interrupted status.
- Turn completion and thread interruption finalize reasoning that did not
  receive an explicit completion event.
- Finalized lifecycle keys reject late streaming deltas and duplicate completions.
- When a provider completes reasoning after Stop, its final text and event range
  update the saved thought while preserving its interrupted status and stop time.
- Completed reasoning details are capped at 32,000 characters.
- Active reasoning remains in the existing `Thinking…` disclosure.

Completed thoughts can appear inside the existing `Worked for…` disclosure.
No new public timeline-row kind or reasoning-specific UI component was added.

There are no database, persisted-event, server/host-daemon wire, CLI,
Plugin SDK, configuration, or documentation contract changes.
`HOST_DAEMON_PROTOCOL_VERSION` remains unchanged.

## How you verified

- Added focused coverage for explicit completion, final-text replacement,
  persisted lifecycle start times, interrupted reasoning, turn-finalization
  fallback, nested/root lifecycle isolation, late-delta rejection, completed
  CLI rendering, and deterministic replay.
- Replayed provider reasoning events through the timeline projection and
  confirmed that active reasoning appears only through `Thinking…`, completed
  reasoning appears once as `Thought for <duration>`, and completed entries
  survive rebuilding the idle timeline.
- Verified the behavior in the isolated development app with Claude Code and
  Pi sessions.
- `@bb/thread-view`: 377 tests passed; server timeline integration tests: 54
  passed; provider parity: 56 passed. Server and thread-view typechecks passed.
- The Stop → late provider completion regression fails on the original code and
  passes with the fix. Doobie verified the saved interrupted thought expands
  after reload in the development app.
[...]

PR #3110, opened 2026-09-05
Keep late thinking on its originating history page

ymichael:
## What was wrong

The shared timeline window closure treated every scoped item that crossed a page boundary as belonging to the newer sequence window. For a completed thinking item whose start was in an older user turn but whose completion arrived after the next prompt began, standard turn pagination therefore separated the lifecycle: the newest page rendered the completion beneath the wrong prompt as “Thought for 0ms,” and loading the older page left the originating turn without its thought. This followed [6e83eb2](https://github.com/get-bb/bb/commit/6e83eb205a9debf8a92e03b6eda8cbd4ee041b7a) from #3066.

A thought that never receives `item/completed` could also disappear from both pages when its original turn moved into older history.

## What changed

- Expose the completion sequence in the existing item-span query.
- Give the shared whole-item closure explicit origin-versus-later page ownership.
- Keep an unfinished thought on its origin page until it receives a newer-page event; then keep the active lifecycle on the newer page without duplication.
- Assign completed items to their originating user segment while preserving existing later-page ownership for byte and details windows.
- Constrain span lookups to item events and distribute their thread/type constraints into each scoped lookup, allowing SQLite to use the existing `(thread_id, turn_id, type, item_id, sequence)` index.
- Guard that multi-turn query plan and the unfinished and late-completed latest/older/details behavior with focused regressions.

There are no database schema or migration, host-daemon wire, CLI, documentation, UI-component, pagination-limit, or mobile changes.

## How you verified

- The focused regression fails before the fixes: late completion appears on the newest page, while a never-completed thought disappears from both pages.
- `pnpm exec turbo run test --filter=@bb/server -- test/services/threads/timeline-in-turn-window.test.ts` — 28/28 passed after rebasing onto current local `origin/main`.
- `pnpm exec turbo run test --filter=@bb/db` — 448/448 passed.
- `pnpm exec turbo run typecheck --filter=@bb/server` and `pnpm exec turbo run typecheck --filter=@bb/db` — passed.
- The committed multi-turn `EXPLAIN QUERY PLAN` regression proves the span lookup uses `events_thread_turn_type_item_sequence_idx`; no new index is created.
- DevBrowser verified the real app history boundary: after loading the older page, the original turn expands through “Worked for 5s” to the interrupted “Thought for 2s,” its text, and “Stopped manually,” with no thought under the next prompt.

> AGENT GENERATED


ymichael:
@slopcop please review the late-completed-thinking pagination fix when you have a moment.

ymichael:
@slopcop another ping for review on the late-completed-thinking pagination fix.

PR #3112, opened 2026-09-05
fix(account-pool): serialize concurrent credential refreshes

smsunarto:
## What was wrong

When parallel requests needed a token refresh, they could refresh the same account twice and race token rotation, risking authentication failures that interrupt otherwise valid sessions.

## What changed

Requests now share one refresh per account, so parallel work reuses the refreshed credential. Other accounts can continue independently.

## How you verified

The duplicate-refresh regression fails before the fix and passes after. All 35 server tests, typecheck, lint, and formatting passed.

Fork branches are stacked; upstream PR diffs are cumulative.

> AGENT GENERATED

---
This is **part 1 of 4 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #3117
- <kbd>&nbsp;3&nbsp;</kbd> #3116
- <kbd>&nbsp;2&nbsp;</kbd> #3115
- <kbd>&nbsp;1&nbsp;</kbd> #3112 👈

### DIFF
--- a/plugins/account-pool/src/hub.ts
+++ b/plugins/account-pool/src/hub.ts
@@ -423,16 +423,18 @@ export class AccountPoolHub {
   ): Promise<AccountSecret> {
     const existing = this.refreshes.get(account.id);
     if (existing !== undefined) return existing;
-    const secret = await this.options.accounts.readSecret(account.id);
-    const refresh = adapter
-      .refreshSecret({
-        account,
-        secret,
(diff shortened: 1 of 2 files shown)


SawyerHood:
Closing as already included in #3117, which was squash-merged into main as 4a5e7557f. Verified that this PR head is included in the top of the stack and that the merged tree matches that stack head. No additional merge is needed.

PR #3115, opened 2026-09-05
fix(account-pool): recover from temporary OAuth refresh outages

smsunarto:
## What was wrong

A brief OAuth refresh outage could leave an account permanently unusable. Sessions then received quota errors even when the account still had a valid access token, requiring manual recovery.

## What changed

Keep sessions running on valid tokens during temporary refresh outages and retry automatically with backoff. If every usable token has expired, return a temporary error instead of permanently disabling the accounts.

## How you verified

Six regressions failed before and pass after. All 86 plugin tests, typecheck, lint, and formatting passed.

Fork branches are stacked; upstream PR diffs are cumulative.

> AGENT GENERATED

---
This is **part 2 of 4 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #3117
- <kbd>&nbsp;3&nbsp;</kbd> #3116
- <kbd>&nbsp;2&nbsp;</kbd> #3115 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #3112

### DIFF
--- a/plugins/account-pool/src/claude-adapter.ts
+++ b/plugins/account-pool/src/claude-adapter.ts
@@ -6,6 +6,7 @@ import {
 } from "./credentials.js";
 import type { ProviderAdapter } from "./provider-adapter.js";
 import {
+  fetchOAuthRefresh,
   filterRequestHeaders,
   mountedUpstreamUrl,
 } from "./provider-adapter.js";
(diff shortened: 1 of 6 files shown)


SawyerHood:
Closing as already included in #3117, which was squash-merged into main as 4a5e7557f. Verified that this PR head is included in the top of the stack and that the merged tree matches that stack head. No additional merge is needed.

PR #3163, opened 2026-09-06
Preview thread storage artifacts inline

### LINKED ISSUE #3158: inline-vis cannot render HTML files from $BB_THREAD_STORAGE
## The workflow

I ask an agent to do a read-only investigation (no workspace mutations) and produce a report or visualization for me to review inline in the thread. `::inline-vis` is very convenient for this, but it only accepts HTML files inside the thread workspace — writing a report file into the workspace isn't appropriate for a read-only session, so the agent writes the HTML to `$BB_THREAD_STORAGE` instead (the correct place for this kind of artifact) and then tries to render it with `::inline-vis{file="..."}`, which is not supported.

## What happens today

`::inline-vis` only accepts a workspace-relative `.html`/`.htm` path. In `plugins/inline-vis/server.ts`, `requireWorkspaceHtmlFile` rejects absolute paths, `..` traversal, and anything that resolves outside the workspace root, and `prepareHtmlPreview` resolves the path against the thread's environment workspace path before reading it. `$BB_THREAD_STORAGE` lives outside that workspace root, so a file written there cannot be referenced by the directive at all — the agent's only options are to open it via a Markdown link or `bb thread open "$BB_THREAD_STORAGE/report.html"`, both of which open it in BB's secondary panel rather than inline in the message.

## What I would expect

`::inline-vis` should be able to render an HTML file from the current thread's `$BB_THREAD_STORAGE` in addition to workspace-relative files, so read-only/investigation sessions have the same inline-rendering option as sessions that touch the workspace. I don't have a strong opinion on the exact directive shape (e.g. a `source="thread-storage"` attribute, a recognized path prefix, or a separate directive) — any of these would work as long as the plugin still validates the resolved path stays confined to that thread's storage directory the same way it currently confines workspace paths.

## Context and alternatives

Current workaround: open the file via a Markdown link or `bb thread open "$BB_THREAD_STORAGE/report.html"`, which opens it in BB's secondary panel instead of rendering it inline in the conversation.

> AGENT GENERATED

IlyaM:
## What was wrong

The inline-vis plugin assumed every HTML artifact belonged to the thread workspace: its RPC always resolved the thread environment, and its frontend always used the worktree file route and workspace-open action. This prevented read-only artifacts stored in thread storage from being rendered inline. Fixes [#3158](https://github.com/get-bb/bb/issues/3158).

## What changed

Added `source="workspace" | "thread-storage"` to the inline-vis directive, with omitted source defaulting to `workspace`.

The backend now resolves thread-storage roots through `bb.sdk.threads.storageLocation()` and preflights files with the returned host and root confinement. The plugin-local RPC returns the validated file and authoritative normalized source.

The frontend uses the existing worktree or thread-storage route based on that returned source. Thread-storage previews retain the sandboxed iframe but omit the workspace-open action.

Updated the inline-vis README, plugin overview, package metadata, and bundled skill. The skill explains that thread-storage artifacts are written beneath `$BB_THREAD_STORAGE` while directive file paths remain storage-relative.

This changes only the plugin-local RPC contract. It does not change the server/daemon protocol and requires no `HOST_DAEMON_PROTOCOL_VERSION` increment.

## How you verified

Added focused backend coverage for thread-storage resolution, host/root confinement, workspace bypass, source defaulting, and invalid-source validation.

Added focused frontend coverage for source-specific routes, segment encoding, iframe sandboxing, RPC validation errors, and suppression of the workspace-open action.

Ran:

`pnpm exec turbo run test typecheck --filter=bb-plugin-inline-vis`

Results:

- 21 tests passed across 2 test files
- Typecheck passed
- 6 Turbo tasks succeeded
- `git diff --check` passed

Fixes #3158

> AGENT GENERATED

### DIFF
--- a/plugins/inline-vis/PLUGIN_OVERVIEW.md
+++ b/plugins/inline-vis/PLUGIN_OVERVIEW.md
@@ -1,21 +1,27 @@
-See an agent's chart, demo, or report in the conversation without opening a side panel. The agent writes an HTML file to the workspace. The plugin shows that file inside the assistant message.
+See an agent's chart, demo, or report in the conversation without opening a side panel. The agent writes an HTML file to the workspace or the thread's storage directory. The plugin shows that file inside the assistant message.
 
 ## What you get
 
 - A live preview card in the message. Scripts and styles in the file run.
 - A default viewport height of 224 pixels. The agent can set a height from 120 to 1200 pixels.
-- A header action that opens the source HTML file in the workspace viewer.
+- A header action that opens workspace source HTML in the workspace viewer. Thread-storage previews omit this action.
 - A clear inline error when the file is missing, too large, or not HTML.
 
 ## How it works
 
-The agent emits a message directive that names a workspace-relative `.html` or `.htm` file:
+The agent emits a message directive that names a source-relative `.html` or `.htm` file. Omitting `source` defaults to the workspace, and explicit `source="workspace"` is equivalent:
 
 ```text
 ::inline-vis{file="charts/out.html" height="480"}
 ```
 
-The plugin confirms the file exists in the thread workspace before it renders. Files must be UTF-8 text with a maximum size of 5 MiB. Relative assets next to the file load as usual.
+Read-only artifacts in the thread's storage directory can be rendered without resolving the workspace:
+
+```text
+::inline-vis{source="thread-storage" file="reports/result.html"}
+```
+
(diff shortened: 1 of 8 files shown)


SawyerHood:
Superseded by #3439, which integrates this PR with #3165 since the two conflicted in every file. The commits there credit @IlyaM as co-author. Thanks for both changes!

PR #3165, opened 2026-09-06
Render Markdown files in inline previews

### LINKED ISSUE #3159: Support rendering Markdown files in inline-vis blocks
## The workflow

An agent finishes a step and wants to show the user a document it just produced or updated in the workspace — a report, a plan, a summary, notes — inline in the chat message, the same way `::inline-vis{file="demo.html"}` already does for HTML. Markdown is the format agentic workflows reach for by default for this kind of output (reports, specs, changelogs, generated docs), so this comes up often.

## What happens today

`inline-vis` only accepts `.html` / `.htm` files. The extension check lives in `plugins/inline-vis/server.ts` (`HTML_EXTENSIONS`), and it rejects anything else with `"file" must end with .html or .htm`. To show a `.md` file inline today, an agent has to either convert it to HTML first or fall back to pasting the raw content into the chat message as text, losing the "here's a live file in the workspace" framing and the "open in sidebar" affordance that `inline-vis` provides for HTML.

## What you would expect

`::inline-vis{file="notes.md"}` (or a similarly-named directive) renders the Markdown file's content as formatted Markdown inline in the message, with the same surrounding chrome inline-vis already has today: the file-path header, the "open in sidebar" button, and the same workspace-relative path / size-limit / traversal validation rules that apply to HTML files now. It does not need to support arbitrary embedded scripts or iframes the way the HTML path does — just Markdown rendering.

Whether this belongs as an added file type on the existing `inline-vis` directive or as a separate directive is an open question; either way, the underlying gap is the same: there is no inline preview path for Markdown artifacts.

## Context and alternatives

- Current workaround: convert Markdown to HTML before emitting `inline-vis`, or just paste the Markdown as message text (which renders as Markdown in chat, but doesn't carry the file-preview framing, sizing, or "open in sidebar" action that inline-vis gives HTML).
- Searched open and closed issues; found no existing request for Markdown support in `inline-vis` (only unrelated inline-vis issues: HTML-from-thread-storage loading, a layout-jump bug, and full-page preview in desktop browser).

> AGENT GENERATED

IlyaM:
## What was wrong

The inline-vis plugin hard-coded HTML extensions in its server validation and always rendered accepted files through an iframe, leaving no path for workspace Markdown documents to use the existing inline preview chrome. This meant agents had to convert Markdown artifacts to HTML or paste their contents directly into messages. Fixes [#3159](https://github.com/get-bb/bb/issues/3159).

## What changed

Generalized inline-vis file validation and its plugin-private RPC to support `.md` and `.markdown` files alongside HTML. The RPC now returns a discriminated HTML or Markdown result, including validated UTF-8 content for Markdown.

Markdown previews use BB's host Markdown renderer inside the existing bounded, scrollable preview card. HTML iframe behavior and sandboxing remain unchanged. Updated the plugin metadata, README, overview, and bundled agent skill with the new supported formats.

The plugin-private RPC was renamed from `prepareHtmlPreview` to `preparePreview`. No server/daemon wire contract changed, so `HOST_DAEMON_PROTOCOL_VERSION` was not incremented. No CLI changes were required.

## How you verified

Added red-first server coverage for Markdown content and extension handling, followed by frontend coverage for host-rendered Markdown, sidebar opening, bounded height, scrolling, and loading-to-ready layout stability.

Ran:

- `pnpm exec turbo run test typecheck --filter=bb-plugin-inline-vis` — 20 tests passed; typecheck passed.
- `bb plugin build plugins/inline-vis` — server and app bundles built successfully.

Fixes #3159

> AGENT GENERATED

### DIFF
--- a/plugins/inline-vis/PLUGIN_OVERVIEW.md
+++ b/plugins/inline-vis/PLUGIN_OVERVIEW.md
@@ -1,23 +1,24 @@
-See an agent's chart, demo, or report in the conversation without opening a side panel. The agent writes an HTML file to the workspace. The plugin shows that file inside the assistant message.
+See an agent's chart, demo, report, or Markdown document in the conversation without opening a side panel. The plugin shows the workspace file inside the assistant message.
 
 ## What you get
 
-- A live preview card in the message. Scripts and styles in the file run.
+- A live HTML preview or formatted Markdown document in the message.
 - A default viewport height of 224 pixels. The agent can set a height from 120 to 1200 pixels.
-- A header action that opens the source HTML file in the workspace viewer.
-- A clear inline error when the file is missing, too large, or not HTML.
+- A header action that opens the source file in the workspace viewer.
+- A clear inline error when the file is missing, too large, or unsupported.
 
 ## How it works
 
-The agent emits a message directive that names a workspace-relative `.html` or `.htm` file:
+The agent emits a message directive that names a workspace-relative `.html`, `.htm`, `.md`, or `.markdown` file:
 
 ```text
 ::inline-vis{file="charts/out.html" height="480"}
+::inline-vis{file="reports/summary.md"}
 ```
 
-The plugin confirms the file exists in the thread workspace before it renders. Files must be UTF-8 text with a maximum size of 5 MiB. Relative assets next to the file load as usual.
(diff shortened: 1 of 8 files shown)


SawyerHood:
Superseded by #3439, which integrates this PR with #3163 since the two conflicted in every file. The commits there credit @IlyaM as co-author. Thanks for both changes!

PR #3216, opened 2026-09-07
Report Cursor terminal PING timeouts as failed turns and reconcile Cursor usage

Flame119052:
## What was wrong

Cursor cloud transport failures (`RetriableError: [unavailable] PING timed out`) arrive over ACP as agent message text with `stopReason: end_turn`, so BB recorded the turn as completed instead of failed: a silent fake-success with no error notification and no retry path. Separately, the Cursor usage mapping dropped the dashboard total (`totalPercentUsed`) and bonus spend, rejected string-encoded percentages, and the sidebar bootstrap had no recovery for transient startup outages.

## What changed

- `packages/provider-bridge-acp/src/bridge/bridge.ts`: track terminal Cursor error text per turn; a turn ending with exactly the PING timeout text now fails legibly with an error notification. Gated to the Cursor dialect, `\\n\\nError:` prefix, 512-char cap, and no trailing recovery text.
- `packages/provider-bridge-acp/src/bridge/provider-maintenance.ts`: surface `Total usage` and `Bonus spend` windows from the live dashboard response; accept string-encoded percentages.
- `apps/app/src/hooks/queries/sidebar-navigation-query.ts`: retry transient/502-504 bootstrap failures with backoff plus polling recovery.
- `plugins/provider-acp`: rename to `provider-acp-opencode-native`, wire Grok/OpenCode native usage maintenance.
- Tests for all of the above.

## How you verified

- `@bb/provider-bridge-acp`: 325 passed (18 files), including 8 new PING-normalization cases and 3 new usage cases.
- Sidebar bootstrap suite: 11 passed.
- `turbo run typecheck --filter=@bb/provider-bridge-acp --filter=@bb/app`: pass.
- `plugins/provider-acp` server suite: 18 passed. (`public-sdk-only` flags pre-existing local build artifacts in the worktree only; unaffected files.)
- Live `cursor-agent acp` probe: connection and tool-use checks pass, so the timeout is transient Cursor-side, not a BB misconfiguration.

> AGENT GENERATED

### DIFF
--- a/apps/app/src/hooks/queries/sidebar-navigation-query.test.tsx
+++ b/apps/app/src/hooks/queries/sidebar-navigation-query.test.tsx
@@ -6,7 +6,8 @@ import {
   type SidebarBootstrapResponse,
 } from "@bb/server-contract";
 import { afterEach, describe, expect, it, vi } from "vitest";
-import { request } from "@/lib/api";
+import { HttpError, request } from "@/lib/api";
+import { shouldRetryTransientReadQuery } from "./query-helpers";
 import {
   MAX_CACHED_SIDEBAR_THREADS_PER_PROJECT,
   SIDEBAR_BOOTSTRAP_CACHE_KEY,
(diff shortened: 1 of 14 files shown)


SawyerHood:

PR #3250, opened 2026-09-08
Unify thinking rows and add a diagnostic events setting

ymichael:
## What was wrong

Live thinking used a separate indicator with italic details, while completed thoughts became generic system rows with monospace details and lost their expansion state. Persisted thoughts also broke activity groups when interleaved with tool work. Routine provider environment resolution also appeared in the conversation unconditionally, and development builds ignored the existing unhandled-event visibility setting.

## What changed

- Give reasoning its own timeline operation kind and a stable identity across completion. Active and completed thoughts share disclosure styling and prose details, and preserve manual expansion within the mounted thread. Completed thought headers use the same dimming as completed tool rows. Only completed thoughts display the AiBrain01 icon; live Thinking has no leading icon. Completed durations use the existing timeline treatment, which hides durations of one second or less.
- Replace `showUnhandledProviderEvents` with `showDiagnosticEvents`, covering environment resolution and unhandled provider events. It defaults to false in all builds; existing saved preferences carry over. Preserve the old setting name as a deprecated API alias for existing SDK clients, normalizing both names at the route boundary. Warnings and model fallback remain visible, and completed warnings/deprecations no longer receive completed-row dimming.
- Update timeline cache options, API/SDK settings, CLI coverage, configuration docs, guide templates, and the CLI skill. The CLI setting is `bb settings general showDiagnosticEvents <true|false>`.
- Bump the plugin SDK to 0.4.49 because the settings and timeline type changes alter its bundled declarations.
- Carry a canonical reasoning disclosure ID separately from the structural row ID so delegation nesting preserves expansion. Add a reasoning lifecycle story and regression coverage. No server/host-daemon wire payload changed, so no daemon protocol bump is needed.

- Register AiBrain01 across shared and mobile icon maps and regenerate the plugin icon registry. Use the existing Bot icon for Tasks agent presets in the sidebar and management panel.
- Allow 60 seconds for the existing Account Pool 4,096-session eviction test after repeated CI timeouts at 20 seconds; retain the workload and assertions.

- Include completed thoughts in activity groups, retaining chronological details while deriving labels only from work. Thoughts-only sequences stay visible, and messages, warnings, and approval boundaries still split groups. Update shared expansion handling and CLI rendering for mixed children.

- Exclude disabled diagnostics in database selection, event/byte budgets, nested-child hydration, and effective timeline cache sequences. Retain legacy Claude model-fallback events.

- Normalize canonical reasoning IDs in provider replay comparisons. Recognize legacy fallback records by their validated event shape in both SQL selection and projection, removing provider-specific branching from this path.

## How you verified

- CI follow-up: all 56 provider parity tests and 238 bridge-protocol tests pass, as do fallback extraction/visibility regressions and the provider-literal architecture guard.

- Review regressions cover old SDK payloads and read-modify-write updates through either setting name, canonical nested reasoning identity and expansion, a hidden 5 MiB environment event, unchanged visible cache sequence, and legacy model fallback. Database suite (449 tests), nested projection suite (74), reasoning UI suite (4), and affected typechecks/lint pass.

- Dimming and reasoning UI tests pass (14 tests); browser confirms completed thought headers use the same 0.4 opacity as completed work.

- Grouping verification: 382 thread-view tests and 254 client-core tests pass; all three reasoning UI tests pass. App, mobile, CLI, thread-view and client-core typechecks and lint pass. Browser fixture confirms grouped thoughts expand and retain expansion when a running exploration bundle becomes a closed exploration/edit step.
- Updated reasoning lifecycle tests pass: live thinking omits the icon, completion adds it, and expansion persists.
- Tasks plugin typecheck passes after switching preset icons to Bot.
- Plugin registry typecheck/tests and the Account Pool eviction test pass locally.
- App lint and the plugin SDK npm version guard pass after fixing the expansion callback dependency and bumping the SDK.
- Turbo typechecks passed for app, server, CLI, database, domain, thread-view, and server-contract.
- Full database and thread-view suites passed; subsequent targeted projection/title tests covered the added stable-identity regression.
- Focused app tests cover expansion through completion, independent subsequent thoughts, icon-free live thinking and icons on completed thoughts, warning readability, and settings cache invalidation. CLI settings tests passed.
- Server timeline-window/cache tests and production/development diagnostic visibility tests passed, including toggling cached results off, on, and off while retaining warnings.
- In an isolated source app, exercised the Settings toggle, verified persistence after reload and through the source CLI, and checked diagnostic visibility in a synthetic thread and expanded turn details. Browser-driven production component checks confirmed expanded Thinking becomes expanded Thought with preserved prose styling.
- `pnpm start:worktree` built and served successfully; server and host-daemon health checks passed. Browser checks used Chromium on macOS, not iOS or a real provider turn.
- `git diff --check` passed and the branch merges cleanly with current main.

> AGENT GENERATED


ymichael:
@slopcop please review this PR.

PR #3254, opened 2026-09-08
Fix collapsed model picker after chat handoff

MateoCerquetella:
## What was wrong

After handing a chat off to the new-thread composer, the model picker could appear to prevent model changes in a short window. Its reasoning and service-tier sections could not shrink, while the model list could shrink to almost zero height. The catalog loaded successfully, but its model rows were clipped and could not receive clicks.

## What changed

Make the desktop picker body one scrollable area for models, reasoning, service tier, and the handoff action. Keep the model list from shrinking, and reset the actual scroll viewport when searching or switching providers. Update picker regression coverage and check that model/provider changes preserve the handoff draft.

## How you verified

- 50 tests passed: `pnpm exec turbo run test --env-mode=loose --filter=@bb/app -- --run src/components/pickers/ModelReasoningPicker.test.tsx src/components/plugin/PluginNewThreadComposer.test.tsx`. A writable `TMPDIR` was supplied because the host's shared `/tmp` was full.
- Reproduced the collapsed list in Chromium at 780 × 437. After the fix, clicked **Handoff to new thread**, selected another model, scrolled to reasoning/service-tier controls, and searched for another model. Selection and the source-thread mention survived reload.
- Captured the comparison below by clicking **Handoff to new thread** on the original picker, then restoring the fix and testing the same handoff draft at the same 780 × 437 viewport. Before: the list was 4 px tall, and attempting to click Terra left Sol selected. After: the list was 189 px tall and clicking Terra changed the selection. Compact-drawer component tests pass; native iOS Safari was not run.

### Before: after Handoff to new thread

The composer contains **Continue from Reply only with ok.**, confirming the handoff. The model list is clipped to 4 px. Attempting to select **5.6-Terra** leaves **5.6-Sol** selected.

Before: handoff draft is present but other models cannot be clicked

### After: a different model selected in the same handoff draft

At the same window size, the model choices are visible and **5.6-Terra** is selected. The source-thread mention stays intact.

After: selecting 5.6-Terra in the handoff draft

### DIFF
--- a/apps/app/src/components/pickers/ModelReasoningPicker.test.tsx
+++ b/apps/app/src/components/pickers/ModelReasoningPicker.test.tsx
@@ -503,7 +503,7 @@ describe("ModelReasoningPicker", () => {
     ).toBe("");
   });
 
-  it("caps the desktop picker and scrolls only the model list", () => {
+  it("keeps desktop models from collapsing and lets all controls scroll", () => {
     renderPicker({ modelOptions: manyCodexModels });
 
     fireEvent.click(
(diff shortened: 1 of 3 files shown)


SawyerHood:
I agree with the bug, but not the fix. This makes using model like pi really hard. I might move the reasoning to be tabs

SawyerHood:
image

See this for context

PR #3255, opened 2026-09-08
Keep rendered diff cards during refresh

ymichael:
## What was wrong

Realtime workspace-change invalidation deleted every cached per-file patch before the refreshed diff table of contents could establish whether its contents had changed. Visible cards therefore regressed from rendered content to loading skeletons, including for filesystem changes that produced the same Git diff.

## What changed

Retain per-file patch cache across realtime workspace invalidations. When the diff table of contents refreshes, refresh visible cached patches in place while retaining their rendered content; replacement patch content is rendered only after it arrives. Initial patch seeding only notifies React when its patch payload actually changed. No wire or CLI changes.

## How you verified

- `pnpm exec turbo run test --filter=@bb/app -- --run src/hooks/realtime-cache-effects.test.ts src/hooks/queries/use-environment-diff-patches.test.tsx`
- `pnpm exec turbo run lint --filter=@bb/app`
- `git diff --check`

Fixes #

> AGENT GENERATED


ymichael:
@slopcop

PR #3261, opened 2026-09-08
Fix duplicate native and injected skills

MPIsaac-Per:
## What was wrong

BB deduplicated only its injected skill catalog. Providers independently discover their native roots, so same-name, byte-identical skills from distinct native and injected paths were both exposed at runtime.

## What changed

`host.list_skills` can now return opt-in whole-skill content hashes. Thread runtime configuration compares same-name injected and provider-native skills, omits an injected copy only when the complete skill content matches, preserves project overrides, and logs conflicting content without hiding it. The hash includes references and excludes registry provenance metadata to match server tree manifests.

The host-daemon protocol is bumped from 183 to 184. Matching server and host-daemon releases are required.

## How you verified

- Focused server tests: 37 passed, covering identical cross-path copies, shared-user acquisition, project overrides, conflicts, and providers without native roots.
- Focused host-daemon tests: 24 passed, including references and registry-metadata hash semantics.
- Host-daemon contract tests: 37 passed.
- `pnpm exec turbo run typecheck --filter=@bb/server --filter=@bb/host-daemon --filter=@bb/host-daemon-contract`

### DIFF
--- a/apps/host-daemon/src/command-discovery.ts
+++ b/apps/host-daemon/src/command-discovery.ts
@@ -10,6 +10,7 @@ import type {
   HostProviderCommand,
   SkillRootKind,
 } from "@bb/host-daemon-contract";
+import { hashInstalledSkillDirectory } from "./injected-skills.js";
 
 const SKILL_FILE_NAME = "SKILL.md";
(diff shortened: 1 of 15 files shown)


SawyerHood:

PR #3271, opened 2026-09-08
feat(git-diff): persist diff view mode and line wrap across reloads

danielbachhuber:
## Human comments

From https://discord.com/channels/1522478522332090528/1545179385194356856/1545395969603014657

## What was wrong

Both git-diff toolbar preferences were component state, so they reset on remount and reload. #3868 had already made an explicit split or unified choice win at every panel width, including resize, breakpoint, and close/reopen transitions, but that choice remained in memory only for the mounted panel.

## What changed

Persist git-diff display mode and line overflow mode in device-local localStorage. This adds reload persistence on top of #3868's explicit-choice-at-every-width rule while keeping compact viewport choices session-only.

`apps/app/src/lib/git-diff-view-preferences.ts` adds validated Jotai `atomWithStorage` preferences, read on initialization with `{ getOnInit: true }`: nullable `bb.thread.gitDiff.displayMode` (`unified | split`) and `bb.thread.gitDiff.lineOverflowMode` (`wrap | scroll`, defaulting to `scroll`). [`apps/app/src/components/secondary-panel/git-diff/useResponsiveGitDiffPanelDisplay.ts`](https://github.com/get-bb/bb/pull/3271/changes#diff-28cd9e28802c7dd45a9959bd40a02922008cfb2e5741048cfaa50f9c9874a6b2) uses the stored display preference and exports `resolveGitDiffDisplayMode`; `apps/app/src/components/secondary-panel/ThreadSecondaryPanel.tsx` uses the stored line-overflow preference for its wrap toggle. `apps/app/src/components/secondary-panel/git-diff/useResponsiveGitDiffPanelDisplay.test.tsx` now has 14 tests.

<details>
<summary>Decisions and scope</summary>

- These are device-local view preferences, so they use Jotai localStorage rather than `AppSettings` or `PUT /settings/general`; server and CLI access are not needed.
- `createNullableLocalStorageEnumStorage` and `createLocalStorageEnumStorage` validate the stored string unions.
- Outside compact viewports, a stored display preference overrides the 760px (`GIT_DIFF_SPLIT_VIEW_MIN_WIDTH_PX`) rule at every width. The width rule supplies the default only when the nullable preference is unset, so a stored `split` can be cramped below 760px.
- Compact viewports (`useIsCompactViewport()`, `max-width: 767px`) use a session-only display choice that defaults to `unified`. A desktop `split` preference does not carry into a phone-width drawer, and a mobile selection does not overwrite the desktop preference. The compact toggle can still select `split` for the session.
- The stored preference replaces main's `hasExplicitDisplayModeRef` and its width-change guard. The hook continues tracking width to supply the fallback when no preference is stored.
- `resolveGitDiffDisplayMode({ isCompactViewport, compactDisplayMode, displayModePreference, isWideEnoughForSplit })` is a pure export so its combinations can be tested without rendering.
- The collapse-all chevron remains unpersisted because its default depends on each diff's file count and delete status. `FilePreview.tsx` and `markdown-preview.tsx` keep independent unpersisted wrap toggles; sharing them is a separate product decision.
- This adds no settings-panel UI and no UI to return a non-compact display mode to unset. Cross-tab sync comes from the storage subscription but is untested. Compact session choices surviving viewport changes within one hook instance are intended but untested. localStorage is per origin, so preferences do not carry between the packaged app and a checkout's dev app on another port.
- After merging main, `git diff upstream/main --stat` shows changes only in the four files above. #3819's `ThreadSecondaryPanel.tsx` changes do not touch display mode or wrap state.

</details>

## How you verified

- `pnpm exec vitest run --config vitest.config.ts src/components/secondary-panel` in `apps/app`: 41 files and 405 tests passed.
- `pnpm exec tsc --noEmit -p .` in `apps/app`: passed.
- `pnpm exec turbo run typecheck lint test --filter=@bb/app`: typecheck passed; lint reported 0 errors and 191 warnings, with none new in these files; tests passed in 551 of 552 files.

`ThreadTerminalView.test.ts` was the remaining failure. It also fails when run alone and this PR does not touch it. The local Node version was v20.19.3, while the repository requires `>=22.19.0`; the likely Node v20 attribution is inferred from the stack trace, where xterm reads a global `navigator`, and is not confirmed on Node 22.

No manual reload check was performed in a running app. Reviewers should toggle both controls and reload within one app instance.

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/secondary-panel/ThreadSecondaryPanel.tsx
+++ b/apps/app/src/components/secondary-panel/ThreadSecondaryPanel.tsx
@@ -49,10 +49,7 @@ import type {
   SecondaryPanelTabReorderHandler,
 } from "./secondaryPanelTab";
 import { useEnvironmentDiffFiles } from "@/hooks/queries/environment-queries";
-import {
-  DEFAULT_CODE_OVERFLOW_MODE,
-  type CodeOverflowMode,
-} from "@/lib/code-overflow-mode";
+import { useGitDiffLineOverflowModePreference } from "@/lib/git-diff-view-preferences";
 import type { DiffPresentation } from "@/components/code/code-rendering";
 import { useGitDiffPanelState } from "./git-diff/useGitDiffPanelState";
 import { useResponsiveGitDiffPanelDisplay } from "./git-diff/useResponsiveGitDiffPanelDisplay";
@@ -267,11 +264,10 @@ export function ThreadSecondaryPanel({
   const {
     gitDiffDisplayMode,
     handleGitDiffDisplayModeChange,
-    handleSecondaryPanelResizeStart,
     handleSecondaryPanelWidthChange,
   } = useResponsiveGitDiffPanelDisplay({ isSecondaryPanelOpen: isOpen });
   const {
-    handleSecondaryPanelDragging: handleResizeDragging,
+    handleSecondaryPanelDragging,
     handleSecondaryPanelResize,
     handleSecondaryPanelResizePointerDownCapture,
     persistedWidthPercent,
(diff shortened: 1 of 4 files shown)


SawyerHood:

PR #3310, opened 2026-09-08
Account Pooler: rebuild Codex WebSocket history from streamed output items

SawyerHood:
## What was wrong

Codex talks to the Account Pooler over its Responses WebSocket transport: after the first request on a socket it sends `previous_response_id` plus only the new input items, and the pooler rebuilds the full history itself before forwarding one HTTPS request upstream. The pooler took the previous response's items from the `response.completed` event, but the ChatGPT Codex backend streams them only as `response.output_item.done` events and sends `"output": []` in `response.completed` (verified with a direct probe through the pooler). Every incremental request therefore lost the previous response. Within a turn the API rejected the orphaned tool output with "No tool call found for custom tool call output" and Codex silently retried on a fresh socket, which is why Codex logs show a reconnect for almost every request. At a turn boundary the request is valid without the previous answer, so the model saw the earlier question unanswered and answered it again before the new one. Codex's own log database (`~/.codex/logs_2.sqlite`) records hundreds of these rejections per day since the WebSocket routing landed in #3056.

## What changed

`plugins/account-pool/src/codex-websocket.ts` accumulates `response.output_item.done` items in stream order for each forwarded response and uses them as the previous output when the next frame carries `previous_response_id`. It falls back to the completed event's `output` only when no items were streamed. This mirrors how Codex builds its own delta baseline. No wire, CLI, or documentation changes.

## How you verified

- New `plugins/account-pool/src/codex-websocket.test.ts` drives the handler with a fake hub: a tool-call continuation within a turn, a follow-up question across a turn boundary, and the completed-output fallback. The first two fail on the previous handler and pass now.
- The existing integration test's fake upstream now streams `response.output_item.done` and sends an empty `output` in `response.completed`, matching the real backend, and its history assertion still holds.
- `pnpm exec turbo run test typecheck lint --filter=bb-plugin-account-pool` passes.

> AGENT GENERATED

### DIFF
--- a/plugins/account-pool/src/codex-websocket.test.ts
+++ b/plugins/account-pool/src/codex-websocket.test.ts
@@ -1,0 +1,147 @@
+import { describe, expect, it, vi } from "vitest";
+import type { ExperimentalPluginWebSocket } from "@get-bb/plugin-sdk";
+import { createCodexWebSocketHandlers } from "./codex-websocket.js";
+import type { AccountPoolHub } from "./hub.js";
+
+function sse(events: ReadonlyArray<Record<string, unknown>>): string {
+  return `${events
+    .map(
+      (event) => `event: ${String(event.type)}\ndata: ${JSON.stringify(event)}`,
+    )
+    .join("\n\n")}\n\ndata: [DONE]\n\n`;
+}
+
+function createFixture(upstreamBodies: string[]) {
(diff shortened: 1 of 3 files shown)


SawyerHood:
Superseded: the same bug was fixed on main in #3307, and a follow-up will remove the WebSocket path from the pooler entirely.

PR #3325, opened 2026-09-09
Fix timeline pagination with snapshot-bound conversation groups

ymichael:
## What was wrong

Timeline pages grouped independent event/byte windows and assigned fragment IDs. The client merged by row ID, so boundaries could change accepted messages to pending, fragment turns, and lose or duplicate content. Reproduced on main before implementation: three of four large copied-corpus walks disagreed with canonical history.

## What changed

Select conversation groups using existing anchors, load their full turn/delegation context, and filter context-only groups out of page ownership. Oversized groups paginate canonical children under stable summary IDs; the client and CLI recursively merge those children. The existing detail endpoint and SDK now support continuation cursors, and the app walks them on expansion.

Opaque cursors bind the history tip, grouping version, status, and display surface. Appends stay outside an existing walk. Pagination continues on a best-effort basis across history edits and suffix replacement: pages loaded before and after an edit may disagree until history is reloaded. A new latest snapshot replaces loaded history because late events can change earlier grouping. Cursors whose anchor event was deleted return HTTP 400, preserving main's behavior. Existing message-edit notifications invalidate request and grouping-context caches. There is no new database table, migration, event trigger or daemon wire change. Main's large-output sidecars remain intact; no rejected streaming-projection code was used.

## How you verified

- Actual Hono endpoint plus actual client merge matched complete canonical content/order on four copied threads: 63,853, 42,033, 32,384, and 22,940 events. The oracle removes one exactly duplicated canonical delegation row in the fourth thread; it does not discard differing content with the same ID.
- Tiny event/512-byte targets, huge turns/details, nested/late events, straddling items, appends between requests, continuation after in-place edits, same-sequence suffix replacement, rejection of deleted anchors, and a stale in-flight app page are covered.
- Turbo: 158 server timeline/public-route tests, 19 client merge tests, 142 database event/migration tests, 30 app query/controller tests, 18 CLI log tests, 104 SDK tests, and 67 server-contract tests. Affected server/client/CLI/app/database/SDK/contract typechecks passed. Server tests and typechecks passed again after rebasing onto `682f0d2c5d`.
- Isolated source browser: paged to all 45 synthetic messages exactly once, expanded the oldest summary into 20 commands, and reloaded. Source CLI `thread log --all --format verbose` returned all 45 messages exactly once. Owned dev processes/store were cleaned up. No provider execution or mobile Safari verification is claimed. The verification inventory has an existing unrelated unmapped `browser` CLI family.

The initial 1.7–5.0 s walks were rejected as too slow. The optimized paired run restores main's original timeline builder, event queries and client merge from `682f0d2c5d` and runs main/fixed/fixed/main per copied thread, with fresh databases and route caches. Correct walks now take **0.72–1.76 s**, versus **0.68–2.15 s** for main: the two largest walks are faster, and the other two are within 6%. All eight fixed walks match canonical content/order. Cold latest requests retain roughly 3 ms and 13 ms overhead on two threads; exact numbers are in the report.

The optimization scopes payload context, selects turn IDs before reading full event bodies, skips already-loaded request lookups, and uses existing selective indexes for turn, child and interruption queries. Only compact request-to-turn IDs and an ordering boundary are cached per exact snapshot; no projected rows or payloads are cached. The profile clears that cache to include its cold construction cost. New tests guard against decoding unrelated history and acceptance after a later conversation boundary. CI's missed optional-contract and old sequence-fragment expectations are now covered and pass locally.

Ordering overlap detection uses a linear scan, with six cases for accepted turns, overlapping alternatives and expired spans. A paired comparison against `51e0dc78d2` preserves canonical results but does not establish another general latency improvement. History-edit invalidation was removed in favor of the best-effort contract above; regression tests require these cursors to continue.

[Full benchmark report](https://github.com/get-bb/bb/blob/bb/timeline-pagination-correctness/docs/timeline-pagination-verification.md). [Pagination lifecycle and SDK contract](https://github.com/get-bb/bb/blob/bb/timeline-pagination-correctness/docs/timeline-pagination.md).

> AGENT GENERATED


ymichael:
@slopcop

ymichael:
@slopcop

PR #3479, opened 2026-09-11
Allow themes to set the terminal font family (#3236)

### LINKED ISSUE #3236: feat(terminal): allow themes to set the terminal font family
## The workflow

I was trying to make bb Monokai use my installed `BerkeleyMono Nerd Font Mono` in the integrated terminal. The palette sets terminal colors, but it cannot supply the font family to the terminal renderer.

## What happens today

Source inspection on `main` at `06aeaa994942ae7527dc49d2268c1f801e8542a0` confirms that `ThreadTerminalView` passes a fixed font stack to xterm. Live visual reproduction is unverified. The local CLI reports bb `0.42.1` on macOS `26.5.2`.

### Why

[`TERMINAL_FONT_FAMILY`](https://github.com/get-bb/bb/blob/06aeaa994942ae7527dc49d2268c1f801e8542a0/apps/app/src/components/thread/terminal/ThreadTerminalView.tsx#L49-L50) starts with:

```ts
'"JetBrainsMono Nerd Font Mono", "MesloLGS NF", "Symbols Nerd Font Mono", ...'
```

[Terminal construction](https://github.com/get-bb/bb/blob/06aeaa994942ae7527dc49d2268c1f801e8542a0/apps/app/src/components/thread/terminal/ThreadTerminalView.tsx#L899-L907) uses `fontFamily: TERMINAL_FONT_FAMILY`. BB also [loads the optional WebGL renderer](https://github.com/get-bb/bb/blob/06aeaa994942ae7527dc49d2268c1f801e8542a0/apps/app/src/components/thread/terminal/ThreadTerminalView.tsx#L932-L939), so overriding `.xterm-rows` with CSS does not configure that renderer's font. The [theme-change effect](https://github.com/get-bb/bb/blob/06aeaa994942ae7527dc49d2268c1f801e8542a0/apps/app/src/components/thread/terminal/ThreadTerminalView.tsx#L1142-L1148) updates `terminal.options.theme` only.

### Proposal

We should let palettes supply a terminal font-family token, for example:

```css
--terminal-font-family: "BerkeleyMono Nerd Font Mono", "Berkeley Mono", monospace;
```

Read it into xterm's `fontFamily` at construction and when the active theme changes. Preserve the current default stack when the token is absent, and refit the terminal after a font change so cell dimensions and PTY sizing stay aligned. It should work with WebGL and the fallback renderer.

## Context and alternatives

Related: #2197 covers independent interface, chat, and code font settings. This request concerns the integrated terminal and palette control of its renderer.

The installed font's name table confirms `BerkeleyMono Nerd Font Mono`, matching Monokai's existing CSS override. Changing that CSS family name alone would not change the constructor option above.

## Checks

- [x] I searched open and closed issues for the same request, including terminal font, xterm font, Berkeley, Monokai, and font-family searches.

> AGENT GENERATED

0xferrous:
## What was wrong

The integrated terminal always used a fixed font stack, so themes could change terminal colors but not the terminal font family. See [#3236](https://github.com/get-bb/bb/issues/3236). Fonts loaded through a delayed CSS import can arrive after the initial `document.fonts.ready` promise resolves, and shared WebGL glyph caches can retain fallback glyphs when font loading leaves cell dimensions unchanged.

## What changed

Added `--font-terminal` with the existing terminal font stack as its default. The terminal reads it at construction and applies active-theme changes. A listener retained for each mounted terminal handles subsequent font loads, remeasures character cells, clears xterm's glyph atlas, and schedules a refit. Cleanup removes the listener and prevents pending callbacks from refreshing a disposed terminal. Updated theming and configuration documentation.

## How you verified

- Focused terminal and theme tests through Turbo: 61 passed, including delayed loading and listener cleanup regressions.
- App typecheck and lint through Turbo passed; lint reported 0 errors.
- Chromium harness using the PR's xterm versions reproduced both font-loading failures before the fix and passed with the updated functions: delayed imports update cell measurements, and two WebGL terminals sharing an atlas redraw loaded-font glyphs.
- Browser verification covered the focused xterm harness; the full BB UI was not exercised.

Fixes #3236

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/thread/terminal/ThreadTerminalView.test.ts
+++ b/apps/app/src/components/thread/terminal/ThreadTerminalView.test.ts
@@ -3,9 +3,11 @@ import { Terminal } from "@xterm/xterm";
 import { describe, expect, it, vi } from "vitest";
 import {
   buildTerminalThemeFromCssColors,
+  applyTerminalFontFamily,
   captureTerminalContextMenuState,
   decodeTerminalOutputBytes,
   encodeTerminalInputChunks,
+  forceTerminalFontMeasurement,
   focusTerminalFromTouchRelease,
   forwardTerminalData,
   loadOptionalTerminalWebglAddon,
@@ -15,6 +17,7 @@ import {
   TERMINAL_ALLOW_PROPOSED_API,
   TERMINAL_FONT_FAMILY,
   TERMINAL_UNICODE_VERSION,
+  resolveTerminalFontFamily,
   writeTerminalOutput,
   updateTerminalTouchFocusGesture,
 } from "./ThreadTerminalView";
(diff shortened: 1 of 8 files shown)


0xferrous:
i've tested it against local build, it worked

AI Disclosure: I don't have much experience with frontend development, but I got gpt sol to review it and the changes from my perspective look fine, only thing I am less experienced to check is the xterm forced font measurement and fitting. I did however built the code and test it to verify that the terminal font customization applies.

PR #3480, opened 2026-09-11
Guard scheduled workflows against forks

fgrehm:
## Human comments

I'm getting daily emails about failed workflow runs on my fork which is a bit annoying. I know I can disable it on my fork but this might be better in the long run and less annoying for others too :smile:

## What was wrong

GitHub Actions runs `schedule:`-triggered workflows in forks too, on the default branch. This repository has three cron workflows (`.github/workflows/publish-bb-app.yml`, `mobile-e2e.yml`, `marketplace-v2-live.yml`), none of which check which repository they run in, so every fork copies the nightly job load: a release-train run that attempts an npm publish it can never succeed at (no `npm-release` OIDC configuration, no macOS signing/notarization or Expo secrets), two long desktop builds plus an EAS/TestFlight submission, a paid-runner iOS simulator E2E suite, and a marketplace validation run. Fork owners need to disable workflows manually to stop this.

## What changed

Added a fork guard so that scheduled runs are skipped outside the canonical repository while manual `workflow_dispatch` (and all push/PR triggers) keep working everywhere:

- `publish-bb-app.yml`: `if: github.event_name != 'schedule' || github.repository == 'get-bb/bb'` on the `publish` and `publish-plugin-sdk` jobs. These two are the only nodes that start on the `schedule` trigger; every other job (`publish-nightly`, `nightly-desktop-*`, `nightly-mobile-ios`, `nightly-desktop-publish`) already requires their upstream jobs to report `success`, so a skipped root skips them transitively.
- `mobile-e2e.yml`: the same repository check added to the existing label-based `if` on the `ios` job.
- `marketplace-v2-live.yml`: same guard on the `validate` job.

No `HOST_DAEMON_PROTOCOL_VERSION`, CLI, or docs changes; workflow-only.

## How you verified

- Parsed all three modified workflows with a YAML parser (no syntax errors).
- Traced job graphs: on a scheduled run in a fork, `publish` and `publish-plugin-sdk` are skipped, and `needs.publish.result == 'success'` (under `!cancelled()`) then skips `nightly-desktop-macos`, `nightly-desktop-linux`, and `nightly-mobile-ios`; `nightly-desktop-publish` needs those, so nothing runs. In the canonical repo `github.repository == 'get-bb/bb'` holds, so nightly behavior is unchanged. On `workflow_dispatch`, the event-name clause short-circuits to true in any repository, so manual dry-run releases and iOS E2E runs still work in forks.
- Not run on GitHub (no actions executed as part of preparing this patch).

> AGENT GENERATED

---

<sub><em>Generated with Omen Alpha (Opencode Go) via Pi</em></sub>

### DIFF
--- a/.github/workflows/marketplace-v2-live.yml
+++ b/.github/workflows/marketplace-v2-live.yml
@@ -11,6 +11,11 @@ permissions:
 jobs:
   validate:
     name: Parse the live v2 marketplace
+    # Scheduled runs are restricted to the canonical repository so that forks,
+    # on which GitHub happily runs cron workflows, do not run the validation
+    # against the upstream marketplace for no benefit. Manual dispatch keeps
+    # working in forks.
+    if: ${{ github.event_name != 'schedule' || github.repository == 'get-bb/bb' }}
     runs-on: blacksmith-4vcpu-ubuntu-2404
     timeout-minutes: 10
 
--- a/.github/workflows/mobile-e2e.yml
+++ b/.github/workflows/mobile-e2e.yml
@@ -53,9 +53,14 @@ env:
 jobs:
   ios:
     name: iOS simulator flows
+    # Scheduled (nightly) runs are restricted to the canonical repository so
+    # that forks, on which GitHub happily runs cron workflows, do not burn
+    # macOS runner time they do not have. Label gating and manual dispatch
+    # keep working in forks.
     if: >-
-      github.event_name != 'pull_request' ||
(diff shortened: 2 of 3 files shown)


SawyerHood (approved):
Reviewed: job graph in publish-bb-app.yml is fully covered by the guard on the two root jobs, canonical-repo schedule behavior unchanged, fork PRs still label-gate correctly. Note: the mobile-e2e guard applies to every event rather than only schedule, which is stricter than the comment says but harmless in the canonical repo.

Issue #2634, opened 2026-08-28
Codex follow-up request is accepted in one turn but tool work is emitted in an unaccepted successor turn


uje-m:
---
title: 'Codex follow-up request is accepted in one turn but tool work is emitted in an unaccepted successor turn'
severity: 'major'
target: 'get-bb/bb'
---

## Summary

On a fresh Codex-provider thread, a later `turn/start` request can be persisted as accepted and completed in turn A, while the provider's actual reasoning/tool call is emitted in a newly started turn B that has no `turn/input/accepted` event. Any consumer that correctly binds effects to the accepted turn must reject the otherwise-valid tool call.

## Reproduction

1. Start a fresh Codex thread with a registered native BB agent tool.
2. Complete one ordinary tool-calling turn successfully.
3. Send a second `turn/start` request that explicitly calls the same tool.
4. Inspect the append-only thread events.

## Observed

- `client/turn/requested` for request R targets a new turn.
- turn A starts and records `turn/input/accepted` for R.
- turn A immediately records `turn/completed` with no work items.
- turn B starts at the same timestamp without `turn/input/accepted`.
- the requested tool call starts/completes in turn B.

This reproduced on multiple fresh threads. The first tool-calling turn did not split.

## Expected

The provider work caused by request R must remain in the turn that records R's input acceptance, or the successor turn must carry an explicit durable correlation to R that effect consumers can validate. A completed/unaccepted successor must not inherit the request implicitly.

## Safety impact

This is not a presentation-only mismatch. Exactly-bound tool consumers must fail closed, so valid user-requested tool work becomes unusable. Weakening consumers to accept a tool call from an unrelated/unaccepted turn would create a cross-turn authority bug.

## Notes

No private plugin, policy, repository, thread identifier, nonce, or product-specific workflow is required to reproduce this event-order defect.

SawyerHood:
Duplicate of #2580.

Consolidating tracking into #2580. Added evidence and scope from this report have been preserved in a consolidation comment there.

Closing this duplicate does not indicate that the underlying work is completed.

> AGENT GENERATED

PR #3503, opened 2026-09-11
Correlate Codex turn dispatches with the turn the turn/start response names

### LINKED ISSUE #2580: Codex bridge fabricates a completed turn when turn/started is delayed

SawyerHood:
## What was wrong

Stacked on #3502. Codex can return `turn/start` before emitting `turn/started`. The bridge discarded the response and fabricated a completed turn after 250 ms, attaching accepted input to that synthetic turn while the real work ran in a second, unaccepted turn (#2580 and consolidated #2634). Compaction's empty response has the same delayed-notification problem.

## What changed

Preserves Ratul Sarna's original #2639 implementation and authorship, with lifecycle and failure-path fixes:

- Open and correlate the turn named by `turn/start`; deduplicate its later native lifecycle notifications. Dedupe is scoped by provider thread, cleared on exit, and bounded to 256 pending lifecycle records per thread.
- Await compaction's actual turn. An idle/system-error report or process exit can settle an unopened dispatch. Register it before sending the request, buffer early status until RPC acceptance, and remove it as soon as its input is accepted. A rejected request cannot fabricate an accepted/completed turn.
- Retry an interrupt once only for the exact `-32600` / `no active turn to interrupt` rejection when that same response-opened turn lacked native activation at the initial attempt. Wait for activation, return the original error on timeout, and avoid retrying a turn that already settled.
- Preserve the legacy zero-work fallback only for a `turn/start` result that names no turn.
- Bump `HOST_DAEMON_PROTOCOL_VERSION` from #3502's 212 to 213. Update lifecycle regressions to use the current test harness after rebasing onto main.

## How you verified

- Nine new failure-path regressions fail against the previous #3503 implementation and pass with these fixes: unrelated interrupt errors, interrupt after native activation, completion-before-error, activation timeout, compact idle/error before response, cross-thread dedupe, and exit cleanup. Additional cases exercise a single retry, activation-before-rejection, delayed terminal notifications, compaction exit, and consecutive requests' tool attribution.
- `pnpm exec turbo run test typecheck --filter=bb-plugin-provider-codex --filter=@bb/provider-bridge-protocol --filter=@bb/provider-parity --filter=@bb/host-daemon-contract` passes on the rebased layer. Codex: 317 tests; protocol: 279; parity: 56; contract: 59.
- Real Codex 0.154.0 via the configured account pool, with fresh empty `CODEX_HOME` directories and workspaces: a relay delays native start notifications 400 ms. Reverting this layer to #3502 reproduces two lifecycles for both an ordinary turn and compaction; the final layer produces one accepted lifecycle. Delaying completion notifications also proves the review fix: the prior bridge sends two interrupts after native activation/real completion; the fixed bridge sends one and preserves the real `-32600` rejection. No BB core or imported store was started, and no credentials were copied.

Fixes #2580
Closes #2639

> AGENT GENERATED

### DIFF
--- a/packages/host-daemon-contract/src/protocol.ts
+++ b/packages/host-daemon-contract/src/protocol.ts
@@ -1,3 +1,3 @@
-export const HOST_DAEMON_PROTOCOL_VERSION = 204 as const;
+export const HOST_DAEMON_PROTOCOL_VERSION = 205 as const;
 
 export const HOST_ARTIFACT_MAX_BYTES = 256 * 1024 * 1024;
--- a/packages/host-daemon-contract/test/contract.test.ts
+++ b/packages/host-daemon-contract/test/contract.test.ts
@@ -999,7 +999,7 @@ const CONTRIBUTED_ENV = [
 
 describe("host-daemon command schemas", () => {
   it("uses the current host-daemon protocol version", () => {
-    expect(HOST_DAEMON_PROTOCOL_VERSION).toBe(204);
+    expect(HOST_DAEMON_PROTOCOL_VERSION).toBe(205);
     expect(HOST_ARTIFACT_MAX_BYTES).toBe(256 * 1024 * 1024);
   });
 
--- a/plugins/provider-codex/src/bridge/bridge.ts
+++ b/plugins/provider-codex/src/bridge/bridge.ts
@@ -60,6 +60,7 @@ import {
 } from "../interactive-requests.js";
 import { parseModelsResponse } from "../models.js";
 import { macOsPermissionPresentation } from "../presentation.js";
+import { codexTurnSchema } from "../schemas.js";
 import {
   resolveCodexInstructionOverrides,
   toCodexDynamicTools,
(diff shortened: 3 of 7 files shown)


ratulsarna:
Yay!

PR #3544, opened 2026-09-11
Recover iOS sidebar swipes after interrupted touch sessions

salemsayed:
## What was wrong

The compact sidebar installs capture listeners before the Send control handles its pointer event. In iOS WKWebView, submitting a follow-up can blur the composer and detach the Send target before the matching terminal touch event reaches the window listener. That leaves `swipeSessionRef` occupied, so every later sidebar gesture is rejected until the page reloads.

## What changed

- Replace unfinished sidebar tracking when a fresh primary pointer or single-touch gesture begins.
- Ignore non-primary pointer starts so a second finger cannot replace a valid gesture.
- Keep the existing edge guard, swipe thresholds, passive vertical-scroll behavior, and target exclusions unchanged.
- Add focused interrupted-session regression coverage, including the detached Send target sequence, missing pointer/touch terminal events, touch identifier reuse, vertical scrolling, and secondary fingers.
- Add a Maestro iOS flow that records video and screenshots before Send and after Send, then asserts that the sidebar still opens.
- Include that flow in the label-gated Mobile E2E default set so its media is uploaded with the workflow artifacts.

There are no server/daemon wire, CLI, SDK, configuration, or protocol changes.

## How you verified

- `pnpm exec turbo run test --filter=@bb/app -- src/components/ui/sidebar.test.tsx src/components/promptbox/PromptBoxInternal.test.tsx`: 166 tests passed. Five new regression cases fail without the implementation change.
- `pnpm exec turbo run typecheck --filter=@bb/app`: passed.
- `pnpm exec turbo run test --filter=@bb/mobile -- src/lib/e2e/mobile-e2e-flows.test.ts src/lib/e2e/ci-run-flows.test.ts`: 4 tests passed.
- `bash -n apps/mobile/e2e/scripts/ci-run-flows.sh`: passed.
- Physical iPhone, TestFlight shell, current source bundle served from an isolated direct server: before the patch, swipe-open worked after reload and stopped immediately after Send; with the patched bundle, swipe-open continued working after multiple sends.
- Upstream CI passed before the Mobile E2E wiring follow-up; the updated CI run is linked on this PR.

Maintainer action for simulator media: add the `mobile-e2e` label. Contributor permissions cannot apply repository labels.

> AGENT GENERATED

### DIFF
--- a/apps/app/src/components/ui/sidebar.test.tsx
+++ b/apps/app/src/components/ui/sidebar.test.tsx
@@ -37,8 +37,8 @@ function settleMobileToggle() {
   });
 }
 
-function createTouch(clientX: number, clientY: number): Touch {
-  return { identifier: 1, clientX, clientY } as Touch;
+function createTouch(clientX: number, clientY: number, identifier = 1): Touch {
+  return { identifier, clientX, clientY } as Touch;
 }
 
 function createTouchList(...touches: Touch[]): TouchList {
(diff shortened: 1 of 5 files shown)


salemsayed:
Follow-up verification against latest upstream `main`:

- Source: clean detached worktree at `d4a7c3d993cbd3bf81ea7a9f18cac7f301d913fc`; none of this PR was present.
- Delivery: production web bundle served from a fresh isolated data directory through a distinct HTTPS origin, avoiding reuse of the patched WebView cache.
- Device: the same physical iPhone/TestFlight shell used for the patched check.
- Result: sidebar swipe-open worked before Send and stopped working immediately after Send again.

The isolated server, Tailscale endpoint, data directory, and temporary worktree were removed after the test.

> AGENT GENERATED

PR #3668, opened 2026-09-14
Treat Cursor ACP terminal errors as failed turns

timigod:
## What was wrong

Cursor ACP (`cursor-agent` ACP mode) catches transport failures, writes them as assistant text (`\n\nError: RetriableError: …`, also `NonRetriableError` / `ActionRequiredError`), and returns `stopReason: end_turn`. bb's ACP translator maps `end_turn` to a completed turn and flushes still-open tools as successful. Observed events: `item/started agentMessage` → `item/agentMessage/delta` (the whole error string) → `item/completed agentMessage` → `turn/completed {"status":"completed"}`, with no `provider/error` or `turn/failed`. Pending `fileChange` items were marked completed in the same second (thread `thr_kfrp7rydm3`, 2026-09-14). Variants included `[canceled] http/2 stream closed`, `Stream ended without turnEnded`, `getaddrinfo ENOTFOUND api2.cursor.sh`, `Connection stalled`, `PING timed out`, `read ETIMEDOUT`, `Service Unavailable`, and `HTTP 502`.

Separately, `provider-retry` only retried `overloaded` and qualifying `rate-limit` failures, so even a correctly failed turn would not recover a dropped connection.

## What changed

- Cursor dialect only: if the current turn's accumulated assistant text is exactly Cursor's terminal emission and the turn ends with `end_turn`, emit `provider.error` with `settlesTurn` (failed turn). Keep completed items; flush still-open items as `failed`. `RetriableError` → `errorInfo.category: connection-failed`; `ActionRequiredError` / `NonRetriableError` → `unknown`. Split chunks are joined; a quoted error inside a longer reply does not match; other ACP dialects are unchanged.
- `provider-retry` retries `connection-failed` with the existing attempt cap (5 total) and exponential delay plus jitter, via `bb.sdk.threads.retry` (original turn, preserved session). Unauthorized, billing, and unknown are not retried. A queued retry is deleted when the thread becomes active, returns to idle (user stop), or a non-retry message is queued.
- Plugin default enabled/disabled is unchanged. Cursor's `enableAgentRetries` is not set.

The running desktop build that showed the false-success events is **0.42.1**. This change is unverified against a live Cursor drop until a build that includes it is released.

## How you verified

- Tests added in `packages/provider-bridge-acp/src/delta-translation.test.ts` and `dialect.test.ts`: single-chunk and split-chunk RetriableError → failed `provider/error`; quoted text does not trigger; generic dialect does not classify; pending fileChange/toolCall close as failed; ActionRequiredError uses `unknown`.
- Tests added in `plugins/provider-retry/server.test.ts`: connection-failed schedules a retry; unauthorized/billing/unknown do not; attempt cap applies; user active/idle/follow-up cancels a queued retry; the plugin's own retry queue event does not.
- `pnpm exec turbo run typecheck lint test --filter=@bb/provider-bridge-acp --filter=bb-plugin-provider-retry`: 320 ACP tests and 22 provider-retry tests passed. `oxfmt` on the touched files. `git diff --check` passed.
- Existing cursor-agent debug logs (`$TMPDIR/cursor-agent-logs-$UID/`, written unless `CURSOR_AGENT_DISABLE_DEBUG_LOG` is set) already contain `[AGENT_ERROR_DIAGNOSTICS]`. Across 40 diagnostic lines on this Mac, every `RetriableError` had `willRetry: false`, `enableAgentRetries: false`, `transportErrorRetries: 0`, `attempt: 0`. The hidden `--debug` flag starts a local log server and was not used. `enableAgentRetries` was not set.
- Live ACP recovery after this lands is unverified until a desktop that includes the change is running. `timigod` is not on `.github/APPROVED_CONTRIBUTORS`; the PR gate may close this until a maintainer runs `/approve @timigod`.

Fixes #

> AGENT GENERATED

Made with [Cursor](https://cursor.com)


timigod:
The branch is on `timigod/bb` (`fix/acp-cursor-terminal-error`). `timigod` is not on `.github/APPROVED_CONTRIBUTORS`, so the PR gate closed this. After `/approve @timigod`, it can be reopened. Do not merge until a maintainer has reviewed.

PR #3681, opened 2026-09-14
Move plugin compilation off the server event loop

ymichael:
## What was wrong

The plugin runtime loaded every server entry, including compatible prebuilt marketplace and bundled artifacts, through JITI. JITI performs synchronous transformation and module loading, so loading a large bundle such as Modal Sandbox could block the server event loop long enough for BB Connect heartbeats to time out and report that the tunnel client was offline.

## What changed

- Compatible prebuilt `dist/server.js` artifacts now use native ESM import and avoid runtime transformation.
- Source and path plugins compile asynchronously through esbuild into an atomic, content-addressed server cache, then the server natively imports the resulting artifact. Cache identity includes plugin source, BB and SDK versions, Node version, toolchain version, and build format.
- Runtime compilation receives the manifest metadata already validated by the server, preserving runtime asset policy without weakening stricter release-build validation.
- BB-owned runtime imports are resolved through native loader hooks while plugin-owned SDK subpaths compile from the plugin's installed SDK. Plugin factory execution remains in-process and retains its existing timeout.
- JITI remains a shipped dependency and can be restored with the persisted `legacyJitiPluginLoader` BB experiment. The selection is read on each plugin load, so changing it affects the next install, update, enable, explicit reload, or server restart; it does not replace an already-running instance.
- Added Settings UI, CLI coverage, configuration docs, and BB guide content for the experiment.
- There are no server/host-daemon wire changes, so `HOST_DAEMON_PROTOCOL_VERSION` is unchanged.

## How you verified

- `pnpm exec turbo run build --filter=bb-app` — 49 tasks passed, including all bundled plugin builds and the packaged BB app.
- `pnpm exec turbo run typecheck --filter=@bb/domain --filter=@bb/db --filter=@bb/app --filter=@bb/cli --filter=@bb/server --filter=@bb/plugin-build --filter=@bb/templates --filter=bb-app` — 12 tasks passed.
- Native import, source compilation cache, cross-plugin dynamic import, prebuilt selection, JITI experiment, system experiment, and runtime logo compatibility tests passed.
- Built-in plugin reconciliation and hot reload tests — 33 tests passed.
- Settings UI — 6 tests passed; settings CLI — 11 tests passed; database experiment defaults — 1 test passed; server build behavior — 6 tests passed.
- Lint completed with zero errors across all affected packages.
- Smoke-tested bundled plugins including Modal Sandbox, Automations, ACP, and Documents, plus marketplace Git and npm install paths.

> AGENT GENERATED


ymichael:
@bb-code-review

ymichael:
Addressed all five review findings in acd2986877:

1. Release builds now place ESM in an explicit module scope, and existing compatible ESM prebuilds inside CommonJS packages are recompiled to a cached CJS runtime artifact.
2. Computed relative dynamic imports are syntax-aware rewritten against their original source module URL.
3. import.meta.url, import.meta.dirname, and import.meta.filename are transformed through esbuild parsing; strings, templates, regexes, and repeated uses remain intact.
4. Runtime cache identity is derived from the compiled artifact, covering the actual bundled graph; static imports escaping the plugin tree are rejected explicitly.
5. Source factories load through reclaimable CJS, are removed from both require.cache and parent.children, mutable resolver roots are owned per loaded plugin, and disk artifacts are capped at four generations.

Verification:
- @bb/server plugin suite: 45 files / 501 tests passed
- @bb/plugin-build: 11 files / 144 passed / 1 skipped
- @bb/server and @bb/plugin-build typechecks passed
- @bb/server, @bb/plugin-build, and @bb/cli production builds passed
- Live worktree server: 30 plugins, zero error statuses; Modal reload succeeds in expected unconfigured state
- Live source fixture verified computed relative import refresh and source-relative import.meta values across reload

@bb-code-review please re-review the latest commit.

PR #3707, opened 2026-09-15
Prevent concurrent launcher config updates from losing changes

### LINKED ISSUE #2322: env.json writers can silently overwrite concurrent updates
## Attempted workflow

A BB CLI process and an authorized plugin both update different keys in the managed `env.json` file.

## Actual result

Each writer can read a whole document, project its change, and atomically rename that projection later. The rename is atomic but not conditional at the commit boundary, so the last writer can silently erase an unrelated concurrent update.

Evidence:

- `packages/bb-app/src/launcher.ts:1183-1211` performs whole-document merge and temp-file rename with no expected revision or shared writer lock.
- `packages/bb-app/src/launcher.ts:1905-1938` routes both `env set` and `env unset` through that read-then-write shape.

## Expected result

BB owns one transactional `env.json` mutation surface used by CLI, server, and authorized plugins. A stale writer receives a conflict and performs no rename or partial mutation.

## Acceptance criteria

- Set and unset share the same transaction path.
- Successful mutation preserves unrelated top-level fields, unrelated environment keys, and private file mode.
- Tests coordinate two real concurrent writers and prove only one stale whole-file projection can commit, the loser receives a conflict, and unrelated keys are never silently lost.
- Plugins can configure non-secret launcher paths without editing `env.json` through a private file protocol.

ymichael:
## What was wrong

Concurrent `bb-app` processes can read the same managed JSON snapshot, change different keys, and atomically replace the file. The later replacement silently loses the other change; an unrelated set can also restore a key removed by unset. Atomic rename prevents partial JSON but does not coordinate writers. Machine enrollment wrote the same `config.json` independently and could participate in the same race.

## What changed

Config/env set and unset, enrollment config patches, and client SSH set/remove now hold a shared cross-process file lock from the mutation read through atomic replacement. Each operation reads current state after acquiring the lock, preserves unrelated fields, writes a private temporary file, and atomically replaces the JSON.

The reusable `@bb/config/file-lock` helper uses `fs-native-extensions` for an advisory kernel lock on a persistent sibling file. Contenders poll asynchronously for up to five seconds. File descriptors release ownership normally and when a process dies; a paused live writer retains ownership, so there is no timestamp, PID, heartbeat, or stale-lock reclamation protocol. The installed dependency footprint is about 1.55 MiB, compared with about 27 MiB for the SQLite-based implementation considered earlier.

The slim enrolled-host artifact includes `fs-native-extensions` and no longer needs `better-sqlite3` for this feature. The normal full `bb-app` package keeps its existing SQLite dependency because its bundled server uses SQLite. `HOST_DAEMON_PROTOCOL_VERSION` is unchanged because no server/daemon wire contract changed. This PR adds no user-facing command, flag, setting, or documentation change.

## How you verified

- Rebased onto current main `79568e97b1`.
- The deterministic real-process suite passes 31/31 cases. It covers env/config set/set and unset/set orderings, client set/remove, machine enrollment versus config, concurrent creation, unrelated fields, `0600` permissions, parsing and I/O failures, SIGKILL recovery at three transaction stages with multiple waiters, same-process contention, SIGSTOP beyond the timeout with an old lock mtime, and abandoned temporary-file cleanup.
- CLI machine-enrollment tests pass 7/7 and server artifact tests pass 15/15.
- Turbo typechecks pass for `@bb/config`, `bb-app`, `@bb/cli`, and `@bb/server`.
- The Turbo enrolled-host build passes. Its tarball was installed outside the checkout, where real `bb-app config set` mutations succeeded under Node 22, 24, and 26. The installed manifest contains `fs-native-extensions@1.5.0` and omits `better-sqlite3`.
- Changed source, fixture, and package files pass formatting. `git diff --check` passes.

All test data is synthetic and lives in temporary directories. Linux package verification is delegated to CI. Coordination requires updated cooperating commands on a local filesystem; older binaries and manual editors do not take the lock. Browser and dev-server verification are N/A for this CLI-only behavior.

Fixes #2322

> AGENT GENERATED


ymichael:
@bb-code-review

ymichael:
@bb-code-review

PR #3738, opened 2026-09-15
Add explicit working directories for automation scripts

### LINKED ISSUE #3616: Automation scripts run in the plugin's scripts directory, so a project-relative script path fails with only an exit code
### Summary

A project-scoped automation whose script is `python3 bin/build-views.py` — a path relative to the
project checkout — fails with a bare `Script exited with code 2`. The script runs with its working
directory set to the automations plugin's own data directory, not the project, so the interpreter
never finds the file. Nothing in `bb automation --help` says where a script runs, and the run list
shows only the exit code, so the first symptom looks like a permissions or interpreter problem. I
expected a project-scoped automation to run in that project.

### Versions and environment

- bb 0.42.1, desktop app (`/Applications/bb.app`), macOS 26.6.2, Node v26.7.0.
- First reproduced against `main` at `4ae0a7c893e9922721e0e367c72aa3a1a2990e7c`; the same line is
  unchanged on `main` at `9e16411141788c67954bceb753bb64ac1dc7057f`.
- No provider involved — this is the script execution path, not an agent run.
- Nothing unusual: local project checkout, default data dir, no bb Connect involvement.

### Steps to reproduce

1. In any project with a local checkout, create a script in the checkout:
   ```sh
   mkdir -p bin && printf 'print("ok")\n' > bin/build-views.py
   ```
2. Create a project-scoped script automation whose command is `python3 bin/build-views.py`.
3. Run it.

Smallest form that still reproduces: any script automation with **any** relative path argument.
Does **not** reproduce with an absolute path — which is the workaround, and is why this reads as
"paths are just broken" rather than "the cwd is somewhere else".

### Expected vs actual

Actual — the run's summary:

```
Script exited with code 2
```

and, only in the full run output:

```
python3: can't open file '/Users/bart/.bb/plugins/automations/scripts/bin/build-views.py'
```

Expected: the script runs in the project's checkout and finds `bin/build-views.py`. Failing that,
the summary should at least say why, instead of an exit code with the cause one level down.

### Evidence

The working directory is set unconditionally to the plugin's own scripts root:

https://github.com/get-bb/bb/blob/9e16411141788c67954bceb753bb64ac1dc7057f/plugins/automations/src/script-runner.ts#L296

```ts
const cwd = scriptsRoot(args.pluginDataDir);
```

It is not load-bearing for resolving the script itself: `resolveAutomationScriptPath` returns an
absolute realpath, and stored scripts live one level deeper
(`<dataDir>/scripts/<automationId>/`), so `<dataDir>/scripts` as cwd plays no role. It appears to
be an arbitrary "somewhere writable".

Second, smaller problem in the same path: stdout and stderr are concatenated into one field on the
run result, so the cause cannot be recovered for the summary even though it was captured.

### What you ruled out

- Not an interpreter or permissions problem: the same script runs with an absolute path.
- Not specific to Python: any relative path in any interpreter behaves the same way.
- Not a duplicate — searched open and closed issues for "automation" in the title and the
  `automations` label; #2692, #2922, #2323, #2299 and #1364 are unrelated. Nothing on `main` up to
  `9e16411` changes the script working directory.
- The auto-pause after three failures is correct behaviour and not part of this report; it just
  makes the missing cause more expensive, because the automation is disabled before you learn why.

Hits anyone writing a project-scoped script automation against files in their repo, which is the
obvious use for the feature. There is a workaround (absolute paths), but it is undiscoverable from
the error. Effort is small; a prototype is linked below.

### Prototype

https://github.com/BartSoj/bb/tree/fix/automations-script-working-directory — one commit on
`9e16411`.

Project-scoped automations run in the project's default local source when it is a directory on the
server host, falling back to the current behaviour otherwise — a project source can name another
host, and `Host` in `@bb/domain` exposes no "this machine" marker, so a remote checkout degrades
to today's behaviour rather than to a confusing `spawn ENOENT`. `stderr` is kept separate on the
run result and its first non-empty line is appended to the exit-code summary. Documented in
`bb automation --help`, the automations skill, its script-runtime reference, the plugin overview
and the in-CLI guide template.

`pnpm exec turbo run test --filter=bb-plugin-automations`: 8 files, 84 tests pass (80 before); the
new cwd test was mutation-checked. `pnpm exec turbo run typecheck --filter=bb-plugin-automations`
passes.

I have not opened a PR: I am not on `.github/APPROVED_CONTRIBUTORS` yet.

> AGENT GENERATED

ymichael:
## What was wrong

Automation scripts ran from the plugin’s shared script storage directory, so project-relative paths failed even when the project had a checkout on the server. Failed runs showed only an exit code despite useful stderr in their output. This addresses [#3616](https://github.com/get-bb/bb/issues/3616) and its [investigation report](https://get-bb.github.io/reports/issues/3616.html).

## What changed

Script executions persist a working-directory policy: `automation-storage` uses shared plugin script storage, `project` uses the project source on the BB server host, and `path` uses an absolute server-host directory. Existing scripts retain storage behavior. New standard-project scripts use the server-local project source when available; Personal and projects without one default to storage. Replacing a script preserves its policy when omitted.

The CLI supports `--working-directory automation-storage|project|<absolute-path>`, including directory-only updates. Details and `bb automation show` display the resolved directory. Unavailable directories fail explicitly, and failed-process summaries include the first sanitized stderr line while retaining raw output. CLI guidance and automation skills document these behaviors.

Primary-host resolution now uses only the persisted local host identity and a valid, non-destroyed database record. It no longer falls back to a sole connected or enrolled remote host. Existing API and helper names are unchanged. Missing or invalid local identity yields `null` in system config and an unavailable error from required-host helpers.

Automations uses the existing `bb.sdk.system.config().primaryHostId`; this PR adds no plugin SDK member or SDK version bump. The integration harness explicitly records its server-local daemon identity instead of relying on the removed fallback. No server/daemon wire fields changed, so `HOST_DAEMON_PROTOCOL_VERSION` is unchanged.

### Follow-up: one resolved working directory across read surfaces

The resolved directory was an optional tri-state field on one response type: a path, `null` when unresolvable, or `undefined` when a caller used the stored mapper. `create`, `update`, `list` and `overview` left it `undefined`, so `create` and `update` printed the stored policy name (`project`) while `show` printed the resolved path. The CLI and the detail panel each carried a three-branch ladder, and the two disagreed on the states neither of them reached.

The response is now split: a summary type for `list`, `overview`, `pause` and `resume`, and a detail type for `get`, `create` and `update` whose script executions carry a required `resolvedWorkingDirectory: string | null`. `create` and `update` resolve it from the project and server host id they already hold, so the three detail surfaces agree and both ladders collapse to one check.

The two copies of the policy resolution (`run.ts` and `service.ts`) are consolidated into `working-directory.ts`. The shared resolver returns the automation storage root rather than `null` for that policy, so `executeStoredScript` takes a concrete directory and only the executor decides that an unresolvable project is a run failure.

### Follow-up: review findings

- `firstStderrLine` split the whole stderr buffer before scanning for the first non-empty line. Stderr is capped at 1 MiB, so a failing script emitting mostly line breaks built up to a million strings synchronously on the server event loop for a detail that almost always comes from the first line. It now walks the buffer with `indexOf` and skips empty lines without slicing them: 1 MiB of line breaks goes from 57ms to 9ms.
- The working-directory path accepted any string within its length bounds, so terminal escape sequences reached a persisted run error and the CLI printed them verbatim. Control characters are now rejected in the path, in the schema for the RPC boundary and in the service for callers that build the input directly. Both are needed: `cli.ts` calls the service directly and never parses through Zod, so the schema refinement alone would have left the CLI path open.

Project *source* paths have the same gap and are not fixed here; they are validated only for shape, so a project path can carry escape sequences into `bb project list`. That is a separate input boundary in `@bb/domain`, filed as #3870 with a reproduction.

## How you verified

- Automations: 96 tests passed and Turbo typecheck passed after switching to system config. Server, plugin SDK, and integration typechecks passed. The SDK version guard reports “Plugin SDK surface unchanged.”
- New regressions cover missing local identity with a connected or disconnected remote host, missing/destroyed local records, disconnected local hosts, required-host errors, and system-config identity/platform behavior.
- Full server run: 2,873 tests passed; 18 failures identified fixtures relying on the removed fallback, and one provisioning test timed out. Updated the affected fixtures to explicitly identify their server-local daemon.
- Reran all seven affected suites: 100 tests passed, including the provisioning test without changes.
- Reproduced the CI custom-model integration failure locally, then passed all 82 integration tests after the harness fix with `pnpm exec turbo run test --filter=@bb/integration-tests -- --maxWorkers=2`. The unrestricted local run hit eight 60-second timeouts; limiting workers resolved them. Formatting and `git diff --check` passed.
- Earlier branch verification included the isolated app: a project script’s `pwd` output matched the directory displayed in the browser and CLI.

### Follow-up verification

- Rebased onto `main`; the only conflicts were the plugin SDK version, resolved to `main`’s value. The guard still reports “Plugin SDK surface unchanged”, so no bump is required.
- 99 automations tests and 4,779 app tests pass, with typecheck and lint clean. The server suite passes 2,919 tests and the integration suite passes all 82 with `--maxWorkers=2`.
- The new assertions were mutation-checked: reverting `update` to the stored mapper fails three tests, disabling the control-character predicate fails one, and breaking the stderr walk fails five.
- End-to-end against a source dev server: `create`, `update` and `show` print the same resolved directory, `list --json` carries no `resolvedWorkingDirectory` while `show --json` does, and all runtime policies behave unchanged — project, storage and explicit-path working directories, the pre-spawn failure for a missing directory, the escape-stripped stderr summary, and the legacy row with no stored policy.
- The unresolvable case was verified end to end on a project with no server-host source: the panel reads “Working directory unavailable” and the run fails with `Project <id> has no source on the bb server host`.
- Control characters were verified rejected through both paths: over RPC (`400`, `"A script working directory path must not contain control characters"`) and through the CLI invoked directly. Note that `pnpm run` mangles control characters in argv, so a CLI repro must not go through a pnpm script.

Fixes #3616

> AGENT GENERATED


ymichael:
@bb-code-review

ymichael:
Addressed all review findings in 9bf1bbae8c:

- resolve `project` only against the actual co-located server host, with operation-time reads for fresh daemon startup
- default Personal and remote-only projects to `legacy`
- reject valueless `--working-directory` flags
- cache project resolution per scheduler sweep
- sanitize persisted stderr summaries while preserving raw output
- display the selected policy in automation details
- add the required experimental SDK surface/version bump and coverage

Focused and adjacent suites, Turbo typecheck/build, lint/format, and real source-app CLI/browser verification pass. The full server suite passed 2,709 tests; its two filesystem-watcher tests also fail under a direct recursive `fs.watch` probe on this host (zero events), independent of this patch.

@bb-code-review please re-review.

PR #3741, opened 2026-09-15
Prevent incidental wakes and reconcile tracked Modal allocations

ymichael:
## What was wrong

Metadata and cleanup commands could wake paused machines, and new-machine checkout setup dropped Git clone progress before it reached the initiating thread's timeline.

Separately, a failed resume can leave Modal compute running while BB records the machine as suspended. The Modal plugin did not keep its own allocation list to detect and clean up that mismatch.

## What changed

- Added an exhaustive wake policy shared by work RPC dispatch and transport retry. Rename, archive, unarchive, stop, cleanup, thread-storage access and goal clearing no longer implicitly wake machines.
- Forward clone output into the initiating thread's provisioning timeline and release its reporter on completion or failure. Coalesce updates to once per second, skip duplicates, and flush the last update at completion or failure. Other threads waiting for the checkout keep their existing behavior.
- The Modal plugin persists allocation names before creation and sandbox IDs afterward, including allocations made during resume. Its existing once-per-minute schedule inspects only tracked allocations. Confirmed stopped entries are pruned; errors and account changes retain them. Pending allocation responses can be recovered by name. Existing machine IDs are imported at plugin startup; debug sandboxes retain their separate ownership policy.
- Added `hosts.experimental_reconcile({hostId})`, `POST /hosts/:id/reconcile`, and `bb machine reconcile`. Core's current state is authoritative: a suspended machine runs the existing coordinated save-and-stop operation; active machines and transitions already in progress are left alone. The API returns HTTP 202 after starting the operation; the CLI polls machine status until completion. The plugin calls this when it finds tracked running compute for a suspended machine. Idle policy separately requests `experimental_suspend`.
- Core adds no periodic provider inspection. Failed-resume state transitions are unchanged. Reconciliation runs only when requested. Allocation checks continue with idle pausing disabled and isolate per-allocation failures.
- A successful Modal stop is not failed by subsequent tracking lookup/deletion errors; retained entries are retried by the next plugin sweep. Provider suspend/resume are documented as idempotent, with regression coverage for preserving saved snapshots and reusing running compute. No provider opt-in flag.
- Updated the Plugin Guide, API audit, CLI guide and plugin skill. No daemon wire change or protocol bump.

Tracking is a list of allocations to check, not a history. Unresolved pending creates remain until their requested lifetime expires and Modal confirms absence. Unmatched running allocations are reported and retained until confirmed gone; the plugin does not terminate another installation's untracked compute.

## How you verified

- Latest server regression suite: 16 tests passed, including accepting reconciliation before provider completion, repeated requests, concurrent resume, clone burst coalescing, initiating-thread-only progress, and existing machine lifecycle behavior.
- Modal plugin: 61 tests passed, including tracking across create/pause/resume/remove, failed lookups, lost create responses, restart recovery, account separation, pruning, per-allocation error isolation, and reconciliation with idle pausing disabled.
- CLI guide coverage: 2 tests passed. SDK: 113 tests passed. Plugin API map: 73 tests passed. Server, CLI, Modal plugin and API-map typechecks passed. `git diff --check` passed.
- Previous clone regression run: all 12 targeted wake/clone/machine-thread tests passed, including timeline output before clone completion and no forwarding to other waiting threads.
- **Real Modal E2E on the rebuilt test server:** created a tracked sandbox, saved a marker, deliberately invalidated its daemon port, paused and resumed. Bootstrap failed after allocation; BB reported suspended while Modal reported running. Confirmed the new allocation was persisted, repaired the test-only fault, and restarted BB before cleanup. The plugin's schedule then called reconciliation, saved a fresh snapshot and stopped compute. A subsequent resume preserved the marker. Calling `bb machine reconcile` on the active machine left it active. Removed the test machine and verified no running test compute remained; the plugin pruned confirmed stopped allocation records.

- Review-fix live verification on a newly rebuilt server with the matching CLI and a real Modal sandbox: create, reconcile while active (unchanged), pause, reconcile while already paused, resume to connected/active, and remove. The earlier failed-resume/restart recovery E2E above was performed before this review-fix commit.

> AGENT GENERATED


ymichael:
@bb-code-review

ymichael:
@bb-code-review

PR #3744, opened 2026-09-15
Avoid replaying migration history on fresh startup

ymichael:
## What was wrong

Every new `bb-app` data directory synchronously replayed the complete 121-migration history before the server could bind its HTTP listener. Retained process samples and server logs from the [recurring macOS package-smoke failure](https://github.com/get-bb/bb/actions/runs/35007793506) show that the server child remained alive and CPU-active while it executed historical create/rebuild/drop work; there was no listener, `EADDRINUSE`, stderr, or process exit. Bounded Intel contention stretched those normally cheap serial operations past the unchanged 60-second health deadline, producing the launcher-only `Starting server` signature even after the port-collision fix in #3459.

## What changed

- Generate a deterministic current-schema SQLite template from the real migration implementation as a cacheable Turbo output.
- Atomically stage and install that template only when the database path is absent, then skip migration replay for that proven fresh-file case.
- Keep existing files and in-memory databases on the full legacy repair, migration, warning, and validation path.
- Remove the wall-clock personal-project row before sealing the template; normal server initialization recreates it at startup. Vacuum and WAL checkpoint cleanup make repeated generation byte-identical and prevent sidecars from entering packages.
- Add coverage for template integrity/current migration history, existing-file preservation, in-memory behavior, and temporary-file cleanup.

`HOST_DAEMON_PROTOCOL_VERSION` remains 209 because this changes only server-local database initialization and adds no server/daemon wire field, type, default, or meaning.

## How you verified

- Red, frozen `423955ea44fdf8f364d77529237951f8d43f4ab1`, Node 22.22.3, bounded Intel load: 32-worker fresh health was 14.235s / 21.797s / 18.692s; at 64 workers the server did not announce its listener until 60.881s, after the health deadline. Migration statements in the retained server log accounted for the pre-listener work.
- Green packaged artifact, same host/runtime/load: 32-worker fresh health was 11.373s / 11.615s / 17.415s; at 64 workers `/health` was observable at 55.308s. No timeout or polling value changed.
- Two forced `@bb/db#generate:fresh-database` runs produced the same SHA-256, `3aeb5a3cdb0d8ef858a65ed0dc7496de69263e5c4109a053c51bea3b25a1b51b`, with no WAL, SHM, or temporary output.
- `pnpm exec turbo run test --filter=@bb/db`: 41 files, 524 tests passed, including the full legacy migration suite.
- `pnpm exec turbo run test --filter=@bb/server -- test/app/skeleton.test.ts test/legacy-automations-export.test.ts`: 2 files, 16 tests passed.
- `pnpm exec turbo run typecheck --filter=@bb/db --filter=@bb/server --filter=bb-app`: 9 tasks passed.
- Node 22.22.3 `pnpm exec turbo run build --filter=bb-app`: 51 tasks passed.
- Node 22.22.3 `pnpm exec turbo run smoke:tarball --filter=bb-app`: passed in 110.2s; full stack 25.0s and daemon join 5.5s.

The full server suite also completed 2,689 tests successfully but had 19 unrelated deadline failures under its own Intel shard oversubscription, one existing timeline micro-benchmark ceiling failure, and the stale ask-user schema expectation fixed on later main by #3740. The affected startup and legacy paths pass when isolated as listed above.


ymichael:
@bb-code-review

ymichael:
Addressed the review findings in 284cdf911a:

- The staged template is chmodded to `0600` before the hard link publishes it as `bb.db`.
- The regression test starts from the generated `0644` template under `umask 077` and verifies the installed database is `0600`.
- Template schema and migration-ledger equality are asserted against a freshly migrated database before the idempotent `migrate()` call.
- Tests that mutate `process.umask()` are now assigned an isolated Vitest worker.

Validation:
- `pnpm exec turbo run test --filter=@bb/db` — 41 files, 525 tests passed.
- `pnpm exec turbo run typecheck --filter=@bb/db` — passed.
- Formatting and `git diff --check` — passed.

@bb-code-review

PR #3749, opened 2026-09-15
Fix watcher recovery and shutdown lifecycle

ymichael:
## What was wrong

The watcher proxy cleared recovery-rescan intent as soon as a replacement child became ready. If that child synchronously lost its IPC channel during subscription replay, child-exit handling restored the intent and the still-running ready handler immediately overwrote it. The next healthy child therefore subscribed without a rescan and missed files created during the outage. Separately, `RootSubscription.dispose()` could wait forever for a native subscription acknowledgment because its pending startup record had no cancellation path.

## What changed

- Keep recovery intent sticky across failed watcher children and clear it only after a healthy child acknowledges a rescan subscription.
- Track each request's child channel so stale acknowledgments and unsubscribes cannot affect a replacement child.
- Make subscription startup abortable and cancel pending proxy records during root disposal.
- Settle pending unsubscribe work when a child exits and recycle a child when native unsubscribe exceeds a bounded timeout.
- Cover consecutive replay failures, pending startup disposal, late in-process startup cleanup, and stuck native unsubscribe.
- No daemon protocol bump: this changes only internal watcher lifecycle behavior and daemon-to-child IPC handling.

## How you verified

- Deterministic replay repro: replay flags changed from `false, true, false` with no event to `false, true, true` with the outage file recovered.
- Deterministic disposal repro: root disposal now settles before backend destruction.
- `pnpm exec turbo run test --filter=@bb/host-watcher -- test/parcel-watcher-proxy.test.ts` (22 passed)
- Root watcher lifecycle tests (27 passed)
- `pnpm exec turbo run test --filter=@bb/host-daemon -- src/watch-manager.test.ts src/app.test.ts` (32 passed)
- `pnpm exec turbo run test --filter=@bb/server -- test/services/plugins/builtin-plugins.test.ts` (34 passed)
- `pnpm exec turbo run typecheck --filter=@bb/host-watcher --filter=@bb/host-daemon --filter=@bb/server`
- Real source app: a file created while the owned watcher child was stopped appeared after automatic recovery without a browser reload; supported daemon shutdown completed in 5.276 seconds.
- Full host-watcher run: 57 passed and one existing host-native FSEvents test failed; a direct raw Parcel diagnostic likewise received no events on this host.

Follow-up to #3730.

> AGENT GENERATED


ymichael:
@bb-code-review

PR #3805, opened 2026-09-16
Derive the compact panel shelf attribute in one place

brsbl:
## What was wrong

The page inset and the sidebar trigger overlay both read the compact secondary-panel presentation store and emit `data-panel-shelf`, but only the inset dropped the attribute while the mobile sidebar drawer is open. The trigger floats above that drawer and is its close affordance, so any state that left the store on `shelf` with the drawer open would hide the control needed to dismiss it.

## What changed

- The sidebar trigger keeps its attribute cleared while the compact sidebar drawer is open, matching the page inset.
- Both consumers derive the attribute from one shared `usePanelShelfState` hook that owns the store subscription and both guards.
- No change to what a user can reach today; the guarded state is currently unreachable through real interaction.

## How you verified

- Remote CI passed on this head. [Run](https://github.com/get-bb/bb/pull/3805/checks)
- Chrome for Testing 151, mobile 390×844 at 2×, on the top of this stack: the shelf hid the trigger, the full presentation kept it, and opening the drawer cleared the attribute on both the trigger and the inset.
- The same pass covered close, reopen, and hard reload with no console errors.
- No before/after screenshots: in every state reachable by interaction this layer renders identically to its base, so the screenshot gate does not apply.

BB-Thread-ID: thr_rhbtb8rhgc

> AGENT GENERATED


brsbl:
Folded into #3807, which now carries all three follow-up commits in one pull request against the parent fix. No content was dropped; its head is unchanged at dba267e12.

PR #3806, opened 2026-09-16
Give the compact shelf hide rule one home

brsbl:
## What was wrong

`COMPACT_SHELF_HIDDEN_PAGE_HEADER_ACTIONS_CLASS` carried a `group-data-[panel-shelf=open]` arm that no writer can produce: the presentation store emits only `closed`, `shelf`, or `full`, and `open` belongs to the separate `data-sidebar-shelf` attribute. Meanwhile the parent layer spelled the same hide rule inline in the trigger overlay's two branches, so the policy existed in three forms and one of them matched nothing.

## What changed

- Rendering is unchanged; the removed selector could never match.
- The dead `panel-shelf=open` arm and its assertion are gone, so a maintainer copying the constant no longer inherits a selector that silently guards nothing.
- Both hide rules now sit beside the presentation union that decides them, leaving one definition per rule.
- The page header re-exports its constant, so existing importers and their test mocks are unaffected.

## How you verified

- Remote CI passed on this head. [Run](https://github.com/get-bb/bb/pull/3806/checks)
- An earlier revision of this layer defined the new constant in the page header module, which two layout tests mock, so the trigger silently lost its hide class and the app shards failed. Moving both constants to the unmocked shelf-state module fixed that.
- The existing chrome-selection test still asserts the live page-header arm is applied.
- Chrome for Testing 151, mobile 390×844 at 2×, on the top of this stack: the trigger hid in the shelf presentation and stayed visible in full and closed, with no console errors.
- No before/after screenshots: this layer changes no rendered output, so the screenshot gate does not apply.

BB-Thread-ID: thr_rhbtb8rhgc

> AGENT GENERATED


brsbl:
Folded into #3807, which now carries all three follow-up commits in one pull request against the parent fix. No content was dropped; its head is unchanged at dba267e12.

PR #3807, opened 2026-09-16
Consolidate the compact panel shelf hide rule

brsbl:
## What was wrong

Three follow-ups from the review of the parent PR, all in the same mechanism. The sidebar trigger overlay read the compact presentation store without the guard for an open sidebar drawer that the page inset applies, so a state leaving the store on `shelf` with the drawer open would hide the drawer's own close affordance. The page header constant carried a `group-data-[panel-shelf=open]` arm that no writer can produce, since the store emits only `closed`, `shelf`, or `full` and `open` belongs to the separate `data-sidebar-shelf` attribute. The new compact test matched a literal Tailwind class and read the store back out through `dataset`.

## What changed

- The sidebar trigger clears its attribute while the compact sidebar drawer is open, matching the page inset.
- Both consumers derive that attribute from one shared `usePanelShelfState` hook holding the store subscription and both guards.
- Both hide-class constants move beside the presentation union that decides them, so each rule has one definition and the dead `panel-shelf=open` arm is gone.
- The compact test states the shelf-only hide rule once through the shared constant instead of asserting a store round-trip.
- Nothing a user can reach today renders differently; the drawer guard covers a state unreachable through real interaction.

## How you verified

- Remote CI passed on this head. [Run](https://github.com/get-bb/bb/pull/3807/checks)
- Chrome for Testing 151, mobile 390×844 at 2×, on this exact head: the shelf hid the trigger, the full presentation kept it, opening the drawer cleared the attribute on both the trigger and the page inset, and close, reopen, and hard reload behaved with no console errors.
- An earlier revision defined the new constant in the page header module, which two layout tests mock, so the trigger silently lost its hide class and the app shards failed. Moving both constants to the unmocked shelf-state module fixed that.
- No before/after screenshots: every state reachable by interaction renders identically to the base, so the screenshot gate does not apply.

BB-Thread-ID: thr_rhbtb8rhgc

> AGENT GENERATED


brsbl:
Folded into #3798, which now carries all four commits as a single pull request. GitHub marked this one merged automatically because its base branch `fix/mobile-right-panel-sidebar-trigger` fast-forwarded to this head (dba267e12); nothing was merged into `main`. Review the combined change in #3798.

PR #3840, opened 2026-09-17
Project CRUD picker and Settings Projects page

elianiva:
Create flow now stages the picked folder, then asks for the project name in a details dialog (path and machine shown read-only).

Sidebar has a clickable Projects entry opening the project list, with a New project row wired to the create dialog.

Settings has a Projects page with rename, delete, and create actions.

Open decision: per-project logo/icon is intentionally out of scope because the server projects table and create/update API have no icon field; that needs a DB migration plus contract change as a follow-up.


elianiva:
fuck my clanker is acting up im so sorry, shouldn't have let it loose :)

PR #3901, opened 2026-09-18
Fix Codex Plan mode and structured questions in BB

maheen-ejaz:
## What was wrong

BB detected the `/plan` composer action but did not pass that mode to Codex app-server. Codex therefore ran the turn in Default mode, where native `request_user_input` is disabled. If Codex did emit a native user-input request, the provider bridge rejected it as unsupported. BB also treated any `/plan` mention in prose as Plan mode, even when the command did not lead the prompt.

## What changed

This PR establishes one explicit contract for Codex turns:

- A leading `/plan` command starts the turn in Codex's native Plan mode.
- Every other turn explicitly uses Default mode, so Plan mode cannot remain sticky.
- A native Codex `request_user_input` request opens BB's structured question interaction; BB returns the selected option and any free-text note to the same Codex turn.
- A `/plan` reference inside ordinary prose does not change modes.

Structured questions remain available across providers and modes through the existing capability-based paths:

| Runtime | Question path |
|---|---|
| Codex Plan mode | Native `request_user_input`, rendered by BB |
| Codex Default mode | BB's `AskUserQuestion` fallback |
| Provider with native question support | Provider-native question tool |
| Provider without native question support | BB's `AskUserQuestion` fallback |

Codex remains declared as not having universally available native questions, so the Default-mode fallback stays enabled. Secret-input requests are rejected because BB's question interaction is not a secret-entry surface.

No host-daemon protocol field changed; `promptMode` already crosses that boundary in the existing execution options.

Observable behavior after this change:

1. Selecting Plan mode produces a native Codex Plan turn.
2. The next ordinary turn runs in Default mode.
3. Questions pause the turn, render in BB, and resume after an answer.
4. Text such as `Can we discuss /plan first?` remains an ordinary prompt.

## How you verified

- All 324 Codex-provider tests pass, including new native Plan transport and user-question round-trip coverage.
- Focused affected tests pass: 12 domain, 76 thread-view, 43 server, and 19 fallback-question tests.
- Affected package typechecks and formatting pass.
- `bb plugin build plugins/provider-codex` succeeds.
- Upstream build, typecheck, lint, server, app, and integration checks pass. Two unrelated CI jobs require a maintainer rerun; details and isolated local reproductions are recorded in PR comments.

> AGENT GENERATED


maheen-ejaz:
The packages shard failed in the unrelated desktop watcher test `apps/desktop/test/server-moved.test.ts:572` (expected callback was not observed). The exact file passes locally: 52/52 tests in 1.03s. All affected Codex, domain, thread-view, server Plan-runtime, and question-interaction tests pass. I do not have upstream admin rights to rerun the failed job; please rerun that shard.\n\n> AGENT GENERATED

maheen-ejaz:
A second unrelated CI failure occurred in Linux package smoke: apt could not download `libfuse2t64` from `mirrors.edge.kernel.org` because the connection timed out. The macOS smoke and all code checks are unaffected. Both failed jobs require an upstream rerun; this fork contributor cannot trigger it.\n\n> AGENT GENERATED

maheen-ejaz:
Closing this proposal because BB's existing AskUserQuestion plugin provides the structured administrator-question workflow we need across the current providers and modes. We are adopting that smaller configuration-only workaround instead of changing the Codex provider bridge.\n\n> AGENT GENERATED

Issue #3566, opened 2026-09-12
Bundled npm in the macOS app is shipped without its own node_modules, so every plugin install fails with Cannot find module 'proc-log'


itizarsa:
## Summary

The npm copied into the macOS app bundle at
`bb.app/Contents/Resources/app.asar.unpacked/node_modules/npm` is missing its
`node_modules/` directory entirely. npm therefore throws before it can do any
work, and every plugin install that shells out to it fails with exit code 7.

## Environment

- bb 0.43.0 (desktop app, macOS, Darwin 25.6.0, arm64)
- Bundled npm: 11.16.0
- Electron Node: v24.15.0

## Reproduction

1. Install bb 0.43.0 on macOS.
2. Install any plugin from the Extensions UI.
3. The install fails with `npm install failed (exit 7)` and a `MODULE_NOT_FOUND`
   stack ending in `npm/lib/cli/exit-handler.js`.

The same failure reproduces without bb by invoking the bundled npm directly:

```
$ node /Applications/bb.app/Contents/Resources/app.asar.unpacked/node_modules/npm/bin/npm-cli.js --version
/Applications/bb.app/Contents/Resources/app.asar.unpacked/node_modules/npm/lib/cli/validate-engines.js:25
    throw err
    ^

Error: Cannot find module 'proc-log'
```

## Cause

`exit-handler.js:1` is `require('proc-log')`. npm ships its dependencies in its
own nested `node_modules/`, but that directory is absent from the bundle:

```
$ ls /Applications/bb.app/Contents/Resources/app.asar.unpacked/node_modules/npm/
bin/  docs/  lib/  man/  LICENSE  index.js  package.json
```

`npm/node_modules/` does not exist, and `proc-log` is not hoisted into the
sibling `app.asar.unpacked/node_modules/` either (95 entries, none of them
npm's dependencies). This looks like a packaging filter that walks into `npm/`
and drops nested `node_modules`, rather than a missing `asarUnpack` rule —
`lib/` and `bin/` made it through intact.

## Impact

No plugin can be installed on a fresh 0.43.0 macOS install. The failure is not
specific to any plugin, and the surfaced error points at npm internals rather
than the actual cause, which makes it hard to diagnose from the UI.

## Workaround

Replace the bundled npm with a working one (breaks the code signature and is
undone by the next update):

```bash
cd /Applications/bb.app/Contents/Resources/app.asar.unpacked/node_modules
mv npm npm.broken
[...]

Issue #3118, opened 2026-09-05
macOS desktop 0.42.0: missing README.md makes daemon auto-update return HTTP 500


ben-vargas:
## Summary

After upgrading a two-Mac setup to bb Desktop 0.42.0, the enrolled execution machine remained offline with `Needs update · daemon protocol 176 · server protocol 180`, even though both desktop apps were already updated. The primary desktop server's `/install/bb-app.tgz` endpoint returns HTTP 500 because host-package generation tries to copy a `README.md` missing from the installed application. The enrolled daemon should download the matching package, restart, and reconnect automatically.

## Versions and environment

- Primary: bb Desktop **0.42.0**, `/Applications/bb.app`, macOS **26.6.2 (25G83)**, arm64; server protocol **180**.
- Enrolled machine: bb Desktop **0.42.0**, macOS **27.0 (26A5425a)**, arm64; separate launchd-managed `bb-app` daemon still at **0.41.0**, protocol **176**.
- Daemon runtime: Node **24.14.0**, npm **11.9.0**.
- Existing enrollment uses `host-daemon --auto-update`, a private npm prefix under `~/.bb-machines/<server>/npm`, and its original pairing/configuration.
- The local desktop targets the primary's custom HTTPS URL over Tailscale, on port **38886**. The enrolled daemon listens locally on **38888**.
- No provider-specific operation is needed to reproduce this.
- Live reproduction used installed release binaries. Source inspection used commit [`7d5de7301544de3073996b4f8a8c33b311ecb2db`](https://github.com/get-bb/bb/commit/7d5de7301544de3073996b4f8a8c33b311ecb2db); a source-built server was not tested.

## Steps to reproduce

The smallest reproduced failure is the download endpoint on the running macOS desktop server; a second machine is not needed to exercise that failure.

1. Run the installed macOS bb Desktop 0.42.0 application as a server.
2. Set `BB_REPRO_SERVER` to that server's reachable URL. For the observed setup this was an existing Tailscale HTTPS URL:

   ```sh
   BB_REPRO_SERVER='https://YOUR-PRIMARY.YOUR-TAILNET.ts.net:38886'
   curl -sS -i "$BB_REPRO_SERVER/install/version"
   curl -sS -i "$BB_REPRO_SERVER/install/bb-app.tgz"
   ```

3. Observe a successful version response followed by HTTP 500 for the package download.
4. On the primary Mac, check the expected source file:

   ```sh
[...]

PR #3914, opened 2026-09-18
Use the bundled npm and npx for server-side package-manager calls

### LINKED ISSUE #3773: macOS Desktop 0.43.1: /install/bb-app.tgz returns HTTP 500 with spawn npm ENOENT, blocking Modal provisioning
## Summary

Creating a thread with a Modal sandbox from bb Desktop 0.43.1 fails during `bb machine setup` because the server's `/install/bb-app.tgz` endpoint returns HTTP 500. The server log records `spawn npm ENOENT` while generating the host artifact. I expected the sandbox to download the matching bb daemon package and complete provisioning.

## Versions and environment

- bb Desktop **0.43.1**, installed at `/Applications/bb.app`
- macOS **15.2** (24C101), arm64
- Bundled `environment-modal-sandbox` plugin 0.1.0
- Default server port `38886` and data directory `~/.bb`
- bb Connect enabled; the local and public routes reproduce identically
- BB launched normally as a macOS desktop application
- Server process PATH: `/usr/bin:/bin:/usr/sbin:/sbin`
- Homebrew npm: `/opt/homebrew/bin/npm`
- Tested release tag: `desktop-v0.43.1`, commit `267938526dfcbc0edb228ce827b5bec202c1af97`

## Steps to reproduce

The artifact endpoint is the smallest reproduction and does not require allocating Modal compute.

1. Launch bb Desktop normally on macOS.
2. Confirm the installed server version:

   ```sh
   curl -sS http://127.0.0.1:38886/install/version
   ```

3. Request the host artifact:

   ```sh
   curl -sS \
     --output /tmp/bb-app.tgz \
     --write-out '%{http_code}\n' \
     http://127.0.0.1:38886/install/bb-app.tgz
   ```

4. Inspect the new entries in `~/.bb/logs/server-stdio.log`.

To observe the product impact:

1. Configure the bundled Modal sandbox plugin with valid credentials.
2. Create a thread using the Modal sandbox environment.
3. Provisioning reaches `bb machine setup` and fails while downloading `/install/bb-app.tgz`.

## Expected vs actual

Expected:

```text
/install/bb-app.tgz returns HTTP 200 with the server-matched bb host artifact.
The Modal machine installs the artifact, connects to BB, and thread provisioning continues.
```

Actual version and account checks:

```text
$ bb --version
0.43.1

$ curl -sS http://127.0.0.1:38886/install/version
{"version":"0.43.1","protocolVersion":207}

$ bb modal account inspect --json
{"available":true,"message":"Connected to Modal (bb-sandboxes)"}
```

Actual artifact response:

```text
$ curl -sS \
    --output /tmp/bb-app.tgz \
    --write-out '%{http_code}\n' \
    http://127.0.0.1:38886/install/bb-app.tgz
500

$ cat /tmp/bb-app.tgz
{"code":"internal_error","message":"Internal server error"}
```

The same request through the account's public `*.getbb.app` route also returns HTTP 500.

Modal provisioning reports:

```text
Machine bootstrap command failed:

  bb machine setup

✗ Could not download the server's bb-app package from
  https://<account>.getbb.app/install/bb-app.tgz (HTTP 500).
```

## Evidence

Two requests made at 2026-09-16 12:43:34 UTC—one against localhost and one through bb Connect—produced matching server errors:

```text
[07:43:34] ERROR: [server] Unhandled server error
    err: {
      "type": "Error",
      "message": "spawn npm ENOENT",
      "stack":
          Error: spawn npm ENOENT
              at ChildProcess._handle.onexit (node:internal/child_process:287:19)
              at onErrorNT (node:internal/child_process:508:16)
              at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
      "errno": -2,
      "code": "ENOENT",
      "syscall": "spawn npm",
      "path": "npm",
      "spawnargs": [
        "pack",
        "--pack-destination",
        "/Users/<user>/.bb/install-cache"
      ],
      "cmd": "npm pack --pack-destination /Users/<user>/.bb/install-cache",
      "stdout": "",
      "stderr": ""
    }
```

The running server is the installed desktop binary:

```text
/Applications/bb.app/Contents/MacOS/bb
/Applications/bb.app/Contents/Resources/app.asar.unpacked/node_modules/bb-app/server/dist/index.js
```

Its environment contains:

```text
PATH=/usr/bin:/bin:/usr/sbin:/sbin
```

The application ships a functional npm 11.16.0. Invoking it directly works even with an empty PATH:

```sh
PATH='' ELECTRON_RUN_AS_NODE=1 \
  /Applications/bb.app/Contents/MacOS/bb \
  /Applications/bb.app/Contents/Resources/app.asar.unpacked/node_modules/npm/bin/npm-cli.js \
  --version
```

```text
11.16.0
```

### Cause hypothesis

This is separate from the observed reproduction, but the release code strongly supports it.

The artifact builder passes bare `"npm"` to its command runner:

https://github.com/get-bb/bb/blob/267938526dfcbc0edb228ce827b5bec202c1af97/apps/server/src/services/install/bb-app-artifact.ts#L269-L274

The default runner passes that command directly to `execFile`, requiring `npm` to be on the desktop server's PATH:

https://github.com/get-bb/bb/blob/267938526dfcbc0edb228ce827b5bec202c1af97/apps/server/src/services/install/bb-app-artifact.ts#L66-L75

The desktop application already ships a working npm, but this artifact-generation path does not resolve or invoke it.

Investigation thread: https://whiny.getbb.app/threads/thr_t5qimadvke

## What you ruled out

- Not invalid Modal credentials: `bb modal account inspect --json` succeeds.
- Not Modal image construction: the failure occurs while downloading BB's daemon artifact during bootstrap.
- Not bb Connect or external reachability: localhost and the public `*.getbb.app` route fail identically.
- Not a transient HTTP failure: the endpoint reproduces consistently and logs a matching exception for every request.
- Not #3118: that issue involved a missing `README.md`; this release logs `spawn npm ENOENT`.
- Not #3566: that issue involved missing dependencies inside bundled npm; the bundled npm in 0.43.1 executes successfully when addressed directly.
- Not a source checkout: reproduction uses the unmodified official Desktop 0.43.1 application.
- Searched open and closed issues and pull requests for `spawn npm ENOENT`, `bb-app.tgz`, Modal sandbox provisioning, npm PATH, and machine setup. No issue with this root cause was found.

High — blocks Modal sandbox provisioning and potentially other fresh machine enrollment from affected macOS desktop installations launched with the normal GUI PATH. There is no supported user-facing workaround and the failure is surfaced only as HTTP 500. Effort is likely Low because the application already bundles a functioning npm and has existing code paths that resolve it explicitly.

> AGENT GENERATED

ymichael:
## What was wrong

A macOS Desktop server launched from Finder inherits `PATH=/usr/bin:/bin:/usr/sbin:/sbin`. Three server-side code paths resolved a package manager through that PATH instead of the npm that bb already bundles, so each failed with `ENOENT` on any machine without Node on the system PATH — which is exactly the population bb bundles npm for.

`bb-app-artifact.ts` passed a bare `"npm"` to `execFile`, so `GET /install/bb-app.tgz` threw `spawn npm ENOENT` and returned HTTP 500, blocking `bb machine setup` for Modal sandboxes and every other fresh machine enrollment (#3773). `server-move/full-artifact.ts` carried `runPackCommand`, a byte-for-byte copy of the same PATH-dependent runner, so a server move failed the same way while packing the full bb-app. `skills/registry-skill-install.ts` spawned bare `npx` and failed with `spawn npx ENOENT`.

## What changed

- `apps/server/src/services/install/bb-app-artifact.ts` — `defaultCommandRunner` resolves the bundled npm CLI and runs it under `process.execPath`, dropping inherited npm script-policy env. This matches `runInstallCommand` in the plugin install service, which already used this pattern. Exported so the one runner can be shared.
- `apps/server/src/services/server-move/full-artifact.ts` — deleted the duplicate `runPackCommand`; `createFullBbAppArtifactService` now defaults `commandRunner` to the shared runner, matching `createBbAppArtifactService`. `environment.ts` no longer wires one explicitly.
- `apps/server/src/services/skills/registry-skill-install.ts` — runs the bundled `npx-cli.js` under `process.execPath` and forwards `ELECTRON_RUN_AS_NODE` through the env allowlist so it works when the server runs under Electron.
- `packages/plugin-build/src/npm-cli.ts` — added `resolveBundledNpxCli` beside the npm resolver; renamed `resolvePluginNpmCli` to `resolveBundledNpmCli` now that it serves more than plugin installs (5 call sites updated).

No wire, CLI, or configuration changes, so no `HOST_DAEMON_PROTOCOL_VERSION` bump.

Deliberately unchanged: `provider-maintenance-kit.ts` (`npmCommand()` installs provider CLIs into the *user's* global prefix and renders a copy-pasteable command; provider processes get a login-shell PATH), `host-daemon/protocol-self-update.ts` (installs into the machine's npm prefix, already prepends `dirname(process.execPath)`), `apps/cli` (runs in the user's terminal against their project), and `install-machine.sh` (runs on the target machine). These need the user's npm by design.

## How you verified

Tests (each verified to fail before the fix and pass after):

- `apps/server/test/app/bb-app-artifact.test.ts` — builds a real artifact with `PATH=""`. Without the fix: `Error: spawn npm ENOENT`.
- `apps/server/test/server-move/full-artifact.test.ts` — same, through `createFullBbAppArtifactService` and its production default runner.
- `apps/server/test/skills/registry-skill-install.test.ts` — asserts the spawned command is `process.execPath` with an existing `npx-cli.js` and a forwarded `ELECTRON_RUN_AS_NODE`. Without the fix: `expected 'npx' to be '/…/node'`.

Suites: `pnpm exec turbo run typecheck lint test --filter='...[origin/main]'` — 97 tasks, all green except one pre-existing parallel-load flake in `bb-plugin-provider-codex` (`cancels an open SSE body after a terminal failure event`), which passes on its own re-run. `@bb/server` 2980 pass, `@bb/plugin-build` 139 pass.

End-to-end against a real packaged install (`npm pack` + `npm install -g --prefix`), running the server under Electron with `ELECTRON_RUN_AS_NODE=1` and `PATH=/usr/bin:/bin:/usr/sbin:/sbin`:

- `GET /install/bb-app.tgz` — pre-fix HTTP 500 reproducing the issue's log verbatim (`"message": "spawn npm ENOENT"`, `spawnargs: ["pack","--pack-destination",…]`); post-fix HTTP 200, byte-identical artifact (`sha256=779a2b46…`) across node/Electron hosts and restricted/full PATH. The downloaded artifact installs and runs (`bb --version` → `0.43.1`).
- `bb machine setup` over the network onto a second Mac — `✓ Downloaded the server's bb-app package` → `✓ Installed` → `✓ Joined successfully` → `● bb machine is ready`.
- A real `bb server move` to that machine — the `Packing bb 0.43.1` step produced `bb-app-full-0.43.1-9fd32ebe….tgz` (47 MB) with no npm on PATH, and the target downloaded, installed, and took over.

Known gap, pre-existing and not a regression: registry skill install now gets past `spawn npx ENOENT` and runs the bundled npx, but the third-party `skills` CLI it fetches carries `#!/usr/bin/env node`, so on a bare PATH it fails one step later with `env: node: No such file or directory`. That is #2343 (the packaged login-shell PATH probe timing out), plus a residual gap for users with no Node at all. Left for a follow-up.

Fixes #3773

> AGENT GENERATED


ymichael:
Added a third commit after reviewing how #3566 broke 0.43.0.

That release shipped bundled npm without its nested `node_modules` because electron-builder dropped it, and it got through because source builds and the bb-app tarball smoke both passed — packaging was the only broken mode. The `afterPack` gate added by #3590 (`smoke-packaged-npm.mjs`) exists to stop exactly that, but it only resolved and ran `bin/npm-cli.js`. This branch adds a runtime dependency on `bin/npx-cli.js`, which had no packaging coverage.

`a0076e5e` resolves both CLIs through `insideArtifact` (realpath + containment, so a missing file fails the gate) and runs `npx --version` through the packaged Electron with an empty PATH, mirroring the npm check. It also adds a missing-file check for `npm-cli.js`, which was previously only path-joined.

Verified against the installed signed release:

- gate passes — `Packaged npm 11.16.0: 65 direct dependencies, 144 bundled packages, 348 CJS/ESM dependency edges; offline plugin dependency install/update passed with Electron 41.7.0 (Node 24.15.0) and empty PATH; packaged resources unchanged.`
- negative control — pointing `npxCli` at a non-existent filename fails the gate with `ENOENT: no such file or directory, realpath .../npm/bin/npx-cli-MISSING.js`
- both bundled CLIs run standalone under the release Electron with `PATH=""`: `npm-cli.js --version` and `npx-cli.js --version` each print `11.16.0`

Caveat worth stating plainly: the end-to-end validation in the PR description used `npm pack` + `npm install -g`, which is the same mode that stayed green while #3566 was broken. A real `turbo run package --filter=@bb/desktop` has not been run on this branch; the gate above is what would catch a packaging regression.

> AGENT GENERATED

ymichael:
Ran the real packaging, closing the gap I flagged above.

`CSC_IDENTITY_AUTO_DISCOVERY=false pnpm exec turbo run package --filter=@bb/desktop` — **51/51 tasks passed**. The `afterPack` gate ran against the freshly built artifact with the Electron that will actually ship:

```
Packaged npm 11.16.0: 65 direct dependencies, 144 bundled packages, 348 CJS/ESM
dependency edges; offline plugin dependency install/update passed with
Electron 44.3.0 (Node 24.20.0) and empty PATH; packaged resources unchanged.
```

Then drove the packaged app itself. Ran its server as Desktop does — `ELECTRON_RUN_AS_NODE=1`, `PATH=/usr/bin:/bin:/usr/sbin:/sbin` — straight out of `release/mac-arm64/bb.app`:

- `GET /install/bb-app.tgz` → **HTTP 200**, 1275210 bytes, `sha256=8b32ecb3…` matching the `x-bb-artifact-sha256` header, 45 entries, zero `spawn npm ENOENT` in the logs. (45 rather than 46 because the packaged app omits `README.md` — the #3118 case, handled.)
- Bundled CLIs in the fresh package, run under empty PATH through the packaged Electron: `npm-cli.js --version` → `11.16.0`, `npx-cli.js --version` → `11.16.0`.
- `npm/node_modules` has its full 118 entries and `bin/` contains `npx-cli.js`, so the #3590 file set survives with this branch applied.
- Registry skill install on the packaged server now fails with `env: node: No such file or directory` rather than `spawn npx ENOENT`, confirming the npx fix takes effect in the packaged app and that the residual gap is only the documented third-party `#!/usr/bin/env node` shebang (#2343 territory).

This was the mode that let #3566 ship — source builds and the tarball smoke passed while packaging was broken. It is now exercised directly.

> AGENT GENERATED

PR #3943, opened 2026-09-19
Fix environment grouping across custom sidebar sections

ymichael:
## What was wrong

With sidebar organization set to Custom and By environment enabled, environment groups were built before custom sections. The first thread in each environment determined the section for the entire group, so other threads sharing that environment appeared in the wrong section or outside their assigned sections.

## What changed

Partition root thread nodes by section before grouping environments in custom organization mode. Threads sharing an environment now group only within their own section, including the unsectioned bucket. Preserve parent/child trees and existing project organization behavior. Add regression coverage for grouping enabled/disabled, separate sections sharing an environment, unsectioned threads, missing section definitions, descendants, and section counts.

## How you verified

- Ran the new tests against the original implementation: two regression cases failed; all 300 client-core tests pass with the fix.
- `pnpm exec turbo run test typecheck --filter=@bb/client-core` — passed.
- `pnpm exec turbo run test --filter=@bb/app -- sidebarDropPreviewPlacement` — all five tests passed.
- Started the branch with `pnpm start:worktree` using a fresh isolated database. Captured before/after browser screenshots using four synthetic idle threads sharing one environment across two sections and the unsectioned bucket.
- Verified source bb CLI section moves, clearing a section, and readback against that server; browser reloads preserved the correct assignments.
- Verified actual browser interactions for grouping off/on, By project, and section collapse/expand. Each fixture thread remained reachable exactly once; restored Custom + By environment for manual QA.

> AGENT GENERATED


ymichael:
Superseded by #3946, which includes this PR's section-aware environment grouping implementation and regression tests alongside worktree group/member dragging. The combined branch passes all 300 client-core tests, 78 focused app tests, and both package typechecks.

